PR #2864, opened 2015-08-14
Fix Canon dual ISO problem

heckflosse:
please test


heckflosse:
When the fix is confirmed, I'll add the dcraw.patch file

PR #2866, opened 2015-08-15
Fix crash when using large output profile saving to uncompressed TIFF

heckflosse:


heckflosse:
For an unknown reason a big diff was created...

Hombre57:
Isn't it related to line endings ?

heckflosse:
because of the Line endings bug I close this pull request and create a new one

PR #2867, opened 2015-08-15
Crash when saving to uncompressed tiff using a huge output profile

heckflosse:


Hombre57:
What are we waiting for pulling this to master ? The patch is small and fix an obvious bug.

Issue #2817, opened 2015-08-11
Add new "perceptual" curve type


Beep6581:
Originally reported on Google Code with ID 2837
```
I'm thinking of adding a new curve type "neutral" with the follwing properties:

- Constant color appearance despite contrast change
    - That is take into account human vision phenomena such as that
      higher contrast requires higher saturation to look the same
- Working space independent, that is regardless of working space the same
  result is had (clipping issues excepted)

I've done this type of work for my camera profiler DCamProf already so I'll see if
I can make a lighter weight realtime version of that. A discussion about how tone curves
affect color apperance can be found at:

http://www.ludd.ltu.se/~torger/dcamprof.html#tone_curves

Why is this needed? RT is designed to use "linear colorimetric" profiles, that is profiles
without any curve applied. This only makes sense if RT can retain the color appearance
when a curve is applied, and today there is no such curve. Weighted standard and luminance
indeed come quite close but it's possible to make an even better result.

Commerical raw converters generally include the curve in the profile itself and then
provide color rendition as a complete package with the curve and all, so the need of
a neutral tone curve is specific to RT through the use of colorimetric profiles.
```

Reported by `torger@ludd.ltu.se` on 2015-07-11 09:24:45

Beep6581:
This is perceptual_speedup_00.patch re-formatted to apply to astyle'd code:
https://gist.github.com/99421e4dee70a77e44a6

heckflosse:
Thanks DrSlony, created pull request https://github.com/Beep6581/RawTherapee/pull/2870

PR #2870, opened 2015-08-17
Speedup for perceptual curve, #2817

heckflosse:


Hombre57:
I've tried your patch and found a slight difference in reds close to the gamut. It's late now, I'd like to test more tomorrow, but is this something already noticed and should we care ? I'll try to report some processing time if you tell me where to put the timer.

heckflosse:
Hombre, Michael also reported differences. There are differences in calculations between ciecam02 from rt and ciecam02 from lcms. So some differences in output are expected. But I don't know which one is 'better' or whether the differences matter at all.

Ingo

Hombre57:
I built the HEAD version in release mode, but don't have numbers to provide (no console window). Anyway, I guess that the speedup is consistent with what you generally achieve :) So :+1 to commit this patch on my side.

heckflosse:
Hombre, for information: Processing time of rgbproc() on my machine for full processing a D800 file
was 26700 ms before changes
is 4700 ms after changes.

Ingo

Hombre57:
well, it's an improvement, indeed :)

Is there any known issue that prevent committing this ?

heckflosse:
Yes, the differences in output. I would be glad to get a review from Anders because he implemented the original version. I also want to try further speedup using partly SSE coding (rt ciecam02 functions are already available for SSE)

Beep6581:
I tested it thoroughly before the github switch and noticed only changes in strong purple hues. The changes were of such a nature that they didn't matter. I'll test head later today. +1 for commit.

Hombre57:
The difference are in deep reds too (when reaching the gamut), e.g. with a rose under bright sunny day. Maybe we should take a deeper look ?

heckflosse:
Ok, two +1 for commit should be enough ;-)
I'll remove StopWatch stuff, clean the code and merge later in the evening

heckflosse:
Ok, I'll push cleaned code later today and then you can merge it or not ;-)

heckflosse:
I cleaned the code and pushed

atorger:
Sorry for not having been able to contribute on this, been choked with other stuff, and I'm entering a period now when I can't contribute as much. I'm sure it's fine though :-), the perceptual curve is no exact science anyway.

heckflosse:
Anders, glad to see you here on github and thank you for commenting. I'll try further speedup using SSE ciecam02 functions later.

Ingo

heckflosse:
For information, https://github.com/Beep6581/RawTherapee/commit/df13fb8b1b7b5fc4f4ee1304cfb974221fa6ff84 should fix the differences bug

heckflosse:
I'm working on an additional speedup for perceptual curve (using SSE). I need some days to complete work.

PR #2875, opened 2015-08-20
Add some new features to the On Preview Objects mechanism

Hombre57:
This patch moves the Coord class from rtgui to rtengine and let us differentiate the mouse button that is used to operate on the on preview widgets. This is the "safe part" of the Spot Removal patch. It'll then simplify the management of this WIP patch if I could commit this part to master.

You'll know that it doesn't brake anything if using the Pipette and the Gradient's On Preview widgets works fine. For the later one, the highlight color is now orange, but I can bring back the yellow color. I know, it's a matter of taste, but I don't plan to let the user chose in Preference for now.


Hombre57:
If no comment, I'll merge this patch tonight. Tested here and works fine.

PR #2960, opened 2015-11-01
Add explicit type conversions to satisfy C++11 requirements.

bobobo1618:
C++11 isn't a fan of narrowing. Just used explicit casts on the few parts of the codebase that relied on it.


Beep6581:
@heckflosse could you take this?

bobobo1618:
Warning: This was originally a pul request from a fork that didn't contain anything else. It's since evolved to contain my (totally separate and unrelated) Halide changes.

I'll fork things soon so they can be handled separately.

Beep6581:
I took a look at the changes, this should not be merged into master.

bobobo1618:
Could you elaborate a bit? This PR is about https://github.com/bobobo1618/RawTherapee/commit/b93c73007ca38c0d49c0a558328806ab7cff35da specifically.

The reason I did this is that `glibmm` 2.4 and `sigc++` 2.6.1 (a `glibmm` dependency) both use C++11 features in their header files. Without the `-std=c++1y` flag on compilation, RawTherapee fails to compile, at least using Clang (the following example is a small snippet):

```
/usr/local/Cellar/libsigc++/2.6.1/include/sigc++-2.0/sigc++/functors/slot.h:737:21: error: expected ';' at end of declaration list
  slot3(slot3&& src) noexcept
                    ^
                    ;
/usr/local/Cellar/libsigc++/2.6.1/include/sigc++-2.0/sigc++/functors/slot.h:835:14: warning: rvalue references are a C++11 extension [-Wc++11-extensions]
  slot4(slot4&& src) noexcept
```

On the other hand, C++11 doesn't allow implicit conversions, so RT fails to compile _with_ the `-std=c++1y` flag a well. This commit simply makes the conversions explicit so that RT is C++11 compatible.

All the \*mm projects and `sigc++` [have moved to C++ 11](http://www.murrayc.com/permalink/2015/07/31/gtkmm-now-uses-c11/) so fixing these things seems necessary, if not now then at some point in the future. And it seems easier to change 15 lines of RawTherapee code than to try to port the RT dependencies.

bobobo1618:
Followed up with split out commit in https://github.com/Beep6581/RawTherapee/pull/2966. Ignore this one now.

Beep6581:
The merge would have done more than just pull in b93c730, that is the reason I closed it.
As for b93c730 itself, of course you are right that C++11 compatibility is needed, better sooner than later, but I don't speak that language so I'll leave it with @heckflosse If its decided that those changes are good and safe then of course we will merge them.

Issue #2934, opened 2015-10-13
Halide


bobobo1618:
Not sure if you guys are aware of this but [Halide](http://halide-lang.org/) seems pretty cool. It's open source, MIT licensed (GPLv3 compatible? I'm not sure) and through a single API can generate code for "x86/SSE, ARM v7/NEON, CUDA, Native Client, and OpenCL".

Given that it enables simpler SSE implementation and easier use of CUDA/OpenCL, it seems like something that could be a good idea for performance.

If I get time to send pull requests implementing/porting image processing to Halide, are they likely to be approved?

heckflosse:
@bobobo1618 you're welcome to send pull requests :-) I will test them on Windows and Linux then

Ingo

bobobo1618:
I just got a bilinear debayer (I'm new to this and wanted something simple) running at 640MP/s on my laptop so I have high hopes for it :)

I'm looking into other demosaicing algorithms though, as [some of](http://www.csee.wvu.edu/~xinl/papers/demosaicing_survey.pdf) the latest academic work has suggested that algorithms like DFPD are pretty promising.  I'm curious how they compare to AMAZE though and couldn't see much literature mentioning it despite it being around for around 7 years. Is anyone aware of any comparisons?

heckflosse:
Hmm, to debayer (independent on method used) 640 MP you need to transfer 640000000 floats = 2560000000 bytes over PCIe bus to graphics card and 1920000000 floats = 7680000000 from graphics card. To do that transfer in one second you have to transfer 10240000000 Bytes = 9756 MBytes in one second which means you need PCIe 3.0 x 16 or PCIe 2.0 x32. And that doesn't include the time to debayer the data.

Ingo

bobobo1618:
I suppose I should elaborate on that number a bit then:
- Debayer was run entirely on CPU. GPU wasn't touched.
- Debayer was pre-compiled (didn't compile it 100 times).
- Debayer was run 100 times, the figure I gave is the average over that time.
- Data was 8-bit (`dcraw -d`).
- I was operating on 16-bit unsigned integers (large enough to contain the sum of two 8-bit integers), not floats.
- Data was in memory prior to running the benchmark.
- Data was 16MP, the 640MP/s I mentioned was the _rate_ (16/(time_to_compute)). I don't actually have a 640MP image to test.
- Halide was using AVX, AVX2 and SSE4.1 instructions, as well as parallelisation.

I've run other tests as well though:
[...]

PR #2965, opened 2015-11-06
Halide

bobobo1618:
Add Halide implementation of [IGD](http://www.eie.polyu.edu.hk/~enyhchan/J-JEI-Low_complexity_color_demosaicing_algorithm_based_on_IG.pdf)  (https://github.com/Beep6581/RawTherapee/issues/2934) to RawTherapee along with build system for generating the code.


heckflosse:
@bobobo1618 : Thanks a lot for this work!!! There are some Issues to solve before merging (Some artifacts, allow to build on Windows and mac, maybe additional issues too). So we won't merge it at current state. But I'm willing to support you (if needed) to solve the remaining issues.

Ingo

bobobo1618:
Mac works as is (that's what I develop using), Linux appears to require a couple of extra libraries to be linked (`-ldl -lpthread`) but I don't know how to use platform specific flags and Windows is a pain so I haven't tried yet.

Do you mean image artifacts? I contacted the authors and they said that they're a result of the way the algorithm handles gradients. I don't think I'll be able to get rid of them. I'd suggest keeping the algorithm as is, just as a way to demonstrate Halide and then port a higher quality algorithm (AMAZE if someone can explain it) to Halide later.

adamreichold:
@bobobo1618 I just wanted to add that I was able to build this using the latest binary distribution from the project and running on Arch Linux amd64. However, I did have to prepend `dl` to link libraries of the RT GUI binary. Also my system does not support OpenCL and building with the default `OPTION_HALIDE_OPENCL=ON` resulted in the program crashing, i.e.
````
Program received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffa9b08700 (LWP 16208)]
0x00007ffff0e0b5f8 in raise () from /usr/lib/libc.so.6
(gdb) bt full
#0  0x00007ffff0e0b5f8 in raise () from /usr/lib/libc.so.6
#1  0x00007ffff0e0ca7a in abort () from /usr/lib/libc.so.6
#2  0x0000000000e17f80 in Halide::Runtime::Internal::default_error_handler(void*, char const*) ()
#3  0x0000000000e1b30a in Halide::Runtime::Internal::OpenCL::load_libopencl(void*) ()
#4  0x0000000000e1de6d in halide_opencl_device_malloc ()
#5  0x0000000000e1a588 in halide_device_malloc ()
#6  0x0000000000e1a465 in halide_copy_to_device ()
#7  0x0000000000dbb7d1 in rtengine::RawImageSource::igd_halide_demosaic (this=0x7fffcc001110, winx=0, winy=0, 
    winw=4284, winh=2844) at /home/adam/RawTherapee/rtengine/igd_halide_demosaic.cpp:84
````
Shouldn't failure to load the OpenCL libraries should result in a CPU fallback being used? With OpenCL being disabled at build time, I however was able to build and run this without immediate problems. Also I wonder if it really is intended that we would ship so much of Halide within the RT source tree, e.g. shouldn't the headers come the system includes?

@heckflosse I personally also like @bobobo1618's proposal to concentrate on the build system and integration issues at first and basically ignore questions of the algorithm's result quality for now as the change should make it easier experimenting with different algorithms but without being bogged down in low-level optimization issues. I also think that the pull request would benefit greatly from being rebased on the current master with some of the commits getting squashed.

adamreichold:
@bobobo1618 Concerning platform-specific flags, you could try to align necessary platform-specific build system changes with Halide's generate CMake manifest, e.g. use a case analysis like
````
if (WIN32)
# Windows
elseif(XCODE)
# Mac OS X
else()
# Unix
endif()
````

bobobo1618:
> Shouldn't failure to load the OpenCL libraries should result in a CPU fallback being used?

That would be ideal but that's up to the runtime and I'm unsure how to test for OpenCL support in C++ (this is the first serious C++ I've really written), particularly across different platforms. I was under the impression OpenCL was always supported though. Doesn't it run on the CPU as well as the GPU? How can it lack support?

> I wonder if it really is intended that we would ship so much of Halide within the RT source tree

I removed the runtime headers (the ones I've included) and switched them to the ones provided by Halide with the binary package (I didn't know it included headers at first). I did the same with the `GenGen.cpp` source. The CMake module has to be included though.

> you could try to align necessary platform-specific build system changes with Halide's generator CMake manifest

I've used that save for `XCODE OR ${CMAKE_SYSTEM_NAME} MATCHES "Darwin"` for Mac, since Mac doesn't always mean XCode.

>  I also think that the pull request would benefit greatly from being rebased on the current master with some of the commits getting squashed.

Done. Squashed the lot.

heckflosse:
There's a simple matter of fact: We can't merge this PR if we don't know whether it builds and works fine on Linux (32/64) and Windows (32/64).

Ingo

bobobo1618:
That's definitely true. I can test Mac and possibly Linux (Fedora) but someone else will have to test Windows.

adamreichold:
> That would be ideal but that's up to the runtime and I'm unsure how to test for OpenCL support in C++ (this is the first serious C++ I've really written), particularly across different platforms.

I'd say its up to the runtime to check, but from the stack trace my guess is that you need to override the default error handler `Halide::Runtime::Internal::default_error_handler` so that is does not call `abort` but just switches to a fallback CPU-only generator.

> I was under the impression OpenCL was always supported though. Doesn't it run on the CPU as well as the GPU? How can it lack support?

At least on Linux it is not, e.g. on my system there is no OpenCL implementation (and no ICD) installed at all, so even trying to load those libraries will fail. And even then, I don't think a GPU-based OpenCL implementation is mandated to provide a CPU-based fallback. Of course, one could make a fallback OpenCL implementation like [POCL](http://pocl.sourceforge.net/) a dependency of RawTherapee, but I don't think this is the best decision. As I understand it, Halide already is the abstraction layer we could use to generate optimized code for different targets and also gives us some control over the trade-offs made for the various code generators. So we probably do not want to have the default code lower our algorithm into OpenCL C to just compile that for the host CPU only losing information along the way. Instead the Halide generator should directly output something making use of all the CPU features even if they for example can not be mapped to OpenCL.

@heckflosse I would be willing to help testing this on Linux/64 (Is Linux/32 still a valid use case for performance critical applications like RT?). As I said above, after disabling OpenCL is does build and run on my box. I will give it a try using POCL and Mesa's R600g OpenCL implementation and also do a code review now that it's rebased as soon as I find some spare time.

Beep6581:
Hi @bobobo1618 are you still willing to maintain this if it passes the requirements?

bobobo1618:
After the concerns @adamreichold raised, I'm not sure if I'm technically able to do so. C++ isn't really my area of expertise and while I was able to get this working now, I'm not confident that I'll be able to keep it working if something changes.

heckflosse:
@adamreichold what do you think? Close or keep it open?

Beep6581:
I will close this PR, but @bobobo1618 this does not mean that we express no interest in it, it only means we don't want an unmergeable PR lingering around. As for your Halide implementation, you could keep it in your repo until the required changes are done some time in the future, or we could even make a halide branch which would simply clone your code if there is any benefit in doing that. For now lets keep using issue #2934.

PR #2935, opened 2015-10-14
Fixes for build with C++11 flags and sigc++ >= 2.5.2

mbajor:

PR #2966, opened 2015-11-06
Add explicit type conversions to satisfy C++11 requirements.

bobobo1618:
Follow up on https://github.com/Beep6581/RawTherapee/pull/2960 but split out from the Halide work.

Seems to be in the same vein as https://github.com/Beep6581/RawTherapee/pull/2935.

Summary of motivation:

 `glibmm` 2.4 and `sigc++` 2.6.1 (a `glibmm` dependency) both use C++11 features in their header files. Without the `-std=c++1y` flag on compilation, RawTherapee fails to compile, at least using Clang (the following example is a small snippet) but likely with `g++` as well:

```
/usr/local/Cellar/libsigc++/2.6.1/include/sigc++-2.0/sigc++/functors/slot.h:737:21: error: expected ';' at end of declaration list
  slot3(slot3&& src) noexcept
                    ^
                    ;
/usr/local/Cellar/libsigc++/2.6.1/include/sigc++-2.0/sigc++/functors/slot.h:835:14: warning: rvalue references are a C++11 extension [-Wc++11-extensions]
  slot4(slot4&& src) noexcept
```

On the other hand, C++11 doesn't allow implicit narrowing conversions, which RT makes (limited) use of, so RT fails to compile _with_ the `-std=c++1y` flag a well. This commit simply makes the conversions explicit so that RT is C++11 compatible.

All the \*mm projects and `sigc++` [have moved to C++ 11](http://www.murrayc.com/permalink/2015/07/31/gtkmm-now-uses-c11/) so fixing these things seems necessary, if not now then at some point in the future.

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -1937,9 +1937,11 @@ void CLASS hasselblad_correct()
         const ushort corners_shift[9] = { 2, 1, 2, 1, 0, 1, 2, 1, 2 };
         for (row = 0; row < bh; row++) {
             const ushort maxdist = bw < bh ? bw/2-1 : bh/2-1;
-            const unsigned corners[9][2] = {{0,0},   {0,bw/2},   {0,bw-1},
-                                            {bh/2,0},{bh/2,bw/2},{bh/2,bw-1},
-                                            {bh-1,0},{bh-1,bw/2},{bh-1,bw-1}};
+            const unsigned bwu = (unsigned)bw;
+            const unsigned bhu = (unsigned)bh;
+            const unsigned corners[9][2] = {{0,0},    {0,bwu/2},    {0,bwu-1},
+                                            {bhu/2,0},{bhu/2,bwu/2},{bhu/2,bwu-1},
+                                            {bhu-1,0},{bhu-1,bwu/2},{bhu-1,bwu-1}};
             for (col = 0; col < bw; col++) {
                 for (i = 0; i < 9; i++) {
                     ushort dist = (ushort)sqrt(abs(corners[i][0] - row) * abs(corners[i][0] - row) + abs(corners[i][1] - col) * abs(corners[i][1] - col));
--- a/rtengine/improcfun.cc
+++ b/rtengine/improcfun.cc
@@ -2004,9 +2004,9 @@ void ImProcFunctions::ciecam_02float (CieImage* ncie, float adap, int begh, int 
         //matrix for current working space
         TMatrix wiprof = iccStore->workingSpaceInverseMatrix (params->icm.working);
         const float wip[3][3] = {
-            {wiprof[0][0], wiprof[0][1], wiprof[0][2]},
-            {wiprof[1][0], wiprof[1][1], wiprof[1][2]},
-            {wiprof[2][0], wiprof[2][1], wiprof[2][2]}
+            {(float)wiprof[0][0], (float)wiprof[0][1], (float)wiprof[0][2]},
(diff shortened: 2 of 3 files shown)


adamreichold:
IMHO, one should try to avoid C-style casts as their interpretation by the compiler is very broad. In these cases, static_cast seems appropriate which also what Clang suggests. Also note that narrowing conversions are still possible with C++11, it is only within initializer lists where it is forbidden, e.g. `int x = 3.5;` compiles but `int x { 3.5 };` does not. (GCC 5.2 for example seems to fall back to C-style array initialization and does compile these initializations.)

Is there a project decision on the use of C++11 language features within the RawTherapee code base itself?

(Of course, writing C++-style casts is often tedious but one could for example use a template like
```
template< typename R, std::size_t N, std::size_t M >
class matrix_cast
{
public:
    using result_type = R[N][M];
private:
    result_type r;
public:
    template< typename S >
    matrix_cast(const S& s)
    {
        for (std::size_t i = 0; i < N; ++i) {
            for (std::size_t j = 0; j < M; ++j) {
                r[i][j] = static_cast< R >(s[i][j]);
            }
        }
    }

    operator const result_type& () const { return r; }
};
```
to reduce the conversion to `const matrix_cast< float, 3, 3 > wip{iccStore->workingSpaceInverseMatrix (params->icm.working)};`.)

heckflosse:
@bobobo1618 I'll take a look at this tomorrow. Because you made changes to dcraw.cc I also have to include a modified dcraw.patch file into your PR.

Ingo

heckflosse:
@bobobo1618 I tested your PR. It compiles and works fine. But I will change one thing and push a new dcraw.patch file after merge.
@adamreichold You're right that it would be better to use c++-style casts but there are so many c-style casts in rt that IMHO it would be better to make a big throw replacing all c-style casts with c++-style casts than replacing one here and one there. When all c-style casts are replaced by c++-style casts we can use -Wold-style-cast / -Werror=old-style-cast to avoid c-style casting in future.

adamreichold:
@heckflosse What you described sounds a good plan and I would like to start working on it as it also sounds like a good way to get more familiar with the code base. But as that might already be useful there: Can we actively use C++11 language features within the RT code base or do we currently restrict ourselves to be able to build -std=c++11? (Actually C++14 might be a better choice for a standard as it is what GCC 6 will be doing and C++14 cleans up some rough edges of C++11.)

Beep6581:
C++11 "was approved by ISO on 12 August 2011, replacing C++03, and superseded by C++14 on 18 August 2014." It would make sense to use C++14 then. But any such code should be strictly optional, and RawTherapee must still be compilable using GCC-4 for a good deal of time to come.

adamreichold:
@Beep6581 Hhhmmm, one could try to do some optional performance enhancements using C++14 code paths, but I don't think that is very useful as the main benefit of C++11 and later is more maintainable code at a high level of abstraction (without overhead) which mainly implies refactoring current code to make use of those language features. So maybe using GCC 4.8 as a baseline for C++11 support but making that mandatory would be a better choice?

adamreichold:
@heckflosse It seems like -Werr=old-style-cast probably won't work as glibmm directly includes glib (which does not guard itself using `extern "C"` when used from a C++ code base) which of course does C-style casts. Even enabling the warning looks pretty hopeless due to that.

heckflosse:
RT win32 builds are still made using gcc 4.6.1
So we should keep an eye on this: https://gcc.gnu.org/projects/cxx0x.html when using c++ language features.

Issue #2942, opened 2015-10-20
Hide JPG files when RAW is present


neilpho:
In the manner of Lightroom it would be good if JPG versions of RAW files were hidden. A badge on the RAW thumbnail would indicate the presence of a matching JPG.

This would help those of us who shoot RAW+JPG.

sguyader:
Can't you simply use the Filter tab, enable Metadata filtering and filter based on "File type"?

neilpho:
I can't because not all images are in RAW+JPG, some are JPG only. Even within a single session.

Lightroom defaults to this behaviour for good reason.

neilpho:
Lightroom treats JPGs as sidecar files, which really they are, when there is a raw version.

adamreichold:
Hello, I just opened pull request #2970 which tries to implement such functionality using the existing filter mechanism. Maybe you could give it a try whether the implementation seem useful w.r.t. this issue?

adamreichold:
@neilpho Since PR #2970 was merged, I think this can be closed?

Beep6581:
Yes, done in https://github.com/Beep6581/RawTherapee/pull/2970

PR #2970, opened 2015-11-08
Add filter to restrict to original files

adamreichold:
This branch adds a new filter option which will restrict the file browser to entries which are original images, i.e. all images are collected by their file name without extension and the entry whose extension is "the most original" is selected as the original of the other ones who are then filtered out.


adamreichold:
If the basic functionality and the integration into the filter panel is considered useful, the branch has two open issues which need resolution before it can be merged: There is currently no icon for the new filter button and it abuses the extension list from the settings to determine what is the most likely original entry, i.e. this list needs to become user sortable at the least or maybe an independent mechanism should be used.

Beep6581:
Hello

The way we do things is to open an issue for a new suggested feature where we discuss how it should work, and then open a pull request once the code is accepted, tested and ready. That makes it easier to follow things now and in the future.

adamreichold:
Hello Beep6581,

this change is directly related to the functionality discussed in issue #2942, so there already is an issue which this pull request tries to address. That it is work-in-progress is so that there is a concrete implementation to discuss and actual code that can be commented on. Restricting that discussion to the format of an issue seems to make getting the code into an acceptable state harder. Otherwise one could just post a link to the branch into the issue instead and would not need the pull request format at all? Of course, if you prefer that style, please close this and the work can continue within issue #2942 only.

Beep6581:
"Otherwise one could just post a link to the branch into the issue instead and would not need the pull request format at all?"
That's what we do, though we're new to git and github.

heckflosse:
@adamreichold I tested your PR. Works fine here.

adamreichold:
Since there were no immediate objections to use the list of extensions to prioritize originals of derived types, I added the necessary buttons to the file browser preferences to make this user-sortable and rebased the branch.

Beep6581:
Hello, what is the status of this? Still WIP?

adamreichold:
Issue #2942 has not seen any feedback whether this change is considered a viable solution but so far there were also no objections against doing this as a filter and using the extension list to determine the originals. I think that the implemented functionality is complete expect for the missing icons for which I would need some advice/help on where to get/how to create them. If you have no objections to merging it, I'd say it is still WIP, but only until the icons are added.

Beep6581:
I can create the icons for you. Everyone seems busy this month, myself included, and I spent a lot of time during the last week on RT so in this coming week I probably won't, but I'll try to test this feature over the course of the week and propose some icons. Take a look at the SVG icons already shipped with RT and let me know if you have any suggestions.

adamreichold:
@Beep6581 Thank you for your offer w.r.t. the missing icons! I had a look at the `tools/source_icons/scalable` folder, and I think that `gtk-add` or `gtk-remove-red` are from some icon theme based on the FDo icon naming specification? Hence I would say this theme's `go-up` and `go-down` icons would be good candidates for the two buttons to sort the file extension list in the preferences dialog. For the filter button, I guess something like a stack of files or images similar to `gtk-copy` would be appropriate to indicate that some entries will be merged resp. stacked, but with more than two items to differentiate it from the copy operation.

Beep6581:
Sorry for taking so long. I locally merged and tested the branch now so I can understand what sort of icons are needed.

> all images are collected by their file name without extension and the entry whose extension is "the most original" is selected

What is considered most original? Is it only the file with the fewest same-format copies, or is it also its position in the "Parsed Extensions" list? And if the latter, is the higher one the preferable one?

Beep6581:
Here are the icons, I included two versions of the filter icon, see which you prefer:
http://filebin.net/vvryhru47f

adamreichold:
@Beep6581 Thank you for creating those icons! I have other obligations today, but will hopefully find time to integrate them into this branch tomorrow and finally remove the WIP tag.

I will also expand the filter button's tool tip. Basically the only thing that decides is the position within the parsed extensions list with higher up meaning preferred. So if your parsed extensions are for example `dng, nef, jpg` and the folder contains `A.dng, A.jpg, B.nef, B.dng`, then the filter would group that to `A.dng, B.dng`.

adamreichold:
@Beep6581 I imported the icons and extended the tool tip. If there are no objections, the code is ready to merge from my perspective.

Beep6581:
Compiles and works fine. Merged, thank you!

PR #2974, opened 2015-11-11
Change type of src to 'signed char' as it contains negative numbers.

innir:
Background is that 'char' is considered 'unsigned' on some architectures which causes a compilation error with C++11. See the [post](https://lists.debian.org/debian-arm/2015/11/msg00001.html) on the debian-arm mailing list for reference.


heckflosse:
PR builds and works fine. Thanks, Rinni :-)

PR #2987, opened 2015-11-24
Fix Sony 12 bit uncompressed bug

heckflosse:
Please test. I'll update dcraw.patch before merge.

Ingo


heckflosse:
I merged to get more tests

PR #2993, opened 2015-11-28
Fix some of the leaks found by starting with GCC's leak sanitizer

adamreichold:
Starting, opening a folder in file browser, then an image in editor and closing RT will produce a rather lengthy log when linking with `-fsanitize=leak` which will intercept calls to `malloc` and `free` as explained in the [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html#index-fsanitize_003dleak-684).

This PR tries to fix three of the reported leaks as they at least seemed to be the most straight-forward:
* The leak in the coordinate adjuster looks genuine.
* The leaked `previewHandler2` looks like an unused variable.
* The leak in `getDefaultTIFFTags` is probably just due to using a static instance of `std::vector<Tag*>` to do a manual return-value-optimization. But since the content of this collection is newly created for each call, the time to copy these 12 pointers is surely dwarfed by time to actually allocate those objects. Hence I opted to to replace the global state by passing the vector by value using named-return-value-optimization.

The cursors allocated in `CursorManager::init` are also reported leaked which could probably prevented by explicitly deleting them in a destructor or using reference-counted managed pointers. But this is probably just a question of style since this looks like explicit global state. The other leaks reported either seem to be rooted in external libraries or involve more complicated object life cycle issues.

### DIFF
--- a/rtexif/rtexif.cc
+++ b/rtexif/rtexif.cc
@@ -2783,17 +2783,11 @@ TagDirectory* ExifManager::parseTIFF (FILE* f, bool skipIgnored)
     return parse (f, 0, skipIgnored);
 }
 
-std::vector<Tag*> ExifManager::defTags;
-
-// forthis: the byte order will be taken from directory "forthis"
-const std::vector<Tag*>& ExifManager::getDefaultTIFFTags (TagDirectory* forthis)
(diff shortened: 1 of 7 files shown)


Beep6581:
It's great that you find and fix bugs :) @he

heckflosse:
@adamreichold Thanks for this fixes :) Merged into master

Issue #2744, opened 2015-08-11
Rendering intent, monitor profile & soft-proofing


Beep6581:
Originally reported on Google Code with ID 2763
```
Rendering intent depends on the particular image and need. If you need accurate color
reproduction, or if you work on a dull image, you need to use the "relative colorimetric"
rendering intent to preserve color accuracy. If you work on a photo of a deep red rose,
you want "perceptual" to not clip the reds. Therefore this option belongs in each PP3
and in the tool panel, not in Preferences.
```

Reported by `entertheyoni` on 2015-05-03 15:51:32

Beep6581:
Rendering intent has been removed, closing.

cipriancraciun:
Wasn't the "rendering intent" supposed to change how the image is **displayed**, based on the current selected monitor profile?

Thus indeed on some photos one would like to have color accuracy, and on some others he prefers to have finer shades, therefore moving it in the processing profile might be OK.

However removing it completely is a bad idea, for the simple reason that now, and I'm assuming that the relativ colorimetric intent is used by default with all, on my poor Lenovo laptop LCD it makes almost impossible to process any photos...

Please add this option back somewhere...

cipriancraciun:
The newer variant that I tested which doesn't permit anymore to choose the rendering intent, is `4.2.366`, and the older one that I previously used and which allows this choice is `4.2.274`.

As expected in `4.2.274`, by using the rendering intent of "perceptual" I can do my normal processing and the monitor displays colors as best it can.

However under `4.2.336`, all the colors that are out of the gamut of my monitor, are crushed to either black or some murky shade of color, which as said makes using Rawtherapee impossible.

As said earlier please revert this commit as it seriously breaks Rawtherapee.

Hombre57:
I second @cipriancraciun comment. Instead of removing it, we should have moved it in the ICM panel and make it work for booth preview image and output image. Rendering intent is a "must have" for this kind of software.

Beep6581:
@cipriancraciun please upload a sample raw file + color profile + PP3 http://filebin.net/

cipriancraciun:
Before you get to the uploaded files I want to give my input on how this rendering intent feature might be best suited for some workflows.

----

[...]

PR #2995, opened 2015-11-29
Add '-std=c++11' if libsigc++ depends on it

adamreichold:
Turns out my comment on #2744 was not correct and `pkg-config --cflags sigc++-2.0` does not include the necessary language support compiler flags. However the [libsigc++ project](http://libsigc.sourceforge.net/) itself states that version 2.5.1 or later definitely depends on C++11 support and it seems does not build without it, at least on a current [Arch Linux](https://projects.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/rawtherapee). Hence I suggest this change to enable the necessary flag if explicitly requested or if such a version of libsigc++ is detected.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -278,10 +278,10 @@ if (OPTION_OMP)
     endif (OPENMP_FOUND)
 endif (OPTION_OMP)
 
-if(USE_EXPERIMENTAL_LANG_VERSIONS)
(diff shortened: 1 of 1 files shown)


Beep6581:
Looks safe and works fine here with GCC-4.9.3, I'll merge.

PR #3006, opened 2015-12-05
Optionally enable link-time optimizations in the build system

adamreichold:
I do not know if any experience with [LTO](https://gcc.gnu.org/onlinedocs/gccint/LTO.html) does exist for the RT code base, but as any complex C++ application it should benefit significantly from LTO due to the importance of inlining. (It also allows one to write better encapsulated code, as one is not as pressed to expose implementation details in the headers for inlining purposes.)
Doing a debug build and running them program yields no immediate crashes, but highly optimized code like that of `rtengine` might show regressions after enabling this, so I put it behind a build-time flag. I also did a release build and developed a DNG using the neutral profile without obvious problems.
Of course, this also depends on a rather recent version of GCC (I tested using version 5.2.0) and it does add quite a bit of non-parallelizable build-time (on my machine, a full debug build goes from 6 to 7 minutes and full release build takes 12 minutes with more than half of that spent for finally linking the executable), but since compiler optimizations are almost free from a developer's perspective, it should be worth it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,6 +91,7 @@ if (APPLE)
 endif (APPLE)
 
 option(USE_EXPERIMENTAL_LANG_VERSIONS "Build RT with -std=c++0x" OFF)
+option(USE_EXPERIMENTAL_OPTIMIZATIONS "Build RT with -flto" OFF)
 option (BUILD_SHARED "Build rawtherapee with shared libraries" OFF)
 option (WITH_BZIP "Build with Bzip2 support" ON)
 option (WITH_MYFILE_MMAP "Build using memory mapped file" ON)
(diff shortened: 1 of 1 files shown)


heckflosse:
@adamreichold , can you please correct the typo in this line?

option(USE_EXPERIMTENAL_OPTIMIZATIONS "Build RT with -flto" OFF)

heckflosse:
I built with -flto on Linux using gcc 4.8.4. Builds and works fine.
On Windows using gcc 4.9.2 link fails http://pastebin.com/pL1zaFtA when -flto is used

adamreichold:
@heckflosse Thank you for pointing out the typo. Concerning the win64 build, searching for references yields only failed attempts to use LTO with MinGW. Since the linker implementation of PE is probably quite different from the ELF one, my guess that using this will be problematic for quite some time to come. I we want to follow this up, I'd say posting to `MinGW-users` would be our best bet?

adamreichold:
Closing this as it is made redundant by #3011.

PR #3007, opened 2015-12-05
[RFC] Use signals and slots instead of listeners

adamreichold:
RT currently uses various helper interfaces to provide coupling of event producers and their interested consumers. It also uses such helper interfaces to provide interface segregation for complicated components.
Since RT uses libsigc++ already, we can use its signals and slots mechanisms to reduce the number of utility types and decrease coupling among the components. This PR make such changes for the `DirBrowser` to open a discussion about whether this should be aimed for throughout the code base.
For the listener and event emission case, replacement by a signal is straight-forward yielding the same performance and type safety with less boiler plate and less coupling.
The case for the slot is probably less clear, since the amount of boiler plate does not decrease and the intention encoded in the explicit interface name is gone. But then again this is intentional since any compatibly typed functor can be used to provide the slot without usage of a predefined utility type.


Hombre57:
Could you explain what a "boiler plate" is? I don't get what it is.

We could of course reuse existing mechanism like signals with your comments in mind, but having utility classes has some advantages too :
1. you can make some methods mandatory. With signals (AFAIK), you can't force a client object to connect to some critical function, while with pure virtual function, you can't build if the method is not implemented (e.g. rtengine.h : BatchProcessingListener)
2. optimization at build time can't be done over signals (that's not critical though, but you're also talking about performance), AFAIK too
3. I haven't checked that, but sometimes Listeners utility class are used for querying the listener with the use of "parameter by reference". That's not pure Listener anymore in this case, and a better name could be used, but I don't think we were very cautious on that. I think that we used "Provider" utility class for such thing, but I can't say if it's properly done in the whole code. I don't know if that's possible with signals.
4. that's not done like that, but we could give a better name than "listener" to the parameter where objects can have only one listener, making the code easier to read.

You example is a good one where converting to signals makes sens, since it's only a one way direction and it handle more easily the "one to many" connections.

adamreichold:
> Could you explain what a "boiler plate" is? I don't get what it is.

For one thing the helper types themselves. Using many small interfaces is definitely a good thing to do, but one has to define a new class for every one. With the signal approach, it is as granular as it gets, each signal has an individual signature just by declaring it.
Also, the one-to-many relationship is implemented manually again and again, e.g.
````
for (size_t i = 0; i < dllisteners.size(); i++) {
    dllisteners[i]->dirSelected (dname);
}
````
instead of
````
dirSelected_ (dname, Glib::ustring());
````

> you can make some methods mandatory. With signals (AFAIK), you can't force a client object to connect to some critical function, while with pure virtual function, you can't build if the method is not implemented (e.g. rtengine.h : BatchProcessingListener)

Yes, signals cannot express everything that a specialized type can. But they also immediately provide other things like adaptors to glue together incompatible signals and slots. It also is possible to connect to many signals of the same type with different methods, whereas I can only have one method called that I inherit via an interface which I then have to multiplex over all callers (e.g. think of a `ButtonPressListener` implemented by a widget containing dozens of buttons).
I would argue that in the end it is all about the degree of coupling that one aims for: In typical GUI, the coupling should be minimal so that it is easy to experiment with different collaborations of the various UI elements and to combine them in different ways. In other places, the contracts should be more strict and an interface class is more appropriate than a signals-and-slots approach. For GUI implementations, it seems to be a sweet spot, e.g. Gtk and Qt use such an approach.

> optimization at build time can't be done over signals (that's not critical though, but you're also talking about performance), AFAIK too

The design of libsigc++ ensures that since version 2.0, using templates and type erasure, only a single virtual function call will be made when invoking a slot even in the presence of adaptors, i.e. while being much more flexible, the performance impact and compile-time optimization possibilities are the same as with an interface type with virtual methods. (They do have an overhead to create and manage the slots though, but that will not impact the UI latency.)

> I haven't checked that, but sometimes Listeners utility class are used for querying the listener with the use of "parameter by reference". That's not pure Listener anymore in this case, and a better name could be used, but I don't think we were very cautious on that. I think that we used "Provider" utility class for such thing, but I can't say if it's properly done in the whole code. I don't know if querying is possible with signals.

One could certainly emit a pointer or a reference holder like `std::ref` using a signal which the invoked slots could then manipulate to pass a value back to the emitting object.  `libsigc++` also provides more in the form of [marshallers](http://libsigc.sourceforge.net/libsigc2/docs/manual/html/ch04s03.html) where each invoked slot just normally returns a value and a helper object is used to suitably combine these result for the emitting object.

> that's not done like that, but we could give a better name than "listener" to the parameter where objects can have only one listener, making the code easier to read.

If we just mean "something that can be called like a function", then using a slot type like `sigc::slot` is for example a simple alternative that does not force the callee to be an instance of something. Also `std::function` and friends can basically do the same when using C++11.

Hombre57:
I can't test this patch in the best conditions since it have few important bugfix missing (done in other branches and already committed), but I think it's ready for commit (minus the trailing underscore in parameters as already discussed), and can certainly be generalized.

heckflosse:
@Hombre57 : I'll test it now

heckflosse:
Oups, it's for gtk3 branch. Can't test it on Win64

adamreichold:
@Hombre57 I rebased the branch onto the latest from `gtk3` and fixed the trailing underscores.

As for generalization, it is as you said: Sometimes a stricter interface, e.g. saying you need to handle of all or none of a set of events using pure virtuals, is more appropriate. However, for quickly wiring widgets together it should make things simpler. What could also be really helpful is if we enable C++11 on `gtk3` and therefore can write things like
````
some_signal.connect([this](SomeType some_arg)
{
    /* Do something simple to 'this' which does not even need a separate method... */
});
````

Beep6581:
Currently all new code is written for master using the requirements of master, and then merged into gtk3 with minimal trouble. Things will become hairy if rtengine starts to differ between the two branches, and then the burden of merging will fall on someone with good C++ knowledge.

adamreichold:
@Beep6581 These changes can certainly be used on `master` as they depend on `libsigc++` only, then developing on master and merging upwards will stay simple.

However, my understanding was that `master` is basically in 'maintenance mode' and all significant changes should be done on `gtk3` which more or less implies that the will diverge. But then, if this is the case, it might be more appropriate to develop all features on `gtk3` and only back port important fixes to `master` as appropriate.

Personally, I also think that allowing a certain amount of intrusive changes on `gtk3` would open up various possibilities to improve the code. For example, with switching to GCC 4.9 and C++11, trying out OpenMP 4.0 features could also allow significant but incompatible improvements.

So if you definitely want to keep the two branches compatible, I would rebase both this and the C++11/GCC 4.9 pull request to `master` and aim anything but Gtk3-specific changes to that branch?

heckflosse:
@adamreichold I tried your PR on Win64 gcc 4.9.2 and got this:
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:223 (pkg_check_modules)


-- checking for module 'glib-2.0>=2.44'
--   package 'glib-2.0>=2.44' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:224 (pkg_check_modules)


-- checking for module 'glibmm-2.4>=2.44'
--   package 'glibmm-2.4>=2.44' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:225 (pkg_check_modules)


-- checking for module 'gtkmm-3.0>=3.16'
--   package 'gtkmm-3.0>=3.16' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:226 (pkg_check_modules)


-- checking for module 'gio-2.0>=2.44'
--   package 'gio-2.0>=2.44' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:227 (pkg_check_modules)


-- checking for module 'giomm-2.4>=2.44'
--   package 'giomm-2.4>=2.44' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:228 (pkg_check_modules)


-- checking for module 'gthread-2.0>=2.44'
--   package 'gthread-2.0>=2.44' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:229 (pkg_check_modules)


-- checking for module 'gobject-2.0>=2.44'
--   package 'gobject-2.0>=2.44' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:230 (pkg_check_modules)


-- checking for module 'sigc++-2.0>=2.3.1'
--   package 'sigc++-2.0>=2.3.1' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:231 (pkg_check_modules)


-- checking for module 'lcms2>=2.6'
--   package 'lcms2>=2.6' not found
CMake Error at C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:279 (message):
  A required package was not found
Call Stack (most recent call first):
  C:/Program Files (x86)/CMake 2.8/share/cmake-2.8/Modules/FindPkgConfig.cmake:333 (_pkg_check_modules_internal)
  CMakeLists.txt:245 (pkg_check_modules)

That means the PR has to be reworked for this at least.

Ingo

Beep6581:
Yes I do agree, the problem is currently there are so few of us active, and I worry that creating this kind of overhead might drive us into a tight spot if even one of us goes into hibernation for a longer period. So far I had no trouble merging master into gtk3 and fixing any small resulting rtgui issues, but when it comes to rtengine I'm clueless.

Due to Gtk3 and GCC5's very limited adoption, I think we must keep supporting Gtk2 and GCC4 in master for a few months at least. But everyone here knows much more than I do about C++ and about any potential future issues which might result from making one branch use something in rtengine incompatible with the other. I wanted to bring that to your attention and let you proceed with that in mind.

adamreichold:
@heckflosse Are you sure this is related to this PR? The two commits do not touch the build system at all, so if you can build `gtk3` you should be able to build this PR.

From the error messages, it looks like CMake 2.8's `FindPkgConfig` module does not understand the `package>=version` syntax used by RT's build system and literally tries to find `package>=version` instead of `package` with at least version `version`. Do you have any other version of CMake on your path resp. system?

@Beep6581 Sorry, I am just not sure I understand you correctly: Agreeing means, that everything but Gtk3-specific stuff should go to `master`? (If so, I will try to rebase my PR on the weekend.)

heckflosse:
@adamreichold Hombre is the only one who can build gtk3 branch on Windows atm

Hombre57:
... which mean that we don't have to care about the CMake errors reported here.

Beep6581:
> that everything but Gtk3-specific stuff should go to master? 

Master needs to keep compatibility with whichever GCC version is needed to create 32-bit Windows builds. IIRC the 32-bit Windows builders use GCC-4.6 and as I don't use Windows I don't know whether such an old GCC is really the latest one available one to make 32-bit builds.

Since we're discussing it, would there be considerable benefits to supporting minimum GCC-4.9 in master? Then we could make a branch for 32-bit Windows builds which would only get backported bug fixes, and make use of whatever GCC-4.9 has to offer in both the master and gtk3 branches. Team, what do you think?

Adam we unanimously decided to accept you to our team, and I don't recall whether I officially welcomed you... If not,  welcome! :)

heckflosse:
@adamreichold @Hombre57 gtk3 branch is not ready for production atm (correct me if I'm wrong). As long as I'm not able (because of lacking compile instructions for rt gtk3 on Win64) to use gtk3 on Win64 I'll make changes for master (gtk2) build. If the changes can be merged into gtk3 build that's fine too.
But not vice versa!

Adam, I already mailed you but here again: Welcome to the team :-)

Ingo

Hombre57:
Welcome aboard Adam!

@heckflosse Gtk3 branch has 2 serious GUI bug still pending (AFAIK), but Adam might find a solution way faster than me. Other than that and theme tuning, it's not far away to be ready for production. Replacing all deprecated functions will take more time.

I'll update the compile text for building the Gtk3 branch on Windows shortly.

@Beep6581 It'd mean that Win32 support is stopped (no new feature on this version, only bugfix) ?

Beep6581:
@ho

adamreichold:
Thank you all for the warm welcome! I hope I will be able to make myself useful here for the foreseeable future...

Concerning the tool chain and language version issues:
* Staying on C++98 and GCC 4.6 is perfectly fine and I don't any see any user-visible feature depending on a newer tool chain. But it is also pretty boring and free software is among other things about having fun with technology, isn't it?
* C++11 and GCC 4.9 seem reasonable, even for `master` as they should be available almost everywhere. And they do open various interesting possibilities, foremost for RT is probably the newer OpenMP version since that technology is heavily used in the engine and version 4 would probably make it simpler to write those highly parallel processing kernels, possibly replacing intrinsics in all but the most performance-critical code paths. Using this to make the core of RT more accessible alone would be worth the change IMHO.
* If `win32` is really stuck on GCC 4.6, than a separate branch to support that platform with bug fixes only sounds sensible to me. But is it really not possible to have a 32-bit MinGW GCC 4.9? (I really do not know since I seldomly work with Windows machines.)
* But then again, if `gtk3` is supposed to become the new `master` at some point, we should make that work accessible to all team members ASAP anyway.

So in summary, I would suggest we increase the baseline to GCC 4.9 for both `master` and `gtk3` to keep code flow between the branches simple, even if that means `win32` going into maintenance mode. And independently of that, we should try to make `gtk3` available to everyone of us and ready for production use.

P.S.: As said above, these changes here are perfectly fine with using C++98 and GCC 4.6 as `libsigc++` provides almost what the C++11 standard library does in terms of dynamic binding.
P.P.S.: Which are the two major issues with the `gtk3` branch?
P.P.P.S.: Any ideas on a simple way to get `win32` and `win64` virtual machines up and running? I would like to try to help with build issues on these platforms, but I do not have them available at all ATM.

adamreichold:
@heckflosse I grabbed a fresh IE11/Win7 32-bit VM from [Modern IE](https://dev.windows.com/en-us/microsoft-edge/tools/vms/linux/) and the gtkmm packages from [MSYS2](https://wiki.gnome.org/Projects/gtkmm/MSWindows) to begin building the `gtk3` branch on `win32` using GCC 5.2. So far it seems to be relatively straight forward as only `libiptcdata` needs to be built from source (and I can even use the same package manager as on my Linux box ;-))

The only thing I needed to do differently when building RT itself was adding `-G "Unix Makefiles"` and `-DCMAKE_MAKE_PROGRAM=mingw32-make` to the CMake command line (and of course disable some sub directories of the `libiptcdata` build). I have only done a debug build so far, but developing a sample file using the neutral profile seems to work fine.

Beep6581:
Once 3028 is merged, do we merge this too?

adamreichold:
@Beep6581 Due to the discussion w.r.t. keeping code flow simple between `master` and `gtk3`, I would rather rebase this to `master` after the simpler stuff like cleaning up `std::auto_ptr` is done.

adamreichold:
Closing this unmerged since I need to rebase it on `master`...

PR #3009, opened 2015-12-05
Add selection of monitor profile and intent

adamreichold:
It is not yet a full resolution of issue #2744 but it is hopefully a part of it and in a state that can be reviewed and merged. (Still missing is the addition of output profile intent parameter for the ICM tab and the ICM processing parameters. And any of the more advanced ideas like better auto-detection of monitor profiles or available intents per-profile.) And there are some issues on which I'd like to ask for advice:
* I am updating the processing parameters with the defaults from the settings immediately after opening an image in the editor by calling `MonitorProfileSelector::reset` from `EditorPanel::open`. Is there are better place to do this once before the processing parameters are created?
* I am abusing the `rtengine::EvOProfile` processing parameter event for the change of the monitor profile. Should I create a more specific event here?
* The combo boxes are pretty large for the bottom tool bar and the relief seems unnecessary but I have no clear idea how to properly adjust the style properties of the widget to change that. Any advice on how this should be done?

The copy-editing of the ICC store is of course not strictly necessary, but since I had to add at least one method, I tried to improve the quality of the surrounding code by adding constness where applicable and made use of early returns instead of arrow programming.


adamreichold:
By the way, we should probably have a look at [colord](http://www.freedesktop.org/software/colord/intro.html) for automatic detection of (monitor) profiles on Linux desktops...

rtengine/improcfun.cc
```diff
@@ -209,21 +209,23 @@ void ImProcFunctions::firstAnalysis (Imagefloat* original, const ProcParams* par
...
 #if !defined(__APPLE__) // No support for monitor profiles on OS X, all data is sRGB
-    Glib::ustring monitorProfile = settings->monitorProfile;
+
 #if defined(WIN32)
 
-    if (settings->autoMonitorProfile) {
-        monitorProfile = iccStore->defaultMonitorProfile;
-    }
+    cmsHPROFILE monitor = settings->autoMonitorProfile
+            ? iccStore->getDefaultMonitorProfile ()
```
Hombre57: This return a Glib::ustring, not a cmsHPROFILE so the build fail here.

Hombre57:
> adamreichold wants to merge 3 commits into Beep6581:master from adamreichold:add-transient-monitor-profile-and-intent

If I understand correctly, this indicate that your branch is related to master, which use Gtk2. I'm then surprised to see that your code is using the "append" method for adding entries to Combobox, which is a method used by gtk3 (AFAIK). That's why it doesn't build here with my Gtk2 setup. Do you have both installed ?

adamreichold:
@Hombre57 I build this with Gtk 2.24.28 and Gtkmm 2.24.4 which provides the `Gtk::ComboBoxText::append` and deprecates `Gtk::ComboBoxText::append_text`. But I also do not know when this method was introduced, so should I replace all `append` by `append_text` so that we are safe w.r.t. to the Gtk version?

Hombre57:
[Bugfix-DefaultMonitorProfile.zip](https://github.com/Beep6581/RawTherapee/files/53012/Bugfix-DefaultMonitorProfile.zip)

This will fix the DefaultMonitorProfile bug, but it's untested because I still can't build this branch.

Hombre57:
> so should I replace all append by append_text so that we are safe w.r.t. to the Gtk version?

Yes please. I converted ```append_text``` to ```append``` when porting to Gtk3. For sure, I don't have this method on GTK2 for Windows (maybe a previous revision).

Hombre57:
I'll answer you initial post's question tomorrow, now it's time for a sleep ;) Bye.

adamreichold:
Rebased with hopefully all `append` calls removed and I added your suggestion to fix and simplify querying the default monitor profile.

Beep6581:
I tested all color management systems in Linux this year - colord, oyranos, whatever else there was. All buggy, none worked with Plasma5, some were dead.
The _ICC_PROFILE X11 atom should be set by any correctly-working CMS, you could check that to pre-select a monitor profile in RT.

Thank you for doing this Adam, I will test today/tomorrow.

Beep6581:
Monitor ICC selection works great!
Intent shows 4 options, but only two are of interest to photographers: perceptual and relative colorimetric.
Adam do you think its possible to check the monitor profile for which intents it supports? If there is no existing function in RT to do it, I could ask LCMS and ArgyllCMS devs whether they have such a thing. Then we could have the intents not supported by the monitor profile grayed-out from the combobox.

adamreichold:
I rebased the branch with following changes:
* Since we only offen perceptive and relative colorimetric, I replaced the intent combo box with a simpler check button just to enable perceptive with relative colorimetric still being the default.
* I tried to make use of the `cmsIsIntentSupported` function from LCMS to check if those intents are actually supported for proofing, but for now it looks like a the ICC profiles in my system folders support all four standard intents. @Beep6581 Does this work as intended for you?

adamreichold:
Rebased again, back to using a combo box for the two available rendering intents. If decided upon, I however think that conversion to a pop-up button should wait until the above issues are resolved and the necessary images are created.

Hombre57:
Would you mind adding this little patch to this branch, it's for the future use of the PopUpButton, I don't want to loose the small changes I made today.

[menuSensitivity.zip](https://github.com/Beep6581/RawTherapee/files/54704/menuSensitivity.zip)

adamreichold:
@Hombre57 Added the patch as a commit in your name. If you use `git format-patch` you can also export the commit message itself, so others can fully replicate your commits from the textual form.

Hombre57:
@adamreichold I have to look out how to do that through Eclipse. I'm not using the command line for Git.

Beep6581:
@adamreichold I merged add-transient-monitor-profile-and-intent from git://github.com/adamreichold/RawTherapee.git but it used two comboboxes while your comment from yesterday states that the latest version uses a checkbox - how can I test the latest version?

adamreichold:
@Beep6581 Because of @Hombre57's comments on issue #2744, I removed the check box again and revert to a combo box, but still only showing the two relevant intents and checking if they are both supported (and selecting the only supported one otherwise).

So I guess you already did test the latest version? In any case, I would recommend just checking the branch out, e.g. either by adding my fork as a separate remote or by just cloning it separately. It seems simpler than merging, especially since I rebased a lot to keep the commit history clean.

Beep6581:
For anyone who wants to test, here is a sample monitor profile I just made for my laptop, it's an XYZ LUT + matrix with gamut mapping for perceptual intent using the perceptual method. Be sure to load the calibration curves (included in the file) as well!
I included a 3D comparison of my profile to sRGB, it only has about 75% sRGB coverage.
http://filebin.net/lhfoe57dl7

Using relative colorimetric intent for RT's preview makes me think saturated colors are clipped
http://i.imgur.com/3wkbDIu.png
but now I can enable perceptual intent and see that it is not so
http://i.imgur.com/56AMtKj.png
(of course the screenshots may look completely wrong on your end)

Beep6581:
The comboboxes need not be wide, since we only need to see the full names when clicking on them. In fact there must be a max width set, or this happens...
http://i.imgur.com/vSC7qNC.jpg

Hombre57:
The intent combo will be replacer by a PopUpButton, which is very small, but I'll do it when the patch will be committed. For the profile combo, yes it needs a max-width.

Hombre57:
@adamreichold I think there must be something wrong when rebasing your branch : the monitor profile is stored in the ProcParams again. I thought that you moved it in options::rtsettings where it is more appropriated, as explained by Beep6581 in issue #2744 .

As for the ```rtengine::EvOProfile``` event, I'd create a new one, because this one throw the M_INIT event, which in ```Crop::update``` does quite a lot of reprocessing. I don't even know if this is legitimate : I know that some tools introduce gamut checking, but the Working Profile should be used in this case, not the output profile. Maybe Jacques could tell us why OUTPUTPROFILE asks for M_INIT ? However I would not make the _processing_ dependent of the monitor profile. So a dedicated event used by a 

On a side note, I think there's something wrong with the OUTPUTPROFILE event (in refreshmap.cc), because to reprocess the initial steps, but not all the following steps ! As you can see in refreshmap.h, each bitfield event start the processing from a point in the pipeline until the end of it. That's not the case for OUTPUTPROFILE and it's an error. OUTPUTPROFILE should be deleted and ALLNORAW should be used instead (beginning the processing from M_INIT... if we really have to start from M_INIT) as well as INPUTPROFILE and WHITEBALANCE which are strictly equal to ALLNORAW.

I'm making a patch with all this changes and more.

adamreichold:
> The intent combo will be replaced by a PopUpButton, which is very small, but I'll do it when the patch will be committed. For the profile combo, yes it needs a max-width.

How do I set the maximum width of the widget? I found border width and wrap width, but they both seem to be unrelated.

> @adamreichold I think there must be something wrong when rebasing your branch : the monitor profile is stored in the ProcParams again. I thought that you moved it in options::rtsettings where it is more appropriate, as explained by Beep6581 in issue #2744 .

The default monitor profile and intent are in `options.rtSettings`, but the currently chosen values are passed via `ProcParams`, but they are not stored on disk. This seems to be necessary since I need to communicate changes of these values to the processing pipelines using parameter change events, hence the use of transient parameters.

> As for the rtengine::EvOProfile event, I'd create a new one, because this one throw the M_INIT event, which in Crop::update does quite a lot of reprocessing. I don't even know if this is legitimate : I know that some tools introduce gamut checking, but the Working Profile should be used in this case, not the output profile. Maybe @Desmis could tell us why OUTPUTPROFILE asks for M_INIT ? However I would not make the processing dependent of the monitor profile. So a dedicated event used by a 

Creating a new event should not be a problem, but choosing or creating the right refresh map is currently beyond my knowledge of how the pipeline is organized. Would just triggering `M_MINUPDATE` be sufficient?

Hombre57:
I'm making a big patch for this branch. Please wait...

Hombre57:
I'm almost done with the patch. It tries to introduce soft-proofing. In fact, it's already there and active by default, if I've understood the concept correctly, in ```ImProcFunctions::lab2monitorRgb```... See also ```ImProcCoordinator::updatePreviewImage``` (look for ```ipf.lab2monitorRgb```).

Now we can deactivate it. I've also added a rendering intent button for the output profile in the ICM tool. It was set to "Relative colorimetic" before.

I hope all this makes sens.

Beep6581:
Cool!
I want to upload a new build to coverity (using gtk-3.16 for now as it would make too much of a mess on my system to install 3.18 yet), I'll wait for this to be merged before I do.

Hombre57:
I think that I'm almost done. I still have to create a dedicated event for the output rendering intent and tweaking the GUI a little bite + testing, but you can already test it. The "system default" on Windows does not work yet however, but it should be easy to fix.

On a side note, @adamreichold , I've refactored your code a little bite to make it more rooky-proof (i.e; for me). I must admit that your C++ coding technique is a bit too high level for me, sorry.

[Adding-softproofing.zip](https://github.com/Beep6581/RawTherapee/files/61903/Adding-softproofing.zip)

Beep6581:
Could you put this on a branch for easier testing?

Hombre57:
It's kind of hijacking Adam's patch, but your words are orders :). See the new "add-monitor-profile-and-softproofing" branch.

adamreichold:
Frankly speaking, I am rather unhappy about how this PR is handled: I initially listed the open issues for the code and asked for help on how to resolve them. Instead of answering these questions, so that I could improve the code and my understanding of the RT code base, the PR was basically hijacked. Since the code is also "refactored" instead of critiqued, I think this PR should be closed and a new one with @Hombre57's last code drop opened.

For what it is worth, I did have a look at the patch and have some questions resp. suggestions:
* While it is awkward to have only some processing parameters not stored persistently, using the processing parameter events to trigger updates but not the processing parameters to store the parameter values seems also problematic. For example, there are now two places where this is tracked: `isColorProfileDirty` and the `todo` mask. More irritating is probably, that we now have a call to `endUpdateParams` without a matching call to `beginUpdateParams`. 
* Using an enumeration everywhere instead of unadorned numbers is definitely an improvement, but I don't see why LCMS' enumeration should be duplicated. (We do pass around bare LCMS profile handles after all.) Also typedef'ing enum is not necessary in C++ and will only pollute the namespace with anonymous enums.
* Replacing `std::auto_ptr` with a bare pointer for the monitor profile selector, i.e. going back to manual memory management, is the wrong direction IMHO. Also replacing the plain members for e.g. the combo box widgets with pointers and the use of `Gtk::manage` seems unnecessary to me and is probably significantly less efficient. Also setting the pointer to the image processor repeatedly instead of just referencing it seems unnecessary.
* The method `getDefaultMonitorProfileStr` is probably better named `getDefaultMonitorProfileName` to indicate what is does, not how it does it, i.e. using a string instead of a profile handle.
* Manual handling of event propagation by blocking and unblocking connections seems like something one would want to avoid as dependent changes should usually stop by themselves in an event-driven system as the system reaches a fixed state and all calls to setters are just no-ops.
* It seems inadvisable to have all these unrelated changes in a single commit instead of logically grouped changes. Also keeping commented-out code around for future reference should not be necessary using version control.

Hombre57:
Damn, I don't know how to delete a branch. Can someone delete my last new branch please ? As usual, I have to refrain myself from doing things myself.

Hombre57:
And don't worry, I won't hijack any of your patch from now on.

Hombre57:
1. yes, isColorProfileDirty may certainly be suppressed.
2. this is to strong typing (?) that value, i.e. let my IDE propose me the values instead of using a generic int
3. that's the kind of C++ trick that I didn't understood. In fact, we're used to make sub-GUI object by making them "Gtk::HBox" that you can simply add to the GUI and that will be deleted by Gtk. Your approach is totally new to this project, but of course totally valid.
4. ok for this
5. that's a possibility, but the event chain is too complex for me and basically, all tools does just that when it has to update its parameters. You'll argue that it's not an excuse :)
6. softproofing is connected to monitor profile and intent handle and was a way overdue functionality, that's why I did that part of the code.

This project deserve to have more "professional" contributor (you seem to be a pro dev, even though google show me you're a physicist, or you have an homonym), and not rooky like me, so there's no problem if you drop my patch, and I'll take care of making things more carefully in the future,

PR #3011, opened 2015-12-05
Mandate C++11 support and GCC version 4.9 or higher on the Gtk3 branch

adamreichold:
This takes up @Hombre57's changes replacing `std::auto_ptr` by `std::unique_ptr` on the `gtk3` branch and adds mandating C++14 support via the build system. It also checks but for now only warns if the GCC version used is lower than 5.0. It also removes all references to `GLIBMM_EXCEPTIONS_ENABLED` as Glib version 3 does not seem to support to disabling exceptions anymore and will unconditionally define this. This actually removes most of our usage of `std::auto_ptr`. The other points of use are extended to make use of `std::make_unique`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,10 +1,16 @@
-if (WIN32)
-  cmake_minimum_required(VERSION 2.8.4)
-else (WIN32)
-  cmake_minimum_required(VERSION 2.6)
-endif (WIN32)
+cmake_minimum_required(VERSION 3.1)
 
 PROJECT(RawTherapee)
+
+# set required C and C++ standards and check GCC version
+SET (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu11")
+SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
+    message(WARNING "RawTherapee should be built using GCC version 4.9 or higher!")
+  endif()
+endif()
 
 # the default target is 'Debug'
 if (CMAKE_BUILD_TYPE STREQUAL "")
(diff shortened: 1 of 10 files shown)


adamreichold:
Of course, C++14 and GCC 5 are pretty hefty requirements, for example the just released OpenSUSE Leap 42.1 does still use GCC 4.9 by default. So maybe C++11 and GCC 4.9 would be a better compromise for us? What do the other developers think? Ingo, what is your opinion on this?

heckflosse:
We have to solve at least one issue before: Currently Win32 builds are made with gcc 4.5.x or 4.6.x !!!
gcc 4.9.x works fine for Win64 builds, but gcc 5.x is not tested for win64. I also remember some issues with gcc 5.x compilations reported by marcin because the dependencies also have to be build using gcc 5.x or you have to use a special link flag (Hombre may remember it, we talked about that in IRC).

In my opinion supporting gcc 5.x is good, but demanding gcc 5.x to build RT is not an option currently.

Ingo

adamreichold:
I'll rebase the branch to C++11 and GCC 4.9 then and drop the last commit...

adamreichold:
The flag you refer to is probably the change of C++ ABI when the final bits and pieces of the standard library for C++11 were finished? If RT is built against the new ABI using a new compiler, the dependencies need to conform to that as well. With GCC 4.9 we would still default to the old ABI, but builders needs to ensure consistency in any case, whether with the new or the old ABI.

heckflosse:
Yes, it's the C++ ABI flag I refered to

Hombre57:
For the record, I'm building RT/Gtk2 and Gtk3 on Win7/64 with gcc5.2. It seems to work fine as of now, but can't tell for a Win/32 version.

Beep6581:
I would appreciate if you could use C++11 and GCC-4.9 for a few months as GCC5 is still not an option for some.

adamreichold:
The [branch](https://github.com/adamreichold/RawTherapee/tree/cpp11-example-code) could be an example of the more or less mechanical changes possible using C++11 within the GUI code if we decide to.
Of course, doing such copy-editing is usually not worth it, but it might show how newly written code using STL container could look like. Also I am not a huge fan of iostreams, but it still looks better than C-style IO IMHO, mostly due to automatic memory management. Also note that these are just localized technical changes, not design changes aimed at fully utilizing the new language features.

Beep6581:
Hello, what is the status here? @adamreichold , @heckflosse ?

adamreichold:
@Beep6581 Since a reasonable compromise that considers building under Windows was found, it can be merged IMHO. The separate branch that shows an example of applying C++11 features is basically just clean-up and can either be merged as well or I could open a separate PR so that it can be reviewed independently.

adamreichold:
This will need to be reworked if #3028 is merged into `master` and `master` into `gtk3`...

gaaned92:
gtk3+PR3011 compiles and builds on WXP. 
Unfortunately, libgdk-3-0.dll calls dwmapi.dll. DWM made a first apparition on Vista and thus is absent from XP.

PR #3028, opened 2015-12-19
Always enable C++11 support and warn if GCC older than 4.9 is used

adamreichold:
As discussed in #3007 and IRC, we'll probably (that is if everyone agrees and MSYS2 works out for the Windows developers) aim for GCC 4.9 and C++11 even on `master` to keep code flow between that branch and `gtk3` straight forward.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,14 @@ endif ()
 
 string (TOUPPER ${CMAKE_BUILD_TYPE} UPPER_CMAKE_BUILD_TYPE)
 
+# Set required C and C++ standards and check GCC version
+SET (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu11")
+SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++11")
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4.9")
+  message(WARNING "RawTherapee should be built using GCC version 4.9 or higher!")
+endif()
+
 # We might want to build using the old C++ ABI, even when using a new GCC version
 if (USE_OLD_CXX_ABI)
   add_definitions (-D_GLIBCXX_USE_CXX11_ABI=0)
@@ -89,6 +97,7 @@ option(USE_EXPERIMENTAL_LANG_VERSIONS "Build RT with -std=c++0x" OFF)
 option (BUILD_SHARED "Build rawtherapee with shared libraries" OFF)
 option (WITH_BZIP "Build with Bzip2 support" ON)
 option (WITH_MYFILE_MMAP "Build using memory mapped file" ON)
+option (WITH_LTO "Build with link-time optimizations" OFF)
 option (OPTION_OMP "Build with OpenMP support" ON)
 option (PROTECT_VECTORS "Protect critical vectors by custom R/W Mutex, recommanded even if your std::vector is thread safe" ON)
 option (STRICT_MUTEX "True (recommended): MyMutex will behave like POSIX Mutex; False: MyMutex will behave like POSIX RecMutex; Note: forced to ON for Debug builds" ON)
@@ -266,17 +275,18 @@ if (WITH_MYFILE_MMAP)
 	add_definitions (-DMYFILE_MMAP)
 endif (WITH_MYFILE_MMAP)
 
+if (WITH_LTO)
+    SET (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -flto")
+    SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -flto")
+    SET (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -flto")
+endif (WITH_LTO)
+
 if (OPTION_OMP)
     find_package(OpenMP)
     if (OPENMP_FOUND)
         set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS} -Werror=unknown-pragmas")
     endif (OPENMP_FOUND)
 endif (OPTION_OMP)
-
-if(USE_EXPERIMENTAL_LANG_VERSIONS OR NOT (SIGC_VERSION VERSION_LESS 2.5.1))
-	SET (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu1x")
-	SET (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=gnu++0x")
-endif ()
 
 # find out whether we are building out of source
 get_filename_component(ABS_SOURCE_DIR "${PROJECT_SOURCE_DIR}" ABSOLUTE)


Beep6581:
+1

Hombre57:
Thanks to your comment on IRC, I installed the gtkmm2 MSYS2 package and am able now to build for both gtkmm2 & 3 on Windows/64 with MSYS2 + gcc 5.2, minus the clearlooks problem (missing). Can't say for building a Win/32 version though... If someone knows which flags to use, I can try.

adamreichold:
@Hombre57 You could try out the new Windows build instructions on RawPedia. (It also describes how to build Clearlooks. We should probably also how to install gdb there?) Should theoretically be as simple as running 'MinGW-w64 Win32 Shell' and following the 32-bit instructions. But I would be glad if someone would verify and improve that documentation independently in any case.

Beep6581:
I will verify and update the instructions if needed within the next few days.

Hombre57:
@adamreichold 
> It also describes how to build Clearlooks

The following command fail : 
```
./configure --prefix=/mingw32 --disable-all --enable-clearlooks
checking for a BSD-compatible install... /usr/bin/install -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /usr/bin/mkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether to enable maintainer-specific portions of Makefiles... no
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.exe
checking for suffix of executables... .exe
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking for style of include used by make... GNU
checking dependency style of gcc... gcc3
checking whether make sets $(MAKE)... (cached) yes
checking whether gcc and cc understand -c and -o together... yes
checking build system type... ./config.guess: unable to guess system type

This script, last modified 2009-12-30, has failed to recognize
the operating system you are using. It is advised that you
download the most up to date version of the config scripts from

  http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD
and
  http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.sub;hb=HEAD

If the version you run (./config.guess) is already up to date, please
send the following data and any information you think might be
pertinent to <config-patches@gnu.org> in order to provide the needed
information to handle your system.

config.guess timestamp = 2009-12-30

uname -m = x86_64
uname -r = 2.4.0(0.292/5/3)
uname -s = MSYS_NT-6.1
uname -v = 2015-11-23 08:33

/usr/bin/uname -p = unknown
/bin/uname -X     =

hostinfo               =
/bin/universe          =
/usr/bin/arch -k       =
/bin/arch              = x86_64
/usr/bin/oslevel       =
/usr/convex/getsysinfo =

UNAME_MACHINE = x86_64
UNAME_RELEASE = 2.4.0(0.292/5/3)
UNAME_SYSTEM  = MSYS_NT-6.1
UNAME_VERSION = 2015-11-23 08:33
configure: error: cannot guess build type; you must specify one
```

adamreichold:
@Hombre57 Could it be you are running this from a 64-bit console? If so, the prefix should probably be `/mingw64` instead of `/mingw32`. I will fix this in the RawPedia article...

Hombre57:
> Could it be you are running this from a 64-bit console?

How can I know ? The icon that start MSYS2 doesn't say anything. This is the command from the icon : 
`C:\Windows\System32\cmd.exe /A /Q /K D:\Jean-Christophe\Developpement\Toolchain\msys64\msys2_shell.bat`

`msys2_shell.bat` contain :
```
:
@echo off

rem ember value of GOTO: is used to know recursion has happened.
if "%1" == "GOTO:" goto %2

if NOT "x%WD%" == "x" set WD=

rem ember command.com only uses the first eight characters of the label.
goto _WindowsNT

start /min %COMSPEC% /e:4096 /c %0 GOTO: _Resume %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
goto EOF

rem ember that we execute here if we recursed.
:_Resume
for %%F in (1 2 3) do shift
if NOT EXIST %WD%msys-2.0.dll set WD=.\usr\bin\

rem ember that we get here even in command.com.
:_WindowsNT

if NOT EXIST %WD%msys-2.0.dll set WD=%~dp0usr\bin\

set MSYSTEM=MSYS
rem To activate windows native symlinks uncomment next line
rem set MSYS=winsymlinks:nativestrict
rem Set debugging program for errors
rem set MSYS=error_start:%WD%../../mingw32/bin/qtcreator.exe^|-debug^|^<process-id^>
set MSYSCON=mintty.exe
if "x%1" == "x-consolez" set MSYSCON=console.exe
if "x%1" == "x-mintty" set MSYSCON=mintty.exe

if "x%MSYSCON%" == "xmintty.exe" goto startmintty
if "x%MSYSCON%" == "xconsole.exe" goto startconsolez

:startmintty
if NOT EXIST %WD%mintty.exe goto startsh
start %WD%mintty -i /msys2.ico /usr/bin/bash --login %*
exit

:startconsolez
cd %WD%..\lib\ConsoleZ
start console -t "MSys2" -r "%*"
exit

:startsh
start %WD%sh --login -i %*
exit

:EOF
```

Anyway, using **/mingw64** doesn't work either, even with updated `config.guess` and `config.sub` files (I have automake-1.15 installed).

adamreichold:
This sounds like you are using the MSYS shell instead of one of the MinGW-w64 shells? After installing MSYS2, you should have three shortcuts in the MSYS2 start menu folder. Two of those are for MinGW-w64, 32 and 64 bits. To make a 64 bit build, you should use the _MinGW-w64 Win64 Shell_ shortcut.

Hombre57:
Ah, ok, sorry for the confusion :-S

sguyader:
On my Windows 7 64 bits, the update of the msys2 packages fails because of
a wrong path set in pacman (//*usr*/share/libalpm/hooks/ : access denied).
I tried an earlier msys2 package (from may 2015) but it was the same.
I found an msys2 install used for building gimp, I'll try this one.

Could it be you are running this from a 64-bit console?

How can I know ? The icon that start MSYS2 doesn't say anything. This is
the command from the icon :
C:\Windows\System32\cmd.exe /A /Q /K
D:\Jean-Christophe\Developpement\Toolchain\msys64\msys2_shell.bat

msys2_shell.bat contain :

:
@echo off

rem ember value of GOTO: is used to know recursion has happened.
if "%1" == "GOTO:" goto %2

if NOT "x%WD%" == "x" set WD=

rem ember command.com only uses the first eight characters of the label.
goto _WindowsNT

start /min %COMSPEC% /e:4096 /c %0 GOTO: _Resume %0 %1 %2 %3 %4 %5 %6 %7 %8 %9
goto EOF

rem ember that we execute here if we recursed.
:_Resume
for %%F in (1 2 3) do shift
if NOT EXIST %WD%msys-2.0.dll set WD=.\usr\bin\

rem ember that we get here even in command.com.
:_WindowsNT

if NOT EXIST %WD%msys-2.0.dll set WD=%~dp0usr\bin\

set MSYSTEM=MSYS
rem To activate windows native symlinks uncomment next line
rem set MSYS=winsymlinks:nativestrict
rem Set debugging program for errors
rem set MSYS=error_start:%WD%../../mingw32/bin/qtcreator.exe^|-debug^|^<process-id^>
set MSYSCON=mintty.exe
if "x%1" == "x-consolez" set MSYSCON=console.exe
if "x%1" == "x-mintty" set MSYSCON=mintty.exe

if "x%MSYSCON%" == "xmintty.exe" goto startmintty
if "x%MSYSCON%" == "xconsole.exe" goto startconsolez

:startmintty
if NOT EXIST %WD%mintty.exe goto startsh
start %WD%mintty -i /msys2.ico /usr/bin/bash --login %*
exit

:startconsolez
cd %WD%..\lib\ConsoleZ
start console -t "MSys2" -r "%*"
exit

:startsh
start %WD%sh --login -i %*
exit

:EOF

Anyway, using */mingw64* doesn't work either, even with updated config.guess
and config.sub files (I have automake-1.15 installed).

—
Reply to this email directly or view it on GitHub
<https://github.com/Beep6581/RawTherapee/pull/3028#issuecomment-166153301>.

gaaned92:
@adamreichold 
on WXP : MSYS2 installation ok. the gdb and diff packages are missing and should be added in the documentation. Very simple.
when building, I get warnings see :  http://www.rawtherapee.com/forum/viewtopic.php?f=10&t=6284#p44570 

Could somebody provide a clean modification in multilangmgr.cc  as the GetUserDefaultLocaleName is missing on wxp

The 536 release build crashes but it is the same with GCC 4.6.1.  
Debug builds run ok.
Relwithdebinfo build crashes when opening an image see BT in above link.

adamreichold:
@gaaned92 Thank you for trying this out! I added `diffutils` and `gdb` to the instructions. I'll also try to get a Windows XP VM to try to diagnose the crash.

As for crashing being the same with GCC 4.6.1, do you mean the current code does crash with a previously working tool chain for?

gaaned92:
@adamreichold : crashing , yes I have problems with old tool chain. I am trying to understand what is the cause. I went back to 7ef35 (4.2.510) and was able to build again a working release build with old tool chain. Now trying to bisect but it is very very slow on my PC. Hope that Santa Claus will sometime bring a new one.

There is also a problem with multilanmgr.cc on XP as the getuserLocalename entry doen't exist. A correction would be welcome

adamreichold:
@gaaned92 I will look into the locale issue when the Windows XP VM is ready. Did you trying building the old version using the new tool chain? (Just to make sure this is a bug introduced in the RT code only, and not a mixture of RT bug and compiler bug.) (You will probably need to remove `add_definitions(-D_GLIBCXX_USE_CXX1_ABI=0)` from `CMakeLists.txt` if that is already present in that version.)

adamreichold:
@gaaned92 I did a clean `relwithdebinfo` build and on pristine Windows XP VM (IE8 on XP from Modern IE) and it seems to work fine here. Opened a RAW file and developed it to JPEG using the neutral profile. (I have been using master, will retry with this PR next.)

Also, I tried to patch the user locale query, i.e. could you try the patch
````
diff --git a/rtgui/multilangmgr.cc b/rtgui/multilangmgr.cc
index 37dcb5e..0d42666 100644
--- a/rtgui/multilangmgr.cc
+++ b/rtgui/multilangmgr.cc
@@ -148,14 +148,12 @@ Glib::ustring MultiLangMgr::getOSUserLanguage()
 #ifdef WIN32
 // When using old versions of MINGW this is not defined
 #ifdef __MINGW64_VERSION_MAJOR
-        WCHAR langRFCU[64] = {0};

-        if (GetUserDefaultLocaleName(langRFCU, 64) != 0 && lstrlenW(langRFCU) >= 2) {
-            // convert UNICODE16 to GTK
-            char langRFC[64];
-            WideCharToMultiByte(CP_UTF8, 0, langRFCU, -1, langRFC, 64, 0, 0);
-            Glib::ustring localRFC = Glib::ustring(langRFC);
+        LCID langID = GetUserDefaultLCID();
+        TCHAR langRFC[64] = {0};

+        if (GetLocaleInfo(langID, LOCALE_SNAME, langRFC, 64) != 0 && lstrlen(langRFC) >= 2) {
+            Glib::ustring localRFC = Glib::ustring(langRFC);
             langName = TranslateRFC2Language(localRFC);
         }
````
which seems to work for me. (I am not proficient with the `win32` API we probably can't push this into the repository as it is, as it does not work with custom locales according to the MSDN documentation. But maybe we can add some build flag to build for releases prior to Windows Vista using this code?)

**Comment by adamreichold** [comment]
@gaaned92 Using this PR also seems to work fine using MSYS2 on 32-bit Windows XP. My guess is that the difference is that my VM has only one (virtual) CPU?

**Comment by adamreichold** [comment]
@gaaned92 Indeed using two virtual CPU, my build crashes as well upon opening an image, albeit in a different place as you posted above. My guess is that we interact badly the OpenMP runtime `libgomp` on platform. Will try the last good build you mentioned above with the new tool chain next...

**Comment by gaaned92** [comment]
@adamreichold 
I don't know what happened but I was no longer able to buid with either the old and new toolchain.
Hopefully and mysteriously, 4.2.538 builds and run ok on old and new toolchain.
I did not made measurements but new toolchain seems to generate faster exe.
Thank you for your efforts. 

Now I will try with this PR and your patch. 
I think we are almost ready to upgrade to msys2 when the warnings will be corrected. warnings

**Comment by adamreichold** [comment]
@gaaned92 I completely forgot, but @heckflosse mentioned this in the forums: One does currently need to add `-mstackrealign` to the compiler flags for reliable optimized `win32` 32-bit builds. And indeed doing so fixes the crashes for me, also on the latest code and using the latest tool chain. I will add a node to the RawPedia article...

**Comment by gaaned92** [comment]
@adamreichold  : I confirm that we need mstackrelign or 32bit builds

**Comment by iliasg** [comment]
I think that -mstackrealign is needed only for winXP32 builds, my winvista32 and later work fine (faster ?) without ..

**Comment by adamreichold** [comment]
@gaaned92 @iliasg Yes, my tests using a 32-bit version of Windows 7 also indicate that `-mstackrealign` is only needed for Windows XP which is also how I tried to describe it in the RawPedia.

**Comment by gaaned92** [comment]
@adamreichold : the patch compiles. After selecting 'use system language' in preferences,  and restarting, RT is still in english. Not very important as langage can be manually selected. If there is no RT modification, I can locally modify Multilangmgr to build for XP.

**Comment by Beep6581** [comment]
We have confirmation that Windows XP 32-bit builds now work with C++11 and have a clear [guide](http://rawpedia.rawtherapee.com/Windows) for installing a toolchain using >=GCC-4.9. @heckflosse @adamreichold ok for commit to master?

**Comment by adamreichold** [comment]
@Beep6581 Very much so from my perspective. I will begin rebasing the dependent work after merging. (For example cleaning up usage of `std::auto_ptr` and porting the signals and slots to master.)

**Comment by sguyader** [comment]
A user reported to me an error while trying to run one of my Win32 builds on Windows XP.
The error message says that the entry point GetUserDefaultLocaleName could not be done in kernel32. 
I tried @adamreichold 's patch, but it failed to compile in multilangmgr.cc.
The user doesn't want to try a debug build, I don't know if it's needed.

**Comment by adamreichold** [comment]
@sguyader A debug build is probably not necessary since the problem seems to be clear, even if the solution is not. Could you paste the compiler error here? Or even better open a separate issue about this, since `GetUserDefaultLocaleName` is pretty much unrelated to using C++11. (The basic problem is assuming that `__MINGW64_VERSION_MAJOR` being defined at compile time means we build for a Windows version that is Vista or newer as far as I understand this.)

**Comment by Beep6581** [comment]
I don't think anyone would object to a CMake option -DFORWINXP32="ON" if that helps.

**Comment by heckflosse** [comment]
Can someone point me to the relevant line of code (I know I touched it some time ago)?

Ingo

**Comment by adamreichold** [comment]
@heckflosse See my above attempt to fix this by using `GetUserDefaultLCID`, i.e. `multilangmgr.cc` line 148 and following.

**Comment by Hombre57** [comment]
@adamreichold I'm still unable to build clearlooks, but using the proper command line does indeed help. Here is where the configure fail : 
```
checking for perl... /mingw64/bin/perl
checking for perl >= 5.8.1... 5.22.0
checking for XML::Parser... configure: error: XML::Parser perl module is required for intltool
```

I don't know where to find this XML::Paser module.

adamreichold:
@Hombre57 I guess this was installed as an indirect dependency of the things mentioned at the beginning of the RawPedia article for me. I think the correct package to install this explicitly is `perl-XML-Parser` which one can find using e.g. `pacman -Ss perl xml` and install by `pacman -S perl-XML-Parser`.

Beep6581:
Hello, I thought we merged this already... @adamreichold could you merge it if its ready?

adamreichold:
@Beep6581 I thought you were waiting for @heckflosse to give his blessings, but it should be ready and it won't change anything on systems with a recent `libsigc++` anyway, so I'll merge.

PR #3040, opened 2015-12-25
Add some example code using C++11 language features

adamreichold:
If pull request #3028 is merged and @Hombre57 agrees to the solution w.r.t. to the `FileChooserLastFolderPersister`, this pull request adds some example of how the use of the `auto` keyword and lambda expression can make the code more readable and more efficient.

### DIFF
--- a/rtgui/batchqueue.cc
+++ b/rtgui/batchqueue.cc
@@ -21,6 +21,7 @@
 #include <cstring>
 #include "../rtengine/rt_math.h"
 
+#include <functional>
 #include <iomanip>
 #include <sstream>
 #include <string>
@@ -37,13 +38,13 @@
 using namespace std;
 using namespace rtengine;
 
-BatchQueue::BatchQueue (FileCatalog* aFileCatalog) : processing(NULL), fileCatalog(aFileCatalog), sequence(0), listener(NULL)
+BatchQueue::BatchQueue (FileCatalog* aFileCatalog) : processing(NULL), fileCatalog(aFileCatalog), sequence(0), listener(NULL),
+    pmenu (new Gtk::Menu ())
 {
 
     location = THLOC_BATCHQUEUE;
 
     int p = 0;
-    pmenu = new Gtk::Menu ();
 
     pmenu->attach (*Gtk::manage(open = new Gtk::MenuItem (M("FILEBROWSER_POPUPOPENINEDITOR"))), 0, 1, p, p + 1);
     p++;
@@ -79,9 +80,9 @@ BatchQueue::BatchQueue (FileCatalog* aFileCatalog) : processing(NULL), fileCatal
     cancel->add_accelerator ("activate", pmenu->get_accel_group(), GDK_Delete, (Gdk::ModifierType)0, Gtk::ACCEL_VISIBLE);
 
     open->signal_activate().connect(sigc::mem_fun(*this, &BatchQueue::openLastSelectedItemInEditor));
-    cancel->signal_activate().connect (sigc::bind(sigc::mem_fun(*this, &BatchQueue::cancelItems), &selected));
-    head->signal_activate().connect (sigc::bind(sigc::mem_fun(*this, &BatchQueue::headItems), &selected));
-    tail->signal_activate().connect (sigc::bind(sigc::mem_fun(*this, &BatchQueue::tailItems), &selected));
+    cancel->signal_activate().connect (std::bind (&BatchQueue::cancelItems, this, selected));
+    head->signal_activate().connect (std::bind (&BatchQueue::headItems, this, selected));
+    tail->signal_activate().connect (std::bind (&BatchQueue::tailItems, this, selected));
     selall->signal_activate().connect (sigc::mem_fun(*this, &BatchQueue::selectAll));
 
     setArrangement (ThumbBrowserBase::TB_Vertical);
@@ -99,20 +100,18 @@ BatchQueue::~BatchQueue ()
     }
 
     fd.clear ();
-
-    delete pmenu;
 }
 
 void BatchQueue::resizeLoadedQueue()
 {
-    // TODO: Check for Linux
 #if PROTECT_VECTORS
     MYWRITERLOCK(l, entryRW);
 #endif
 
-    for (size_t i = 0; i < fd.size(); i++) {
-        fd.at(i)->resize(getThumbnailHeight());
-    }
+    const auto height = getThumbnailHeight ();
+
+    for (const auto entry : fd)
+        entry->resize(height);
 }
 
 // Reduce the max size of a thumb, since thumb is processed synchronously on adding to queue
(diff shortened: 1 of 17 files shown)


adamreichold:
As a probably more controversial proposal, I'd suggest removing the `safegtk` module as IMHO, it is not useful since it makes it impossible to do proper error handling.

Generally, errors should be handled where it is clear how these errors can be handled. Just printing an error message to standard output is probably not something users will ever notice and hence should only be a last resort. Also exceptions make it possible to delay error handling up to a level in the call chain where the necessary knowledge of what do in case of a failure is available.

Also the ability to disable the use of exceptions in `glibmm` was [disabled](https://git.gnome.org/browse/glibmm/tree/NEWS#n2660) in release 2.25.1 and hence `GLIBMM_EXCEPTIONS_ENABLED` is now always defined, so there is no more need to have `safegtk` abstract away the difference in error handling.

Hence if there are no objections, I would begin preparing a pull request to remove that module and placing its reusable parts into other modules as appropriate?

adamreichold:
Also if this is merged into master and the work w.r.t to `safegtk` is done, all parts of pull request #3011 will be resolved otherwise.

Hombre57:
PR ok for me, including the change in safegtk.

Hombre57:
Correction : we should take care of Win32 release for some time, and ensure that exceptions are supported on this platform, MSYS2 only provide gtkmm2.24. I don't know if exceptions are enabled in the provided package. If so, then we can get ride of safegtk I presume.

adamreichold:
@Hombre57 I have begun to prepare a branch [`remove-safegtk`](https://github.com/adamreichold/RawTherapee/tree/remove-safegtk) to remove the `safegtk` wrapper, but it is not yet ready for merging as I need check up on the shortcut directory problem and it does surely need more review and testing before it can be merged (and at +1808-2360 it is admittedly uncomfortably large).

Concerning the MSYS2 build, I think there is a misunderstanding: The flag to disable exceptions was removed from `glibmm` not `gtkmm` version 2.25.1 and MSYS2 currently provides version 2.46.1, so exceptions should definitely be enabled and just removing the non-exception code paths from `safegtk` might be a useful first step until the above branch is ready.

Hombre57:
Erm... yes, I misread it.

adamreichold:
Is anyone still planning to review this? (I would count @Hombre57's review as positive after the mix-up w.r.t. `glibmm` and `gtkmm` was resolved?) Are there questions w.r.t. to the new language features? Otherwise, I would propose merging it at the end of the week since it will remove quite a few compilation warnings and some dead code?

rtgui/batchqueue.cc
```diff
@@ -99,20 +100,18 @@ BatchQueue::~BatchQueue ()
...
 #if PROTECT_VECTORS
     MYWRITERLOCK(l, entryRW);
 #endif
 
-    for (size_t i = 0; i < fd.size(); i++) {
-        fd.at(i)->resize(getThumbnailHeight());
-    }
+    const auto height = getThumbnailHeight ();
+
+    for (const auto entry : fd)
```
Hombre57: Is it equivalent to the gone syntax ? Interesting !

On a side note, we have an official astyle format (provided in the root directory tree). The style says to add {} even on single line 'if' block. I know it's annoying, but I some point, someone will launch astyle and update the whole  code.

adamreichold: Yes, this is definitely one of the most useful small-scale additions. The syntax is almost equivalent, with a some caveats: `for(type var : container)` will basically expand to
````
for (auto iterator = std::begin(container), iterator != std::end(container); ++iterator) {
    type var = *iterator;
}
````
which should usually be more efficient than e.g. using indices. But it also means that one should usually bind `var` by reference, i.e. use `for(const auto& var : container)` as to not make a copy of `*iterator`. In this particular case, the entries are pointers so copying them is the same as taking a reference. (Also factoring out the common value of `getThumbnailHeight` is unrelated.)

Concerning the if clauses, this is the same as with the trailing underscores for member variables: I saw it in various places and hoped to be consistent by adopting it, but now I see that this is not the case. I will stick to braces for single line if clauses from now on, astyle will fix the rest.

File: rtgui/batchqueue.cc [hunk 2]
```diff
@@ -161,328 +160,220 @@ bool BatchQueue::keyPressed (GdkEventKey* event)
...
-                                    jpegQuality = atoi(currLine.substr(0, pos).c_str());
-                                    currLine = currLine.substr(pos + 1);
-
-                                    // Looking for the jpeg subsampling
-                                    pos = currLine.find('|');
-
-                                    if (pos != Glib::ustring::npos) {
-                                        jpegSubSamp = atoi(currLine.substr(0, pos).c_str());
-                                        currLine = currLine.substr(pos + 1);
+            while (std::getline(line, column, '|')) {
```
Hombre57: This one will definitely go in my memo book :)

rtgui/batchqueue.cc
```diff
@@ -509,107 +400,106 @@ int cancelItemUI (void* data)
...
-                for (pos = fd.begin(); pos != fd.end(); pos++)
-                    if (!(*pos)->processing) {
-                        fd.insert (pos, entry);
-                        break;
-                    }
-            }
+            fd.erase (pos);
+
+            // find the first item that is not under processing
+            const auto newPos = std::find_if (fd.begin (), fd.end (), [] (const ThumbBrowserEntryBase* fdEntry) { return !fdEntry->processing; });
```
Hombre57: Nice one too, using Lambda function.

Wouldn't it be interesting to add a special "//SAMPLE: brief explanation" comment above the key elements that you want to show so that we could make a global search for //SAMPLE and pick in the result list the piece of code that will make our life easier ? I'd add one above this line.

rtgui/batchqueue.cc
adamreichold: Yeah, one does splitting using a separator so often, even the C++ standard library had to add something to make it simpler. :-D

Hombre57:
:+1: for committing. I have no question so far, I made some search about C++11 while reviewing your patch, but no doubt that they will come if I can't reproduce them.

I also saw that C++11 introduce a multithreading model. I suppose that we should stick to using Glib one, right ?

rtgui/batchqueue.cc
adamreichold: I will make a note to provide some instances of this, but I would prefer to do that in a separate pull request as to not inflate the diff even more. But I do agree that STL algorithms is definitely one of the best places to use lambdas.

adamreichold:
@Hombre57 Yes, C++11 does introduce a multi-threaded memory model and also some standard library support for threads, locks and condition variables. However, the implementation of these parts of the standard library was often lacking on Windows, especially in MinGW-derived implementations. I also find it still somewhat incomplete, i.e. reader-writer locks only appeared with C++14. Hence I would say we should stick with Glib's threading primitives for now.

PR #3059, opened 2016-01-04
Fix enforcing the cache size limit and modernize the cache manager

adamreichold:
Due to the method `CacheManager::applyCacheSizeLimitation` not using the MD5-derived part of the file name, it would not actually delete anything, and hence not enforce the cache size limit upon closing the application and hence the cache grows without bounds. The reason seems to be the high amount of redundancy in the cache manager's implementation.
This pull request tries to fix that error and also aims to remove its root cause by reworking the cache manager to add optional error reporting, reduce the amount of code duplication and make use of current language features.
(The diff is quite large since this is basically coming from the `remove-safegtk` branch. Review might be easier on the resulting version of the source file as Git itself reports an 81% rewrite factor.)

### DIFF
--- a/rtengine/safegtk.cc
+++ b/rtengine/safegtk.cc
@@ -150,59 +150,6 @@ Glib::RefPtr<Gio::FileInfo> safe_next_file (Glib::RefPtr<Gio::FileEnumerator> &d
 # define SAFE_ENUMERATOR_CODE_END   } if (error.get())  printf ("%s\n", error->what().c_str());}while (0)
 #endif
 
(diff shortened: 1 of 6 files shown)


Beep6581:
It builds and runs fine here.

gaaned92:
It compiles, build and runs fine on WXP. Display of cached thumbnails seems faster.

adamreichold:
Considering the positive tests and the discussed grace period of one to two days, I'll merge this tomorrow evening (CET)?

Issue #1668, opened 2015-08-11
thumbnail.cc uses Glib::Mutex recusively


Beep6581:
Originally reported on Google Code with ID 1684
```
thumbnail.cc uses Glib::Mutex recusively, for example getThumbnailSize() locks the mutex,
and then calls  getProcParams() which itself locks the mutex.

The Glibmm docs say that this should not work:
"Glib::Mutex is not recursive, i.e. a thread will deadlock, if it already has locked
the mutex while calling lock(). Use Glib::RecMutex instead, if you need recursive mutexes."

It seems that while in the current version of glibmm on windows it works without a
deadlock, it does deadlock on linux (hence it being ifdef'ed out).

It would be wise to follow the advice of the docs, and switch the code to use RecMutex
(if recursiveness is needed). The attached patch does this. It works for me on linux,
but i have not tested on windows.

This was noticed while investigating issue 1675, but i think is really a separate issue.
There a plenty of other bits of code using Glib::Mutex, i have not checked those yet.

Click on Preferences > About > Version and paste all that here.
current hg b681d69af34d


```

Reported by `samtygier` on 2013-01-16 23:03:31

<hr>
 * *Attachment: [swtich_to_recmutex.diff](https://storage.googleapis.com/google-code-attachments/rawtherapee/issue-1684/comment-0/swtich_to_recmutex.diff)*

Beep6581:
@Hombre57 what's the status here?

Hombre57:
IIRC, I saw a change in gcc that would make both Linux and Windows pthread implementation works the same, maybe since gcc 5, but I can't find that information again. We can still close this issue, since the workaround shouldn't introduce any speed penalty. If one want to look in that again, he may simplify the MyMutex classes. However we should keep them for debugging purpose.

PR #3062, opened 2016-01-06
Clean up thread utilities

adamreichold:
The thread utilities are currently used to abstract away some differences in the implementation of the Glib threading primitives on POSIX systems and Windows, however interface and implementation and hardly separate and so it is hard to read or reason about.

This is not something I would try to merge without review, but this pull request tries to improve the situation somewhat and also aims fix a few shortcomings of the current implementation:
* The instrumentation of the strict implementation of mutexes is definitely nice for the programmer, but we don't need to pay the performance cost of a recursive mutex in release builds where crashing and freezing is equally bad IMHO. (For debug builds nothing is changed and violating strict locking discipline will drop one into the debugger immediately.)
* The custom read-write lock implementation actually had a race condition insofar the static lock counters where updated concurrently without synchronization. (Only the specific locks mutex was taken, but they all share the same static counter.) Since file and line number already provides a lot of options to debug issues IMHO, this pull requests removes these counters. It also makes the `acquire` and `release` methods separately compiled as they seem too large to be obvious candidates for inlining. (If there are any performance regressions, the non-tracing implementation can certainly be inlined again.)
* It removes the `PROTECT_VECTORS` option and always locks the shared data structures. I do not know the full back story on this, but I think a program is either correct without locking or it is not. I also never heard of a thread-safe implementation of `std::vector` on any major platform, so I think this only clutters up the code and would be one more configuration to test or which users could report bugs against.


adamreichold:
I also find the implementation of the read-write lock adventurous insofar readers a double-counted as writers instead of the other way around? But I did not opt to change anything in the algorithm as to concentrate on the maintainability and performance issues.

gaaned92:
builds and runs on WXP. no noticeable impact

Hombre57:
I made this class to help find location of freeze, and created MyMutex to solve a difference in platform implementation that was behaving differently (see issue #1668).

I don't have the knowledge of reviewing your patch, so I trust you and the other devs on this. But I agree that PROTECT_VECTORS is pointless.

adamreichold:
@Hombre57 The working of `MyMutex` - i.e. resolving the difference between the POSIX and Windows implementations of `Glib::Threads::Mutex` - would still be the same if `STRICT_MUTEX = 0`. If `STRICT_MUTEX = 1` (which however seems to be default), then `MyMutex` already is non-recursive on both platforms.
This patch mainly does away with paying a recursive mutex in the implementation to raise `SIGINT` on failing to following locking discipline. (Which will only benefit the POSIX platforms since their default Glib mutex was non-recursive to being with to explain @gaaned92's observation.) For debug builds, everything basically stays as it is, but hopefully with an easier to maintain implementation.

Beep6581:
Please ding me when merging so I will upload to coverity.

Beep6581:
Compilation of `adamreichold/clean-up-threadutils` after merging into master is successful and the build runs fine. Gentoo, GCC-5.3.0.

Beep6581:
@adamreichold do we merge?

adamreichold:
@Beep6581 I would feel better if there was some definite yes or no from a code review and I also don't think time is pressing for this change. If nobody currently has time to review, I'd suggest to upload to Coverity without this change. If we are feeling adventurous we can merge now and handle any fallout as it happens since it did see testing from at least two people besides me.

Beep6581:
It's true that we're not pressed for time, but I also know from history that waiting longer won't get it more testing, it will just go stale. As such, if @heckflosse gives the thumbs up, I would go ahead with the merge so that all users can help catch issues (which I'm happy to see is becoming easier all the time, now that we regularly ship debug builds which include gdb.exe).

heckflosse:
Looks ok for me though build of the branch fails:

```
/batchqueue.cc:221:25: error: variable 'std::ofstream file' has initializer but incomplete type
     std::ofstream file (fileName, std::ios::trunc);
```

adamreichold:
@heckflosse That is the same issue we had on master due to the missing `#include <fstream>`. I rebased the branch on the current master which should fix this among other things.

heckflosse:
@adamreichold Thank you! Builds and works fine. :+1: for merge

Beep6581:
Done. If things explode, blame me :)

PR #3063, opened 2016-01-06
Use signals and slots instead of listeners

adamreichold:
This is an updated (or rather rebased) version of pull request #3007 which targets `master` instead of `gtk3`. The benefits and limitations of a signals-and-slots approach to wiring together widgets are discussed there and as reasoned there, while this should not be used everywhere indiscriminately, it could decouple and simplify some interactions in the UI code.


Beep6581:
Compiles and runs fine, tested the file browser, recent folders, places, etc.

PR #3066, opened 2016-01-06
Merge master into gtk3

adamreichold:
This pull request is the result of merging `master` into `gtk3`. The conflicts seem to be simple enough, expect for `PopUpCommon` which saw significant incompatible changes. @Hombre57 Could you have a look at the result of merging before this is pushed?


Hombre57:
Yes, but how should I proceed ?

Beep6581:
Merged successfully and compiles and runs fine.

PR #3073, opened 2016-01-08
Simplify the multiple language manager for better maintainability

adamreichold:
I recently looked into the multiple language manager to make it work on Windows XP again and noticed the various tricks to work around old MinGW versions. Also that the language mappings were expanded into code branches and the very common `M` method was not inlined (`getStr` is not a good candidate for inlining as it is called recursively). This pull request tries to improve on these observations and provide a reduction of line count while doing so without changing the interface provided by the module.

### DIFF
--- a/rtgui/multilangmgr.cc
+++ b/rtgui/multilangmgr.cc
@@ -19,6 +19,7 @@
 #include "multilangmgr.h"
 
 #include <fstream>
+#include <regex>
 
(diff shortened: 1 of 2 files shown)


gaaned92:
W32 : builds aand runs ok

Beep6581:
Getting old, merging.

PR #3080, opened 2016-01-10
Reworked editor bottom toolbar

Hombre57:


Beep6581:
Compiles and works well.

rtgui/editorpanel.cc
```diff
@@ -46,6 +46,8 @@ class EditorPanel::MonitorProfileSelector
     void prepareProfileBox ()
     {
         profileBox.set_size_request (100, -1);
+        profileBox.setPreferredWidth(90, 150);
```
adamreichold: I was under the impression we need have a space before parentheses for function and constructor calls, e.g. `profileBox.setPreferredWidth (90, 150);`? I mess it up myself all the time, but the code added here seems inconsistent w.r.t. to this?

rtgui/editorpanel.cc
```diff
@@ -355,56 +354,51 @@ EditorPanel::EditorPanel (FilePanel* filePanel)
...
-    progressLabel = Gtk::manage (new Gtk::ProgressBar());
+    progressLabel = Gtk::manage (new MyProgressBar(300));
     progressLabel->set_show_text(true);
-    setExpandAlignProperties(progressLabel, false, false, Gtk::ALIGN_CENTER, Gtk::ALIGN_CENTER);
+    setExpandAlignProperties(progressLabel, true, false, Gtk::ALIGN_START, Gtk::ALIGN_CENTER);
     progressLabel->set_fraction(0.0);
 
     // TODO is this needed? What for?
-    if (cssProvider) {
+    /*if (cssProvider) {
```
adamreichold: If the code is commented out, why not fully delete it? We have version control so nothing is really lost and commented-out code only clutters things up and breaks when its surroundings are changed.

rtgui/editorpanel.cc
```diff
@@ -890,7 +900,7 @@ void EditorPanel::setProgressState (bool inProcessing)
 struct spparams {
     double val;
     Glib::ustring str;
-    Gtk::ProgressBar *pProgress;
+    MyProgressBar *pProgress;
     Glib::RefPtr<Gtk::CssProvider> cssProvider;
```
adamreichold: If it isn't necessary above, is the `cssProvider` necessary here?

rtgui/guiutils.cc
```diff
@@ -982,19 +982,37 @@ bool MyComboBoxText::on_scroll_event (GdkEventScroll* event)
     return false;
 }
 
+void MyComboBoxText::setPreferredWidth (int minimum_width, int natural_width)
+{
+    if (natural_width==-1 && minimum_width==-1) {
```
adamreichold: Spaces around the `==` operator seem advisable IMHO.

rtgui/guiutils.cc
```diff
@@ -1010,13 +1028,31 @@ bool MyComboBox::on_scroll_event (GdkEventScroll* event)
...
 
+void MyComboBox::setPreferredWidth (int minimum_width, int natural_width)
+{
+    if (natural_width==-1 && minimum_width==-1) {
+        naturalWidth = minimumWidth = 70;
+    } else if (natural_width==-1) {
+        naturalWidth =  minimumWidth = minimum_width;
+    } else if (minimum_width==-1) {
+        naturalWidth = natural_width;
+        minimumWidth = rtengine::max(naturalWidth/2, 20);
```
adamreichold: What is wrong with `std::max`? Would be one less cross-dependency into `rtengine`.

rtgui/guiutils.h
```diff
@@ -430,6 +436,22 @@ class MyImageMenuItem : public Gtk::MenuItem
     const Gtk::Label* getLabel () const;
 };
 
+class MyProgressBar : public Gtk::ProgressBar
```
adamreichold: The `guiutils` package is becoming quite large. What do you think about splitting it up, e.g. making a `mywidgets` subfolder of `rtgui` with separate headers or something like that?

rtgui/zoompanel.cc
```diff
@@ -41,34 +41,51 @@ ZoomPanel::ZoomPanel (ImageArea* iarea) : iarea(iarea)
...
-    pack_start (*zoom11, Gtk::PACK_SHRINK, 0);
+    setExpandAlignProperties(zoom11, false, false, Gtk::ALIGN_CENTER, Gtk::ALIGN_CENTER);
+
+    attach_next_to (*zoomOut, Gtk::POS_RIGHT, 1, 1);
+    attach_next_to (*zoomIn, Gtk::POS_RIGHT, 1, 1);
+    attach_next_to (*zoomFit, Gtk::POS_RIGHT, 1, 1);
+    attach_next_to (*zoomFitCrop, Gtk::POS_RIGHT, 1, 1);
+    attach_next_to (*zoom11, Gtk::POS_RIGHT, 1, 1);
+
+    /*
```
adamreichold: Personally, I rather dislike commented-out code...

adamreichold:
Builds and works fine here. Comments on some technicalities inline...

rtgui/editorpanel.cc
Hombre57: Nope, removed.

rtgui/editorpanel.cc
Hombre57: I don't saw this requirement anywhere, so maybe it was only an impression ? Sometimes I align some functions in header files, but it's purely cosmetic. I ran editorpanel.cc through astyle and it hasn't changed that, so this doesn't seem to be important. Some devs add one, but I don't. I'm using the autocompletion which doesn't add space here neither, but I don't think it's important.

rtgui/guiutils.h
Hombre57: Yes, I was about to create a `widgets` subdir for our custom widgets. `MyCombobox`, `MyProgressBar`, etc... are more utility class than new widgets, so should we move them to `widgets`, create a separate `utility` subdir ?

rtgui/editorpanel.cc
adamreichold: Well I agree that is probably not important (actually I strongly dislike it as I am not used to it ;-)), but it is definitely all over the code and I did try to be consistent with that up to now. While that is probably irrelevant to this pull request, I think we should fix one way and make `astyle` apply that for us instead of leaving it untouched.

rtgui/guiutils.h
adamreichold: Things named `utility` or `miscellaneous` tend to be dumping grounds for all kinds of code, hence I'd suggest to use `widgets`. (And even if the `My*` do only small extensions to their base classes, I think it is important to think of them as fully-fledged widgets that only happen to inherit most of the interface and implementation as even small extension can significantly change their behaviour.)

rtgui/editorpanel.cc
Hombre57: As I don't really care of that, you can open an issue for adding this to rawtherapee.astylerec, but it will touch really a lot of files and should be applied in a "critical ~~section~~ moment".

adamreichold: Even if you don't care, could you try to stay consistent, i.e. add a space before the parentheses of a function call? I'd say it is of similar importance as the trailing underscore.

rtgui/guiutils.h
Hombre57: Things like this should be done after the Gtk2 branches are dropped I think, otherwise I'll let you merge it with the Gtk3 branch ;) There's no hurry.

PR #3086, opened 2016-01-11
Dcraw decode speedup

heckflosse:


adamreichold:
@heckflosse I am sorry, that I have not found the time yet to review since. But can comment on why you deleted the branch before merging anything? Did these or similar changes go upstream? Thanks!

heckflosse:
@adamreichold Unfortunately changes won't go upstream. I want to upgrade to newest dcraw version first. Currently we're one behind latest. Then I'll probably apply some of the changes.

Issue #3088, opened 2016-01-12
Filebrowser's History panel broken


Hombre57:
Description:
  1. Click on a single or multiple images in  the File Browser tab.
  2. Make some adjustments
  3. Select the first entry of the History to go back to initial state
  4. Select another image to change the selection

The selected images are updated to the N-1 state (or to the last entry, haven't checked).

A fix is coming.

Hombre57:
See PR #3089 .

PR #3089, opened 2016-01-12
Solving issue #3088 : "Filebrowser's History panel broken"

Hombre57:
The `clean` method does throw a `selection_change` event, which was not the case with Gtk2.

### DIFF
--- a/rtgui/history.cc
+++ b/rtgui/history.cc
@@ -155,10 +155,9 @@ History::History (bool bookmarkSupport) : blistener(NULL), tpc (NULL), bmnum (1)
 void History::initHistory ()
 {
 
-    selchangehist.block(true);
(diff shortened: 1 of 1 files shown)


rtgui/history.cc
```diff
@@ -155,8 +155,10 @@ History::History (bool bookmarkSupport) : blistener(NULL), tpc (NULL), bmnum (1)
 void History::initHistory ()
 {
 
+    selchangehist.block(true);
```
adamreichold: The `ConnectionBlocker` helper class is available on `gtk3` and should be used here IMHO.

Hombre57: Ah, I missed that change. I'll use it then and will commit, since the fix is trivial and no one complain.

PR #3096, opened 2016-01-14
Rldeconv speedup

heckflosse:
I want to merge this into master after getting a review and removing the benchmarking code.

Ingo


rtengine/gauss.cc
```diff
@@ -0,0 +1,1293 @@
...
+ *  but WITHOUT ANY WARRANTY; without even the implied warranty of
+ *  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ *  GNU General Public License for more details.
+ *
+ *  You should have received a copy of the GNU General Public License
+ *  along with RawTherapee.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include "gauss.h"
+#include <cmath>
+#include <stdlib.h>
```
adamreichold: I think including `cstdlib` would be more idiomatic for C++ code.

rtengine/gauss.h
```diff
@@ -19,797 +19,12 @@
...
-#include <cstdlib>
-#include <cstring>
-#include <cmath>
-#include "opthelper.h"
-#include "stdio.h"
-#include "boxblur.h"
-// classical filtering if the support window is small:
+#include <string.h>
+enum eGaussType {GAUSS_STANDARD, GAUSS_MULT, GAUSS_DIV};
+#define GAUSS_SKIP 0.25
```
adamreichold: I don't think these constants need to be in the header, as their usage seems to be limited to `gauss.cc`.  There I would also recommend to use `constexpr auto` instead of `#define` to get proper type checking without overhead.

I am also do not think it is necessary to include `string.h` (or `cstring` for that matter) and `NULL` is probably better replaced by `nullptr` in any case.

rtengine/gauss.cc
```diff
@@ -0,0 +1,1293 @@
...
+            } else { // large sigma only with double precision
+                gaussHorizontal<T> (src, dst, W, H, sigma);
+                gaussVertical<T>   (dst, dst, W, H, sigma);
+            }
+#endif
+        }
+    }
+}
+
+void gaussianBlur(float** src, float** dst, const int W, const int H, const double sigma, float *buffer, eGaussType gausstype, float** buffer2)
```
adamreichold: Since `gaussianBlur` is the only function that should be visible in other translation units, everything else in the source file should probably go into the anonymous namespace to tell the compiler and linker that all generated code should have static visibility which will improve optimization possibilities.

rtengine/PF_correct_RT.cc
```diff
@@ -1341,7 +1341,7 @@ SSEFUNCTION void ImProcFunctions::BadpixelsLab(LabImage * src, LabImage * dst, d
 #ifdef __SSE2__
         int j;
 //  vfloat2 sincosvalv;
-//  __m128 piidv = _mm_set1_ps(piid);
```
adamreichold: If it is commented-out, maybe it could go away completely if we are touching it?

rtengine/helpersse2.h
```diff
@@ -117,19 +117,22 @@ static INLINE vfloat vcast_vf_f(float f)
 
 static INLINE vfloat vaddf(vfloat x, vfloat y)
 {
-    return _mm_add_ps(x, y);
```
adamreichold: I think some comments why the intrinsics were removed here would be in order. If I remember correctly, you did this due to the compiler not emitting FMA instructions?

rtengine/impulse_denoise.h
```diff
@@ -42,14 +44,14 @@ SSEFUNCTION void ImProcFunctions::impulse_nr (LabImage* lab, double thresh)
...
-    float ** impish = new float *[height];
+    char ** impish = new char *[height];
+    impish[0] = new char [width * height];
 
-    for (int i = 0; i < height; i++) {
-        lpf[i] = new float [width];
-        //memset (lpf[i], 0, width*sizeof(float));
-        impish[i] = new float [width];
-        //memset (impish[i], 0, width*sizeof(unsigned short));
+    for (int i = 1; i < height; i++) {
```
adamreichold: Nice trick to fix the array-of-arrays problem without changing the users. Maybe this merits a separate helper class that does this transformation and manages the memory using constructors and destructors?

rtengine/impulse_denoise.h
```diff
@@ -188,10 +180,6 @@ SSEFUNCTION void ImProcFunctions::impulse_nr (LabImage* lab, double thresh)
 
                 for (i1 = max(0, i - 2); i1 <= min(i + 2, height - 1); i1++ )
                     for (j1 = 0; j1 <= j + 2; j1++ ) {
-                        if (i1 == i && j1 == j) {
```
adamreichold: Just out of curiousity: Was it not worth it? Branchiness increased pipeline stalls?

rtengine/impulse_denoise.h
```diff
@@ -277,19 +257,15 @@ SSEFUNCTION void ImProcFunctions::impulse_nr (LabImage* lab, double thresh)
...
 
-    for (int i = 0; i < height; i++) {
-        delete [] lpf[i];
-        delete [] impish[i];
-    }
-
-    delete [] lpf;
-    delete [] impish;
+    delete [] lpf[0];
+    delete [] impish[0];
```
adamreichold: You still have two calls to `new`, so you should also have two calls to `delete`, i.e. think `delete [] lpf;` and friends are missing as an array of pointers also needs memory. Again, I would suggest a separate helper class that implicitly or explicitly converts to `float**` (or `T**` if you want to go template on it).

rtengine/ipsharpen.cc
```diff
@@ -183,31 +144,32 @@ void ImProcFunctions::deconvsharpening (LabImage* lab, float** b2, SharpeningPar
...
             }
-
     } // end parallel
 
-    for (int i = 0; i < H; i++) {
-        delete [] tmpI[i];
-    }
+    delete [] tmpI[0];
 
-    delete [] tmpI;
```
adamreichold: Again, I don't think this can be left out.

adamreichold:
I do like the clean-up w.r.t. to the Gaussian blur and `MLmicrocontrast`. I am still somewhat lukewarm about the manual loop unrolling even though I know that it is a standard technique. But since it is so standard, I would rather prefer the compiler to do it, maybe using pragmas or some other form of explicit control, as the code becomes rather redundant this way. Do we know if more modern compilers are able to better unroll and vectorize the previous code?

rtengine/impulse_denoise.h
heckflosse: That was already covered by the following:

if (impish[i1][j1]) {
                            continue;
                        }

rtengine/impulse_denoise.h
heckflosse: Good catch. Will fix it.

rtengine/ipsharpen.cc
heckflosse: Again good catch. I'll fix that too.

rtengine/gauss.cc
heckflosse: Fixed.

rtengine/gauss.h
heckflosse: Fixed.

rtengine/gauss.cc
heckflosse: Done.

rtengine/PF_correct_RT.cc
heckflosse: It was touced by replacing all _mm_set1_ps( with F2V( in that file. I'll treat this function in a separate issue.

rtengine/helpersse2.h
heckflosse: Yes, compiler (at least gcc) emits better code when not using intrinsics. For example:
_mm_add_ps(_mm_mul_ps(a,b),c) results in exactly these instructions, whereas a*b+c is compiled to an FMA instruction when compiled for chips with that feature.
I'll add some comments.

heckflosse:
Adam, thanks a lot for the review :+1: 

about the manual vectorization:
In my experience, gcc (without pragma omp simd) is able to vectorize 'simple' loops. As soon as there is some non trivial stuff inside a loop, it won't be vectorized at all. If a loop gets vectorized by gcc, I'm not able to write faster code manually!
Maybe OpenMp 4.0 #pragma omp simd does a better job here.

About manual loop unrolling to get better usage of CPU cache:
I'm not aware of how to tell the compiler to do that automatically.
I also would prefer an automatic optimal solution for unrolling and vectorization...

Ingo

rtengine/impulse_denoise.h
heckflosse: We should review this in a separate Issue I guess. Maybe we could also use array2d here...

heckflosse:
I pushed some changes to the PR. They compiled fine, but didn't link due to my recent change to new msys2 toolchain. If someone else could test and merge them (in case the changes are ok) I would be glad.
@adamreichold or @Beep6581 ^^^

Ingo

Beep6581:
@heckflosse this is a rebase of your commits on top of master including the ipretinex.cc fix:
https://gist.github.com/3f12c8cfd6de4c012cd4
Currently compilation fails:
```
/home/drslony/repo-rt/rtengine/shmap.cc: In member function ‘void rtengine::SHMap::updateL(float**, double, bool, int)’:
/home/drslony/repo-rt/rtengine/shmap.cc:236:26: error: expected primary-expression before ‘float’
             gaussianBlur<float> (map, map, W, H, radius);
                          ^
```

If I remove the <float>, it succeeds, compiles and Retinex works well. I haven't tested anything else.

Beep6581:
I compared patched vs unpatched.
Neutral + S/H sharp mask off + Sharpening (RL) + INR + Defringe
Same + Retinex
`compare` shows differences http://i.imgur.com/DF15Gi3.jpg
but I see no differences by eyeballing except for one issue: there is a bar running all along the top edge in the patched versions
http://i.imgur.com/l1SwVEq.png

heckflosse:
@Beep6581 Thanks for testing. There are three things which cause the differences:
1) The top line difference was caused by a bug in deconv-sharpening, fixed
2) the vertical stripes were caused by three typos in gauss.cc, fixed
3) the remaining difference is caused by a more accurate calculation of the YvV factors in new version.
Old version calculated them with double precision in gaussverticalsse, but with single precision in gausshorizontalsse. New version uses double precision to calculate this factors in all cases.

Because I've some difficulties to merge the current branch into master, I'll create a new one, which is based on master and can be applied easily.

Ingo

rtengine/impulse_denoise.h
adamreichold: Agreed, reusing existing code should be preferred.

heckflosse:
I created https://github.com/Beep6581/RawTherapee/tree/gaussmerge to replace this one. Above mentioned Issues are solved.

Ingo

Issue #3100, opened 2016-01-16
Display Problems with several GTK-Themes


naanoo:
You seem to try to override the system's gtk theme. But not all the properties are set.

That causes some problems with several themes.

See:
https://github.com/horst3180/arc-theme/issues/376

Warm regards,

Sebastian

horst3180:
More specifically the inbuilt themes need to set `GtkTreeView::even_row_color` and `GtkTreeView::odd_row_color` similar to [this](https://github.com/horst3180/arc-theme/blob/4d092abce315926a33442d20f27e0f82f4939ae3/common/gtk-2.0/main.rc#L63).

Beep6581:
Can you reproduce it on a recent version? Which version? (number + hash) Which version of Gtk3 does it use?

naanoo:
I use the RAWtherapee and GTK versions supplied by Ubuntu 15.10 by default.

*RAWTherapee:*
Version: 4.2.0
Changeset: c81d9026d25c

*GTK:*
Version: 3.14.5-1+deb8u1 0

Beep6581:
That's very old and unsupported. Probably already fixed. Closing. Please only open bug reports if you have reproduced the problem with the latest code.
http://rawtherapee.com/downloads
http://rawpedia.rawtherapee.com/Linux

horst3180:
No, it's not fixed.
I can reproduce this with the latest version from `master` (https://github.com/Beep6581/RawTherapee/commit/5ff270bf51d16a13147f1849fff0936ce528530f). I think only the GTK2 version is affected by this bug.
I managed to fix it by patching the inbuilt themes like I described above. I can send you a pull request.

Beep6581:
Please send a PR if it fixes master.

Beep6581:
Fixed by @horst3180 
https://github.com/Beep6581/RawTherapee/commit/94725dedae4c210f5d15eda980f4e01a9eeb817a

PR #3102, opened 2016-01-17
Prevent treeview background color from being changed by the system theme

horst3180:
This fixes the issue outlined in https://github.com/Beep6581/RawTherapee/issues/3100 by setting `GtkTreeView::even_row_color` and `GtkTreeView::odd_row_color`.
This prevents the system's default theme from overwriting these values.


adamreichold:
Tested on Linux using Gtk+ 2.24.29 with XFCE theme 2.10.1 using system theme and some of the built-in themes with out any obvious ill effects. +1 for merging.

Beep6581:
Thank you @horst3180

PR #3106, opened 2016-01-18
Cleaned gauss code and included some speedups

heckflosse:
This replaces the rldeconv_speedup PR in the hope merging this will be easier

Ingo


heckflosse:
I also added a small speedup (only 6%) for Defringe when Defringe curve is used.

PR #3124, opened 2016-01-27
Improve handling of slow CLUT parsing

adamreichold:
We now ask the user if he wants to continue scanning or abort the scan right now instead of always completing the scan after the first notification.


Beep6581:
Works well, I think that solves it.

PR #3125, opened 2016-01-28
limit number of file attributes

heckflosse:
Limiting the number of fetched file attributes significantly reduces start time and time to change folder in file browser when working with folders containing a lot of files (at least on windows).

### DIFF
--- a/rtengine/safegtk.cc
+++ b/rtengine/safegtk.cc
@@ -111,15 +111,15 @@ Glib::RefPtr<Gio::FileInfo> safe_next_file (Glib::RefPtr<Gio::FileEnumerator> &d
     return info;
 }
 
-# define SAFE_ENUMERATOR_CODE_START \
-                do{try {    if ((dirList = dir->enumerate_children ())) \
+# define SAFE_ENUMERATOR_CODE_START(attributes) \
+                do{try {    if ((dirList = dir->enumerate_children ((attributes)))) \
                         for (Glib::RefPtr<Gio::FileInfo> info = safe_next_file(dirList); info; info = safe_next_file(dirList)) {
 
 # define SAFE_ENUMERATOR_CODE_END \
                 }}  catch (Glib::Exception& ex) {   printf ("%s\n", ex.what().c_str()); }}while(0)
 
 /*
- * safe_build_file_list can now filter out at the source all files that doesn't have the extensions specified (if provided)
+ * safe_build_file_list can now filter out at the source all files that don't have the extensions specified (if provided)
  */
 void safe_build_file_list (Glib::RefPtr<Gio::File> &dir, std::vector<Glib::ustring> &names, const Glib::ustring &directory, const std::vector<Glib::ustring> *extensions)
 {
@@ -127,7 +127,7 @@ void safe_build_file_list (Glib::RefPtr<Gio::File> &dir, std::vector<Glib::ustri
 
     if (dir) {
         if (!extensions) {
-            SAFE_ENUMERATOR_CODE_START
+            SAFE_ENUMERATOR_CODE_START("standard::name")
             names.push_back (Glib::build_filename (directory, info->get_name()));
             SAFE_ENUMERATOR_CODE_END;
         } else {
@@ -138,7 +138,7 @@ void safe_build_file_list (Glib::RefPtr<Gio::File> &dir, std::vector<Glib::ustri
                 lcExtensions.push_back ((*extensions)[i].lowercase());
             }
 
-            SAFE_ENUMERATOR_CODE_START
+            SAFE_ENUMERATOR_CODE_START("standard::name")
             // convert the current filename to lowercase in a new ustring
             Glib::ustring fname = Glib::ustring(info->get_name()).lowercase();
 
@@ -174,7 +174,7 @@ void safe_build_subdir_list (Glib::RefPtr<Gio::File> &dir, std::vector<Glib::ust
             return;
         }
 
-        SAFE_ENUMERATOR_CODE_START
+        SAFE_ENUMERATOR_CODE_START("standard::name,standard::type,standard::is_hidden")
 
         if (info->get_file_type() == Gio::FILE_TYPE_DIRECTORY && (!info->is_hidden() || add_hidden)) {
             subDirs.push_back (info->get_name());
--- a/rtgui/cachemanager.cc
+++ b/rtgui/cachemanager.cc
@@ -351,7 +351,7 @@ void CacheManager::applyCacheSizeLimitation () const
         const auto dirName = Glib::build_filename (baseDir, "data");
         const auto dir = Gio::File::create_for_path (dirName);
 
-        auto enumerator = dir->enumerate_children ("standard::*,time::*");
+        auto enumerator = dir->enumerate_children ("standard::name,time::modified");
 
         while (auto file = enumerator->next_file ()) {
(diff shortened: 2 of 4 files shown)


heckflosse:
I added https://github.com/Beep6581/RawTherapee/commit/ccd4be9e10c26feb358a350771540ef0bd3a2991 to avoid double build of folder list in dirbrowser to save some processing time when starting rt, using recent folders etc...

@adamreichold Can you take a look please? Maybe there's a better solution disconnecting the listeners instead of using semaphor variables?

Ingo

heckflosse:
I want to add some measurements:

Before change (master branch): 3327 ms to start rt with a folder containing 249 files
First commit reduced time to: 314 ms
Second commit reduced time to : 234 ms

That's on my Win64 8-core 4Ghz Bulldozer with 32 GB RAM and RAID-0 of three 1TB-Drives

On my old laptop (Sabayon Linux64, dual core 2.3Ghz with 3 GB RAM and a slow 320 GB drive you can simply divide above times by 10 (really, divide, not multiply, Linux file system seems to be much faster)

Ingo

adamreichold:
@heckflosse I have this one on my to-do list, but I will probably not be able to so until tomorrow evening.

heckflosse:
Ok, thank you

adamreichold:
Tested the cache manager part on Linux and it does seem to work as intended with the correct sorting by modification time. Concerning the speed changes, I don't think the Windows filesystems are that slow, it is rather the `stat`-like API that are problematic on Windows. But limiting the amount of attributes fetched is certainly a nice way of solving that problem.

Concerning the `safegtk` change, I would personally prefer if every commit to that module would remove something from there, but the change itself seems correct AFAIU.

As for suppressing the unnecessary invocation of `DirBrowser::row_expanded`, I think the idiomatic way to go about that would be to store the connection object and block that, e.g. using the `ConnectionBlocker` RAII helper class, e.g. something like
````
diff --git a/rtgui/dirbrowser.cc b/rtgui/dirbrowser.cc
index 9a11bf7..cfccdb0 100644
--- a/rtgui/dirbrowser.cc
+++ b/rtgui/dirbrowser.cc
@@ -42,8 +42,7 @@ struct DirNameComparator {
 DirBrowser::DirBrowser () : dirTreeModel(),
     dtColumns(),
     tvc(M("DIRBROWSER_FOLDERS")),
-    expandSuccess(false),
-    ignoreExpandAction(false)
+    expandSuccess(false)
 #ifdef WIN32
     , volumes(0)
 #endif
@@ -107,7 +106,7 @@ void DirBrowser::fillDirTree ()
     crt.property_ypad() = 0;
     render_pb->property_ypad() = 0;
 
-    dirtree->signal_row_expanded().connect(sigc::mem_fun(*this, &DirBrowser::row_expanded));
+    rowExpandedConn = dirtree->signal_row_expanded().connect(sigc::mem_fun(*this, &DirBrowser::row_expanded));
     dirtree->signal_row_activated().connect(sigc::mem_fun(*this, &DirBrowser::row_activated));
     dirTreeModel->signal_sort_column_changed().connect(sigc::mem_fun(*this, &DirBrowser::on_sort_column_changed));
 }
@@ -239,9 +238,6 @@ void DirBrowser::on_sort_column_changed() const
 
 void DirBrowser::row_expanded (const Gtk::TreeModel::iterator& iter, const Gtk::TreeModel::Path& path)
 {
-    if(ignoreExpandAction)
-        return;
-
     expandSuccess = false;
 
     // We will disable model's sorting because it decreases speed of inserting new items
@@ -412,9 +408,10 @@ Gtk::TreePath DirBrowser::expandToDir (const Glib::ustring& absDirPath)
     free(dcpy);
 
     path.up ();
-    ignoreExpandAction = true;
-    dirtree->expand_to_path (path);
-    ignoreExpandAction = false;
+    {
+        ConnectionBlocker blocker(rowExpandedConn);
+        dirtree->expand_to_path (path);
+    }
 
     return path;
 }
diff --git a/rtgui/dirbrowser.h b/rtgui/dirbrowser.h
index 5a0505f..e378b7f 100644
--- a/rtgui/dirbrowser.h
+++ b/rtgui/dirbrowser.h
@@ -80,7 +80,7 @@ private:
     Glib::RefPtr<Gdk::Pixbuf> iremovable;
 
     bool expandSuccess;
-    bool ignoreExpandAction;
+    sigc::connection rowExpandedConn;
 
 #ifdef WIN32
     int volumes;
````
However, I think that this does help with readability in this case as someone reading the `row_expanded` definition will not necessary find the related blocking of the invoking connection, but one will always find the explicit return in the current approach. In any case, I'd say the logic of `expandToDir` is rather convoluted and would merit a more broad refactoring even though this is probably not intended in this performance fix.

In any case, whether doing it with the explicit flag or by blocking the connection, the current change seems to have the unintended side effect, that it will add a dummy node called `foo` to the chosen directory node which seems to be caused by `DirBrowser` adding dummy rows in various places to ensure non-empty children sets which are then removed again in `row_expanded` using `dirTreeModel->erase`. Again, I currently see no easy fix, but would suggest a more thorough refactoring of the directory browser code.

heckflosse:
@adamreichold Can you tell me how to reproduce the 'foo'? On my Windows system I don't get 'foo' with my changes. Building on Linux now to check...

adamreichold:
@heckflosse On the Linux build, I did not have to do anything special to get the `foo` node after the changes, but the code is also full of conditional compilation which is one of the reasons I'd suggest refactoring, so it will probably behave quite differently on these two platforms.

heckflosse:
@adamreichold I tested the branch (without your suggested patch) on my Sabayon laptop. Again no 'foo'. Strange...

adamreichold:
@heckflosse It does not seem to be related to how `row_expanded` is blocked but it does seem to happen only if you start up having a directory _without_ subdirectories selected.

heckflosse:
@adamreichold Thanks for the hint. 'Startup' was the key. Reproduced

heckflosse:
@adamreichold I had a look at the foo stuff and indeed a refactoring of the dirbrowser.* code should be done. If you don't mint I would like to push my first commit https://github.com/Beep6581/RawTherapee/commit/26fcc3c80183d0a5e038191efcd9d95a63955fa0 to master. Any objections?

adamreichold:
@heckflosse Yes, the first commit definitely is useful and I am for merging it.

> However, I think that this does help with readability in this case as someone reading the `row_expanded` definition will not necessary find the related blocking of the invoking connection, but one will always find the explicit return in the current approach.

I just noticed that I mistyped there: I meant that the patch does **not** improve readability in this case even though it is idiomatic w.r.t. Gtk+. Otherwise, the second part of the sentence sounds kind of funny. I mean, we should really make stuff harder to find... ;-)

heckflosse:
@adamreichold I know that you mistyped it, because it didn't make sense :)

heckflosse:
@adamreichold what's the standard procedure to merge one commit of a branch into master but not another one? Is there a standard procedure for doing this at all? I lack git experience for this kind of stuff (still have to learn).

adamreichold:
@heckflosse I think you have several options: Interactively rebasing the branch to drop the second commit, merging the hash directly with a branch head associated to it or cherry-picking the commit. Personally, I would opt to cherry-pick the commit since it is a single one without any dependencies on other changes.

heckflosse:
I cherry-picked the first commit

PR #3138, opened 2016-02-06
Coverity - Try to solved iplab2rgb ipwavelet partialpasted retinex

Beep6581:


rtengine/iplab2rgb.cc
```diff
@@ -386,6 +386,14 @@ Image16* ImProcFunctions::lab2rgb16b (LabImage* lab, int cx, int cy, int cw, int
 
     Image16* image = new Image16 (cw, ch);
     float p1, p2, p3, p4, p5, p6; //primaries
+    //default primaries
```
adamreichold: I think making the ProPhoto branch of the if-else-if-else statements into a pure else so that we always take one of these branches would be preferable to duplicating these values.

rtgui/partialpastedlg.h
```diff
@@ -67,7 +67,7 @@ class PartialPasteDlg : public Gtk::Dialog
 
     // options in color:
     Gtk::CheckButton* icm;
-    Gtk::CheckButton* gam;
```
adamreichold: Since version control takes care of reviving stuff, I suggest removing the variable instead of commenting it out if it is not used anymore.

rtengine/ipwavelet.cc
```diff
@@ -2032,7 +2032,8 @@ void ImProcFunctions::WaveletcontAllL(LabImage * labco, float ** varhue, float *
 
                         if(cp.lip3 && cp.lipp) {
                             // comparaison betwwen pixel and neighbours
-                            float kneigh, somm;
```
adamreichold: Again, making taking one of the initialization branches seems preferable. Rewriting this using the ternary `?:` operation and giving those variables the type `const auto` might be advisable as well.

rtengine/ipwavelet.cc
```diff
@@ -3191,7 +3192,9 @@ void ImProcFunctions::ContAllL (float *koeLi[12], float *maxkoeLi, bool lipschit
             float scale = 1.f;
             float scale2 = 1.f;
 
-            float LL100, LL100res, LL100init, kH[maxlvl];
```
adamreichold: Which Coverity defect is intended to be fixed here? Did you target this one?
````
CID 157589 (#1 of 1): Uninitialized scalar variable (UNINIT)271. uninit_use: Using uninitialized value kH[level].
````
This at least is not solved by allocating the memory dynamically which will only slow things down and lead to new bugs if it isn't freed properly.

Also the Coverity report is a bit tricky here IMHO: It tells us that if `maxlvl == 0`, the variable `kH` will never be set because the for loop initializing it will never be entered. But then again it will be an array of length zero and there will be nothing to initialize inside. In summary, I think the correct fix would be to ensure that the parameters `level` and `maxlvl` make sense, e.g. fulfill relations like `0 <= level < maxlvl` immediately upon entering the function.

Desmis:
@adamreichold 

Please, can you make these changes. Thank you :)

adamreichold:
@Desmis I made the necessary adjustments and tried to clean up some other stuff that looked problematic. Please retest.

While this might be a delicate subject, I feel I need to add a more general comment: I made the changes because I care strongly about the long-term maintainability of this code base. Hence I find it rather irresponsible if you contribute code to it and then do not try to fix these problems yourself. To me, it feels like you sometimes do not really care about the more technical parts of working on a large-scale computer program.
Therefore I would like to ask that you invest as much effort into the engineering and maintenance aspects of this project as into the underlying algorithms. Both aspects pose difficult problems that need solutions if this program is to be useful for a lot of people and a long time.

Desmis:
I understand and agree on the principle, and thank you for your work :)

I just tested, and all seems to work well.

And another time, excuse my bad english! And I hope (with this bad english) no misunderstandings.

However, and this is what I have already said many times, I have no computer training, and many changes you just made for me are "dark" (I do not see too much difference, and if you say it's better, then it's true).

Moreover, as I also said, I am "messy", looking for algorithms that work, and I do not always do the household behind. (There is the code you touched that I did not participate at all!).
Fortunately for some time, there is "Astylert" that allows a standard presentation (but sure do not correct code).

I do not think that whether or not a willingness to invest, but a question of skills and learning; if I asked you your participation is that in some cases I do not know how (see my remarks on collective learning and social regulation...still no reply or response).

On the other hand, for some time, and this is not specifically for you, the new algorithms that I propose, are not tested or whether quickly or superficially that results in caricatures.

Since the episode of "merge Retinex" earlier this year, I feel like a leper...but, believe it or not, there was no will to do evil or anything like that, but just get something to solve complex (the amount of work that represented the update of the 3 branches was huge for me...perhaps there is other way..but I don't know).

Of course, you made me notice that there was an ongoing "PR", certainly I looked on the net (for PR), but precisely, I do not see clearly how it works...including this PR!!


All this disturbs me (despite appearances) much, disturbs my sleep, whereas this development should be a hobby (it was until recent months), for an "old man" (I am near of 70).
I am a scientist, not a computer. In some cases, I feel like a monkey, copying parts of code! A project as "RT" must have operating rules-based regulation, collective learning process, and necessary cooperation

I ask myself to abandon the project, or at least put aside for a while.

Thank you for this sincerity that I like

adamreichold:
> Moreover, as I also said, I am "messy", looking for algorithms that work, and I do not always do the household behind. (There is the code you touched that I did not participate at all!).
Fortunately for some time, there is "Astylert" that allows a standard presentation (but sure do not correct code).

This is why we do code review. Everybody is messy to a degree, i.e. most programmers allow themselves to cut corners, but will immediately criticize the same behavior when reviewing code written by someone else.

> I do not think that whether or not a willingness to invest, but a question of skills and learning; if I asked you your participation is that in some cases I do not know how (see my remarks on collective learning and social regulation...still no reply or response).

Nobody said you are on your own here, we are all learning one thing or the other. If something is unclear, just ask and the person in question will try to explain itself. In my personal experience, someone who cannot explain himself has not really understood his subject.

> All this disturbs me (despite appearances) much, disturbs my sleep, whereas this development should be a hobby (it was until recent months), for an "old man" (I am near of 70).
I am a scientist, not a computer. In some cases, I feel like a monkey, copying parts of code! A project as "RT" must have operating rules-based regulation, collective learning process, and necessary cooperation

Since this is spare time effort for probably all of us, there is no pressure involved and nobody needs to lose sleep over anything we do around here. This pull request (just a fancy name for a process of code review, i.e. peer review for source code) can take years and nobody could force us to hasten things. We have the time to explain ourselves and we have the time to try again if we are unhappy with current results.

However, a rule I would derive from this situation would be that if you want to merge changes and somebody reviews them, then either you falsify their critique or fix what they remarked. (This sounds pretty close to publishing a paper to me: If there is a hole in a proof you want to publish, you either fix it or withdraw the submission.)

> I ask myself to abandon the project, or at least put aside for a while.

I think that would be the worst possible outcome and certainly not what I intended here. I would rather suggest we try to find a mode of working together that benefits the project.

Trying to work out if and how we want to make use of pull requests seems to be good starting point. Even though I would personally prefer just doing a lot of pull requests and figuring it out along way to formulating a fixed set of rules beforehand.

adamreichold:
@Beep6581 As anything else would be quite schizophrenic by now, I am for merging... ;-) Are you satisfied with the review?

Beep6581:
Immensely :)

Beep6581:
@Desmis PR = Pull Request, this page explains it: https://help.github.com/articles/using-pull-requests/

Desmis:
@Beep6581 
I'll read, thank you :)

PR #3139, opened 2016-02-07
Remove freeArray2 helper function

adamreichold:
This helper function is currently used incorrectly AFAIU and we are replacing array of arrays temporary by jagged arrays wherever we find them, so this branch helps these two aims by adding a new `JaggedArray<T>` helper type to allocate such a structure that will be automatically freed, is noncopyable and converts to `T**` implicitly. This will hopefully help us to clean up various temporary buffer allocations in the future.

### DIFF
--- a/rtengine/demosaic_algos.cc
+++ b/rtengine/demosaic_algos.cc
@@ -21,6 +21,7 @@
 
 #include "rawimagesource.h"
 #include "rawimagesource_i.h"
+#include "jaggedarray.h"
 #include "median.h"
 #include "rawimage.h"
 #include "mytime.h"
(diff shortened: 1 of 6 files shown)


adamreichold:
Rebased to resolve merge conflict with D1x fixes...

heckflosse:
:+1: for merge. But in case we want to use it at more places in RT we may need a fallback solution like the one implemented [here](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/cieimage.cc#L2) too

adamreichold:
@heckflosse I agree about the fallback for possibly very large allocations, but I would probably want to put that in a separate class to keep `JaggedArray<T>` as simple as possible. I made a note in any case...

PR #3142, opened 2016-02-08
Modified Preview Canvas

adamreichold:
- Now the Preview can show free space around the image (the image's
corner will coincide with the center of the preview area)
- Editing objects can now be manipulated in this free space
- The editing mechanism has been split : it was completely handled in
rtengine before, now rtengine still handle the pipette's data provider,
but rtgui now handle the objects data provider.
- Bugfix: when using coarse rotate in the Editor panel, the Gradient
widgets are now correctly displayed

### DIFF
--- a/rtgui/edit.h
+++ b/rtgui/edit.h
@@ -153,8 +153,6 @@ private:
 
     // Used to draw the objects where the color correspond to the object's ID, in order to find the correct object when hovering
     Cairo::RefPtr<Cairo::ImageSurface> objectMap;
-    // If more than 254 objects has to be handled, objectMap2 contains the "upper part" of the 16 bit int value. objectMap2 will be NULL otherwise.
-    Cairo::RefPtr<Cairo::ImageSurface> objectMap2;
     ObjectMode objectMode;
 
 protected:
@@ -164,32 +162,19 @@ protected:
     EditDataProvider* dataProvider;
 
     void createBuffer(int width, int height);
-    void resize(int newWidth, int newHeight, EditSubscriber* newSubscriber);
+    void resize(int newWidth, int newHeight);
     void flush();
     EditSubscriber *getEditSubscriber ();
 
 public:
-    ObjectMOBuffer(EditDataProvider *dataProvider);
+    explicit ObjectMOBuffer (EditDataProvider *dataProvider);
     ~ObjectMOBuffer();
 
-    EditDataProvider* getDataProvider()
-    {
-        return dataProvider;
-    }
-    void setObjectMode(ObjectMode newType);
-    ObjectMode getObjectMode()
-    {
-        return objectMode;
-    }
+    EditDataProvider* getDataProvider ();
+    void setObjectMode (ObjectMode newType);
+    ObjectMode getObjectMode ();
 
-    Cairo::RefPtr<Cairo::ImageSurface> &getObjectMap ()
-    {
-        return objectMap;
-    }
-    Cairo::RefPtr<Cairo::ImageSurface> &getObjectMap2()
-    {
-        return objectMap2;
-    }
+    Cairo::RefPtr<Cairo::ImageSurface>& getObjectMap ();
 
     // return true if the buffer has been allocated
     bool bufferCreated();
@@ -238,36 +223,16 @@ class RGBColor
     double b;
 
 public:
-    RGBColor () : r(0.), g(0.), b(0.) {}
-    explicit RGBColor (double r, double g, double b) : r(r), g(g), b(b) {}
-    explicit RGBColor (char r, char g, char b) : r(double(r) / 255.), g(double(g) / 255.), b(double(b) / 255.) {}
+    RGBColor ();
+    explicit RGBColor (double r, double g, double b);
+    explicit RGBColor (char r, char g, char b);
 
-    void setColor(double r, double g, double b)
-    {
-        this->r = r;
-        this->g = g;
-        this->b = b;
-    }
+    void setColor (double r, double g, double b);
+    void setColor (char r, char g, char b);
 
-    void setColor(char r, char g, char b)
-    {
-        this->r = double(r) / 255.;
-        this->g = double(g) / 255.;
-        this->b = double(b) / 255.;
-    }
-
-    double getR()
-    {
-        return r;
-    }
-    double getG()
-    {
-        return g;
-    }
-    double getB()
-    {
-        return b;
-    }
+    double getR ();
+    double getG ();
+    double getB ();
 };
 
 class RGBAColor : public RGBColor
@@ -275,26 +240,14 @@ class RGBAColor : public RGBColor
     double a;
 
 public:
-    RGBAColor () : RGBColor(0., 0., 0.), a(0.) {}
-    explicit RGBAColor (double r, double g, double b, double a) : RGBColor(r, g, b), a(a) {}
-    explicit RGBAColor (char r, char g, char b, char a) : RGBColor(r, g, b), a(double(a) / 255.) {}
+    RGBAColor ();
+    explicit RGBAColor (double r, double g, double b, double a);
+    explicit RGBAColor (char r, char g, char b, char a);
 
-    void setColor(double r, double g, double b, double a)
-    {
-        RGBColor::setColor(r, g, b);
-        this->a = a;
-    }
+    void setColor (double r, double g, double b, double a);
+    void setColor (char r, char g, char b, char a);
 
-    void setColor(char r, char g, char b, char a)
-    {
-        RGBColor::setColor(r, g, b);
-        this->a = double(a) / 255.;
-    }
-
-    double getA()
-    {
-        return a;
-    }
+    double getA ();
 };
 
 /// @brief Displayable and MouseOver geometry base class
@@ -345,85 +298,29 @@ public:
     Datum datum;
     State state;  // set by the Subscriber
 
-    Geometry () : innerLineColor(char(255), char(255), char(255)), outerLineColor(char(0), char(0), char(0)), flags(F_VISIBLE | F_HOVERABLE | F_AUTO_COLOR), innerLineWidth(1.5f), datum(IMAGE), state(NORMAL) {}
+    Geometry ();
     virtual ~Geometry() {}
 
-    void     setInnerLineColor     (double r, double g, double b)
-    {
-        innerLineColor.setColor(r, g, b);
-        flags &= ~F_AUTO_COLOR;
-    }
-    void     setInnerLineColor     (char   r, char   g, char   b)
-    {
-        innerLineColor.setColor(r, g, b);
-        flags &= ~F_AUTO_COLOR;
-    }
-    RGBColor getInnerLineColor     ();
-    void     setOuterLineColor     (double r, double g, double b)
-    {
-        outerLineColor.setColor(r, g, b);
-        flags &= ~F_AUTO_COLOR;
-    }
-    void     setOuterLineColor     (char   r, char   g, char   b)
-    {
-        outerLineColor.setColor(r, g, b);
-        flags &= ~F_AUTO_COLOR;
-    }
-    RGBColor getOuterLineColor     ();
-    double   getOuterLineWidth     ()
-    {
-        return double(innerLineWidth) + 2.;
-    }
-    double   getMouseOverLineWidth ()
-    {
-        return getOuterLineWidth() + 2.;
-    }
-    void     setAutoColor          (bool aColor)
-    {
-        if (aColor) {
-            flags |= F_AUTO_COLOR;
-        } else {
-            flags &= ~F_AUTO_COLOR;
-        }
-    }
-    bool     isVisible             ()
-    {
-        return flags & F_VISIBLE;
-    }
-    void     setVisible            (bool visible)
-    {
-        if (visible) {
-            flags |= F_VISIBLE;
-        } else {
-            flags &= ~F_VISIBLE;
-        }
-    }
-    bool     isHoverable           ()
-    {
-        return flags & F_HOVERABLE;
-    }
-    void     setHoverable          (bool visible)
-    {
-        if (visible) {
-            flags |= F_HOVERABLE;
-        } else {
-            flags &= ~F_HOVERABLE;
-        }
-    }
+    void setInnerLineColor (double r, double g, double b);
+    void setInnerLineColor (char r, char g, char b);
+    RGBColor getInnerLineColor ();
+    void setOuterLineColor (double r, double g, double b);
+    void setOuterLineColor (char r, char g, char b);
+    RGBColor getOuterLineColor ();
+    double getOuterLineWidth ();
+    double getMouseOverLineWidth ();
+    void setAutoColor (bool aColor);
+    bool isVisible ();
+    void setVisible (bool visible);
+    bool isHoverable ();
+    void setHoverable (bool visible);
 
     // setActive will enable/disable the visible and hoverable flags in one shot!
-    void     setActive             (bool active)
-    {
-        if (active) {
-            flags |= (F_VISIBLE | F_HOVERABLE);
-        } else {
-            flags &= ~(F_VISIBLE | F_HOVERABLE);
-        }
-    }
+    void setActive (bool active);
 
     virtual void drawOuterGeometry (Cairo::RefPtr<Cairo::Context> &cr, ObjectMOBuffer *parent, EditCoordSystem &coordSystem) = 0;
     virtual void drawInnerGeometry (Cairo::RefPtr<Cairo::Context> &cr, ObjectMOBuffer *parent, EditCoordSystem &coordSystem) = 0;
-    virtual void drawToMOChannel   (Cairo::RefPtr<Cairo::Context> &cr, Cairo::RefPtr<Cairo::Context> &cr2, unsigned short id, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem) = 0;
+    virtual void drawToMOChannel   (Cairo::RefPtr<Cairo::Context> &cr, unsigned short id, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem) = 0;
 };
 
 class Circle : public Geometry
@@ -434,13 +331,13 @@ public:
     bool filled;
     bool radiusInImageSpace; /// If true, the radius depend on the image scale; if false, it is a fixed 'screen' size
 
-    Circle () : center(100, 100), radius(10), filled(false), radiusInImageSpace(false) {}
-    Circle (rtengine::Coord &center, int radius, bool filled = false, bool radiusInImageSpace = false) : center(center), radius(radius), filled(filled), radiusInImageSpace(radiusInImageSpace) {}
-    Circle (int centerX, int centerY, int radius, bool filled = false, bool radiusInImageSpace = false) : center(centerX, centerY), radius(radius), filled(filled), radiusInImageSpace(radiusInImageSpace) {}
+    Circle ();
+    Circle (rtengine::Coord& center, int radius, bool filled = false, bool radiusInImageSpace = false);
+    Circle (int centerX, int centerY, int radius, bool filled = false, bool radiusInImageSpace = false);
 
     void drawOuterGeometry (Cairo::RefPtr<Cairo::Context> &cr, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem);
     void drawInnerGeometry (Cairo::RefPtr<Cairo::Context> &cr, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem);
-    void drawToMOChannel   (Cairo::RefPtr<Cairo::Context> &cr, Cairo::RefPtr<Cairo::Context> &cr2, unsigned short id, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem);
+    void drawToMOChannel   (Cairo::RefPtr<Cairo::Context> &cr, unsigned short id, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem);
 };
 
 class Line : public Geometry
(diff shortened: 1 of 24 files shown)


rtgui/toolpanelcoord.cc
```diff
@@ -336,6 +336,24 @@ void ToolPanelCoordinator::panelChanged (rtengine::ProcEvent event, const Glib::
...
+        tr = TR_R180;
+    } else if (params->coarse.rotate == 270) {
+        tr = TR_R270;
+    }
+
+    // Update "on preview" geometry
+    if (event == rtengine::EvPhotoLoaded || event == rtengine::EvProfileChanged || event == rtengine::EvHistoryBrowsed || event == rtengine::EvCTRotate) {
+        // updating the "on preview" geometry
+        int fw, fh;
+        rtengine::ImageSource *ii = (rtengine::ImageSource*)ipc->getInitialImage();
```
adamreichold: Not sure what is going on: Either the return type of `StagedImageProcessor::getInitialImage` is wrong or we can't actually assume that the initial image will always be an instance of `ImageSource`. In any case, we should not use a C-style cast. Since it is down casting, we should probably use `dynamic_cast`, e.g.
````
if (const auto ii = dynamic_cast<rtengine::ImageSource*>(ipc->getInitialImage())) { /* ... */ }
````
but as I said, if we are sure about the type, the return type of the method should probably be changed. (At least the only implementation seems to return an instance of `ImageSource`.)

File: rtgui/previewwindow.cc [hunk 2]
```diff
@@ -199,6 +199,8 @@ bool PreviewWindow::on_motion_notify_event (GdkEventMotion* event)
 
         if (isMoving) {
             mainCropWin->remoteMove ((event->x - press_x) / zoom, (event->y - press_y) / zoom);
```
adamreichold: Just out of curiosity: Was this a bug? Could I notice that this was missing in the current `master` branch?

rtgui/previewhandler.cc
```diff
@@ -204,27 +204,19 @@ Glib::RefPtr<Gdk::Pixbuf> PreviewHandler::getRoughImage (int x, int y, int w, in
             h = image->getHeight() * totalZoom;
         }
 
-        int ix = x * zoom;
-        int iy = y * zoom;
+        x *= zoom;
+        y *= zoom;
 
-        if (ix < 0) {
```
adamreichold: Don't we need to check for `x < 0` and `y < 0` anymore? `x = std::max(x, 0)` and `y = std::max(y, 0)` should do the trick.

rtgui/previewhandler.cc
```diff
@@ -204,27 +204,19 @@ Glib::RefPtr<Gdk::Pixbuf> PreviewHandler::getRoughImage (int x, int y, int w, in
...
-            ix = 0;
+        if ((x + w) / totalZoom > previewImg->get_width()) {
```
adamreichold: I would try to avoid depending on algebraic identities with floating-point arithmetic, i.e. I  would prefer something like
````
w = std::min(w, previewImage->get_width() * totalZoom - x);
````
or even
````
w = std::min(w, std::max(previewImage->get_width() * totalZoom - x, 0));
````
if we are being pedantic.

File: rtgui/imagearea.cc [hunk 5]
```diff
@@ -618,19 +611,18 @@ void ImageArea::initialImageArrived (CropWindow* cw)
...
         } else {
             int w, h;
             mainCropWindow->cropHandler.getFullImageSize(w, h);
 
-            if(w == fullImageWidth && h == fullImageHeight) { // && mainCropWindow->getZoom() != mainCropWindow->getZoomFitVal()) {
-                int x, y;
-                mainCropWindow->getCropPosition(x, y);
-                mainCropWindow->setCropPosition(x, y, false);
+            if(w != fullImageWidth || h != fullImageHeight) {
+                printf("zoomFit !\n");
```
adamreichold: Probably left over from debugging? (More generally: `printf`-style debugging has its uses, but generally shouldn't break points do this much better?)

rtgui/imagearea.cc
```diff
@@ -170,10 +169,11 @@ CropWindow* ImageArea::getCropWindow (int x, int y)
 
     CropWindow* cw = mainCropWindow;
 
-    for (std::list<CropWindow*>::iterator i = cropWins.begin(); i != cropWins.end(); i++)
-        if ((*i)->isInside (x, y)) {
-            return *i;
+    for (auto cropWin : cropWins) {
```
adamreichold: You definitely like the ranged for loop syntax. Well, I like it too! :-D

rtgui/gradient.h
```diff
@@ -42,12 +42,11 @@ class Gradient : public ToolParamBlock, public AdjusterListener, public Foldable
     void setDefaults    (const rtengine::procparams::ProcParams* defParams, const ParamsEdited* pedited = NULL);
     void setBatchMode   (bool batchMode);
 
-    void updateGeometry (int centerX_, int centerY_, double feather_, double degree_);
-
     void adjusterChanged (Adjuster* a, double newval);
     void enabledChanged  ();
     void setAdjusterBehavior (bool degreeadd, bool featheradd, bool strengthadd, bool centeradd);
     void trimValues          (rtengine::procparams::ProcParams* pp);
+    void updateGeometry  (const int centerX_, const int centerY_, const double feather_, const double degree_, const int fullWidth=-1, const int fullHeight=-1);
```
adamreichold: Should we resolve the name collision of the `centerX` parameter with the `centerX` member using explicit `this->centerX` where necessary inside the methods definition?

rtgui/gradient.cc
```diff
@@ -136,88 +136,99 @@ void Gradient::read (const ProcParams* pp, const ParamsEdited* pedited)
...
-        currLine->begin += origin;
-        polCoord1.set(350.f, float(-degree_    ));
-        currLine->end.setFromPolar  (polCoord1 + polCoord2);
-        currLine->end   += origin;
-        // update circle's position
-        currCircle = static_cast<Circle*>(visibleGeometry.at(4));
-        currCircle->center = origin;
-        currCircle = static_cast<Circle*>(mouseOverGeometry.at(4));
-        currCircle->center = origin;
+        if (!imW || !imH) {
```
adamreichold: I know `ImageArea::getImageSize` nulls them, but I would feel better if `imW` and `imH` were properly initialized to zero, especially since `getImageSize` is a virtual method and might collect further overrides.

rtgui/gradient.cc
```diff
@@ -136,88 +136,99 @@ void Gradient::read (const ProcParams* pp, const ParamsEdited* pedited)
...
+            return;
+        }
     }
+
+    PolarCoord polCoord1, polCoord2;
+    double decay = feather_ * sqrt(double(imW) * double(imW) + double(imH) * double(imH)) / 200.;
```
adamreichold: This looks like I would love be of type `const auto`? Also `sqrt(x * x + y * y)` could probaly be a useful helper function? Maybe something for `rt_math.h` like
````
template<typename T>
T norm1(const T& x, const T& y)
{
    return std::abs(x) + std::abs(y);
}
template<typename T>
T norm2(const T& x, const T& y)
{
    return std::sqrt(x * x + y * y);
}
template< typename T >
T norminf(const T& x, const T& y)
{
}
```` 
which can be called like
````
const auto decay = feather_ * rtengine::norm2<double>(imW, imH) / 200.0;
````
(Please also note that constructor-style casts like `T(x)` are basically equivalent to C-style casts, i.e. equally problematic.)

File: rtgui/gradient.cc [hunk 10]
```diff
@@ -399,7 +410,7 @@ bool Gradient::button1Pressed(int modifierKey)
 {
     EditDataProvider *provider = getEditProvider();
 
-    if (!(modifierKey & GDK_CONTROL_MASK)) {
+    if (!(modifierKey & GDK_CONTROL_MASK) && lastObject > -1) {
```
adamreichold: If `lastObject > - 1` should always hold, I think it would be preferable to have
````
if (lastObject < 0) {
    return;
}
````
or even
````
assert(lastObject > -1);
````
at the beginning of the method.

File: rtgui/gradient.cc [hunk 11]
```diff
@@ -136,88 +136,99 @@ void Gradient::read (const ProcParams* pp, const ParamsEdited* pedited)
...
+    currLine->begin.setFromPolar(polCoord1 + polCoord2);
+    currLine->begin += origin;
+    polCoord1.set(350.f, float(-degree_    ));
+    currLine->end.setFromPolar  (polCoord1 + polCoord2);
+    currLine->end   += origin;
+    // update circle's position
+    currCircle = static_cast<Circle*>(visibleGeometry.at(4));
+    currCircle->center = origin;
+    currCircle = static_cast<Circle*>(mouseOverGeometry.at(4));
+    currCircle->center = origin;
```
adamreichold: Obviously a lot of duplication here, I am not sure if you prefer it since it is a bit more indirect, but one way to reuse parts of the code is to declare local lambda expressions like
````
    const auto updateLine = [&](Geometry* geometry, const double radius, const double begin, const double end)
    {
        const auto line = static_cast<Line*>(geometry);
        line->begin.setFromPolar(PolarCoord(radius, -degree_ + begin));
        line->begin += origin;
        line->end.setFromPolar(PolarCoord(radius, -degree_ + end));
        line->end += origin;
    };

    const auto updateLineWithDecay = [&](Geometry* geometry, const double radius, const double begin, const double end)
    {
        const auto line = static_cast<Line*>(geometry);
        line->begin.setFromPolar(PolarCoord(radius, -degree_ + begin) + PolarCoord(decay, -degree_ + end));
        line->begin += origin;
        line->end.setFromPolar(PolarCoord(radius, -degree_) + PolarCoord(decay, -degree_ + end));
        line->end += origin;
    };

    const auto updateCircle = [&](Geometry* geometry)
    {
        const auto circle = static_cast<Circle*>(geometry);
        circle->center += origin;
    };

    // update horizontal line
    updateLine (visibleGeometry.at(0), 1500.f, 180.f, 0.f);
    updateLine (mouseOverGeometry.at(0), 1500.f, 180.f, 0.f);
    // update vertical line
    updateLine (visibleGeometry.at(1), 700.f, 90.f, 270.f);
    updateLine (mouseOverGeometry.at(1), 700.f, 90.f, 270.f);
    // update upper feather line
    updateLineWithDecay (visibleGeometry.at(2), 350.f, 180.f, 270.f);
    updateLineWithDecay (mouseOverGeometry.at(2), 350.f, 180.f, 270.f);
    // update lower feather line
    updateLineWithDecay (visibleGeometry.at(3), 350.f, 180.f, 90.f);
    updateLineWithDecay (mouseOverGeometry.at(3), 350.f, 180.f, 90.f);
    // update circle's position
    updateCircle (visibleGeometry.at(4));
    updateCircle (mouseOverGeometry.at(4));
````
and let the compiler take care of inlining them and optimizing away the temporaries. (For this to work, it is necessary to fix `PolarCoord` at least like this
````
diff --git a/rtengine/coord.h b/rtengine/coord.h
index 0b134ce..359fb3a 100644
--- a/rtengine/coord.h
+++ b/rtengine/coord.h
@@ -187,23 +187,23 @@ public:
     {
         radius *= scale;
     }
-    PolarCoord operator+(PolarCoord & rhs)
+    PolarCoord operator+ (const PolarCoord& rhs) const
     {
+        PolarCoord result;
         Coord thisCoord, rhsCoord;
         thisCoord.setFromPolar(*this);
         rhsCoord.setFromPolar(rhs);
         thisCoord += rhsCoord;
-        PolarCoord result;
         result.setFromCartesian(thisCoord);
         return result;
     }
-    PolarCoord operator-(PolarCoord & rhs)
+    PolarCoord operator- (const PolarCoord& rhs) const
     {
+        PolarCoord result;
         Coord thisCoord, rhsCoord;
         thisCoord.setFromPolar(*this);
         rhsCoord.setFromPolar(rhs);
         thisCoord -= rhsCoord;
-        PolarCoord result;
         result.setFromCartesian(thisCoord);
         return result;
     }
```
but that class does not follow best practices in any case for which I'll make a note.)

rtgui/edit.h
```diff
@@ -152,6 +145,59 @@ class EditBuffer;
...
+    // instead of pointing to the EditSubscriber directly
+    EditDataProvider* dataProvider;
+
+    void createBuffer(int width, int height);
+    void resize(int newWidth, int newHeight, EditSubscriber* newSubscriber);
+    void flush();
+    EditSubscriber *getEditSubscriber ();
+
+public:
+    ObjectMOBuffer(EditDataProvider *dataProvider);
```
adamreichold: Single argument constructors should usually be declared `explicit` to avoid unintended implicit conversions.

rtgui/edit.h
```diff
@@ -152,6 +145,59 @@ class EditBuffer;
...
+    ~ObjectMOBuffer();
+
+    EditDataProvider* getDataProvider()
```
adamreichold: Personally, I find it preferable to move even inline method definitions outside of the class body (and add the `inline` modifier) so that a class contains only declarations a one can have a better overview. (Most IDE offer that as an automatic refactoring nowadays. Move definition out of class body or something like it.)

rtgui/edit.h
```diff
@@ -152,6 +145,59 @@ class EditBuffer;
...
+    {
+        return dataProvider;
+    }
+    void setObjectMode(ObjectMode newType);
+    ObjectMode getObjectMode()
+    {
+        return objectMode;
+    }
+
+    Cairo::RefPtr<Cairo::ImageSurface> &getObjectMap ()
```
adamreichold: If everybody has non-const reference access to the field, there is no use in declaring a getter since its type will also be unchangeable later on. Either access should be const or the fields can just become public as well.

rtgui/edit.h
```diff
@@ -152,6 +145,59 @@ class EditBuffer;
...
+    ObjectMode objectMode;
+
+protected:
+
+    // To avoid duplicated information, we points to a EditDataProvider that contains the current EditSubscriber
+    // instead of pointing to the EditSubscriber directly
+    EditDataProvider* dataProvider;
+
+    void createBuffer(int width, int height);
+    void resize(int newWidth, int newHeight, EditSubscriber* newSubscriber);
```
adamreichold: The third parameter looks like it wants a separate method. Even if a size and subscriber change often happen together, it looks dubious IMHO.

rtgui/edit.h
```diff
@@ -152,6 +145,59 @@ class EditBuffer;
  *
  */
 
+
+
+class ObjectMOBuffer
+{
+private:
+
+    // Used to draw the objects where the color correspond to the object's ID, in order to find the correct object when hovering
```
adamreichold: This sounds interesting! It is almost certainly out of scope for this change, but isn't this representation relatively wasteful? I.e. what about using an polymorphic `contains` method that checks each geometry item in reverse drawing order if it contains the given coordinate? Should be relatively straight-forward for most geometric primitives like rectangles, circles and polygons and could probably be cached as well. In any case, why not upgrade the surface format instead of drawing twice to a second one if more objects need to be tracked, e.g. using `FORMAT_RGB16_565` and transforming the object identifier into a suitable color?

adamreichold:
@Hombre57 Added some technical comments after beginning to read through from the end. Will hopefully be able to continue during the week. Haven't done any proper testing yet.

rtgui/toolpanelcoord.cc
Hombre57: You're right. I've found an even better solution, replacing :
```
rtengine::ImageSource *ii = (rtengine::ImageSource*)ipc->getInitialImage()->getImageSource();
ii->getFullSize (fw, fh, tr);
```
by
```
ipc->getInitialImage()->getImageSource()->getFullSize (fw, fh, tr);
```
Ok, it doesn't check for NULL pointers, but we're assuming that it's non-NULL.

rtgui/previewwindow.cc
Hombre57: Nope, it wasn't a bug but a change for the new mechanism, since it works incrementally now, it has to set the new 'press' position at each move.

rtgui/previewhandler.cc
Hombre57: No, we don't need to check that anymore. `x` and `y` are `cropX` an `cropY` from the caller which are obtained through `cropHandler.getPosition`. The returned values (by reference) of this function can't be negative (`CropHandler::cropX` and `CropHandler::cropY` are trimmed in `CropHandler::compDim`).

rtgui/previewhandler.cc
Hombre57: Ok, replaced by
```
w = rtengine::LIM<int>(w, 0, int(double(previewImg->get_width()) * totalZoom) - x);
```
I hope you don't mind the manual cast, it let me sleep at night better ;)

rtgui/imagearea.cc
Hombre57: Yes, it's a left over, deleted.

As you said, printf has its uses. I'm not very comfortable with debugging on Windows using Linux tools. It's slow, and sometimes hang.

rtgui/imagearea.cc
Hombre57: iterator, with templates, and the 2 things that I'm not used to at all in C++, so this syntax is very convenient.

For the `auto` type, it's certainly useful (especially here), but one of the downside is that you don't know the type of your variable by visually inspecting the code. Hopefully Eclipse show us the real type when hovering the variable.

rtgui/gradient.h
Hombre57: Done.

rtgui/gradient.cc
Hombre57: Done. Should I initialize all parameters that I know will be set in a function, even non virtual ones ?

rtgui/gradient.cc
Hombre57: Methods added, thanks for the hint.

Why are this kind of cast problematic for basic types ? How do you cast then ?

rtgui/gradient.cc
Hombre57: Done

rtgui/edit.h
Hombre57: My intention was initially to handle lots of complex shape like curves with different thickness on screen. As you might have understood or seen, the MouseOver objects are usually thicker than their visual representation. Of course, we can still handle the thickness difference for basic shapes, but I don't know how to deal with curves or with complex shapes like polylines.

I don't think that the memory consumption of this mechanism should be a concern, I'm more concerned by the performance and the versatility of this tool. Updating the ObjectMOBuffer takes time, then we only read the value under the cursor. But in your opinion, should we switch to the `contains` method ?

About using `FORMAT_RGB16_565`, yes, that should be better. Like the dumb guy I am, I was looking for an FORMAT_A16, and found a silly trick to circumvent its absence :)

rtgui/gradient.cc
Hombre57: Done. This use case is very instructive.

rtgui/edit.h
Hombre57: I thought that C++ was a strongly typed language. I can understand for basic types, but how can a pointer be implicitly converted without stopping the build process ?

rtgui/toolpanelcoord.cc
adamreichold: Agreed, definitely better than any cast. And assume non-null is usually alright since dereferencing null pointers will give a "clean and understandable" crash.

rtgui/previewhandler.cc
adamreichold: I see. (Asserts might still be useful to get a clean crash in debugging builds when this invariant is violated due to an error in other parts of the code, but it is not strictly necessary.)

rtgui/previewhandler.cc
adamreichold: I don't mind the casts, but the inner one should not be necessary as `int * double` should automatically result in `double`.

rtgui/imagearea.cc
adamreichold: Yeah, that is the reason that for example Java explicitly decided against something like `auto`. AFAIK, the idea of the C++ standard committee was that the concrete type shouldn't matter to the programmer. It should even change automatically if the right hand side of the assignment changes. If there is a problem, it should be caught later on due to type constraints.
At least with `auto` the will never be narrowing, loss of precision or unintended wasteful implicit casts. But for example, for me personally using `auto` somehow felt wrong in the beginning and only after using it for a while I began to feel comfortable about it.

rtgui/gradient.cc
adamreichold: The performance impact should be minimal and since it does reduce coupling since we don't need to inspect both parts at once, I would recommend always initializing yes. (Usually I would recommend returning `std::pair` or `std::tuple` even more, so that one can just write `const auto size = dataProvider->getImageSize();`. If it is used often enough, a simple `struct Size { int width; int height; };` might also be useful. Returning values by reference or pointer should usually not be necessary.)

adamreichold: C++11 can also be pretty Pythonesque, e.g. one can write
````
int imH, imW;
std::tie(imH, imW) = dataProvider->getImageSize();
````
to unpack a `std::tuple` into individual values with no runtime overhead.

rtgui/gradient.cc
adamreichold: The C-style and similarly constructor style casts are usually considered problematic because they can mean a lot of things depending on the types involved while the new style casts better specify intent, i.e. you only want some kind of conversion but not another, e.g. C-style casts will go as far as a `reinterpret_cast` and just a different view on the same memory even if you intented a value conversion.
With the primitive types, it is usually alright since there are no questions of memory layout and class hierarchies involved, but as with most rules, it is better to never use them instead of having to carefully examine each case. (Consider for example the huge difference between a primitive type and a pointer to a primitive type.) So they would usually be cast using a `static_cast`, i.e. a conversion that is fully determined a compile time.
But actually, I think the most important rule is to never cast at all, i.e. depend on automatic widening where applicable and use helper functions that specify intent where you want to narrow, i.e. if I want to drop the fractional part by truncation and do not need to handle overflow, i.e. one could define something like
````
int truncate(const float x)
{
   assert(x > std::numeric_limits<int>::min() && x < std::numeric_limits<int>::max());
   return static_cast<int>(x);
}
````
and use it instead of a plain cast.

rtgui/edit.h
adamreichold: The problem is again that it is problematic as the intent is unclear and the code becomes hard to follow, e.g. with the implicit constructor something like
````
buffer = dataProvider;
````
with `buffer` being of type `ObjectMOBuffer` is legal and would implicitly create a new instance of `ObjectMOBuffer` and assign it to `buffer`. At least to me, it would be very much unclear if that was actually intended or rather is a typo.
Implicitly constructing objects with a complicated life cycle that containing other objects is usually a bad thing. Implicit constructors of course do make sense in other places, e.g. implicitly constructing a `std::string` from a `const char*`, but they are definitely problematic as a default.

rtgui/edit.h
adamreichold: I think whether we should use a `contains` method is definitely out of scope for this change here and I guess we should not decide on it without trying it out first. As you mention, the polygon case could become complicated even though standard algorithms exist. The usage of `FORMAT_RGB16_565` might still be a sensible change and performance improvement.

In the long run, I'd say this really looks like it wants to be a general purpose scene graph library, c.f. [QGraphcisScene](https://doc.qt.io/qt-5/qgraphicsscene.html) or [Papyrus](http://libpapyrus.sourceforge.net/index.html). So we should probably look into working with an existing library instead of rolling our own.

rtgui/edit.h
Hombre57: Ok, I then discovered the existence of Ctrl+Shift+T under Eclipse. So handy that I converted all inline functions of edit.h

rtgui/edit.h
Hombre57: I'll use FORMAT_RGB16_565 in the next commit then.

For the general purpose scene graph library, I could have used one if I knew their existence at the time of adding this feature, but the classes already exist in RT now, we're not bound to a third party, and it has the extra possibility to define a different associated "mouse over" geometry than the displayed one (e.g. the central "mouse over" spot of the Gradient widget is far bigger than the displayed one).

rtgui/edit.h
Hombre57: Done.

PR #3159, opened 2016-02-18
Fixed 'LensModel' not extracted for Sony cameras

dschlegel:
When extracting Exif data from a RAW file shot with a Sony camera the lens name/type was not extracted and displayed even though it was set.

Original:
bug

Fixed:
fixed


Beep6581:
I confirm the patch makes passive lenses show up as "Unknown" in the "i" dialog while electronic ones show their correct name.

heckflosse:
I close this because it is fixed with ad32009. @dschlegel Thanks for your contribution :+1:

PR #3164, opened 2016-02-21
Remove SafeGtk and SafeKeyFile modules

adamreichold:
I know this is a big one with few short-term benefits, but I also think this will not get better lying around in my fork of the repository and will improve the code base quality in the long run. Hence here is the pull request to fully remove the SafeGtk and SafeKeyFile modules which embody some questionable choices on error handling, often have only a single place of use and generally add an extra layer of problems on top of the Glib and/or OS-specific API.

The PR itself was just rebased on `master` and the code churn is of course rather large, but at least in the end a small reduction in code size is achieved. (More would probably be possible with more aggressive refactoring at the places of use, but that would make the PR even harder to review.) I also tried to group the changes into reasonably large commits to help with reviewing them. If I can do anything else to make this hefty change easier to digest, please let me know.


heckflosse:
@adamreichold 
I found a bug: Clear cache, restart rt => no thumbnails!

adamreichold:
@heckflosse Thanks for spotting. It seems to be due the write path for some key files trying to load a possibly already existing key file and depending on exceptions from that load being swallowed which should be fixed by the last commit.

rtengine/dfmanager.cc
```diff
@@ -263,13 +262,21 @@ void dfInfo::updateBadPixelList( RawImage *df )
...
-    if( dir && !dir->query_exists()) {
+    auto dir = Gio::File::create_for_path (pathname);
+    if (!dir || !dir->query_exists()) {
         return;
     }
 
-    safe_build_file_list (dir, names, pathname);
+    try {
+
+        auto enumerator = dir->enumerate_children ();
```
heckflosse: I suggest to use `auto enumerator = dir->enumerate_children ("standard::name");`

rtengine/ffmanager.cc
```diff
@@ -209,13 +208,21 @@ void ffInfo::updateRawImage()
...
-    if( dir && !dir->query_exists()) {
+    auto dir = Gio::File::create_for_path (pathname);
+    if (!dir || !dir->query_exists()) {
         return;
     }
 
-    safe_build_file_list (dir, names, pathname);
+    try {
+
+        auto enumerator = dir->enumerate_children ();
```
heckflosse: I suggest to use `auto enumerator = dir->enumerate_children ("standard::name");`

rtgui/filecatalog.cc
```diff
@@ -538,11 +541,44 @@ void FileCatalog::closeDir ()
...
+    std::set<Glib::ustring> extensions;
+    for (const auto& parsedExt : options.parsedExtensions) {
+        extensions.emplace (parsedExt.lowercase ());
+    }
+
+    try {
+
+        auto dir = Gio::File::create_for_path (selectedDirectory);
+
+        auto enumerator = dir->enumerate_children ();
```
heckflosse: I suggest to use `auto enumerator = dir->enumerate_children ("standard::name");`

heckflosse:
@adamreichold Fix confirmed. Thumbs are back :)

adamreichold:
@heckflosse Changed the usage of `enumerate_children` to request the minimal set of attributes. I did make sure to apply where you changed them last, but those calls stepped aside of the merge conflicts and hence I missed them.
The repetition involved is something that makes me nervous since it might imply a helper method, on the other we can now tune each call to `enumerate_children` to use exactly the number of attributes it requires.

heckflosse:
@adamreichold Thank you :)

heckflosse:
@adamreichold Because #3170 also touches rtengine/imagedata.cc I'll wait with merging #3170 until you merged this one.

adamreichold:
@heckflosse Ok, since you mention merging. We already had some fallout here (the cache clearing issue), so I do expect more in principle, but I also think that we probably will not find it in this code review. Should I merge and then try to handle any subsequent bug reports ASAP or do we give this more time to mature? (Questionable if already other code changes are basically blocked by it.) What is @Beep6581's opinion of this?

Beep6581:
My opinion succumbs to your expertise.
I can confirm that it builds and so far runs fine (tested several PP3s and saving) in Linux/GCC-5.3.0. It won't get any testing outside of our small circle if we don't merge, so :+1: for that.

heckflosse:
:+1:

adamreichold:
Ok, so I'll try to be brave and await my punishment in form of the oncoming deluge of bug reports... :-)

PR #3165, opened 2016-02-21
Add build option to enable gprof-based profiling instrumentation

adamreichold:
This pull request adds a build option called `WITH_PROF` which makes it easy to enable [gprof-based](https://sourceware.org/binutils/docs/gprof/) profiling instrumentation which rather useful to spot performance bottlenecks without the prohibitive overhead of using `callgrind` for runtime profiling (but of course also without the per instruction cycle resolution).

To use it, one just runs the instrumented binary through whatever functions should be tested and exits normally. This should produce a `gmon.out` file in the program's working directory which can then be analyzed using the `gprof` command line utility or e.g. the profiling support of Eclipse CDT.


heckflosse:
:+1:

PR #3166, opened 2016-02-21
Clean up the Coord and PolarCoord helper classes

adamreichold:
As discussed in the canvas review @Hombre57, the `Coord` and `PolarCoord` helper classes used by the GUI do not follow standard practices for C++ values types which prevents some standard usage patterns. This pull request makes the necessary changes and also tries to remove some code duplication in the implementation without any changes to the interface.

The only thing I found puzzling was the choice of default values, i.e. `(-1,-1)` instead of `(0,0)` and `(1.0, 0.0)` instead of `(0.0, 0.0)`. I did change them in this pull request since zero is often a bit faster as it is usually a special value in hardware too and I would say that it usually also the less surprising value.

Related to that is the special case of converting a zero Cartesian coordinate to its polar form which would then yield a non-zero vector. Is that exception relied upon anywhere? For now, this PR removes it since it is not what an assignment operator should usually be doing, e.g. ignoring rounding errors, `PolarCoord(Coord(foo)) == foo` and `Coord(PolarCoord(bar)) == bar` should hold which would not be the case if `PolarCoord(Coord(0,0)) == PolarCoord(1.0, 0.0) == Coord(1, 0)`.

### DIFF
--- a/rtengine/coord.cc
+++ b/rtengine/coord.cc
@@ -19,21 +19,49 @@
 
 #include "coord.h"
 
+#include "rt_math.h"
+
 namespace rtengine
 {
 
-void Coord::setFromPolar(PolarCoord polar)
+Coord& Coord::operator= (const PolarCoord& other)
 {
-    while (polar.angle <   0.f) {
-        polar.angle += 360.f;
-    }
+    const auto radius = other.radius;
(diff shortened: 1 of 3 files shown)


Hombre57:
Everything looks fine (and better), could you commit ?

PR #3167, opened 2016-02-21
Add a build-time option to easily enable run-time sanitizers

adamreichold:
Another small thing that might help us to find problems before our users do are the various run-time sanitizers that recent versions of Clang and GCC offer. Especially as they - like `cppcheck` but in contrast with Coverity - can be executed locally by each developer as often as necessary.

This pull request adds a build time option `WITH_SAN` which can be used to enable a specific kind of sanitizer, e.g. using [GCC version 5.3.0](https://gcc.gnu.org/onlinedocs/gcc-5.3.0/gcc/Debugging-Options.html#index-fsanitize_003daddress-648), at least the following variants should be available:
* `WITH_SAN=leak` allows one to find memory leaks with very little run-time overhead compared to running fully emulated under Valgrind,
* `WITH_SAN=address` allows one to find general memory access problems like use-after-free or out-of-bounds accesses still with significantly less overhead compared to Valgrind,
* `WITH_SAN=thread` allows one to find race conditions when several threads execute concurrently catching hard to trace unpredictable failure modes,
* `WITH_SAN=undefined` allows one to find instances of undefined behaviour which often yield unintended compiler optimizations and hard to find bugs.


Beep6581:
Compiles and runs fine, tried with -DWITH_SAN="leak" and it reported a whole bunch of leaks. +1 for merge.

PR #3170, opened 2016-02-22
Fix Sony LensID issue

heckflosse:
There is already a [pull request](https://github.com/Beep6581/RawTherapee/pull/3159) for this issue from @dschlegel, but it doesn't solve the Issue for some Sony cams. I don't know whether this pr will solve it for all Sony cams but it does solve it for most of them at least.
@dschlegel Thanks for pointing out this bug. I hope you don't mind that I made a pr which solves it for more Sony cams than your pr.
Ingo

### DIFF
--- a/rtengine/imagedata.cc
+++ b/rtengine/imagedata.cc
@@ -276,7 +276,7 @@ void ImageData::extractInfo ()
             }
 
         }
-         else if (root->findTag("MakerNote")) {
+        if (lens == "Unknown" && root->findTag("MakerNote")) {
             rtexif::TagDirectory* mnote = root->findTag("MakerNote")->getDirectory();
 
             if (mnote && !make.compare (0, 5, "NIKON")) {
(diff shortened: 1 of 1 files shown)


Beep6581:
I confirm the patch makes passive lenses show up as "Unknown" in the "i" dialog while electronic ones show their correct name.

dschlegel:
@heckflosse  Your PR seems to work better, so I don't mind at all.

But if `DNGLensInfo` is set on a RAW file shot with a Fujifilm camera the previously extracted `LensModel` ([line 267](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/imagedata.cc#L267)) will be overwritten in [line 438](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/imagedata.cc#L438) by the value of `DNGLensInfo`.

It's a minor problem but I wanted to point it out.

heckflosse:
@dschlegel Thanks for pointing this out. I'll fix it

PR #3172, opened 2016-02-24
Reduce start time and memory usage by removing redundant lookup table…

heckflosse:
…s and fillig remaining lookup tables in parellel

I'll make some further changes tomorrow

### DIFF
--- a/rtengine/LUT.h
+++ b/rtengine/LUT.h
@@ -52,6 +52,8 @@
  *          LUTf stands for LUT<float>
  *          LUTi stands for LUT<int>
  *          LUTu stands for LUT<unsigned int>
+ *          LUTd stands for LUT<double>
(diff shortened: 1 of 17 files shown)


adamreichold:
Certainly a good change not just for performance but also for maintainability. I wonder if the gamma-related look-up tables from `rtthumbnail.cc` shouldn't also be moved to `color.cc` to have everything in one place? (Also not sure why these are dynamically allocated?) There also seems to be a leftover `BENCHFUN` in `init.cc`?

heckflosse:
@adamreichold I followed your suggestions about rtthumbnail.cc, made some additional small changes and removed benchmark code

heckflosse:
Any objections to merge?

adamreichold:
Sorry, forgot the thumbs up... +1 for merge from me.

PR #3197, opened 2016-03-06
Enhanced edit mecanism

Hombre57:
Make some changes and add OPIcons to the Edit mechanism, for the upcoming Spot Removal tool.

This is an intermediate patch to avoid too much conflicts regarding the merge of the Gtk3 branch on master, and my Spot Removal branch.

To check this PR, just check if the on preview manipulations (gradient tool widget, crop, wb, curve pipette) are still working as usual.

I non one complains, I'll merge on Monday.


rtgui/cropwindow.cc
```diff
@@ -66,20 +66,19 @@ ZoomStep zoomSteps[] = {
 #define ZOOM11INDEX  13
 
 CropWindow::CropWindow (ImageArea* parent, rtengine::StagedImageProcessor* ipc_, bool isLowUpdatePriority_, bool isDetailWindow)
-    : ObjectMOBuffer(parent), onResizeArea(false), deleted(false), fitZoomEnabled(true), fitZoom(false), isLowUpdatePriority(isLowUpdatePriority_),
-      backColor(options.bgcolor), decorated(true), titleHeight(30),
-      sideBorderWidth(3), lowerBorderWidth(3), upperBorderWidth(1), sepWidth(2),
-      xpos(30), ypos(30), imgX(-1), imgY(-1), imgW(1), imgH(1), iarea(parent),
-      cropZoom(0), cropgl(NULL), pmlistener(NULL), observedCropWin(NULL), ipc(ipc_), isFlawnOver(false)
+    : ObjectMOBuffer(parent), state(SNormal), press_x(0), press_y(0), action_x(0), action_y(0), pickedObject(-1), pickModifierKey(0), rot_deg(0), onResizeArea(false), deleted(false),
```
adamreichold: With C++11, the simple initializations can go directly into the header which usually makes the logic of the constructors if they have more readable or removes the need to have them completely. (Some people like to say that the rule of three became the rule of five is actually the rule of zero, i.e. with proper choice of types most classes do not need any explicit constructors resp. destructors at all.)

rtgui/cropwindow.cc
```diff
@@ -272,210 +269,209 @@ void CropWindow::buttonPress (int button, int type, int bstate, int x, int y)
 {
 
     bool needRedraw = true;  // most common case ; not redrawing are exceptions
+    EditSubscriber *editSubscriber = iarea->getCurrSubscriber();
```
adamreichold: Could use `const auto` instead of spelling out the type.

rtgui/cropwindow.cc
```diff
@@ -272,210 +269,209 @@ void CropWindow::buttonPress (int button, int type, int bstate, int x, int y)
...
 
-    if (button == 1 && type == GDK_2BUTTON_PRESS && onArea (CropImage, x, y) && (state == SNormal || state == SCropImgMove)) {
-        if (fitZoomEnabled) {
-            if (fitZoom) {
-                state = SNormal;
-                zoomVersion = exposeVersion;
-                screenCoordToImage (x, y, action_x, action_y);
-                changeZoom (ZOOM11INDEX, true, action_x, action_y);
-                fitZoom = false;
+    if (button == 1) {
```
adamreichold: The logic that follows look hard to read and hence brittle to maintain. I am not sure how to go about it exactly, but maybe the repetitive parts could be change to be data-driven instead of logic-driven? For example have a list of possible crop areas and the associated values the various variables should have if the event happened inside that area?

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
     }
 }
 
-EditSubscriber::EditSubscriber (EditType editType) : ID(EUID_None), editingType(editType), bufferType(BT_SINGLEPLANE_FLOAT), provider(NULL), dragging(false) {}
+void OPIcon::drivenPointToRectangle(rtengine::Coord &pos,
```
adamreichold: If `pos` is never modified, could it be bound as `const Coord&` instead of `Coord&`?

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
...
+    bottomRight.x = topLeft.x + W - 1;
+    bottomRight.y = topLeft.y + H - 1;
+}
+
+OPIcon::OPIcon(Cairo::RefPtr<Cairo::ImageSurface> *normal,
+               Cairo::RefPtr<Cairo::ImageSurface> *active,
+               Cairo::RefPtr<Cairo::ImageSurface> *prelight,
+               Cairo::RefPtr<Cairo::ImageSurface> *dragged,
+               Cairo::RefPtr<Cairo::ImageSurface> *insensitive,
+               enum DrivenPoint drivenPoint) :
```
adamreichold: Prefixing enumerated types with `enum` is not necessary in C++.

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
...
+        topLeft.x = pos.x;
+        topLeft.y = pos.y - H / 2;
+        break;
+    }
+
+    bottomRight.x = topLeft.x + W - 1;
+    bottomRight.y = topLeft.y + H - 1;
+}
+
+OPIcon::OPIcon(Cairo::RefPtr<Cairo::ImageSurface> *normal,
```
adamreichold: As those are never modified and `RefPtr` itself already indicates whether it contains a valid pointer (same code as now, just test it like a Boolean), I think these would usually be of type `const Cairo::RefPtr<Cairo::ImageSurface>&`. (Passing in an empty `RefPtr` using its default constructor via `{}` should not much less efficient but the interface becomes safer.)

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
...
+    if (!draggedImage.empty()) {
+        draggedImg = Cairo::ImageSurface::create_from_png( RTImage::findIconAbsolutePath(draggedImage) );
+    }
+
+    if (!insensitiveImage.empty()) {
+        insensitiveImg = Cairo::ImageSurface::create_from_png( RTImage::findIconAbsolutePath(insensitiveImage) );
+    }
+}
+
+Cairo::RefPtr<Cairo::ImageSurface> OPIcon::getNormalImg()
```
adamreichold: I guess these should be `const` and they could return `const Cairo::RefPtr<Cairo::ImageSurface>&` to elide unnecessary copying.

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
...
+Cairo::RefPtr<Cairo::ImageSurface> OPIcon::getDraggedImg()
+{
+    return draggedImg;
+}
+Cairo::RefPtr<Cairo::ImageSurface> OPIcon::getInsensitiveImg()
+{
+    return insensitiveImg;
+}
+
+void OPIcon::drawImage(Cairo::RefPtr<Cairo::ImageSurface> img,
```
adamreichold: I don't we need to modify or copy the `RefPtr` here, we just want to modify the image itself. So the appropriate type seems to be `const Cairo::RefPtr<Cairo::ImageSurface>&` again? (Copying valid reference-counted pointers usually involves atomic operations on its control block which can become a performance burden.)

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
...
+    rtengine::Coord tl, br; // Coordinate of the rectangle in the CropBuffer coordinate system
+    drivenPointToRectangle(pos, tl, br, imgW, imgH);
+
+    cr->set_source(img, tl.x, tl.y);
+    cr->set_line_width(0.);
+    cr->rectangle(tl.x, tl.y, imgW, imgH);
+    cr->fill();
+}
+
+void OPIcon::drawMOImage(Cairo::RefPtr<Cairo::ImageSurface> img, Cairo::RefPtr<Cairo::Context> &cr,
```
adamreichold: Same as above. Also `const Cairo::RefPtr<Cairo::Context>&` seems appropriate. (The `const` that woudl be prevent the context from being modified would be `Cairo::RefPtr<const Cairo::Context>`.)

rtgui/edit.cc
```diff
@@ -747,7 +747,294 @@ void Rectangle::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned shor
...
+}
+
+void OPIcon::drawToMOChannel(Cairo::RefPtr<Cairo::Context> &cr, unsigned short id,
+                             ObjectMOBuffer *objectBuffer, EditCoordSystem &coordSystem)
+{
+    if (flags & F_HOVERABLE) {
+        // Here we will handle fallback solutions
+        State tmpState = state;
+
+        if (tmpState == INSENSITIVE) {
```
adamreichold: Are these not exclusive? I.e. shouldn't this be `if-else-if-else` or even `switch-case`?

rtgui/edit.cc
```diff
@@ -791,24 +1078,29 @@ void EditSubscriber::switchOffEditMode()
     unsubscribe();
 }
 
-EditUniqueID EditSubscriber::getEditID()
+const EditUniqueID EditSubscriber::getEditID()
```
adamreichold: The `const` before the return type should be superfluous for integral types as they are always immutable as rvalues, e.g. when they are returned from a function. I think GCC even warns about this at `-Wall`?

rtgui/edit.h
```diff
@@ -381,9 +392,49 @@ class Rectangle : public Geometry
...
-    void drawToMOChannel (Cairo::RefPtr<Cairo::Context> &cr, unsigned short id, ObjectMOBuffer *pipetteBuffer, EditCoordSystem &coordSystem);
+    void drawOuterGeometry (Cairo::RefPtr<Cairo::Context> &cr, ObjectMOBuffer *objectBuffer, EditCoordSystem &coordSystem);
+    void drawInnerGeometry (Cairo::RefPtr<Cairo::Context> &cr, ObjectMOBuffer *objectBuffer, EditCoordSystem &coordSystem);
+    void drawToMOChannel (Cairo::RefPtr<Cairo::Context> &cr, unsigned short id, ObjectMOBuffer *objectBuffer, EditCoordSystem &coordSystem);
+};
+
+class OPIcon : public Geometry    // OP stands for "On Preview"
+{
+
+protected:
```
adamreichold: Why `protected` instead of `private`? I think one usually starts at the most restrictive protection level and only releases at little as possible as needed.

Beep6581:
I tested all widgets you listed, couldn't find any differences between this and `gtk3`, and no crashes either.

rtgui/cropwindow.cc
Hombre57: I don't really understand what you mean : there is not simple initialization here, so why do you want to put that in the header ?

rtgui/cropwindow.cc
Hombre57: This code has been done before you jumped into the project. I'll change to `auto` here but won't spend time to hide all pointer type introduced (where it applies in this patch). I'll try to think about doing it for future ones.

rtgui/cropwindow.cc
Hombre57: You're probably right, I spent quite some time on this function, but it is actually working and because of the deadline, would like to commit as is, unless you want me to change that now.

Please note that we have to check for the mouse button, the area, the modifier keys and the current tool mode. Maybe I could rewrite this function so it is sorted by tool mode first, then area, then mouse buttons, then modifier key ? The actual implementation does handle the priority in a very specific manner.

rtgui/edit.cc
Hombre57: Sure, done.

Just for my information, is it worth to define all variable as `const` if they are not supposed to be modified ? I don't really know what are the benefit of using `const` in fact, but I guess it is related to optimization and pedantic API ? e.g. for `OPIcon::drawMOImage`, should I switch `img` and `id` to `const` ?

rtgui/edit.cc
Hombre57: Do you mean that I have to write e.g. `const Cairo::RefPtr<Cairo::ImageSurface> &insensitive = Cairo::RefPtr ()` to replace the `= NULL`default ?

rtgui/edit.cc
Hombre57: No, because we test tmpState, which can be modified inside an `if`statement.

rtgui/cropwindow.cc
adamreichold: Stuff like `state(SNormal)` can be put directly into the header like `ImgEditState state = SNormal;` with the same effect. Initializing member variables at the point of can be more readable and makes it harder to forget something. Of course it is not applicable for initializations that depend on constructor arguments and such. But it is just a suggestion, nothing really important.

rtgui/edit.cc
adamreichold: Ah, I see, this is clever. Maybe you want to add a comment?

rtgui/cropwindow.cc
adamreichold: Well, it probably isn't worth it to try to make this data-driven: At the end of the day, this wants to use widgets or a scene graph or something that allows to dispatch the events and handle them locally within a separate object. This is just how GUI code looked like before OOP. Maybe we can someday find the time to integrate this with the rest of the scene graph infrastructure.

rtgui/cropwindow.cc
adamreichold: I see that the code is older, I just meant that if we touch stuff anyway, we can modify as if we would have written it today.

rtgui/edit.cc
adamreichold: Declaring function arguments as `const` is mostly about safety: You cannot accidentally modify what you declare constant whereas with a normal function argument, you usually get a copy which you can modify even though that is usually not intended. With references as in this case, it is even worse since any changes made to the object would actually be visible to the caller. The general rule of thumb is to be as free of side effects as possible to make reasoning about the code easier.

rtgui/edit.cc
adamreichold: It just be enough to write `const Cairo::RefPtr<Cairo::ImageSurface> &insensitive = {}` to specify a default-constructed default argument. On the other hand, you can probably also initialize the `RefPtr` with `nullptr` to get an invalid/empty one.

rtgui/cropwindow.cc
Hombre57: Oh, I didn't knew this was possible :). It's good to know, but it somewhat contradict what said some times ago about inline functions, so that they should be defined out of the class definition to keep it clean, or (I assumed) to hide class implementation.

adamreichold: You are right that this could work in the opposite direction as moving inline method out of the class body. But considering that you cannot hide the data members in C++ (without resorting to an indirection like a pimpl) anyway, that part of the implementation is always exposed anyways. And I would only suggest to use this to initialize simple types, not complicated objects, which would otherwise just clutter up the constructors.

Hombre57:
I'll try to rebase and merge this branch in the next hours. Please forgive me if something is flawed after that, it's my first try (and that not really intuitive). If I fail to rebase, I'll push all the successive commits.

Hombre57:
I give up, rebasing is not for me :(

Merging pull request now.

Hombre57:
@adamreichold 
Now that the prototype of OPIcon uses `{}` as default value, how can I use this ctor easily if I want to ignore `prelight` and `dragged` but want to set `insensitive` ?

adamreichold:
@Hombre57 You can also use `{}` to default construct the argument, e.g. if you want to set only normal, active and insensitive, you would call the constructor like `OPIcon(normal, active, {}, {}, insensitive)`. Since the compiler knows that an argument of `Cairo::RefPtr<Cairo::ImageSurface>` is expected, it will call a fitting constructor using the initializer list given. For an empty list, this will be the default constructor which will create an empty `Cairo::RefPtr`.

PR #3199, opened 2016-03-06
About 3x speed for raw false colour suppression

heckflosse:


heckflosse:
Branch stays open for xtrans speedups

PR #3228, opened 2016-03-20
Speedup for xtrans demosaic

heckflosse:


heckflosse:
Branch stays open for further improvements

Issue #1853, opened 2015-08-11
RTTI - migrate to compressed TIFF format


Beep6581:
Originally reported on Google Code with ID 1869

```
Re: http://www.luminous-landscape.com/forum/index.php?topic=77666.0

"The lossy-compressed DNG is also the internal format used for Smart Previews in Lr
5 Beta"

This is quite nice, actually. If RTTI format is also lossy DNG it will offer
1. saving disk space
2. larger thumbnail size for more convenient editing in File Browser
3. could improve performance on some systems where IO is bottleneck

Visually lossy and lossless DNG are very close.
```

Reported by `michaelezra000` on 2013-05-11 20:32:40

Hombre57:
Just to keep it under the hand, this issue should remain opened. Even if we don't go to lossy-compressed DNG or TIFF or whatever is compressed, I think it will bring preformance to have smaller thumbnail files, and reduce the HD footprint (which can be quite large even with today's capacity).

@adamreichold If you feel confident with file formats (more than me), this issue waits for you or for someone wanting to bring some performance to the File Browser.

Beep6581:
Agreed, I've helped some users who had cache folders over 10GB!
There is #1895 for OpenEXR support. If we're talking about adding a new format, and the requirement of the new format is high bit depth (that's why we're not using JPEG, right?), then maybe OpenEXR is the way to go? It supports lossless and lossy compression. Kill two birds with one stone.

Hombre57:
Yes, it has to support high bit depth. Having OpenEXR would be great ! But if that's not possible, I'd go TIFF or DNG. Darktable already implements OpenEXR, maybe we could get some code from them ;)

adamreichold:
@Hombre57 Yes, I would like to try to work on that. Should I assign myself to this then? (You'll take care of the ICM preview properties tab then?)

@Beep6581 When removing in the `safegtk` API in my `remove-safegtk` branch, I think I noticed that the method `CacheManager::applyCacheSizeLimitation` is broken insofar it does not use the 'CacheManager::getCacheFileName` method to determine the file to delete, therefore missing the MD5-dependent part.
I have not yet reported this as a separate issue, because I could not yet properly verify this and I though it could be fixed along the way. But since that branch does depend on the C++11 example code and is itself pretty large, maybe I should isolate these changes for a separate pull request?

Hombre57:
> Should I assign myself to this then?

Yes, if you're okay.

[...]

PR #3234, opened 2016-03-27
Compress thumbnails

adamreichold:
As per my assignment of #1853, I have been looking into compression of the thumbnail images. This pull request does the following in that regard:
- It first changes the methods used to write the raw `Image8/16/float` data to use GIO streams instead C-style I/O to allow for transparent stream layering.
- It then uses GIO's converter streams to extend the custom RTTI format by Zlib-compressed variants.
- Finally it replaces the custom RTTI format by standard TIFF images using Deflate compression and difference prediction.

The resulting sizes on the thumbnails for a small set of RAW and JPEG files are:

| Method | Size | Reduction |
| --- | --- | --- |
| RTTI baseline | 26844 | - |
| RTTI Deflate | 23232 | 13.5% |
| TIFF Deflate & predictor | 20936 | 22.0% |
| TIFF LZMA & predictor | 19880 | 26.0% |

The advantage of simply compressing the RTTI files is that migration is completely transparent, i.e. every thumbnail written is just compressed but existing thumbnails need not be touched. Also the necessary code changes are minimal. Of course the size reduction is also quite modest at around ten percent.

Using TIFF yields the benefit of having a standard format for thumbnails which is much more future proof IMHO. Together with difference prediction, the size reduction is also much more pronounced at more than twenty percent. However, this implies that all thumbnails will need to be recreated immediately.

Finally, using LZMA instead of Deflate compression for the TIFF images (by just replacing COMPRESSION_DEFLATE by COMPRESSION_LZMA) shows how easy it may be to change these parameters in the future after adopting TIFF once, but also yields only a modest improvement in size reduction while needing noticeably more processing time. (Deflate compression does not seem to have any noticeable impact on runtime.)

Hence if no problems are found, I would suggest merging as is, i.e. switching to TIFF but for now using just Deflate compression.


adamreichold:
I also tried LZW compression, but for 16-bit image data this usually increased file size instead of reducing it.
The options for lossy compression using [LibTIFF](http://www.libtiff.org/) seem to be quite limited at the moment, e.g. JPEG compression but only for 8-bit image data.
JPEG2000 compression seems to available lossy and lossless and for all image data types we currently need, but seems to be incompatible with the scanline-based I/O interface. In any case, this might be worthwhile to investigate in the future.

adamreichold:
Concerning [OpenEXR](http://openexr.com/), while this seems applicable at first, its sample formats seem to be limited to 16-bit and 32-bit floating point and 32-bit integer. So the rather common 16-bit integer format currently used for our thumbnails does not fit to well into this scheme. Also it would add an additional dependency which the above approaches do not have.

adamreichold:
I would be rather grateful if someone with a larger and more diverse data set (and maybe a faster computer ;-)) could repeat the above measurements so that we have better data to base our decisions on. Thanks!

adamreichold:
Amended the last commit to avoid naming confusion between the methods `writeTiff` (now named `writeData` again) and the full export `saveTIFF`.

Hombre57:
Nice improvements even with 26% reduction "only".

Before RT went open source and the pipeline migrated to 32 fp internal data, the thumbnail file format was already jpeg. It would certainly save a lot more disk space (and thumbnail load time) to have a lossy file format. The scanline method is not mandatory, it's just a way to copy the data between Image8/16/float to ImageIO. We could easily write a function to write data by tiles I guess, we just have to know what kind of data arrangement we have on both side.

For the numbers, I'll make some test later today.

adamreichold:
> The scanline method is not mandatory, it's just a way to copy the data between Image8/16/float to ImageIO. We could easily write a function to write data by tiles I guess, we just have to know what kind of data arrangement we have on both side.

Obviously, it is just that the scanline-based interface currently aligns best with out internal representation and hence is easiest to program which is why I started with it.

Concerning JPEG2000 compression inside TIFF: It will probably not be that easy or even desirable as this seems to mean that a full JPEG2000 stream (created using some other library like JasPer or OpenJPEG) is just embedded inside a TIFF container. In that case, it might probably be better to read and write JPEG2000 files directly...

Hombre57:
> Concerning JPEG2000 compression inside TIFF: It will probably not be that easy or even desirable as this seems to mean that a full JPEG2000 stream (created using some other library like JasPer or OpenJPEG) is just embedded inside a TIFF container. In that case, it might probably be better to read and write JPEG2000 files directly...

That was unclear, but that's what I had in mind.

adamreichold:
@Hombre57 Concerning the usage of JPEG2000: Support for floating point components seems to be problematic. OpenJPEG version 2 seems to have support for up to 32-bit integer components, but while the specification includes floating point components and some commercial libraries implement it, the free software projects do not yet do so.

Since I have not found this yet: Where do we generate floating point image data for thumbnails? The code
````
if (tpp->thumbImg) {
    delete tpp->thumbImg;
}

tpp->thumbImg = NULL;
tpp->thumbImg = resizeTo<Image16>(w, h, TI_Bilinear, tmpImg);
delete tmpImg;
````
from `Thumbnail::loadFromRaw` seems to imply that we will only ever persist 8-bit or 16-bit image data?
If that is the case, JPEG2000 using OpenJPEG looks like it could provide another 75% reduction even with lossless compression. (But the code to achieve this will be significantly more complicated and would introduce a new dependency so that I would probably like to defer that to a separate pull request.)

> That was unclear, but that's what I had in mind.

Why do you want to put the JPEG2000 stream inside a TIFF container instead of just using the JPEG2000 container format?

**Comment by adamreichold** [comment]
On second thoughts, [JasPer](https://www.ece.uvic.ca/~frodo/jasper/) might actually be a little bit simpler to use (and also seems to be used more widely) if we can restrict ourselves to 8- and 16-bit components...

**Comment by Hombre57** [comment]
> Why do you want to put the JPEG2000 stream inside a TIFF container instead of just using the JPEG2000 container format?

i.e. I had in mind generating jpeg files directly.

> Where do we generate floating point image data for thumbnails?

For standard images (i.e. non-raw), the code on line 116 of `rtthumbnail.cc` is 
```
tpp->thumbImg = resizeToSameType(w, h, TI_Bilinear, img);
```

At that time, we assumed that raw images would have up to 16 bit integer data, but some DNG files and perhaps modern raw files could have higher bit depth, I don't know. In this case, we should update the code.

The thumbnails data we are caching is the post-demozaicing image, without further processing. So it makes sense to keep the same bit depth than the original file.

For HDR images coming from OpenEXR files (request from issue #1895) or from HDRMerge, we have to stick to 32 bit data.

I think that most files processed with RT will be 16 bit maximum, but still... Does it harm to handle 2 thumbnail file type depending on the file format (just asking) ?

**Comment by adamreichold** [comment]
> I think that most files processed with RT will be 16 bit maximum, but still... Does it harm to handle 2 thumbnail file type depending on the file format (just asking) ?

Only in terms of complexity and code bloat. Our users probably don't care as long as it is fast and frugal...

**Comment by adamreichold** [comment]
Just for reference, I must have miscalculated the sizes when trying this using ImageMagick: Now that I have implemented writing 8-bit and 16-bit thumbnails using JasPer, the size of the above file set using lossless compression is `17400`, i.e. a `35.2%` reduction.
Lossly compression can of course bring this much lower, e.g. a size of `10464` meaning a `61.0%` reduction at rate `0.5`. But then we have to start tuning this or making it configurable and so...
Of course, when trying to make this useful the main complication now becomes handling two possible formats instead of one...

**Comment by adamreichold** [comment]
So I still think losslessly compressed TIFF is good first step and maybe using TIFF as a generic container is actually the simplest solution to handling several formats as the next step after that?

**Comment by Hombre57** [comment]
Ok, fine.

**Comment by iliasg** [comment]
I don't understand why do we need 16bit thumbnails !!! ???
8bit gamma encoded *.tif is absolutely fine for the purpose .. and I think 8bit data are more efficiently compressed with deflate ;)

**Comment by adamreichold** [comment]
@iliasg AFAIU, the thumbnails that are stored on disk are used as the basis to apply similar processing as in the preview area to actually preview the chosen settings in file browser. Hence reducing their dynamic range would significantly alter the processing results.

**Comment by iliasg** [comment]
@adamreichold , 
I think the 8bit depth would only be a problem with linear data.
If the data are 8bit gamma encoded it's hard to believe that we will have problems.
I don't know exactly which RT tools/algorithms (and at which order) are applied for thubs creation but if someone can make a list I could try to replicate with normal "save image" i.e tiff, resamled at thubs size, at 8bit gamma 2.0
I choose gamma 2.0 instead of sRGB gamma https://en.wikipedia.org/wiki/SRGB#The_forward_transformation_.28CIE_xyY_or_CIE_XYZ_to_sRGB.29 because the last is linear at the darks and gives less density at the dark values (i.e. suspected for posterization), 
.. and I choose tiff instead of jpeg because with jpegs and integer arithmetics, there is one more quantization step (convert from RGB to Y'CbCr https://en.wikipedia.org/wiki/JPEG#Color_space_transformation ) which could result in increased posterization

**Comment by Hombre57** [comment]
Some data for 221 file on the same body :
Before patch : 140 082 911 bytes
After patch, DEFLATE method : 111 183 586 bytes  (-20.63%)
After patch, LZMA method : 106 587 160 bytes (-23.91%)

DEFLATE is already good, and it's a step in the right direction. Nice works!

There should be only one correction before committing : if the TIFF file doesn't exist (e.g. if you install a new Release version or if you clean up the cache), it'll trigger a "File doesn't exist" alert for each file where a pp3 file exist, which is quite annoying. Could you test the file existence first ? i.e. adding this on line 133 of iimage.cc :
```
    if (!Glib::file_test(fname, Glib::FILE_TEST_EXISTS)) {
        return nullptr;
    }
```

**Comment by adamreichold** [comment]
@Hombre57 I fixed the warning issued by libTIFF but in a slightly different way: I `open` the file myself and pass on the file descriptor to libTIFF's `TIFFFdOpen` function. This way, I only access the file once instead of call `stat` or something similar to test its existence and then having libTIFF open it. This saves the possibly expensive extra system call and prevents the possible race condition of deleting the file between the calls.

I also moved the routines into the `ImageIO` module where they seem to fit better which also allows us to make proper use of polymorphism. As an aside: Do you know why virtual inheritance is used for the various image classes?

I also noticed that I did not yet support wide character file names on Windows and hence tried to implement that as well. It does build on Windows XP but maybe you could also give it a short test using a non-latin1 file name? Thanks!

If there are no further issues, I'll merge this tonight?

**Comment by Hombre57** [comment]
> As an aside: Do you know why virtual inheritance is used for the various image classes?

Which virtual inheritance are you referring to ?

Theres still a problem when trying to load non-existing TIFF thumbnails : there's a per file alert saying *Cannot read TIFF header*. I guess that `_wopen` does not return the value you expect. Also, `_wopen` has a `__MINGW_ATTRIB_DEPRECATED_SEC_WARN` flag in the include file. Don't know if it matters.

**Comment by morosakubek** [comment]
With enabled HaldCLUT gets a strange effect.

**Comment by adamreichold** [comment]
@morosakubek Can you elaborate? And did you check that the effect was not present with the RTTI thumbnails?

**Comment by adamreichold** [comment]
@Hombre57 Indeed `_wopen` is deprecated and I amended the last commit replacing it by `_wsopen_s`. However, I am currently having trouble getting my builds on Windows XP working (but it seems completely unrelated to this), hence could you give it a try? I will try to fix my builds in the mean time...

**Comment by Hombre57** [comment]
Still the same with `_wsopen_s`

**Comment by adamreichold** [comment]
@Hombre57 Thanks for trying. Taking a hint from `g_fopen_withBinaryAndLock` when I obviously need it, I switched over to using `CreateFileW` together with the `OPEN_EXISTING` flag.

However I still can't test because my Windows problems persist. Did you ever have the problem of illegal instructions crashing your build? I set up a pristine build environment under Windows XP which worked fine in the past but produces corrupted binaries now...

**Comment by adamreichold** [comment]
@Hombre57 Mysteriously, my builds started working again even though I do not have a clue what I changed... In any case, it seems the warning messages are coming from the write side?

**Comment by adamreichold** [comment]
@Hombre57 Now I completely confused: I removed everything but passing Unicode strings to `TIFFOpenW` and now the error messages are gone here. But isn't that what we started with? Does it work for you now as well?

**Comment by Hombre57** [comment]
Just tried now and no, it still doesn't work, i get the *Cannot read TIFF header* alerts. I don't know if it's on reading or writing the files though. All I can say is that the tiff files are created as long as I click OK on the alert windows. I'm investigating as well...

**Comment by Hombre57** [comment]
The readtiff function seems to work fine, the file exist when it tries to load it, but the error seems to say that the tiff file is corrupted, it can't find the Header section.

**Comment by adamreichold** [comment]
@Hombre57 `readTiff` should not be there anymore in last version? I amended that commit again and now it should be `TIFFOpenU` which does nothing but form Unicode text if necessary... With `readTiff` I had similar errors as I guess `TIFFFdOpen` does not work reliably with the Windows low-level file handling emulation.

**Comment by Hombre57** [comment]
Keep calm :

tifferror

The tiff file doesn't exist here.

**Comment by Hombre57** [comment]
> readtiff should not be there anymore in last version?

You're just amending faster than I'm testing :)

**Comment by Hombre57** [comment]
In thumbail.cc :
```
    succ = succ && tpp->readImage (getCacheFileName ("images", ".tif"));
```

Here it tries to open the cached image. Failing to open it is something that is expected, so it will generate it. But `tpp->readImage` will end up with `ImageIO::readThumbnail`, which should handle to non existence of the file properly.

**Comment by adamreichold** [comment]
@Hombre57 But it does it handle it properly: If the file does not exist, opening it fails and we return an indication of that to the sender. It is only libTIFF's insistence on doing its own error handling by showing a dialog to the user instead of leaving that to the application that is annoying here. Hence I pushed another commit that registers a Unix-like error handler with the library, printing errors to `stderr` (and also warnings if `options.rtSettings.verbose` is true) instead of show dialogs.

**Comment by Hombre57** [comment]
It's working fine now, no more alert window. Special chars like `çéèà` are correctly handled as well.

**Comment by morosakubek** [comment]
@adamreichold This happens when you use images with enabled HaldCLUT
screen

**Comment by adamreichold** [comment]
@morosakubek And this is specific to this branch and does not happen on `master`?

**Comment by morosakubek** [comment]
@adamreichold yes, from your branch "compress-thumbnails"

**Comment by adamreichold** [comment]
@morosakubek I do not doubt that the screenshot was made using this branch. I only have a hard understanding how the way we store the thumbnails should interact with HaldCLUT which is why I am asking what the current behaviour on the `master` branch is.

**Comment by adamreichold** [comment]
@Hombre57 Can you reproduce this? I tried both this branch and `master` without any visible difference when using the film simulation.

**Comment by Hombre57** [comment]
I had similar wrong thumbnail rendering, but I don't think it is related to HaldCLUT. This is how it occured here :
- delete the cache/images folder
- entering in a directory containing pp3 files that was created before HaldCLUT was added to RT (I don't know if it matter, but that was the case)
- all thumbnails are extracted with the Quick method, and displayed properly
- all thumbnails that has a sidecar file are being processed, and the display is broken as shown above, all of them
- I'm selecting all the edited file and change a parameter (e.g. toggling Resize On and Off) with the Batch Editor to force a refresh, now they are displayed correctly.

I didn't had time to work on RT today, so can't say about being specific to this branch or not, I'll have more time tomorrow.

**Comment by Hombre57** [comment]
Okay, I've investigated more : in fact, the steps above are incomplete. And I confirm that it is specific to this issue.

On first step, you have to delete the *data* and *aehistogram* as well as *images*. I don't know if they have to be both deleted but at least you'll be able to reproduce. Then follow the already given steps. I confirm that HaldClut has no influence on this bug.

Also look at the *images* folder while it build the tif files, you'll certainly understand what's going on. The problem is that you're saving as thumbnail tif files the data that are extracted from the Quick loading, i.e. 8 bits data. This 8 bits tif file is then loaded to process the image of a 12 or 14 bits raw file, which is why it is severely overexposed. At the end of the process, the edited files (i.e. having a pp3 sidecar file) are overwritten by a 16 bit tif files, as it should have been from the beginning. The unedited files are still 8 bits.

I'll be back in 30mn to check any new comment, but am mostly off for today.

**Comment by adamreichold** [comment]
> Thinking at it again, I think this statement is wrong. Having 8 bits tif file for unedited image should be correct. The problem may be that RT should update the tif thumbnail image before the full process, not after.

Sorry for the delay, I was otherwise engaged. If the above is the case, it can't be specific to this change as I did not change what is stored or when. The data in those `Image8/16/float` instance is byte for byte the same as before the change. And the modified code also does not determine what kind of data is stored or when. So if there is a problem with the order of loading and saving thumbnails, it should be present in `master` as well.

**Comment by Beep6581** [comment]
I compiled `master`, deleted the cache, ran it, applied some PP3s to some files, took this screenshot:
scrot_2016-04-03 161440

I compiled `adamreichold/compress-thumbnails`, ran it, loaded the same folder, took this screenshot:
scrot_2016-04-03 161516

I deleted the cache, ran it, took this screenshot:
scrot_2016-04-03 161547

The artifacts are obvious, but there's also a difference in the thumbs of the photos which don't have a profile applied. It looks as if the embedded JPEGs are used in the first and third screenshots, while something else is used in the second one.

**Comment by adamreichold** [comment]
@Beep6581 Thanks for testing. So it seems the issue is something different altogether or I made a stupid mistake somewhere and changed something I did not intend to change. Could you provide me with a image/pp3 pair that triggers the issue since I have yet to reproduce it locally?

**Comment by adamreichold** [comment]
@Beep6581 Also when you write deleting the cache, do you mean the whole cache folder or the subdirectories @Hombre57 mentions?

**Comment by adamreichold** [comment]
@Beep6581 Finally, I have seen it after enabling `internalThumbIfUntouched` and deleting the complete cache folder. It seems the internal thumbnails options is becoming active if the clear thumbnails function from the preferences is used which is why toggling it and doing that had no effect. Now I can find out what went wrong... Sorry about the delay.

**Comment by adamreichold** [comment]
@Beep6581 What I understand so far: The difference between the first and second screenshot is due to the fact if just the thumbnail image could not be found, RT does not consider this an initial thumbnail and will never use the embedded preview even if no stored preview is available (and also even for untouched images without profiles). This is true for `master` as well and can be checked by just deleting the `images` subdirectory of the cache.

The distortions happen whenever `Thumbnail::upgradeThumbImage` is called even though I have currently no idea why that is this case as indeed `tpp` is of type `Image16` at this point. Also, just switching to a different folder and back yields correct thumbnails, i.e. `Thumbnail::processThumbImage` always works no matter whether it performs quick or full processing. Since this implies that the image files written and read are correct, I must changed something that was relied upon below the call to `Thumbnail::upgradeThumbImage`. Any ideas?

**Comment by Hombre57** [comment]
I don't really understand how the update process is working, so I put some printf to trace the process, and put the result in this PDF file. In all case, this is the trace I get when starting RT in a folder containing only one image, so it's not polluted by asynchronous update of several images.

The only difference I can see is that in this branch, we're calling `refreshQuickThumbImages` before `_loadThumbnail`. `refreshQuickThumbImages` is what will start a full processing (`upgrade = 1`). I may be wrong, but it could be a race condition because the thumbnails takes less time to save.

Hope this helps...

[Compress thumbnails - trace.pdf](https://github.com/Beep6581/RawTherapee/files/201489/Compress.thumbnails.-.trace.pdf)

**Comment by Beep6581** [comment]
@adamreichold 
> Could you provide me with a image/pp3 pair that triggers the issue since I have yet to reproduce it locally?
Every image in the screenshot used a different PP3. Some were stock PP3s, the ones we ship, others were e.g. neutral + white-point correction=0.25. As you can see, the problem appeared in every image which had an associated PP3, so you can really choose any random PP3 to reproduce.
http://filebin.net/d5ze0c4p9q

> when you write deleting the cache, do you mean the whole cache folder or the subdirectories
The whole `~/.cache/RawTherapee*` folder.

**Comment by adamreichold** [comment]
@Hombre57 Thank you for the intense testing and the detailed trace! I think you might be right about the data race, but then I do not know which one... :-(
The thread sanitizer is very unhappy when just opening RT in a folder containing a single image. Unfortunately it also slows things down considerably so that I am still waiting for my first full thumbnail and will have to continue to look into this tomorrow, but [here](http://filebin.net/vsey85rvs7) is the log up to that point for anybody who wants to have a look...

**Comment by adamreichold** [comment]
From reading the log, I think the thread sanitizer is also not completely compatible with OpenMP, but in any case, I think
````
==================
WARNING: ThreadSanitizer: data race (pid=2839)
  Write of size 1 at 0x7d880009488b by thread T5 (mutexes: write M14584):
    #0 Thumbnail::_generateThumbnailImage() /home/adam/Projekte/RawTherapee/rtgui/thumbnail.cc:102 (rawtherapee+0x000000bf9033)
    #1 Thumbnail::upgradeThumbImage(rtengine::procparams::ProcParams const&, int, double&) /home/adam/Projekte/RawTherapee/rtgui/thumbnail.cc:600 (rawtherapee+0x000000bfc355)
    #2 ThumbImageUpdater::Impl::processNextJob() /home/adam/Projekte/RawTherapee/rtgui/thumbimageupdater.cc:164 (rawtherapee+0x000000920ca5)
    #3 sigc::bound_mem_functor0<void, ThumbImageUpdater::Impl>::operator()() const /usr/include/sigc++-2.0/sigc++/functors/mem_fun.h:1989 (rawtherapee+0x00000092251c)
    #4 sigc::adaptor_functor<sigc::bound_mem_functor0<void, ThumbImageUpdater::Impl> >::operator()() const /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:256 (rawtherapee+0x000000922376)
    #5 sigc::internal::slot_call0<sigc::bound_mem_functor0<void, ThumbImageUpdater::Impl>, void>::call_it(sigc::internal::slot_rep*) /usr/include/sigc++-2.0/sigc++/functors/slot.h:115 (rawtherapee+0x000000922041)
    #6 <null> <null> (libglibmm-2.4.so.1+0x00000005cca1)

  Previous read of size 1 at 0x7d880009488b by thread T6 (mutexes: write M5232):
    #0 FileCatalog::previewReady(int, FileBrowserEntry*) /home/adam/Projekte/RawTherapee/rtgui/filecatalog.cc:734 (rawtherapee+0x000000a73349)
    #1 PreviewLoader::Impl::processNextJob() /home/adam/Projekte/RawTherapee/rtgui/previewloader.cc:131 (rawtherapee+0x000000a87fa2)
    #2 sigc::bound_mem_functor0<void, PreviewLoader::Impl>::operator()() const /usr/include/sigc++-2.0/sigc++/functors/mem_fun.h:1989 (rawtherapee+0x000000a8a33e)
    #3 sigc::adaptor_functor<sigc::bound_mem_functor0<void, PreviewLoader::Impl> >::operator()() const /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:256 (rawtherapee+0x000000a89fa6)
    #4 sigc::internal::slot_call0<sigc::bound_mem_functor0<void, PreviewLoader::Impl>, void>::call_it(sigc::internal::slot_rep*) /usr/include/sigc++-2.0/sigc++/functors/slot.h:115 (rawtherapee+0x000000a89bd3)
    #5 <null> <null> (libglibmm-2.4.so.1+0x00000005cca1)

  Location is heap block of size 6120 at 0x7d8800094800 allocated by thread T6:
    #0 operator new(unsigned long) /build/gcc-multilib/src/gcc-5-20160209/libsanitizer/tsan/tsan_interceptors.cc:571 (libtsan.so.0+0x000000025993)
    #1 CacheManager::getEntry(Glib::ustring const&) /home/adam/Projekte/RawTherapee/rtgui/cachemanager.cc:114 (rawtherapee+0x0000008f99af)
    #2 PreviewLoader::Impl::processNextJob() /home/adam/Projekte/RawTherapee/rtgui/previewloader.cc:126 (rawtherapee+0x000000a87ef0)
    #3 sigc::bound_mem_functor0<void, PreviewLoader::Impl>::operator()() const /usr/include/sigc++-2.0/sigc++/functors/mem_fun.h:1989 (rawtherapee+0x000000a8a33e)
    #4 sigc::adaptor_functor<sigc::bound_mem_functor0<void, PreviewLoader::Impl> >::operator()() const /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:256 (rawtherapee+0x000000a89fa6)
    #5 sigc::internal::slot_call0<sigc::bound_mem_functor0<void, PreviewLoader::Impl>, void>::call_it(sigc::internal::slot_rep*) /usr/include/sigc++-2.0/sigc++/functors/slot.h:115 (rawtherapee+0x000000a89bd3)
    #6 <null> <null> (libglibmm-2.4.so.1+0x00000005cca1)

  Mutex M14584 (0x7d0c00035fd0) created at:
    #0 pthread_mutex_init /build/gcc-multilib/src/gcc-5-20160209/libsanitizer/tsan/tsan_interceptors.cc:1079 (libtsan.so.0+0x000000028015)
    #1 <null> <null> (libglib-2.0.so.0+0x00000008de67)
    #2 Thumbnail::Thumbnail(CacheManager*, Glib::ustring const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) /home/adam/Projekte/RawTherapee/rtgui/thumbnail.cc:67 (rawtherapee+0x000000bf8b48)
    #3 CacheManager::getEntry(Glib::ustring const&) /home/adam/Projekte/RawTherapee/rtgui/cachemanager.cc:114 (rawtherapee+0x0000008f99d2)
    #4 PreviewLoader::Impl::processNextJob() /home/adam/Projekte/RawTherapee/rtgui/previewloader.cc:126 (rawtherapee+0x000000a87ef0)
    #5 sigc::bound_mem_functor0<void, PreviewLoader::Impl>::operator()() const /usr/include/sigc++-2.0/sigc++/functors/mem_fun.h:1989 (rawtherapee+0x000000a8a33e)
    #6 sigc::adaptor_functor<sigc::bound_mem_functor0<void, PreviewLoader::Impl> >::operator()() const /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:256 (rawtherapee+0x000000a89fa6)
    #7 sigc::internal::slot_call0<sigc::bound_mem_functor0<void, PreviewLoader::Impl>, void>::call_it(sigc::internal::slot_rep*) /usr/include/sigc++-2.0/sigc++/functors/slot.h:115 (rawtherapee+0x000000a89bd3)
    #8 <null> <null> (libglibmm-2.4.so.1+0x00000005cca1)

  Mutex M5232 (0x7d0c00053010) created at:
    #0 pthread_mutex_init /build/gcc-multilib/src/gcc-5-20160209/libsanitizer/tsan/tsan_interceptors.cc:1079 (libtsan.so.0+0x000000028015)
    #1 <null> <null> (libglib-2.0.so.0+0x00000008de67)
    #2 FileCatalog::FileCatalog(CoarsePanel*, ToolBar*, FilePanel*) /home/adam/Projekte/RawTherapee/rtgui/filecatalog.cc:54 (rawtherapee+0x000000a65754)
    #3 FilePanel::FilePanel() /home/adam/Projekte/RawTherapee/rtgui/filepanel.cc:56 (rawtherapee+0x00000085a2f1)
    #4 RTWindow::RTWindow() /home/adam/Projekte/RawTherapee/rtgui/rtwindow.cc:166 (rawtherapee+0x0000008332ac)
    #5 main /home/adam/Projekte/RawTherapee/rtgui/main.cc:347 (rawtherapee+0x000000ae7c0b)

  Thread T5 'pool' (tid=2868, running) created by main thread at:
    #0 pthread_create /build/gcc-multilib/src/gcc-5-20160209/libsanitizer/tsan/tsan_interceptors.cc:895 (libtsan.so.0+0x0000000274d7)
    #1 <null> <null> (libglib-2.0.so.0+0x00000008e40f)
    #2 ToolPanelCoordinator::dirSelected(Glib::ustring const&, Glib::ustring const&) /home/adam/Projekte/RawTherapee/rtgui/toolpanelcoord.cc:901 (rawtherapee+0x000000c3ac5c)
    #3 sigc::bound_mem_functor2<void, BatchToolPanelCoordinator, Glib::ustring const&, Glib::ustring const&>::operator()(Glib::ustring const&, Glib::ustring const&) const <null> (rawtherapee+0x0000008682ee)
    #4 sigc::adaptor_functor<sigc::bound_mem_functor2<void, BatchToolPanelCoordinator, Glib::ustring const&, Glib::ustring const&> >::deduce_result_type<Glib::ustring const&, Glib::ustring const&, void, void, void, void, void>::type sigc::adaptor_functor<sigc::bound_mem_functor2<void, BatchToolPanelCoordinator, Glib::ustring const&, Glib::ustring const&> >::operator()<Glib::ustring const&, Glib::ustring const&>(Glib::ustring const&, Glib::ustring const&) const <null> (rawtherapee+0x000000866c19)
    #5 sigc::internal::slot_call2<sigc::bound_mem_functor2<void, BatchToolPanelCoordinator, Glib::ustring const&, Glib::ustring const&>, void, Glib::ustring const&, Glib::ustring const&>::call_it(sigc::internal::slot_rep*, Glib::ustring const&, Glib::ustring const&) <null> (rawtherapee+0x000000864ea2)
    #6 sigc::internal::signal_emit2<void, Glib::ustring const&, Glib::ustring const&, sigc::nil>::emit(sigc::internal::signal_impl*, Glib::ustring const&, Glib::ustring const&) /usr/include/sigc++-2.0/sigc++/signal.h:1262 (rawtherapee+0x000000a35c85)
    #7 sigc::signal2<void, Glib::ustring const&, Glib::ustring const&, sigc::nil>::emit(Glib::ustring const&, Glib::ustring const&) const /usr/include/sigc++-2.0/sigc++/signal.h:2985 (rawtherapee+0x000000a35299)
    #8 sigc::signal2<void, Glib::ustring const&, Glib::ustring const&, sigc::nil>::operator()(Glib::ustring const&, Glib::ustring const&) const /usr/include/sigc++-2.0/sigc++/signal.h:2993 (rawtherapee+0x000000a34a39)
    #9 DirBrowser::open(Glib::ustring const&, Glib::ustring const&) /home/adam/Projekte/RawTherapee/rtgui/dirbrowser.cc:473 (rawtherapee+0x000000a32443)
    #10 FilePanel::init() /home/adam/Projekte/RawTherapee/rtgui/filepanel.cc:189 (rawtherapee+0x00000085c5ea)
    #11 FilePanelInitUI(void*) /home/adam/Projekte/RawTherapee/rtgui/filepanel.cc:27 (rawtherapee+0x000000857d02)
    #12 g_main_context_dispatch <null> (libglib-2.0.so.0+0x000000049c39)
    #13 __libc_start_main <null> (libc.so.6+0x00000002070f)

  Thread T6 'pool' (tid=2864, running) created by main thread at:
    #0 pthread_create /build/gcc-multilib/src/gcc-5-20160209/libsanitizer/tsan/tsan_interceptors.cc:895 (libtsan.so.0+0x0000000274d7)
    #1 <null> <null> (libglib-2.0.so.0+0x00000008e40f)
    #2 main /home/adam/Projekte/RawTherapee/rtgui/main.cc:347 (rawtherapee+0x000000ae7c0b)

SUMMARY: ThreadSanitizer: data race /home/adam/Projekte/RawTherapee/rtgui/thumbnail.cc:102 Thumbnail::_generateThumbnailImage()
==================
````
is currently the best fit to @Hombre57's observation.

**Comment by Hombre57** [comment]
Adding the following line 602 of rtgui/thumbnail.cc will solve the problem :
```
    if (cfs.supported) {
        _loadThumbnail (false);
    }
```

I don't know if the test is mandatory, but I've replicated the same code than in `Thumbnail::_loadThumbnail` where is calls `_generateThumbnailImage` and `_loadThumbnail` right after. Depending on files and CPU power (i.e. race condition), RT might show an empty thumbnail until the full process is done because the lock is acquired before the quick image can display itself. That's what happened when debuging with all the printf here, but is working fine once cleaned up.

With this change, `_loadThumbnail` is called twice per image when the cache is empty, and twice too (like before, see my PDF above) with a valid cache.

Last note : you can also remove ` && firstTrial` from `_loadThumbnail`, since the test is already done. Nothing important though.

adamreichold:
@Hombre57 Thank you for pouring your limited spare time into this! I have implemented your suggestions (adding the additional call to `_loadThumbnail` and removing the `firstTrail` dance) and it does seem to work now. But frankly, I think we swapped one kind of non-deterministic behaviour against another but I would be lying if I claimed to understand this completely...
AFAIU this currently, we are actually synchronizing concurrent access to the cache via the file system which seems at least problematic. And also this means that this worked for now only by chance as this extra call should have been there all along?

@Beep6581 @morosakubek Can you give the latest version another try? Thank you. Please also have a closer look at the painting of the thumbnails in the file browser and the batch queue as I also found unnecessary copies of the image data in those components which I tried to avoid by always passing around `IImage8*` instead of bare `guint8*`.

Beep6581:
Hey! I'd love to but I'm leaving for Scotland in the morning and must pack... I'll be back and able to test next week.

Hombre57:
@adamreichold  Please stand by, still investigating... `_loadThumbnail` is the only place where we read the tif file, so I'd like to see if we can avoid reading it again from disk since `loadFromRaw` populates the `Image16` correctly. `tpp->thumbImg` must be deleted somewhere, we should avoid this I think.

adamreichold:
@Hombre57 Are you sure that `tpp->thumbImg` is deleted? When debugging this, it looked like the image was still there but the auto exposure histogram was different. Wouldn't this fit with overexposure?

Hombre57:
> Are you sure that tpp->thumbImg is deleted?

No, it was just guessing.

> When debugging this, it looked like the image was still there but the auto exposure histogram was different. Wouldn't this fit with overexposure?

Probably. I'm done with this patch, I can't really understand the call chain that bring this erroneous result. If you can fix this Histogram bug (if any), then my patch above is useless, I suppose.

Hombre57:
> When debugging this, it looked like the image was still there but the auto exposure histogram was different. Wouldn't this fit with overexposure?

One more thing : I rarely use Auto-exposure, and the images are still overexposed, so I'm not 100% sure that the Histogram is the culprit, but just fix that and we'll know.

adamreichold:
@Hombre57 You are right, the artifacts are completely independent of the auto-exposure settings... I also completely agree that the call chain here is beyond messy and the only proper solution might be to rewrite... But maybe we should take a few days off from this and have look again later, but for now your changes are the only thing that apparently makes this work even though I do not really understand why. :-\

heckflosse:
@adamreichold @Hombre57 : I'll recall that I've stumbled upon the Auto exposure histogram stuff some months ago. I'll take a look at it.

Ingo

heckflosse:
I also would like to reduce the accesses to RT cache from two source files (cachemanager.cc and thumbnail.cc) to one source file... But that's nothing new so most probably not related to the artifacts.

Hombre57:
[off-topic, but you have typed the "rewrite" word :) ]
Just for your information : speaking of performance, the Batch Editor Panel is actually optimized for low memory consumption, i.e. if you are in a slider adjustment operation, RT loads the tif file of aaaaalll the selected images (one after the other, but multi-threaded), process them and flush them at each update of the thumbnails, no matter that you have selected 5 files or 500 files. This process makes adjustments through the Batch Editor quite slow. But it's the safer option.
[/off-topic]

heckflosse:
@Hombre57 Is your last statement true for master or only for this branch? I just tried in 'master' selecting three files out of a folder containing 250 files and adjusting exposure slider. Can't confirm it being slow.

Ingo

Hombre57:
@heckflosse True for both branch. I'm testing here with 17 files : while adjusting the Exposure slider, I can clearly see that the thumbnails are updated one after the other (I'd say it takes 0.6s per update for the 17 images). For my machine and considering the size of the images, it should be in the 0.25s range or less if it didn't had to reload the tif images each time. But that's another issue, I'm just saying to keep that in mind if the update mechanism had to be updated.

@adamreichold @heckflosse I just discovered that if you select an image and select the "Profile operation  /  Reset profil" menu, the erroneous thumbnails come back.

Also, if you look closely at the images, you'll see that they are composed of primary colors only, i.e. the components are either 0 or 255. This looks more like a range conversion problem than overexposure. But still, I agree that there's no apparent reason why it happen. Ok for setting this back for some days.

heckflosse:
@Hombre57 You modified 17 files from file browser, ok. But does the time modifying this 17 files really depends on the total amount of the files in the folder as you mentioned before?

Hombre57:
Huh ! I was wrong, the values are not either 0 or 255, there are some intermediate values, but what's most surprising is that the very first column is correct!! Just capture the thumbnail image and zoom in by 400%.

Hombre57:
@heckflosse I didn't said that. I said it was depending on the number of *selected* file, i.e. edited files. We could assume a threshold where editing 20 files at the same time does not flush the image data (until you switch tab or something like that, like for the Inspector tool), while editing more images would flush the image data at each update.

Hombre57:
For a 250x250 pix thumbnail, 20 thumb of float data would represent 15MB. This could be a performance enhancement in the future.

heckflosse:
@Hombre57 The 'no matter that you have selected 5 files or 500 files' was a quote from your post...
Anyway, I'll investigate further and try to help getting a good solution (btw: I still think that we don't need to save the auto exposure histogram (32 kb of data) for each thumb. It should be enough to calculate the auto wb factors from histogram and save them (much less data) into data area of rt-cache instead (which would also mean one less cache file)

heckflosse:
@Hombre57 Now I understood your 'no matter...' It was a bit misleading. Never mind

heckflosse:
I had a look at the auto exposure histogram cache. We only need it to be able to set Auto levels clip % from file browser. For me that seems waste of disk space...

Beep6581:
@adamreichold hello, what is the status here, should we test and merge?

heckflosse:
What shall we do with this stale pr?

agriggio:
seems that @adamreichold is not around anymore, is he? In that case, probably this should wait at least until 5.1 is out?

Beep6581:
I've been in touch with Adam, he said that he enjoyed working on RT and would like to restart doing so, but he won't be able to do that in the near to intermediate future.

heckflosse:
I think, we should close this. Any objections?

PR #3239, opened 2016-04-03
Merge xtrans speedups into master

heckflosse:


heckflosse:
Branch stays open for further speedups

Issue #2704, opened 2015-08-11
Serialze read of tiff files


Beep6581:
Originally reported on Google Code with ID 2722

```
from here: http://rawtherapee.com/forum/viewtopic.php?f=1&t=5949

When you open a folder with tiffs first time in rt, rt has to read all the tiffs to
build the thumbs. It turned out, that reading tiffs in parallel is not the best idea.

For reference: 
http://rawtherapee.com/forum/viewtopic.php?f=1&t=5949#p41299
http://rawtherapee.com/forum/viewtopic.php?f=1&t=5949#p41334

This patch serializes read of tiff (and other non-raw) files.
It only serializes the read part, not the thumb generation part.

Would be great if someone could test this on Linux too.

Ingo
```

Reported by `heckflosse@i-weyrich.de` on 2015-03-24 11:48:32

<hr>
- _Attachment: [serialize_tiff_for_thumbs_reads.patch](https://storage.googleapis.com/google-code-attachments/rawtherapee/issue-2722/comment-0/serialize_tiff_for_thumbs_reads.patch)_

Issue #2259, opened 2015-08-11
Black-and-White images still have some color


Beep6581:
Originally reported on Google Code with ID 2275

```
As reported in the forum, black-and-white images (using the Black-and-White tool or
-saturation or -chroma) are not completely black-and-white - instead of each pixel
having the same R, G and B values, sometimes there are tiny differences. You can use
imagemagick's "identify -verbose" command on a saved image to get info.
http://rawtherapee.com/forum/viewtopic.php?f=1&t=5307#p36122

Opened issue for a professional opinion :]
```

Reported by `entertheyoni` on 2014-03-03 13:07:45

PR #3244, opened 2016-04-07
Save 8 bit b/w images using PseudoGrey Plus

Floessie:
I rediscovered [Pat David's article on PseudoGrey](http://blog.patdavid.net/2012/06/true-pseudogrey-in-gimp.html) recently and came across PseudoGrey Plus on Rich Franzen's site. I implemented a little stand-alone converter first and tried it successfully on 16 bit TIFFs exported from RT. Now here's an implementation directly for RT. It might be a nice (though small) USP for what is already an excellent piece of software.

[PseudoGrey Plus](http://r0k.us/graphics/PseudoGreyPlus.html) by Rich Franzen is an algorithm to store 3110 levels of gray in 24 bit color images. This is simply done by adding 1 or 2 to the R, G, and/or B values to adjust the pixel's luminance. It's a subtle but perceivable effect and helps to prevent banding.

This implementation adds a new method `isBW()` to `IImage`, which is only implemented for `Image16` and defaults to `false`. It's only called once per saving and has a shortcut to determine that the image is *not* b/w.

The minimally invasive core algorithm was added to `rtengine/imageio.cc` and is only called when saving an 8 bit image that is b/w.

Please note that when saving as JPEG, compression squashes the effect and a quality of 95+ and 4:4:4 subsampling is recommended. Many PseudoGrey sample image on the net suffer from this and do not really show the benefit.

### DIFF
--- a/rtengine/iimage.h
+++ b/rtengine/iimage.h
@@ -98,6 +98,10 @@ public:
         return "unknown";
     }
 
+    virtual bool isBW () const
+    {
+        return false;
+    }
 };
 
 template <>
--- a/rtengine/iimage.h
+++ b/rtengine/iimage.h
@@ -67,41 +67,30 @@ public:
     // parameters that will never be used, replaced by the subclasses r, g and b parameters!
     // they are still necessary to implement operator() in this parent class
     virtual ~ImageDatas() {}
-    virtual void allocate (int W, int H) {}
-    virtual void rotate (int deg) {}
+    virtual void allocate (int W, int H);
+    virtual void rotate (int deg);
     // free the memory allocated for the image data without deleting the object.
-    virtual void flushData ()
-    {
-        allocate(0, 0);
-    }
+    virtual void flushData ();
 
-    virtual void hflip () {}
-    virtual void vflip () {}
+    virtual void hflip ();
+    virtual void vflip ();
 
     // Read the raw dump of the data
-    void readData  (FILE *fh) {}
+    void readData  (FILE *fh);
     // Write a raw dump of the data
-    void writeData (FILE *fh) {}
+    void writeData (FILE *fh);
 
-    virtual void normalizeInt (int srcMinVal, int srcMaxVal) {};
-    virtual void normalizeFloat (float srcMinVal, float srcMaxVal) {};
-    virtual void computeHistogramAutoWB (double &avg_r, double &avg_g, double &avg_b, int &n, LUTu &histogram, int compression) {}
+    virtual void normalizeInt (int srcMinVal, int srcMaxVal);
+    virtual void normalizeFloat (float srcMinVal, float srcMaxVal);
+    virtual void computeHistogramAutoWB (double &avg_r, double &avg_g, double &avg_b, int &n, LUTu &histogram, int compression);
     virtual void getSpotWBData (double &reds, double &greens, double &blues, int &rn, int &gn, int &bn,
                                 std::vector<Coord2D> &red, std::vector<Coord2D> &green, std::vector<Coord2D> &blue,
-                                int tran) {}
-    virtual void getAutoWBMultipliers (double &rm, double &gm, double &bm)
-    {
-        rm = gm = bm = 1.0;
-    }
-    virtual const char* getType () const
-    {
-        return "unknown";
-    }
+                                int tran);
+    virtual void getAutoWBMultipliers (double &rm, double &gm, double &bm);
 
-    virtual bool isBW () const
-    {
-        return false;
-    }
+    virtual const char* getType () const;
+
+    virtual bool isBW () const;
 };
 
 template <>
--- a/rtengine/image16.cc
+++ b/rtengine/image16.cc
@@ -284,6 +284,46 @@ void Image16::getStdImage (ColorTemp ctemp, int tran, Imagefloat* image, Preview
 #undef GCLIP
 }
 
+bool
+Image16::isBW() const
+{
+    if (!height || !width) {
+        return false;
+    }
+
+    bool res = true;
+
+    for ( int w = 0, h = height / 2; res && w < width; ++w ) {
+        res =
+            res
+            && (
+                r(h, w) == g(h, w)
+                || r(h, w) == b(h, w)
+            );
+    }
+
+    if (!res) {
+        return false;
+    }
+
+#ifdef _OPENMP
+    #pragma omp parallel for reduction(&&:res)
+#endif
+
+    for ( int h = 0; h < height; ++h ) {
+        for ( int w = 0; w < width; ++w ) {
+            res =
+                res
+                && (
+                    r(h, w) == g(h, w)
+                    || r(h, w) == b(h, w)
+                );
+        }
+    }
+
+    return res;
+}
+
 Image8*
 Image16::to8()
 {
--- a/rtengine/image16.cc
+++ b/rtengine/image16.cc
@@ -291,27 +291,19 @@ Image16::isBW() const
         return false;
     }
 
+    for (int w = 0, h = height / 2; w < width; ++w) {
+        if (r(h, w) != g(h, w) || r(h, w) != b(h, w)) {
+            return false;
+        }
+    }
+
     bool res = true;
-
-    for ( int w = 0, h = height / 2; res && w < width; ++w ) {
-        res =
-            res
-            && (
-                r(h, w) == g(h, w)
-                || r(h, w) == b(h, w)
-            );
-    }
-
-    if (!res) {
-        return false;
-    }
 
 #ifdef _OPENMP
     #pragma omp parallel for reduction(&&:res)
 #endif
-
-    for ( int h = 0; h < height; ++h ) {
-        for ( int w = 0; w < width; ++w ) {
+    for (int h = 0; h < height; ++h) {
+        for (int w = 0; w < width; ++w) {
             res =
                 res
                 && (
--- a/rtengine/imageio.cc
+++ b/rtengine/imageio.cc
@@ -23,6 +23,7 @@
 #include <tiffio.h>
 #include <cstdio>
 #include <cstring>
+#include <cstdint>
 #include <memory>
 #include <fcntl.h>
 #include <libiptcdata/iptc-jpeg.h>
@@ -85,18 +86,11 @@ Glib::ustring to_utf8 (const std::string& str)
     }
 }
 
-unsigned short addSat (unsigned short a, unsigned short b)
-{
-    unsigned short res = a + b;
-    res |= -(res < a);
-    return res;
-}
-
-void scanlineToPseudoGrey (const unsigned short *in16, unsigned char *out8, unsigned long width)
+void scanlineToPseudoGrey (const std::uint16_t *in16, unsigned char *out8, unsigned long width)
 {
     // For details see http://r0k.us/graphics/PseudoGreyPlus.html
 
-    static const unsigned short plusses[16][3] = {
+    static const std::uint16_t plusses[16][3] = {
         {0 << 8, 0 << 8, 0 << 8},
         {0 << 8, 0 << 8, 1 << 8},
         {0 << 8, 0 << 8, 2 << 8},
@@ -116,9 +110,9 @@ void scanlineToPseudoGrey (const unsigned short *in16, unsigned char *out8, unsi
     };
 
     for (unsigned long x = 0; x < width * 3; x += 3) {
-        out8[x + 0] = addSat(in16[x + 0], plusses[in16[x + 0] >> 4 & 0x0F][0]) >> 8;
-        out8[x + 1] = addSat(in16[x + 1], plusses[in16[x + 1] >> 4 & 0x0F][1]) >> 8;
-        out8[x + 2] = addSat(in16[x + 2], plusses[in16[x + 2] >> 4 & 0x0F][2]) >> 8;
+        out8[x + 0] = adds(in16[x + 0], plusses[in16[x + 0] >> 4 & 0x0F][0]) >> 8;
+        out8[x + 1] = adds(in16[x + 1], plusses[in16[x + 1] >> 4 & 0x0F][1]) >> 8;
+        out8[x + 2] = adds(in16[x + 2], plusses[in16[x + 2] >> 4 & 0x0F][2]) >> 8;
     }
 }
 
@@ -954,7 +948,7 @@ int ImageIO::loadPPMFromMemory(const char* buffer, int width, int height, bool s
     return IMIO_SUCCESS;
 }
 
-int ImageIO::savePNG  (Glib::ustring fname, int compression, volatile int bps)
+int ImageIO::savePNG  (Glib::ustring fname, int compression, int bps)
 {
 
     FILE *file = g_fopen_withBinaryAndLock (fname);
@@ -1008,9 +1002,9 @@ int ImageIO::savePNG  (Glib::ustring fname, int compression, volatile int bps)
 
     const bool do_pseudogrey = bps == 8 && isBW();
 
-    const std::unique_ptr<unsigned short[]> row16(
+    const std::unique_ptr<std::uint16_t[]> row16(
         do_pseudogrey
-            ? new unsigned short[width * 3]
+            ? new std::uint16_t[width * 3]
             : nullptr
     );
 
@@ -1199,9 +1193,9 @@ int ImageIO::saveJPEG (Glib::ustring fname, int quality, int subSamp)
 
     const bool do_pseudogrey = isBW();
 
-    const std::unique_ptr<unsigned short[]> row16(
+    const std::unique_ptr<std::uint16_t[]> row16(
         do_pseudogrey
-            ? new unsigned short[width * 3]
+            ? new std::uint16_t[width * 3]
             : nullptr
     );
 
@@ -1277,9 +1271,9 @@ int ImageIO::saveTIFF (Glib::ustring fname, int bps, bool uncompressed)
 
     const bool do_pseudogrey = bps == 8 && isBW();
 
-    const std::unique_ptr<unsigned short[]> row16(
+    const std::unique_ptr<std::uint16_t[]> row16(
         do_pseudogrey
-            ? new unsigned short[width * 3]
+            ? new std::uint16_t[width * 3]
             : nullptr
     );
 
--- a/rtengine/imageio.cc
+++ b/rtengine/imageio.cc
@@ -23,6 +23,7 @@
 #include <tiffio.h>
 #include <cstdio>
 #include <cstring>
+#include <memory>
 #include <fcntl.h>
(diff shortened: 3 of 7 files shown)


rtengine/imageio.cc
```diff
@@ -83,6 +85,43 @@ Glib::ustring to_utf8 (const std::string& str)
...
 }
 
+unsigned short addSat (unsigned short a, unsigned short b)
+{
+    unsigned short res = a + b;
+    res |= -(res < a);
+    return res;
+}
+
+void scanlineToPseudoGrey (const unsigned short *in16, unsigned char *out8, unsigned long width)
```
adamreichold: Wouldn't it be more appropriate to use the fixed-width integer types from the `cstdint` header here?

rtengine/imageio.cc
```diff
@@ -83,6 +85,43 @@ Glib::ustring to_utf8 (const std::string& str)
...
+        {2 << 8, 0 << 8, 2 << 8},
+        {2 << 8, 0 << 8, 2 << 8},
+        {0 << 8, 1 << 8, 0 << 8},
+        {0 << 8, 1 << 8, 1 << 8},
+        {0 << 8, 1 << 8, 2 << 8},
+        {0 << 8, 1 << 8, 2 << 8},
+        {1 << 8, 1 << 8, 0 << 8}
+    };
+
+    for (unsigned long x = 0; x < width * 3; x += 3) {
```
adamreichold: Could probably benefit from using pointer arithmetic instead of indexing? But maybe the compiler will do that itself...

rtengine/imageio.cc
```diff
@@ -83,6 +85,43 @@ Glib::ustring to_utf8 (const std::string& str)
     }
 }
 
+unsigned short addSat (unsigned short a, unsigned short b)
```
adamreichold: Isn't there an GCC intrinsic for saturated addition? I think it was SSE or MMX. In any case, this probably wants to go into a separate header otherwise. Maybe `rt_math.h`?

rtengine/iimage.h
```diff
@@ -98,6 +98,10 @@ class ImageDatas : virtual public ImageDimensions
         return "unknown";
     }
 
+    virtual bool isBW () const
```
adamreichold: I know it is currently done everywhere, but I think inlining the virtuals implies vague linkage? Shouldn't it just be moved into `iimage.cc`?

rtengine/image16.cc
```diff
@@ -284,6 +284,46 @@ void Image16::getStdImage (ColorTemp ctemp, int tran, Imagefloat* image, Preview
...
+    }
+
+    bool res = true;
+
+    for ( int w = 0, h = height / 2; res && w < width; ++w ) {
+        res =
+            res
+            && (
+                r(h, w) == g(h, w)
+                || r(h, w) == b(h, w)
```
adamreichold: Maybe I am misreading this but I would have expected `r == g && r == b` instead of `r == g || r == b`? Also the loop would probably look simpler with an early return?

rtengine/imageio.cc
```diff
@@ -964,14 +1003,26 @@ int ImageIO::savePNG  (Glib::ustring fname, int compression, volatile int bps)
     png_set_IHDR(png, info, width, height, bps, PNG_COLOR_TYPE_RGB,
                  PNG_INTERLACE_NONE, PNG_COMPRESSION_TYPE_DEFAULT, PNG_FILTER_TYPE_BASE);
 
-
     int rowlen = width * 3 * bps / 8;
     unsigned char *row = new unsigned char [rowlen];
 
+    const bool do_pseudogrey = bps == 8 && isBW();
```
adamreichold: It has of course nothing to do with this change, but why the heck is `bps` volatile?

rtengine/image16.cc
heckflosse: @adamreichold I completely agree with you that an early return could reduce the processing time of this loop to about zero in case the image is not b&w!
for ( int w = 0; w < width && res; ++w ) {
would do the job near optimal for this case.
I didn't measure processing time, but for the case of a real b&w image where the loop has to run for all pixels, avoiding the logical operators (which usually introduce branches) should be preferred. I try to post a faster loop soon

rtengine/imageio.cc
heckflosse: @adamreichold IIRC there is an intrinsic for saturated addition. I'll have a look

rtengine/imageio.cc
Floessie: Of course. But as I saw no `<cstdint>` include but the use of `unsigned short` I went for the latter.

rtengine/imageio.cc
Floessie: I guess that's equivalent these days and for readability I prefer indexing.

rtengine/imageio.cc
Floessie: @heckflosse The whole loop could well be manually vectorized with `_mm_adds_pu16`. I didn't have a look at the assemby yet, but my guess is, the auto-vectorizer came to the same conclusion.

rtengine/iimage.h
Floessie: Inlining virtuals is always not so smart (except for the d'tor) and is not my style, either. I tried to adapt to the surrounding code.

But why would vague linkage bother something other than a library? Time to link?

rtengine/image16.cc
Floessie: @adamreichold Yeah, good catch. That's definitely a bug. I'll fix that. And go for the early return as well.

rtengine/imageio.cc
adamreichold: So now reason to repeat mistakes of the past? I would really prefer this to use fixed-width types.

rtengine/imageio.cc
adamreichold: Agreed, let's have the compiler take care of that one.

rtengine/imageio.cc
Floessie: :) There are still things left to improve in RT and you are well on to it.

rtengine/iimage.h
adamreichold: I guess for the same reason we use the anonymous namespace, to make the problem we pose to the linker as simple as possible by presenting the most restricted set of symbols we can work with so that it can concentrate on producing the most efficient binary.
But for RT in particular, yeah definitely also build time. :-( A full debug build already needs north of six minutes on my machine, so I a crave for every compile and link time reduction. And this only gets worse with LTO enabled which however is sadly missed for an "organically grown" code base like RT's.

rtengine/imageio.cc
adamreichold: I think we should really check the generated assembly here (but only here) with different versions of GCC and then use that knowledge to decide whether saturated addition should get the @heckflosse treatment.
If GCC 4.9 is already properly vectorizing the loop then there is probably no use to it, but I'd still argue that saturated addition would fit well into `rt_math.h` or a similar generic header.

rtengine/imageio.cc
Floessie: If it's okay for you, then it's okay for me. :) I'll change that to the better.

rtengine/image16.cc
adamreichold: @heckflosse I am not sure if there really is a difference after an optimizing compiler and branch prediction are through with this, for me the early return is strictly about readability. As always, the measurements decide if there is a signal to be measured here.

rtengine/iimage.h
Floessie: Agreed. Though my workstation is a bit faster. ;)

Why isn't LTO enabled by default? I had positive experience with it at work, where we go for `-Os`.

rtengine/image16.cc
heckflosse: @adamreichold My idea was to include the early return check in the inner loop of parallel region. That should break the loop quite early for non b&w images and avoids extra code for checking the middle line. Of course it could happen that one core gets an area with only b&w pixels and but that's really unlikely for non b&w images.

heckflosse: I tested an [arithmetic version of the loop](https://gist.github.com/heckflosse/efb2b3cb5ac229764618a23a2040cf0b). It's about 12 % faster, but as we are in 16 bit integer space here it's already very fast in original version. On my machine for a full size 50 Mp output it needs 32 ms in original version and 28 ms in arithmetic version.

rtengine/imageio.cc
heckflosse: The tricky part in a vectorized version will be the access to the 'plusses' lookup table which also is the reason why (according to current processing time) it does not get auto vectorized. But I have to look at the assembly to verify

rtengine/rt_math.h
```diff
@@ -101,11 +102,23 @@ T norm2(const T& x, const T& y)
...
+template<typename T>
 T norminf(const T& x, const T& y)
 {
     return std::max(std::abs(x), std::abs(y));
 }
 
+template<
+    typename T,
+    typename = typename std::enable_if<std::is_integral<T>::value>::type,
+    typename = typename std::enable_if<std::is_unsigned<T>::value>::type
```
adamreichold: Nice one! :-)

rtengine/iimage.h
adamreichold: @Floessie I agree that LTO should be on by default. I also had favorable experiences using GCC and LTO in embedded projects. But I also had GCC fail horribly on me, actually especially using `-Os`, by removing stuff but actually needing it later. It seemed to depend very much on the GCC version at the time and hopefully everything is now fine.
I think a good start would be to ask package maintainers to enable it where recent GCC versions are used? For example we could ask @mbajor if he could enable it for his packages on OpenSUSE 42.1 when building with GCC 5.3? Just didn't find the time yet...

heckflosse:
I had a look at processing times. 'scanlineToPseudoGrey' adds about 10 to 15% to the processing time of the jpeg save loop. I think that's fine as it is now. There are still more important parts to speed up in rt ;-)

Ingi

heckflosse:
Currently it seems like pseudogray implementation is mandatory for 8-bit per channel output. We really should have to make it optional for backwards compatibility and for those who want really monochrome output. Any Ideas where to place the gui control (checkbox) are welcome

adamreichold:
@heckflosse @Floessie AFAIU, the only consumers that would see reduced quality due to this are people you are doing numerical processing with the resulting images, i.e. scientific applications. But I doubt that all the image processing heuristics employed within RT make it suitable for that task anyway.
Is there any downside to doing this with images intended for human consumption? If so, I think the performance & quality tab of the preferences dialog would be suitable. But even then, I would probably enable it by default.
While we are there: There is an option to "serialize" TIFF image I/O which maps to `options.serializeTiffRead` in the code which leads to a static mutex being taken in `ImageIO::loadTIFF`. Isn't this `PROTECT_VECTORS` all over again? Meaning either the code is correct without locking or it isn't. Or are there really different libTIFF variants in use, some of which are thread-safe and some are not?

Floessie:
@heckflosse: I have to hook up with @adamreichold in this matter. But if there has to be a checkbox, I would expect to see it in the "Save" dialogs below or besides the target format selection and only enabled for 8 bit formats. Or in the place, @adamreichold mentioned.

heckflosse:
@adamreichold About the serialized tiff read: https://github.com/Beep6581/RawTherapee/issues/2704

heckflosse:
@adamreichold @Floessie There was an Issue some time ago about b&w images not being completely monochrome. For that reason we have [this force b&w code](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/simpleprocess.cc#L1377). For me an option in preferences would do it. I'm also fine with enabling it by default.

adamreichold:
@heckflosse The forum links from #2704 seem to be dead so I can follow up on the story. Was it about performance or about correctness? And from the comments I wonder under what circumstances one would disable the setting?

adamreichold:
@heckflosse Besides the question on in the intention of the issue about images not being mathematically grayscale and the quite heavy handed solution of forcing the pixel values, shouldn't we then rather extend the B&W parameters with an explicit force option which would enable the mention code and disable the pseudo-gray code as well?
But again without that earlier discussion it is hard for me to decide whether we really should do that forcing at all. What were the problems that led to the forcing? Coloured artifacts from other processing steps? The need to numerically post-process?

heckflosse:
@adamreichold I don't know about the reasons for the b&w forcing. But without it, pseudobw isBW() check would not be possible ;-) For that reason, your suggested solution (disabling pseudogray when explicit force option is set) is not possible too.
I'll try to find the Issue which led to the b&w force code.

The SerializeTiffread was about performance. When opening a folder full of tiff files or opening more than one tiff at once in METM, reading one tiff per thread from HD is much slower than the serialized read. On SSD that's not a problem. For that reason I made it optional.

heckflosse:
@adamreichold I found the [Issue](https://github.com/Beep6581/RawTherapee/issues/2259)
Currently 'pseudogray' only works with b&w tool in rt (because only for b&w tool b&w is forced). It would be nice to have pseudogray also for b&w film simulations and the other methods to generate a b&w image.

Floessie:
@heckflosse: An option in preferences sounds like a sane solution.

adamreichold:
@heckflosse While `isBW` does indeed need perfect grayscale input, I rather was thinking of hooking this directly into the B&W effect, i.e. if B&W is enabled but forcing is not, image is always saved using pseudo-gray, if B&W and forcing are enabled, pseudo-gray is not used, and if B&W is disabled pseudo gray is not used (or it could still be done using `isBW` but RT will probably never produce perfectly grayscale images without B&W and forcing enabled anyway?).

One open issue I see here is whether the pseudo-gray algorithm will work properly with practically but not mathematical grayscale input?

Concerning the method of grayscale coercion, I was wondering if we should use something that is perceptibly weighted like [`qGray`](https://doc.qt.io/qt-4.8/qcolor.html#qGray) instead of just setting `r = b = g`?

Concerning the serialized TIFF read, in other I/O heavy processing tools, I have seen options to limit the number of I/O threads used. We could use something similar, at least controlling the number of concurrent I/O tasks we perform, as even SSD have only so many ports internally and also spinning disks can sustain concurrent streams in some RAID configurations? But that's probably a wholly different topic...

adamreichold:
@heckflosse If we make pseudo-gray an explicit choice based on the processing parameters we could also enable it for B&W film simulation as long as we are able to determine if the HaldCLUT profile in use is actually simulating B&W. So we would probably be more flexible in general compared to use something like `isBW`.

Floessie:
@adamreichold: As to the open issue of mathematical grayscale input: PseudoGrey takes only 12 bits into account. The lowest nibble is discarded anyway. So relaxing `isBW()` (or adding the number of significant bits as a parameter) could help.

@heckflosse: If the outcome of film simultation is not absolutely `R == B == G`, how do we know, this wasn't intended by the HaldCLUT? One could test the HaldCLUT for `isBW()`, but that's too extensive for my taste. Why not stick to the `isBW()` metric (perhaps relaxed to significant 12 bits) to keep things simple. PseudoGrey isn't that much of a benefit to kick up a storm.

heckflosse:
@Floessie Testing HaldCLUT for isBW() could be done [without much overhead here](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/clutstore.cc#L185) after loading the file and before colour space conversion. At this point we should know whether the HaldCLUT is intended to be b&w. Processing time of this check should be quite low because size of HaldCLUT files usually is less than 3 Mp. Additionally when checking HaldCLUT for isBW() and not checking the image we want to be b&w for isBW() but making pseudo-gray based on processing parameters (as suggested by @adamreichold) instead we would save more processing time than we need to check isBW() for the HaldCLUT.
@adamreichold About the qGray. AFAIK the current force bw code is made to eliminate very small differences between channels in b&w images. So setting r = b = g should do it. But maybe I'm wrong here.

Ingo

Beep6581:
Hello

As stated already, some users explicitly asked that RawTherapee's grayscale images use R=G=B, something to do with printing, possibly the printer using color ink instead of black when it detects colors in what looks like a grayscale image.

I don't mind the merge, but:
* it would have to be optional, I would expect to see such an option in the Save As window (Grayscale/PseudoGray) where it can be used on a per-image basis rather than in Preferences,
* it would be nice to see a convincing argument for its use. Does having 1786 or 3110 "unique" colors/shades of pseudo-gray vs 256 really matter? I don't see banding in an 8-bit R=G=B grayscale image on the monitor. If the usefulness only comes out when strongly manipulating curves or levels in GIMP, would not salting the image with slight noise in GIMP accomplish a similar result? Who would want to strongly manipulate an 8-bit TIFF image anyway? This won't benefit users saving JPEG for web. It would be nice to see an actual example rather than text.

@adamreichold
> The forum links from #2704 seem to be dead so I cannot follow up on the story.
Try this: http://rawtherapee.com/oldforum/viewtopic.php?f=1&t=5949

There are more ways of achieving B&W than just using the B&W tool. HaldCLUT has already been mentioned, but there's also desaturation via the RGB Saturation slider, via the L*a*b* Saturation slider and CC/CL curves, etc. Also remember that you can use Film Simulation to achieve B&W which you then subsequently color-tone... Putting this in the Save As window seems more simple to me.

Floessie:
Hello DrSlony,

[here](http://filebin.net/nwtbxcphk8) are some files for you to judge. The questions you pose are absolutely justified. Depending on the image I perceived a little effect, but whether or not that's enough for integrating the feature into RT is up to you, of course. Perhaps we could ask Pat, what he thinks about PseudoGrey as he has obviously tested it.

When I first wrote the code, I wondered why RT didn't save grayscale as 8 bit files instead of 24 bit ones and had the impression RT could do more with the 24 bits. As I had read about PseudoGrey on Pat's site long before pixls.us started, I thought, that could be beneficial. Well, later it turned out, that you actively clamp grayscale to R=G=B for the reasons you mentioned above, so I do no longer know if PseudoGrey really has a stand here.

Best,
Flössie

Beep6581:
I took a look at those files. A colorcube analysis shows that the PG version has 3271 unique colors while GS has 255. Despite the numbers, I see no visual difference.

Setting the white point to 80 out of 255 (31%) in a Levels tool while working in 8-bit precision leads to this:
GS: 
scrot_2016-04-13 233937

PG: 
scrot_2016-04-13 233946

Doing the same in 32-bit floating point mode, setting Levels white point to 31%, leads to:
GS: 
scrot_2016-04-13 234947

PG: 
scrot_2016-04-13 235002

Again no visual difference, with the same hair comb in the histogram. As a result, I see no reason for implementing this.

Floessie:
Thanks for your thorough analysis and @adamreichold and @heckflosse for looking into it! So PseudoGrey seemingly doesn't live up to its promise.

See you,
Flössie

Beep6581:
I will run one more test on a downscaled image later today, just because I'm very curious, but it seems like a case of trying very hard to find a problem for a solution instead of a solution which helps with a common problem.

Issue #2567, opened 2015-08-11
Bugfix and Enhancement for Clut


Beep6581:
Originally reported on Google Code with ID 2584

```
In http://rawtherapee.com/forum/posting.php?mode=reply&f=3&t=5746#pr39814
Flössie posted a patch to fix a bug in clut handling, which also increases size of
clut cache from 1 to 8. Thanks a lot for the patch, Flössie :-)

I'll make some changes to this patch to allow the user to set the size of clut cache
in preferences and post a new patch soon.

Ingo
```

Reported by `heckflosse@i-weyrich.de` on 2014-11-20 13:14:20

PR #3256, opened 2016-04-18
HaldCLUT revision and speedup

Floessie:
Here is my revision of the HaldCLUT core which tries to clean up the `clutstore.*` files by integrating a real LRU cache in `cache.h`, using `std::shared_ptr<>` and other C++11 features, and turning the `CLUTStore` class into a real singleton plus there is some speedup and memory reduction by using `Image16` instead of `Imagefloat` to store the CLUT.

I was unable to benchmark the implementation in RT (could need a lesson from @heckflosse), but proved it in my `clutbench` project.

While comparing the output, I realized that every saved file (8 bit uncompressed TIFF in this case) based on the same PP3 ("Default", so independent from film simulation) differs when using multiple threads, but not with `OMP_NUM_THREADS=1`, so there seem to be races somewhere in the pipeline.

difference_with_8_threads

Apart from that, the switch to `Image16` yields only very small differences, and I hope, they are not crucial. I'm not really sure, where they come from, as I didn't see them in my `clutbench` project.

If you give me the green light, I would further de-virtualize the `HaldCLUT` class and get rid of its `CLUT` base, which isn't needed at the moment and probably never will.

Comments welcome (looking at @adamreichold ;) ).

Best,
Flössie

### DIFF
--- a/rtengine/clutstore.cc
+++ b/rtengine/clutstore.cc
@@ -3,440 +3,233 @@
 #include "stdimagesource.h"
 #include "../rtgui/options.h"
 
-rtengine::CLUTStore clutStore;
-
-using namespace rtengine;
-
-const float MAXVAL8 = 255.;
-
-CLUTStore::CLUTStore()
+namespace
 {
-}
 
-CLUT* CLUTStore::getClut( const Glib::ustring& filename )
+std::unique_ptr<rtengine::Imagefloat> loadFile(
+    const Glib::ustring& filename,
+    const Glib::ustring& working_color_space,
+    unsigned int& clut_level
+)
 {
-    CLUT *result = 0;
-    m_mutex.lock();
-    Cluts::iterator cluts_it = m_cluts.find(filename);
+    std::unique_ptr<rtengine::Imagefloat> result;
 
-    if (cluts_it == m_cluts.end()) {
-        if (m_cluts.size() >= options.clutCacheSize) {
-            // Evict a "random" entry from cache
-            Cluts::iterator victim_it = m_cluts.begin();
+    rtengine::StdImageSource img_src;
 
-            if (--victim_it->second.first == -1) {
-                delete victim_it->second.second;
-                m_cluts.erase(victim_it);
-            }
-        }
-
-        cluts_it = m_cluts.insert(std::make_pair(filename, std::make_pair(0, new HaldCLUT))).first;
-        cluts_it->second.second->load( filename );
-    }
-
-    if (cluts_it->second.second->isValid()) {
-        result = cluts_it->second.second;
-        ++cluts_it->second.first;
-    } else {
-        delete cluts_it->second.second;
-        m_cluts.erase(cluts_it);
-    }
-
-    m_mutex.unlock();
-
-    return result;
-}
-
-void CLUTStore::releaseClut( const CLUT* clut )
-{
-    m_mutex.lock();
-
-    for (Cluts::iterator cluts_it = m_cluts.begin(); cluts_it != m_cluts.end(); ++cluts_it) {
-        if (cluts_it->second.second == clut) {
-            if (--cluts_it->second.first == -1) {
-                delete cluts_it->second.second;
-                m_cluts.erase(cluts_it);
-            }
-
-            break;
-        }
-    }
-
-    m_mutex.unlock();
-}
-
-void CLUTStore::clearCache()
-{
-    m_mutex.lock();
-
-    for (Cluts::iterator cluts_it = m_cluts.begin(); cluts_it != m_cluts.end();) {
-        if (--cluts_it->second.first == -1) {
-            delete cluts_it->second.second;
-            Cluts::iterator tmp = cluts_it;
-            ++cluts_it;
-            m_cluts.erase(tmp);
-        } else {
-            ++cluts_it;
-        }
-    }
-
-    m_mutex.unlock();
-}
-
-void rtengine::splitClutFilename( Glib::ustring filename, Glib::ustring &name, Glib::ustring &extension, Glib::ustring &profileName )
-{
-    filename = Glib::path_get_basename( filename );
-    name = filename;
-    //remove dirs
-    size_t lastSlashPos = filename.find_last_of( "/" );
-
-    if ( lastSlashPos == Glib::ustring::npos ) {
-        lastSlashPos = filename.find_last_of( "\\" );
-    }
-
-    size_t lastDotPos = filename.find_last_of( '.' );
-
-    if ( lastDotPos != Glib::ustring::npos ) {
-        name = filename.substr( 0, lastDotPos );
-        extension = filename.substr( lastDotPos + 1, Glib::ustring::npos );
-    }
-
-    profileName = "sRGB"; // sRGB by default
-    static std::vector<Glib::ustring> workingProfiles = rtengine::getWorkingProfiles();
-
-    for ( std::vector<Glib::ustring>::iterator it = workingProfiles.begin(); it != workingProfiles.end(); ++it ) {
-        Glib::ustring &currentProfile = *it;
-
-        if ( std::search( name.rbegin(), name.rend(), currentProfile.rbegin(), currentProfile.rend() ) == name.rbegin() ) {
-            profileName = currentProfile;
-            name = name.substr( 0, name.size() - currentProfile.size() );
-            break;
-        }
-    }
-}
-
-//::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
-
-HaldCLUT::HaldCLUT()
-    :   m_clutImage( 0 ),
-        m_level (0),
-        m_profile( "sRGB" )
-{
-}
-
-HaldCLUT::~HaldCLUT()
-{
-    if ( m_clutImage ) {
-        m_clutImage->free();
-        m_clutImage = 0;
-    }
-}
-
-void HaldCLUT::load( Glib::ustring filename )
-{
-    m_clutImage = loadFile( filename, "", m_level );
-    Glib::ustring name, ext;
-    splitClutFilename( filename, name, ext, m_profile );
-
-    if ( m_clutImage ) {
-        m_filename = filename;
-    }
-}
-
-Glib::ustring HaldCLUT::profile() const
-{
-    return m_profile;
-}
-
-Imagefloat* HaldCLUT::loadFile( Glib::ustring filename, Glib::ustring workingColorSpace, int &outLevel )
-{
-    Imagefloat *result = 0;
-    StdImageSource imgSrc;
-
-    if ( !Glib::file_test( filename, Glib::FILE_TEST_EXISTS ) || imgSrc.load(filename) ) {
+    if (!Glib::file_test(filename, Glib::FILE_TEST_EXISTS) || img_src.load(filename)) {
         return result;
     }
 
     int fw, fh;
-    imgSrc.getFullSize (fw, fh, TR_NONE);
+    img_src.getFullSize(fw, fh, TR_NONE);
 
     bool valid = false;
 
-    //test on Hald format, copypasted from http://www.quelsolaar.com/technology/clut.html
-    if ( fw == fh ) {
-        outLevel = 1;
-
-        for(; outLevel * outLevel * outLevel < fw; outLevel++);
-
-        if( !( outLevel * outLevel * outLevel > fw ) ) {
+    if (fw == fh) {
+        unsigned int level = 1;
+        while (level * level * level < fw) {
+            ++level;
+        }
+        if (level * level * level == fw && level > 1) {
+            clut_level = level;
             valid = true;
         }
     }
 
-    if ( valid ) {
-        ColorTemp currWB = imgSrc.getWB();
-        Imagefloat* baseImg = new Imagefloat (fw, fh);
-        PreviewProps pp (0, 0, fw, fh, 1);
+    if (valid) {
+        rtengine::ColorTemp curr_wb = img_src.getWB();
+        result = std::unique_ptr<rtengine::Imagefloat>(new rtengine::Imagefloat(fw, fh));
+        const PreviewProps pp(0, 0, fw, fh, 1);
 
-        procparams::ColorManagementParams icm;
-        icm.working = workingColorSpace;
+        rtengine::procparams::ColorManagementParams icm;
+        icm.working = working_color_space;
 
-        imgSrc.getImage (currWB, TR_NONE, baseImg, pp, procparams::ToneCurveParams(), icm, procparams::RAWParams());
+        img_src.getImage(curr_wb, TR_NONE, result.get(), pp, rtengine::procparams::ToneCurveParams(), icm, rtengine::procparams::RAWParams());
 
-        if ( !workingColorSpace.empty() ) {
-            imgSrc.convertColorSpace(baseImg, icm, currWB);
+        if (!working_color_space.empty()) {
+            img_src.convertColorSpace(result.get(), icm, curr_wb);
         }
-
-        result = baseImg;
     }
 
     return result;
 }
 
-void HaldCLUT::loadClut( Imagefloat *img, RawClut &outClut )
+inline void posToXy(unsigned int pos, unsigned int width, unsigned int (&x)[2], unsigned int (&y)[2])
 {
-    img->normalizeFloatTo1();
-    int y_size = img->getH();
-    int x_size = img->getW();
-    outClut.resize( x_size * y_size * 3 );
-    int clutIdx = 0;
+    x[0] = pos % width;
+    y[0] = pos / width;
+    x[1] = (pos + 1) % width;
+    y[1] = (pos + 1) / width;
+}
 
-    //int level = m_level * m_level;  (unused)
-    for(int y = 0; y < y_size; y++) {
-        for(int x = 0; x < x_size; x++) {
-            outClut[ clutIdx * 3     ] = img->r( y, x ) * MAXVAL8;
-            outClut[ clutIdx * 3 + 1 ] = img->g( y, x ) * MAXVAL8;
-            outClut[ clutIdx * 3 + 2 ] = img->b( y, x ) * MAXVAL8;
+}
 
-            ++clutIdx;
+void rtengine::CLUT::splitClutFilename(
+    const Glib::ustring& filename,
+    Glib::ustring& name,
+    Glib::ustring& extension,
+    Glib::ustring& profile_name
+)
+{
+    Glib::ustring basename = Glib::path_get_basename(filename);
+
+    Glib::ustring::size_type last_slash_pos = basename.rfind('/');
+    if (last_slash_pos == Glib::ustring::npos) {
+        last_slash_pos = basename.rfind('\\');
+    }
+
+    const Glib::ustring::size_type last_dot_pos = basename.rfind('.');
+
+    if (last_dot_pos != Glib::ustring::npos) {
+        name.assign(basename, 0, last_dot_pos);
+        extension.assign(basename, last_dot_pos + 1, Glib::ustring::npos);
+    } else {
+        name = basename;
+    }
+
+    profile_name = "sRGB";
+
+    for (const auto& working_profile : rtengine::getWorkingProfiles()) {
+        if ( std::search( name.rbegin(), name.rend(), working_profile.rbegin(), working_profile.rend() ) == name.rbegin() ) {
+            profile_name = working_profile;
+            name.erase(name.size() - working_profile.size());
+            break;
         }
     }
 }
 
-Imagefloat* HaldCLUT::generateIdentImage( int level )
+rtengine::HaldCLUT::HaldCLUT() :
+    clut_level(0),
+    clut_profile("sRGB")
 {
-    int imageWidth = level * level * level;
-    Imagefloat *resultImg = new Imagefloat( imageWidth, imageWidth );
+}
 
-    int cubeSideSize = level * level;
-    float step = MAXVALF / (cubeSideSize - 1);
-    int pos = 0;
+rtengine::HaldCLUT::~HaldCLUT()
+{
+}
 
-    for( int b = 0; b < cubeSideSize; ++b ) {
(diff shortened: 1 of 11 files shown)


rtengine/clutstore.cc
```diff
@@ -3,440 +3,233 @@
...
-
-    r = rr * (float)(level - 1) - red;
-    g = gg * (float)(level - 1) - green;
-    b = bb * (float)(level - 1) - blue;
+    unsigned int x[2];
+    unsigned int y[2];
+    posToXy(color, clut_image->getWidth(), x, y);
 
-    color = red + green * level + blue * level * level;
+    float tmp1[4] __attribute__((aligned(16)));
```
heckflosse: please use ALIGNED16 instead of __attribute__((aligned(16)))
ALIGNED16 is defined in rtengine/opthelper.h, so you may need to include that file

heckflosse:
@Floessie Thanks for your work! 
About the races: Do you get them also with neutral + HaldClut ?
About benchmarking the implementation: processing of Hald is integrated into the tiled rgb-processing (rgbproc in improcfun.cc). Means, if you want to compare your implementation to the old one you have to measure processing of a whole file without hald giving time A. Then measure without your changes processing a whole file with hald giving time B. Then measure again processing a whole file with hald and with your changes giving time C.
To measure rgpbroc function you simply have to include this two lines
`#define BENCHMARK`
`#include "StopWatch.h"`
at the top of `improcfun.cc`
and
`BENCHFUN`
after the opening bracket of function `rgbproc`

Ingo

heckflosse:
I just benchmarked your fork against the original code. I could see no speedup at all. But if it produces the same output (which I didn't try yet) and needs half of space in cache to store the hald files, that would be interesting.

heckflosse:
@Floessie I checked the output of master and of your branch for differences caused by races using extreme brightness curves on the absolute difference of two consecutive processing using same code base and neutral profile + clut. What I detected are very very small differences in output between master branch and your fork, but they are only visible by using the above mentioned method, not by eyeball. So still, if your fork reduces memory footprint of clut cache by a reasonable amount and also cleans the code a bit (judging this part should be done by @adamreichold) I'm not against merging it.
But what's more interesting for me is to find out the reasons for the races you got using default profile. 

Ingo

Floessie:
> please use ALIGNED16 instead of __attribute__((aligned(16)))
ALIGNED16 is defined in rtengine/opthelper.h, so you may need to include that file

Okay, no problem. There are many instances of `__attribute__((aligned(16)))` in other places, so I wasn't aware of `ALIGNED16`.

> About the races: Do you get them also with neutral + HaldClut ?

The only thing I additionally tested last evening was neutral only, and there I get *no* differences. I'll try to break it down a bit as time permits.

> I just benchmarked your fork against the original code. I could see no speedup at all.

That's shocking, as [clutbench](https://github.com/Floessie/clutbench) shows a significant speedup. For benchmarking, I placed a `MyTime` timer [here](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/improcfun.cc#L4339l), and a `printf` at the end, which wrote values around 1500us with the original code and 1100us with the new one.

>I checked the output of master and of your branch for differences caused by races using extreme brightness curves on the absolute difference of two consecutive processing using same code base and neutral profile + clut. ...
> But what's more interesting for me is to find out the reasons for the races you got using default profile.

Again, I must emphazise: I saw those races on the master branch with the "Default" profile enabled. It has nothing to do with my branch or film simulation. Interesting enough, that you don't see them. I'm testing on a brand new Skylake Xeon, if that matters. Ahh, and what I also did, was closing RT between the two consecutive processings. I'll try to narrow it down for you.

heckflosse:
@Floessie About the `__attribute__((aligned(16)))`: I didn't replace them everywhere in code with `ALIGNED16`. The reason for the `ALIGNED16` is, that `__attribute__((aligned(16)))` leads to crashes on Win32Xp machines sometimes. That's why `ALIGNED16` is defined empty for windows 32 bit builds in `rtengine/opthelper.h`. WinXp32 does not align correctly especially when using OpenMP. But the compiler assumes that it is aligned when you use `__attribute__((aligned(16)))` and generates code which relies on the alignment which then leads to crashes. The other instances of `__attribute__((aligned(16)))` are already inside `#if defined( __SSE2__ ) && defined( __x86_64__ )`, so it doesn't matter there, but I will replace them with next commit for the sake of consistency
About the speedup: I'll measure it again.

heckflosse:
@Floessie I measured another time using a D800 File neutral + AutoExposure + Clut
master branch: 1115 ms
your fork: 1084 ms
without clut: 441 ms
so clut processing in master branch is 1115 ms - 441 ms = 674 ms
and in your fork it's 1084 ms - 441 ms = 643 ms
which is a speedup of about 5%

heckflosse:
@Floessie Another thing: Clut processing has a lot of a*b + (1-a)*c interpolations. There's this function https://github.com/Beep6581/RawTherapee/blob/master/rtengine/rt_math.h#L82 which does this calculation a bit faster. Maybe we should try to use it in clut processing too?

heckflosse:
@Floessie Integer division and modulo is quite expensive. Replacing the posToXy function with [this one](https://gist.github.com/heckflosse/fd6820f229bd0b3a2e9ca323290fe55e) gives an additional speedup.

Floessie:
Hi Ingo,

thanks for you hints. I will try them as soon as possible and do some benchmarking the way you proposed. I'm still wondering, why it's not as fast as expected. In `clutbench` I chose a pixel based layout for my image class, whereas `Image16` stores pixels in planes. I'll experiment a bit.

As to the race: I narrowed it down to RAW CA auto-correction! It's independent of architecture (tested Skylake and Sandy Bridge) and compiler version (tested 4.9.2 and 5.3.1), and you don't need to close RT between the save operations. I tested it with a CR2 and amaze as well as lmmse.

Best,
Flössie!

heckflosse:
@Floessie My 8-core machine is a bit limited by memory bandwidth. Maybe that's the reason I don't get the same speedup as you.
I just tested RAW CA auto-correction on two different files (one cr2 and one nef) and cannot reproduce the race.

heckflosse:
@Floessie I tested using the intp function for the interpolations. It seems that gcc perfectly optimizes the (1-x) stuff and I got no speedup by using the intp function. The change for the posToXy function gives a nice about 10% speedup to your fork on my machine.

Floessie:
> I just tested RAW CA auto-correction on two different files (one cr2 and one nef) and cannot reproduce the race.

Okay, perhaps we are misunderstanding each other. So here is exactly how to reproduce (as mentioned seen on different machines and builds):

1. Open a CR2 in a folder, containing only that very CR2.
2. Chose "Neutral".
3. Enable "Sensor with Bayer Matrix / Chromatic Aberration / Auto-correction".
4. Save current image as 8b uncompressed TIFF names "test1.tif".
5. Wait until saved.
6. Save current image as 8b uncompressed TIFF names "test2.tif".
7. Quit RT.
8. `convert test1.tif test2.tif -compose difference -composite -normalize diff.tiff`

diff

HTH,
Flössie!

heckflosse:
@Floessie Again no differences. Maybe OS specific? I'm on WIn7/64

Beep6581:
> convert test1.tif test2.tif -compose difference -composite -normalize diff.tiff
Better (or "other way"):
`compare test1.tif test2.tif diff.tif`

Floessie:
> Again no differences. Maybe OS specific? I'm on Win7/64.

Beware! ;) I'm on Debian Stretch and Jessie amd64.

> Btw: we should open an issue for this and not hijack your pr ;-)

Will you or shall I?

> `compare test1.tif test2.tif diff.tif`

That's way shorter, thanks, DrSlony!

heckflosse:
@Floessie Please open the Issue, as you can reproduce it. Maybe post a link to a raw file and pp3 file too.

heckflosse:
Hmm, isn't there a possible buffer overrun when you access the last value of a clut (in master and in Floesies branch)? Assume you access last value of clut. The interpolation will access the last+1 value then which overruns the buffer. Or do I miss something?
Ingo

Floessie:
> Hmm, isn't there a possible buffer overrun when you access the last value of a clut (in master and in Floessies branch)? Assume you access last value of clut. The interpolation will access the last+1 value then which overruns the buffer. Or do I miss something?

No, you must be right. On the other hand, I did a quick check with `CHECK_BOUNDS` and ASAN on master, and they found no problem (apart from ASAN failing early due to [this](https://github.com/Beep6581/RawTherapee/blob/master/rtgui/options.cc#L501), which can and should be fixed easily).

I'll file the bugs this evening.

Best,
Flössie

Floessie:
@heckflosse: Last evening I did some tests using `BENCHFUN` with the same results as yours: No speedup with my new implementation, slight speedup with your `posToXy()`. I'll investigate further on why it doesn't scale as expected.

As to your optimization: I always thought, with division you get the modulo for free, at least for x86's `IDIV` I know for sure. Why doesn't that apply here?

heckflosse:
@Floessie You're right about the modulo. I updated the [posToXy Function](https://gist.github.com/heckflosse/fd6820f229bd0b3a2e9ca323290fe55e/edit) with the fastest solution I could reach. I also changed last line to reduce dependencies.

About the scaling: Probably the bad scaling is caused by memory bandwidth. In the current implementation accesses to clut very likely will result in cache misses for first value of an interpolation pair even for access to L3 cache because clut is quite large (about 17 MB for a 1728x1728 clut when hold in memory as short integers).
The second value of interpolation pair is for free (most likely, if not always, you get it from L1 cache).
The way you store the clut in clutbench implementation (as rgb triplets) leads to much more cache hits.
We really should use this layout to hold the clut in memory in rt too!

Floessie:
@heckflosse: Okay. I'm really excited about the new `posToXy()`, but the link seems to be dead.

> The way you store the clut in clutbench implementation (as rgb triplets) leads to much more cache hits.
We really should use this layout to hold the clut in memory in rt too!

That's my next experiment. :) That would also fix the out-of-bounds access, if I make my internal CLUT one column wider and copy the last column.

Shall I go RGBx instead of RGB for better alignment, or RGB for more cache hits? If I go RGBx, I could perhaps reuse my SSE2 implementation from `clutbench`...

heckflosse:
@Floessie The old link still works fine here but nevertheless [here is a new one](https://gist.github.com/heckflosse/0654b76a181b882da2b76f3bde8920fe). But with changed layout we won't need it at all. I would go for aligned RGBx.

Floessie:
@heckflosse: Thanks for the new link, the one in your second last post has an `/edit` at the end, that's probably the culprit.

Floessie:
@heckflosse: Now here is the new HaldCLUT storage. I get a speedup of 23% from the original implementation. I also added a SSE2 version, but it seems to be slightly slower. It would be nice, if you could benchmark both of them, and if the vectorized version is slower with you, too, I'll remove it with my next commit.

heckflosse:
@Floessie Thanks, 23% speedup already sounds very good. I'm compiling your fork now. I'll benchmark both of them and also have a look at the code.

heckflosse:
@Floessie I confirm the about 23 % speedup of your sse-branch compared to master-branch (at my system it's a 22% speedup, but the difference doesn't matter). But there's still some potential. I tried some things and reached 28 % speedup now and there are still some things which could lead to additional speedup.

Floessie:
@heckflosse: That conforms to my observations: 23% plain, 22% with SSE2.

> But there's still some potential. I tried some things and reached 28 % speedup now and there are still some things which could lead to additional speedup.

Wow, really looking forward to your wizardry. :)

Could you also have a look at the conversion loop in line 62? Shall I use OMP and have the index calculated in each iteration?

heckflosse:
@Floessie [Here's](https://gist.github.com/heckflosse/20ae9bd5fa8f235ee1eb88a0829f7661) a bit faster version. I got rid off the _mm_set_ps(0.0f, blue, green, red) and thus also of the implicit conversions of red green blue to floating point, used SSE4.1 for the getClutValue function (optionally), replaced the posToIndex by simple additions and also used the PERMUTEPS macro instead of _mm_shuffle_ps. The changes are only in SSE2 part of clutstore.cc. Didn't test non SSE2 part yet. I didn't beautify my changes because I just wanted to give you some ideas for further speedups. One thing which I didn't implement is a wrapper function which gets three lines of red green blue and then calls getrgb for each pixel of the line. Doing this we should be able to avoid the three stores and loads of SSE registers on stack at start and end of getrgb (these are per pixel, so maybe expensive).
About 'line 62': I didn't check yet, but I would expect that the possible gain of using OMP in the load function would be mostly hidden by the time to load the clut data from disk.

Floessie:
> I got rid of the _mm_set_ps(0.0f, blue, green, red) and thus also of the implicit conversions of red green blue to floating point, used SSE4.1 for the getClutValue function (optionally), replaced the posToIndex by simple additions and also used the PERMUTEPS macro instead of _mm_shuffle_ps.

These also make the code more readable. Very nice, thank you! I'll also switch from `__m128` to `vfloat` while I'm at it.

> One thing which I didn't implement is a wrapper function which gets three lines of red green blue and then calls getrgb for each pixel of the line. Doing this we should be able to avoid the three stores and loads of SSE registers on stack at start and end of getrgb (these are per pixel, so maybe expensive).

You mean a wrapper in `improcfun.cc`?

heckflosse:
@Floessie I mean a function in clutstore.cc which gets passed the addresses of linebuffers for red, green and blue and the width of the line. This function then would call the getrbg function per pixel of the line in the hope it will be inlined and get rid of the stack stores and loads. Of course we would have to use per line processing inside the tiled processing of imrocfun.cc too, but whis way we also could vectorize the colorspace transformations, which eat a bit of processing time.

Floessie:
Now, with `getRGB()` per line I'm 46% faster (SSE) than the original implementation. Next could be moving `filmSimCorrectedStrength` into `getRGB()` and apply the strength vectorized.

@heckflosse: I need some advice, where to move the `out_rgbx` allocation, to hoist it out of the hot loops but have still one instance per OMP thread. And what about the color space conversions?

Floessie:
Now it's 48%.

heckflosse:
@Floessie Great :+1: For the per line vectorized color space conversions we have to make vectorized versions of rgbxyz and xyz2rgb (for one of them there is already a SSE version iirc)

heckflosse:
@Floessie I would have tried to allocate the line buffer on stack (which is only an integer add compared to memory allocation, especially because width of tile is small and so size of buffer is small too), but I didn't measure it, maybe it's pointless.

heckflosse:
@Floessie For readability I would suggest to use the _MM_SHUFFLE(z, y, x, w) macro instead of using 0x55... Do you agree?

Floessie:
@heckflosse: Definitely! I wasn't aware of `_MM_SHUFFLE()`, but that's a great way of improving readability. I was about to have `out_rgbx` on the stack, but changed my mind as a colleague of mine once hit a 2MB (?) stack limit on Windows. I'd like to keep it that way. I'll have a look at `rgbxyz()` and `xyz2rgb()` after moving `filmSimCorrectedStrength` into `getRGB()`. Also on my agenda is removing the unneccessary `CLUT` base class, so that `HaldCLUT` doesn't need any virtual functions.

heckflosse:
@Floessie Tile width in rgbproc is 112, meaning 3 * 4 * 112 = 1344 bytes allocation an stack per thread. That should not be a problem at all. In case it turns out to be a problem, we can use -Wl,--stack,16777216 ti increase stack size to 16MB if needed ;-) With modern machines having a lot of ram that should be no problem.
The 'filmSimCorrectedStrength' is also a use case for the intp (vintpf) function.

Ingo

Floessie:
@heckflosse Okay, I moved `out_rgbx` onto the stack. I'm now at 51%. Next is the color space conversion. Is it okay to use the `VECTLENSP` define? I couldn't find a place where it is used at the moment, but it would be handy to generalize the conversion loops for SSE and AVX...

heckflosse:
@Floessie Sure you can use the VECTLENSP define. I didn't wrote AVX code in past because on my machine it slows down compared to SSE code when using all cores. My Bulldozer has 8 cores which share 4 AVX units, but 8 SSE units.

Floessie:
@heckflosse Okay, color space conversion vectorized. Now at 59%.

I was wondering why there are no `F2V`, `LVF`, and `STVF` defines for AVX. And how can I enable an AVX build?

Then again, my C++ eyes stumbled upon those many declarations for the same thing. Take `intpf()` and `vintpf()` or the `float` and `vfloat` versions of `rgbxyz()`: They only differ in their parameter type, thus `intpf<vfloat>()` and `vintpf()` are equivalent. The same in `color.h`, where `static inline` could be elegantly replaced by `template<typename T>`. Is there any reason, it's coded like it is?

What about the still unvectorized gammas? Can they be vectorized, too? They seem to have LUT nature, that might be cumbersome.

heckflosse:
@Floessie There are no F2V ... defines for AVX because I never made them for the reason I mentioned above. AVX should be automatically enabled when you make a native build on your (hopefully supporting AVX) machine. To enable AVX in sleef is a bit more difficult because it seems that defining ENABLE_AVX collides with the SSE part of sleef.
A different approach to make use of AVX could be to use #pragma omp simd. There are cases where it works good and cases where it fails miserably (maybe because I'm not experienced in using it).

About the intpf vintpf and so on there are mainly two reason for the current coding:
1) I was lazy ;-)
2) I'm not a template specialist (means I know how to code a template, but not how to restrict it to a datatype for example, though that's not the case here. I just mention it because I have that case in my upcoming speedup thumb processing changes)

About the unvectorized gammas:
In first step we could make a Color::gamma_srgbclipped function which uses gamma2curve[] instead of gammatab_srgb[]. Doing this we can delete the 
CLIP<float>
from
CLIP<float>( Color::gamma_srgb( sourceR ) );
Explanation:
gamma2curve[] has the same content as gammatab_srgb[] but clips automatically. Both do range checks, so by using gamma2curve[] instead of gammatab_srgb[] we can eliminate the two per pixel and channel instructions introduced by CLIP<float>.
In a second step we could make a working version of [LUT vfloat operator[](vfloat)](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/LUT.h#L285)

Floessie:
@heckflosse So, I do have AVX enabled and `VECTLENSP` is 8, but `LVF` loads only 4 floats at once? That wouldn't work out, won't it? What about having `F2V_MAX` (or something like that) and friends which load `VECTLENS` floats?

> I just mention it because I have that case in my upcoming speedup thumb processing changes.

I'd like to help you out, if you like. I'm neither a specialist, but have some experience with C++11's `std::enable_if()`.

> In first step we could make a Color::gamma_srgbclipped function which uses gamma2curve[] instead of gammatab_srgb[].

Phew, I'll take a look if time permits.

heckflosse:
@Floessie [Here](https://gist.github.com/heckflosse/1e1cd4df04759e35077947016f7e9aba) is a patch for your fork with the Color::gamma_srgbclipped related changes. Only first step, no vectorized version of that part yet.
I'll answer your other questions later.

Beep6581:
Wouldn't your work and discussion fit better in an issue instead of in a pull request?

Floessie:
Don't know. As far as I'm concerned, I'd like to finish this PR by cleaning up some things and call it a day. With Ingo's kind help I reached the goal I wanted to reach: We've a speedup of 1.59 and a memory reduction by 1/3 now. And the code is significantly cleaner. I'm content.

heckflosse:
@Floessie About the AVX coding we really should open an Issue for discussion before we do any AVX coding.
About the speedup and code cleaning of HaldCLUT. I'm also content with your result :+1: Though I would like to get my small patch from 3 posts above also included because it leads to a small additional speedup ;-)
Maybe we should also open a new Issue to change the coding from methods which only differ in parameter type to templates.
About your offer to help me out with some template related stuff: I greatly appreciate that offer :+1: 

Ingo

Floessie:
@heckflosse Ready to merge from my side. In case you wonder about the removed `inline` from `getClutValue()`: It makes no difference (as expected), checked with `objdump`.

With your `gamma_srgbclipped` patch we finally have a speedup of 1.62. :+1:

heckflosse:
@Floessie Can you change line 232 of clutstore.cc to
`STVF(*out_rgbx, vintpf(_mm_load_ps1(&strength), v_out, v_in));`
please? Otherwise it will crash on Win32 systems.
Ingo

heckflosse:
@Floessie One additional suggestion. Because you reduced the memory usage of clut store we could increase the max number of cached cluts by factor 1.5 in [preferences.cc](https://github.com/Beep6581/RawTherapee/blob/master/rtgui/preferences.cc#L574)

Floessie:
@heckflosse Good points. Done.

heckflosse:
@Floessie I just made another [small speedup](https://gist.github.com/heckflosse/b02756da998c2a5a62cb8600244ca4f2). Would be great if you could include it before I merge ;-)

Ingo

Floessie:
@heckflosse Here it is. Hoisting the `strength` load out of the loop was absolutely needed, but I can see no additional speedup with GCC 5.3.1. Maybe the compiler was smart enough before.

heckflosse:
@Floessie I got a speedup from 717 ms to 700 ms (measured whole rgbproc() processing time for a D800 file 7 times and medianed) GCC 5.3.0

Beep6581:
Thank you for this! Using it feels much faster now.

Floessie:
@Beep6581 That was a fun session with Ingo, and nice that you can feel the difference. I had this on my agenda ever since I fixed the race in `clutstore.cc` [some time](https://github.com/Beep6581/RawTherapee/issues/2567) ago. And next time I won't forget to run `astyle`. ;)

@heckflosse Thanks a lot for your help on this. There was a lot to learn from you. My final tests resulted in 275ms for whole `rgbproc()` and a 600D (18MP) file, medianed over 5 samples, so perhaps `F2V` hasn't that big penalty on Skylake.

Best,
Flössie

heckflosse:
@Floessie It was really a joy to speed up this part of rt processing with your great work. And I'm glad that I could help a bit :)

Issue #3154, opened 2016-02-17
Crash while loading Fuji XTrans files


pi99y:
Hi, I get quite frequent crashes that seem to be caused by older pp3 files, but it's really hard to pinpoint the issue, since it's a bit "random"...

Tested on:
- OS: Windows 8.1 64bit, 8GB ram
- RT:  RawTherapee_WinVista_64_Gtk3_Debug_4.2.730.zip Changeset: f5d5083be68f7cf487a688dcab37b91455fc9e6e
- Fuji X-Pro1 XTrans RAF files

I open up a folder with a at least a few images, try to load a file and it sometimes crashes. It can either crash while still in (or returning to) "file browser" view or already in the "editing view", but seems like it always crashes while either "loading", "demosaicing" or "processing" - not later on while editing (if i can get to that stage).

It seems not to be tied to specific image since it sometimes crashes on one file, some other time on another file, but nothing really consistant.

I tried to narrow it down, but I can only say that it appears that at least two things help while reproducing:
* thumbnails of images are not yet created (It's easier to reproduce if I clear the cache before opening RT) Since this mostly happens during the thumbnails are being generated it helps to have many images in the folder - so you have more time to reproduce it. This is probably also why I can't reproduce this with a single file (that crashed before).

* images have old pp3 files present (it doesn't happen if I create new pp3's with the current RT version, although  the general "profile" I load has roughly the same preferences - since that's the one I always start with)
PP3 files in this case say "AppVersion=4.1.80 Version=321".

* it could be raf/xtrans specific (can't test other)

Most of the time it crashed I get the debug report saying "pure virtual method called, terminate called without an active exception" but not always (sometimes there is no error message in the debug window)

Also note that the "failed to fetch network locations" error you see in the screenshot pops up every time I start RT and seems not to be related to this issue.

Screenshot: https://i.imgur.com/PMbtxiM.jpg
PP3: http://filebin.net/tvof90mays
(I'm not missing any files that are referred to in a profile, just the RT version is newer)

heckflosse:
@pi99y It's related to xtrans files. I can reproduce it in Release build of master branch. It's not related to old pp3 files (at least the crash I get).

sguyader:
I do not seem to get this kind of problems with xtrans files from my
Fujifilm X-T1.

heckflosse:
[...]

PR #3276, opened 2016-05-04
Fix buffer overrun when applying HaldCLUT (#3154)

Floessie:
Ingo found a buffer overrun due to an inverted mask when deciding
whether to take the SSE path or not. This fix applies his suggested
pattern for border cases. Kudos to @heckflosse.


heckflosse:
@Floessie Thanks a lot :+1:

Floessie:
Wow, that was quick! :)

heckflosse:
:) I had a look and think it's good now. I use this pattern for almost all SSE coding in rt (though there are still some parts left which should be adapted to this pattern) because it's understandable, correct for border cases, doesn't blow up code and also is easily expandable to AVX coding

Floessie:
Yep, it's brilliant and should even be faster than my (wrong) tile based pattern! :)

PR #3291, opened 2016-05-17
Fix std::out_of_range exception with empty HaldCLUT basename

Floessie:
User **jgschaefer** [reported an error on pixls.us](https://discuss.pixls.us/t/rt-build-from-git-crash-on-launch-debian-testing-64-bit/1425)
which could be traced down to an empty basename for a HaldCLUT. The
original implementation did not throw an exception due to the use of
`std::string::substr()` instead of `std::string::erase()`, but silently
assigned the first working profile to `profile_name`.

### DIFF
--- a/rtengine/clutstore.cc
+++ b/rtengine/clutstore.cc
@@ -283,11 +283,16 @@ void rtengine::HaldCLUT::splitClutFilename(
 
     profile_name = "sRGB";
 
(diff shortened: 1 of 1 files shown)


heckflosse:
@Floessie Thanks a lot for the patch. Merged.

PR #3293, opened 2016-05-18
Fix build warnings

Floessie:
Fix two warnings when building special configurations:
- ODR violation when building with LTO enabled
- Narrowing conversion when building for PPC

LTO seems to really speed things up, btw. :)


heckflosse:
Floessie, thank you!
I can't build on Windows with LTO. Also can' test on PowerPC. But it builds fine on windows => Merge.

Issue #3290, opened 2016-05-17
Unable to compile under Linux


eduperez:
I tried to compile the GTK3 branch under Fedora 23, following the instructions at RawPedia, but there seems to be something wrong with the `cmake` file:

`git clone https://github.com/Beep6581/RawTherapee ~/repo-rt`
`cd ~/repo-rt`
`git checkout remotes/origin/gtk3`
`rm -rf build`
`make clean`
`./clean`
`mkdir build`
`cd build`
```
cmake -DCMAKE_CXX_FLAGS="-std=c++11 -Wno-deprecated-declarations -Wno-unused-result" \
>       -DWITH_LTO="OFF" \
>       -DCMAKE_BUILD_TYPE="release" \
>       -DPROC_TARGET_NUMBER="2" \
>       -DBUILD_BUNDLE="ON" \
>       -DBINDIR="." \
>       -DDATADIR="." \
>       -DCACHE_NAME_SUFFIX="4" ..
```
`make -j6 install`
```
CMake Error at /home/eperez/repo-rt/AboutThisBuild.cmake:27 (string):
  string sub-command REGEX, mode REPLACE needs at least 6 arguments total to
  command.


CMakeFiles/AboutFile.dir/build.make:57: recipe for target 'CMakeFiles/AboutFile' failed
make[2]: *** [CMakeFiles/AboutFile] Error 1
CMakeFiles/Makefile2:67: recipe for target 'CMakeFiles/AboutFile.dir/all' failed
make[1]: *** [CMakeFiles/AboutFile.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2

```

I changed all instances of `${GIT_VERSION_WHOLE}` to `"${GIT_VERSION_WHOLE}"` in the `string` calls, and this seems to have fixed the issue.

ClaesGitHub:
Hi!
I use another checkout terminology. Try

> git checkout gtk3

It works for me.

eduperez:
@ClaesGitHub : Tried that, but got the same result; thanks anyway.

aserebryakov:
@eduperez, what CMake version do you use?

ClaesGitHub:
@eduperez I had to build a new Virtualbox today for some experiments, and I fed it with a very minimal Fedora 23. Then I compiled the gtk3 branch of RT - and **it worked just fine**. I followed the standard instructions for Fedora on Rawpedia.  (My cmake version is 3.4.1.)

[...]

PR #3294, opened 2016-05-19
Use "find_program" to search for git executable

eduperez:
CMake's "find_file" can return non-executable files or folders, thus "find_program" must be used to ensure that only executable files are detected.
Tested on Linux only.
Fixes issue #3290.


heckflosse:
@eduperez Thanks for the fix. Works fine on Windows too.

PR #3343, opened 2016-06-11
Dcp speedup

heckflosse:
I want to merge this branch after removing Stopwatches and before making further cleanups to get more tests.
@Floessie Are you ok with that procedure?


Floessie:
@heckflosse I'm currently testing your branch, and would like to push some additional cleanups to it. Okay?

Floessie:
@heckflosse Couldn't resist. :)

heckflosse:
@Floessie I tested on Linux using gcc 4.9.3 and on Win64 using gcc 5.3.0
Builds and works fine. Thank you

PR #3350, opened 2016-06-17
Speedup for save to 8bit formats (mainly for 8bit tiff and j…

heckflosse:
…peg)


heckflosse:
I meant SSSE3, not SSS3, stupid typo...

PR #3364, opened 2016-07-01
Add milliseconds to queue params file, fixes #3363

### LINKED ISSUE #3363: Quick adding same image to queue several times and then deleting the middle entry from queue leads to wrong queue content
To reproduce:

Add the same image 8 times to queue by pressing add to queue button 8 times as fast as you can.
Go to queue tab and remove 4th entry.
Close rt
Start rt
Queue should have 7 entries now, but has less than 7

heckflosse:


Beep6581:
Fix confirmed.

PR #3366, opened 2016-07-01
Updated rtexif

Beep6581:
This branch contains all classes in rtexif updated to match exiftool-10.20.
Mostly additions, but a few corrections as well, for example [SADriveMode()](https://github.com/Beep6581/RawTherapee/blob/master/rtexif/sonyminoltaattribs.cc#L1212) in sonyminoltaattribs.cc contained indexes in hex notation but they were written in decimal form.


heckflosse:
I tested and didn't detect any issues. :+1:  for merge

PR #3375, opened 2016-07-06
Fix cppcheck performance hints

heckflosse:


aserebryakov:
Doesn't compiler optimize the `i++` and `++i` on optimization level higher that `O0`?

heckflosse:
@aserebryakov I wanted to get rid of all the hints to make checking easier in future.

Floessie:
@aserebryakov @heckflosse This is only true for primitive types, not for `operator++()` as used in iterators, thus preinc should generally be prefered.

@heckflosse [This](https://github.com/Beep6581/RawTherapee/blob/c8779c04f52c3f4525646df5a459491168d8c8e4/rtengine/shmap.cc#L230) is a bit odd.

Very good job, Ingo. I can offer to run clang-tidy on RT (e.g. to do the `nullptr` conversion automatically), but first I have to finish the median branch (not much left) and then I'm on vacation for three weeks. If RT5 is scheduled for this summer this will be too tight, otherwise there might be enough time left to do such cleanups.

heckflosse:
@Floessie I'll change the one you mentioned.
There are also another 3 hints which I didn't fix yet

- [number 1](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/imagedata.cc#L165) : `Inefficient usage of string::find() in condition; string::compare() would be faster.`

- [number 2](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/iccstore.cc#L249) : `Buffer 'nd' is being written before its old content has been used.`

- [number 3](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/iccstore.cc#L636) : `Buffer 'oprof' is being written before its old content has been used.`

number 2 and 3 are clear for me, number 1 is not. Maybe you can enlighten me?

Floessie:
@heckflosse A very strange expression, indeed. `i` doesn't seem to be referenced later, so the assignment should be dropped. But `i = model.find("FILE VERSION")` will always be true, and false only, if "FILE VERSION" matches at position 0. Don't know what's meant by that, and I guess `cppcheck` doesn't know either.

heckflosse:
@Floessie I'll let number 1 as it is now...

heckflosse:
@Floessie @Floessie `But  i = model.find("FILE VERSION")  will always be true, and false only, if "FILE VERSION" matches at position 0`. Now I at least understand what cppcheck meant by `string:compare would be faster` :)

PR #3376, opened 2016-07-07
Improve processing time to read camconst.json by almost factor 2. Add…

heckflosse:
…itionally make it future-proof by reducing O(n²) copying amount in reallocate phase to O(n)

### DIFF
--- a/rtengine/camconst.cc
+++ b/rtengine/camconst.cc
@@ -644,8 +644,7 @@ CameraConstantsStore::parse_camera_constants_file(Glib::ustring filename_)
             Glib::ustring make_model(ji->valuestring);
             make_model = make_model.uppercase();
 
-            std::pair<std::map<Glib::ustring, CameraConst *>::iterator, bool> ret;
-            ret = mCameraConstants.insert(std::pair<Glib::ustring, CameraConst *>(make_model, cc));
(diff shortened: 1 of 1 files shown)


rtengine/camconst.cc
```diff
@@ -637,18 +643,17 @@ CameraConstantsStore::parse_camera_constants_file(Glib::ustring filename_)
 
             Glib::ustring make_model(ji->valuestring);
             make_model = make_model.uppercase();
-            std::map<Glib::ustring, CameraConst *>::iterator existingccIter = mCameraConstants.find(make_model);
 
-            if (existingccIter == mCameraConstants.end()) {
-                // add the new CamConst to the map
-                mCameraConstants.insert(std::pair<Glib::ustring, CameraConst *>(make_model, cc));
+            std::pair<std::map<Glib::ustring, CameraConst *>::iterator, bool> ret;
+            ret = mCameraConstants.insert(std::pair<Glib::ustring, CameraConst *>(make_model, cc));
```
Floessie: @heckflosse May I suggest one small change here? This is a place where `auto` really shines. So you could convert it to:
```c++
const auto ret = mCameraConstants.emplace(make_model, cc);
```
Best,
Flössie

heckflosse:
@Floessie Thanks for the suggestion! I changed it.

PR #3377, opened 2016-07-08
Median revision

Floessie:
This is a cleanup and speedup for the median functions conceived by @heckflosse and me.

@Beep6581 Could you take the time to test this branch, as I'm unsure if my merge of `rtengine/ipwavelet.cc` went all well.

I'll be off at the end of next week for three weeks, so it would be nice if we could finish this within the days left. :)

Best
Flössie


Beep6581:
I compared 3b734e4a7a48917c1c285d9efcfec5170616739a vs c0c82abb3210355bcff4e1584f23c3d7a3e84ce0 using NR with median denoising, all methods for the 3x3 type and then all types for the weighted method. No differences in output found.

heckflosse:
@Floessie There is still an old macro version of median in rtengine/ffmanager.cc. Shall I change it or will you do? It's there becaused I moved it there after you started your rework.

heckflosse:
@Floessie in rtengine/FTblockDN.cc there is this code for each kind of median:

part 1:

```
                int j = 0;

                for (; j < border; ++j) {
                    medianOut[i][j] = medianIn[i][j];
                }
```

part 2: the median part which depends on median kind

part 3:
```
                for(; j < width; ++j) {
                    medianOut[i][j] = medianIn[i][j];
                }
```

Can't we move part 1 and 3 outside the median kind condition? Or do I miss some thing?

Ingo

Floessie:
@Beep6581 Thanks for testing so quickly!

@heckflosse I've already converted the code in `rtengine/ffmanager.cc` during the merge. I know, I should have split it out to another commit, but I forgot that during the merge.

You're absolutely right about `rtengine/FTblockDN.cc`. I'll fix that.

Floessie:
> I've already converted the code in rtengine/ffmanager.cc during the merge. I know, I should have split it out to another commit, but I forgot that during the merge.

Actually, I had it split out and not yet committed. It's now part of 7f66eb5.

heckflosse:
@Floessie Great, thank you. There's one really nitpicking thing which could be fixed: It's a double include of `median.h` in `rtengine/rawimagesource.cc`. :+1:

Floessie:
@heckflosse However you found that one... :+1: 

Now a lot of (partly automated) cleanups. Please test.

Thanks
Flössie

Floessie:
@heckflosse There is another median9 [here](https://github.com/Beep6581/RawTherapee/blob/median-rework/rtengine/dcraw.cc#L4960). Shall we convert that, too? My renaming of `ULIM` also changed `dcraw.cc`. Do I have to revert that or remove the define completely, as it's not used in that file?

heckflosse:
@Floessie It's your decision whether you modify dcraw.cc or not. But if you modify it, you have to generate the new dcraw.patch file too. Here is the [how to create dcraw.patch file](https://github.com/Beep6581/RawTherapee/blob/master/rtengine/dcraw.patch.README). For modifications of dcraw.cc you only have to execute this part:
`diff -u dcraw.c dcraw.cc > dcraw.patch`
I looked at the `clang-tidy` changes. Looks good, but you should astyle the file again as the indentation seems different to our astyle format.
The median9 in dcraw::median_filter can stay as is because we don't use that function at all.

Ingo

Floessie:
> Looks good, but you should astyle the file again as the indentation seems different to our astyle format.

Very diplomatically. ;) It wasn't indented at all. Should be fixed now. I also reverted dcraw's `ULIM` and refreshed the patch. Thanks for the hints.

heckflosse:
@Floessie Can you remove rtengine/colorclip.h from repository too with you final push? It's currently unused and we can restore it from history in case we need it in future. I found it by searching for median.h includes

Floessie:
@Beep6581 @heckflosse Shall I merge now?

heckflosse:
@Floessie Sorry, I just detected that my last push conflicts with your PR. I'll revert it now, then you can push the PR

heckflosse:
@Floessie I reverted my last commit. Now you can merge. I'll revert the revert afterwards

Beep6581:
@Floessie please do :)

Floessie:
@heckflosse Ingo, please go ahead and re-revert your patch for #3379. Thanks for waiting, but I would have re-merged master, so no problem if you commit conflicting. Anyway, this procedure was much easier for me. :)

@Beep6581 You're welcome. Again, this was fun working together. :+1: Thanks for testing and managing. :)

heckflosse:
@Floessie Thanks, patch for #3379 pushed again

PR #3401, opened 2016-08-23
Colourtoning: Speedup and improvement for Automatic Saturation Protection

heckflosse:
When using Colourtoning Automatic Saturation Protection the function ImProcFunctions::moyeqt is called.
I made an about 7x speedup for this function (processing time of the function for a D800 NEF reduced from 323 ms to 42 ms).
Additionally in past the function was called twice in preview mode. One time for whole image area (navigator and pan background) and one time for the preview area. When zoomed in the values returned by the function were not from full image which could lead to differences between preview and final output. Now the values are always from full image.

Ingo

### DIFF
--- a/rtengine/color.h
+++ b/rtengine/color.h
@@ -231,6 +231,19 @@ public:
     */
     static void rgb2hsv (float r, float g, float b, float &h, float &s, float &v);
 
+    static inline float rgb2s(float r, float g, float b) // fast version if only saturation is needed
+    {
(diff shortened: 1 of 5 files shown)


heckflosse:
@Desmis Jacques, any objections to merge this pr?

Ingo

Desmis:
@heckflosse 
Ingo

No problem :)

Beep6581:
Tested, all fine :+1:

PR #3415, opened 2016-09-03
Add full support of c++11 to the Eclipse IDE

Hombre57:
...(i.e. including when building the index), with the help of
-DCMAKE_ECLIPSE_VERSION="4.6.0" (set to your version, of course) on the
command line. Without this, the declared symbols are wrong and c++11 is
not supported by the IDE (vs the build makefiles).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,13 @@ else (WIN32)
   cmake_minimum_required(VERSION 2.6)
 endif (WIN32)
 
+# must stay before the 'project' command
+if(${CMAKE_EXTRA_GENERATOR} MATCHES "Eclipse CDT4")
+  set(CMAKE_CXX_COMPILER_ARG1 "-std=c++11" CACHE STRING "C++ version for eclipse" FORCE)
+  # users building with Eclipse should set CMAKE_ECLIPSE_VERSION through the command line to their current version of Eclipse
+  #set(CMAKE_ECLIPSE_VERSION "4.6.0" CACHE STRING "Eclipse version" FORCE)
+endif()
+
 PROJECT(RawTherapee)
 
 # the default target is 'Debug'


Hombre57:
@Floessie @Beep6581 @heckflosse I've found why my IDE doesn't provide support of C++11 while I'm able to build RT. There seem to be a bug in cmake, but found a workaround here https://public.kitware.com/Bug/bug_relationship_graph.php?bug_id=15316&graph=relation . If you use Eclipse, could you test it and/or give me a go for merging ?

Beep6581:
I don't use any IDE for RT.

Which version of CMake are you using? I'm on 3.5.2.

Hombre57:
I'm using the one provided by MSYS2, i.e. 3.4.1 (didn't checked since 1 month ago).

Beep6581:
@Hombre57 by the way, this is the new issue for the same bug https://gitlab.kitware.com/cmake/cmake/issues/14962

Floessie:
@Hombre57 Can you give me instructions, on how to use it? I normally run `cmake` from the command line and integrate my build directory as a normal Makefile project in Eclipse. To benefit from your change, I guess I'll have to set it up differently...

Hombre57:
@Floessie Ah, you're using one of the Makefile generator I suppose ? I'm using the Eclipse one, but don't know which one is better. I never had any real issue, except some minor things that I can explain here.

I'm using the following command line options to generate the Eclipse project (but close Eclipse first doing so) : `-G "Eclipse CDT4 - MinGW Makefiles" -D_ECLIPSE_VERSION=4.6.0 -D CMAKE_ECLIPSE_MAKE_ARGUMENTS="-j8 -C D:\Dev\WorkspaceCPP\RTManagement\build\Debug"`

I don't know to do it otherwise, but I'm using `j8` as command line argument for the building process to use all the cores and use the `-C` argument to specify where to find the makefiles. But you might have mastered that part already ;)

Once done, I copy the `.project` and `.cproject` to where Eclipse expect to find them (if you want the full process including importing the code with EGit first, this is another story). Then I'm firing up Eclipse and press F5 to update the project. It will report and error, don't know why this hasn't been solved already, about virtual directories (`[ ... ]`) that I delete right away.

The content of the main cmakefile will fully enable C++11, even for the Eclipse's indexer. It's all about adding `CMAKE_CXX_COMPILER_ARG1 "-std=c++11"` together with the `_ECLIPSE_VERSION` parameter (if it doesn't work, try the `CMAKE_ECLIPSE_VERSION` parameter instead). I don't know if/how it can be used by the Makefile generator though.

heckflosse:
Compiles fine here. :+1: for merge

Floessie:
Yep, works. :+1:

Hombre57:
Patch committed.

PR #3434, opened 2016-09-20
About 15% speedup for function lab2rgb16 which is used when saving 16 bit output

heckflosse:


rtengine/rt_math.h
```diff
@@ -101,5 +101,17 @@ inline T norminf(const T& x, const T& y)
     return std::max(std::abs(x), std::abs(y));
 }
 
+inline int float2uint16range(float d) // clips input to [0;65535] and rounds
+{
+    d = CLIP(d); // clip to [0;65535]
+#ifdef __SSE2__ // this only works in IEEE 754 maths. For simplicity I restricted it to SSE2. We can enhance it later, but we have to take care of endianness then.
+    d += 12582912.f;
+    return reinterpret_cast<int&>(d);
```
Floessie: @heckflosse Great adaption of the Lua double code! I was wondering, why you cast to `int&` here instead of `int`. What's the reason here?

Floessie:
@heckflosse Aha, you had a look at the ImageMagick code ... :wink:

heckflosse:
@Floessie The first (with the reinterpret_cast) was a copypasta ;) adapted to float without thinking much about it.
The second (+0.5f) came to my mind not by looking at the ImageMagick code, but by our discussion about lround() in the other Issue ;) So, in fact, I got that idea from you!
But not a bad idea to look into  ImageMagick code....

heckflosse:
Though it's possible to make an SSE version for both loops I modified, that would only give an additional ~5 % speedup because the bottleneck is the lcms conversion.
If there are no objections, I would like to merge now.
@Floessie because I made changes to rtengine/iplab2rgb.cc which you also probably will modify in a different issue, I'll wait for your ok/not ok

Ingo

Floessie:
@heckflosse No problem, just go ahead, I'll merge.

Thanks for the flowers, but it's in the ImageMagick code right [here](https://github.com/ImageMagick/ImageMagick/blob/master/MagickCore/quantum-private.h#L375) a few lines above the code I first referenced. :)

Flössie

PR #3438, opened 2016-09-26
Fix "-Wignored-attributes" with vector arguments in `rtengine/median.h`

Floessie:
See [GCC bug 69884](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69884).

### DIFF
--- a/rtengine/median.h
+++ b/rtengine/median.h
@@ -24,6 +24,10 @@
 
 #include "opthelper.h"
 
+#if defined __GNUC__ && __GNUC__>=6 && defined __SSE2__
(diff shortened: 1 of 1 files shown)


Floessie:
@mbajor Marcin, are you doing GCC 6 builds without SSE2? If so, could you also test this change? I'm going to merge this tomorrow anyway, as I'm confident it'll work, but a quick countercheck won't hurt. :)

mbajor:
@Floessie 
I don't have any problems with GCC 6.x builds without -msse2 flag except #3440

Floessie:
@mbajor Thanks for checking!

PR #3443, opened 2016-09-30
Fix buffer overflow in dcraw

innir:
Apply patch from https://github.com/LibRaw/LibRaw/commit/89d065424f09b788f443734d44857289489ca9e2, this fixes CVE-2015-8366.


heckflosse:
@innir Thanks for the pr. :+1: for merge from me. Any objections from other devs?

Floessie:
@heckflosse :+1: from me, too. Better one check too much than too few. :wink: 

@innir Philip, thanks for uploading the latest GTK3 version to unstable. If we wanted to have RT5 in Stretch, what's the deadline?

Beep6581:
Compiles and works fine here.

Issue #3440, opened 2016-09-29
 gcc 6.2.1 build failure


mbajor:
I can't build packages (from master branch, gtk3 builds fine) for some distributions with gcc 6.2.1 (with and without msse2)
e.g. openSUSE Tumbleweed, openSUSE Factory PowerPC (ppc64, ppc64le)

There is no problem with build for openSUSE Factory PowerPC(ppc) and Arch. Both have gcc 6.2.1.

```
[  317s] /home/abuild/rpmbuild/BUILD/rawtherapee-4.2.1071/rtgui/dirbrowser.cc:62:64: error: no match for 'operator<<' (operand types are 'std::basic_ostream<char>' and 'const Glib::RefPtr<Gio::File>')
[  317s]              std::cerr << "Failed to list subdirectories of \"" << dir << "\": " << exception.what () << std::endl;
```

```
[  461s] /home/abuild/rpmbuild/BUILD/rawtherapee-4.2.1071/rtgui/dirbrowser.cc:62:23: error: cannot bind 'std::basic_ostream<char>' lvalue to 'std::basic_ostream<char>&&'
[  461s]              std::cerr << "Failed to list subdirectories of \"" << dir << "\": " << exception.what () << std::endl;
```

[Sample log file
](https://build.opensuse.org/public/build/home:rawtherapee/openSUSE_Tumbleweed/i586/rawtherapee/_log)

Hombre57:
I'm at work so I can't commit a patch right now, but I've fixed that in the lockable-color-picker branch by replacing `dir` by `dir->get_basename()` where the error occur.

Floessie:
@mbajor As it doesn't seem GCC related, do the gtkmm versions differ between those builds that work and those that don't?

mbajor:
@Floessie There are more differences
e.g.
[openSUSE Factory PowerPC ppc (build success)
](http://pastebin.com/J3qJL7dJ) + [full log](https://build.opensuse.org/public/build/home:rawtherapee/openSUSE_Factory_PowerPC/ppc/rawtherapee/_log)

[openSUSE Factory PowerPC ppc64 (build failed)
](http://pastebin.com/nx5pL5F1) + [full log](https://build.opensuse.org/public/build/home:rawtherapee/openSUSE_Factory_PowerPC/ppc64/rawtherapee/_log)

 |openSUSE Factory PowerPC ppc (build success) | openSUSE Factory PowerPC ppc64 (build failed)
[...]

PR #3444, opened 2016-10-01
Fix incompatibility with glibmm 2.50 (#3440)

Floessie:
Kudos to @Hombre57 for the suggestion.

### DIFF
--- a/rtgui/dirbrowser.cc
+++ b/rtgui/dirbrowser.cc
@@ -59,7 +59,7 @@ std::vector<Glib::ustring> listSubDirs (const Glib::RefPtr<Gio::File>& dir, bool
     } catch (const Glib::Exception& exception) {
 
         if (options.rtSettings.verbose) {
-            std::cerr << "Failed to list subdirectories of \"" << dir << "\": " << exception.what () << std::endl;
+            std::cerr << "Failed to list subdirectories of \"" << dir->get_basename() << "\": " << exception.what () << std::endl;
         }
 
     }


Floessie:
@Beep6581 @heckflosse Could (one of) you have a look, if it also compiles with an older version of glibmm-2.4?

Beep6581:
I don't have access to 2.50 yet.

Floessie:
@Beep6581 That's what I meant. :smile:

Beep6581:
Compiled and runs fine using glibmm-2.46.4.

innir:
The gkt3 branch uses `dir->get_parse_name()` (see https://github.com/Beep6581/RawTherapee/blob/gtk3/rtgui/dirbrowser.cc#L62). What's the difference? And shouldn't it be the same in both cases?

Floessie:
@innir Good point, although I don't think it really matters, as it's about an error message on `stderr`. But yes, that's an inconsistency.

@Hombre57 As you did the fix in the GTK3 branch and suggested the solution here: What is closer to the original message? `get_parse_name()` or `get_basename()`?

Beep6581:
I need to choose one or the other when merging, so I'd like to know too.

Hombre57:
@Floessie @Beep6581 Well, you're right, `get_basename` is a better option as it gives the full path instead of the last folder only. I didn't cared too much wen fixing it, sorry.

Floessie:
@Beep6581 I'll care about this one directly on master tonight.

Floessie:
@Hombre57 I guess you meant `get_parse_name()` and corrected that in the master branch. Now master and gtk3 are the same in this regard.

Hombre57:
I probably need some hollyday :-/

Floessie:
@Hombre57 Probably everyone needs some time to breathe when RT5 is out.

PR #3467, opened 2016-10-18
Minor fix

kant:


README.md
```diff
@@ -29,7 +29,7 @@ Download source code tarballs:
 http://rawtherapee.com/shared/source/
 
 Git handbook:
-http://git-scm.com/book/en/
+https://git-scm.com/book/en/v2
```
Beep6581: Hello
Changing to https should be fine, but please remove the /v2. There is no need for it, and it points users to an old version when v3 comes along.

PR #3490, opened 2016-10-31
better support for Pentax K-70 and XiaoYI M1

heckflosse:


iliasg:
Ingo, I cannot complile RT (yet ..) so if you tried at least one pentax DNG we should be fine .. I am a bit anxious about the name because RT reports "RICOH PENTAX" as maker ..

heckflosse:
@iliasg I placed a printf in code before I made the change. dcraw reports 'Pentax' as maker

PR #3502, opened 2016-11-08
Pentax: ISO > 65535 was reported as 65535 in info dialog

heckflosse:
@Hombre57 Can you please test the branch with a ISO > 65535 DNG file from your camera? I could test it with a PEF only.

### DIFF
--- a/rtengine/imagedata.cc
+++ b/rtengine/imagedata.cc
@@ -440,6 +440,16 @@ void ImageData::extractInfo ()
                         }
                     }
                 } else if (mnote && (!make.compare (0, 6, "PENTAX") || (!make.compare (0, 5, "RICOH") && !model.compare (0, 6, "PENTAX")))) {
+                    // ISO at max value supported, check manufacturer specific
+                    if (iso_speed == 65535 || iso_speed == 0) {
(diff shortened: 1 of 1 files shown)


rtengine/imagedata.cc
```diff
@@ -440,6 +440,16 @@ void ImageData::extractInfo ()
                         }
                     }
                 } else if (mnote && (!make.compare (0, 6, "PENTAX") || (!make.compare (0, 5, "RICOH") && !model.compare (0, 6, "PENTAX")))) {
+                    // ISO at max value supported, check manufacturer specific
+                    if (iso_speed == 65535 || iso_speed == 0) {
+                        rtexif::Tag* baseIsoTag = mnote->getTag("ISO");
+                        if (baseIsoTag) {
+                            std::string isoData = baseIsoTag->valueToString();
+                            if (isoData.size() > 1) {
+                                iso_speed = atoi(isoData.c_str());
```
Floessie: Pro tip: Use [`stoi()`](http://www.cplusplus.com/reference/string/stoi/) instead of `atoi()`. :wink:

Hombre57:
Tested with a DNG file from the K-1, it works fine. Thanks !

PR #3563, opened 2016-12-30
Printer profile added in Preferences for soft-proofing

Hombre57:
The output profiles now retain only 'Display' device class as well as
the Monitor profile - and they both have to have rgb color space, the
printer profile retain only 'Output' device class.

MacOS can use soft-proofing but due to undocumented feature of OSX, and
since Cairo assume that the image data are sRGB, the monitor profile is
forced to 'RT_sRGB'. A warning message replace the combobox for this OS.


Hombre57:
@Beep6581 Do we need to repeat `monitor` and `printer` in the labels, since they are in `Monitor` and `Printer` frame ? I think there's a useless repetition and would prefer to keep the label as short as possible (TLDR syndrom).

Beep6581:
@Hombre57 if you're asking about `PREFERENCES_MONPROFILE` and `PREFERENCES_PRTPROFILE` then remember that one sets the **default** which you can still change, while the other is not a default choice but **the** choice. They could be shortened to just "Default profile" and "Profile". Your choice.

Hombre57:
Changes committed, now the preview image in the Editor tab(s) reflects the new soft-proofing options. The soft-proofing button is disabled if "None" is selected as Printer profile.

Beep6581:
All good, enabling/disabling of the soft-proofing button works well. +1 for merge.

Issue #3492, opened 2016-11-01
Compile support for system libraries


mattiaverga:
I'm packaging Rawtherapee for Fedora and distribution guidelines requires to use libraries already present in the system, if possible.
It is possible to compile Rawtherapee using the klt library already present as standalone package in Fedora, instead of using the bundled version?
Thanks.

sharkcz:
see pull request #3573

Beep6581:
Seems to have been implemented in RawTherapee 5.2.

PR #3573, opened 2017-01-01
use system KLT library if available

sharkcz:
It should solve issue #3492

### DIFF
--- a/rtengine/CMakeLists.txt
+++ b/rtengine/CMakeLists.txt
@@ -20,11 +20,17 @@ set (RTENGINESOURCEFILES colortemp.cc curves.cc flatcurves.cc diagonalcurves.cc 
     dirpyr_equalizer.cc
     calc_distort.cc lcp.cc dcp.cc ipretinex.cc
     cJSON.c camconst.cc
-    klt/convolve.cc klt/error.cc klt/klt.cc klt/klt_util.cc klt/pnmio.cc klt/pyramid.cc klt/selectGoodFeatures.cc
-    klt/storeFeatures.cc klt/trackFeatures.cc klt/writeFeatures.cc
     clutstore.cc
     ciecam02.cc
     )
+
+if (NOT KLT_FOUND)
+    set (RTENGINESOURCEFILES ${RTENGINESOURCEFILES}
+	klt/convolve.cc klt/error.cc klt/klt.cc klt/klt_util.cc klt/pnmio.cc klt/pyramid.cc klt/selectGoodFeatures.cc
+	klt/storeFeatures.cc klt/trackFeatures.cc klt/writeFeatures.cc
+    )
+    set (KLT_LIBRARIES)
+endif ()
 
 include_directories (BEFORE "${CMAKE_CURRENT_BINARY_DIR}")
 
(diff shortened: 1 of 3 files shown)


heckflosse:
@sharkcz Thanks for the pull request. Currently it fails building on travis-ci...

sharkcz:
Yup, I noticed the failure, although it built OK (with and without KLT) on my Fedora 24. I will take a look ...

Hombre57:
@sharkcz @heckflosse I guess that your version or klt is more recent than our ? If yes, why should we keep an outdated lib in our source tree ? Is this something generally available on most if not all distrib ?

Beep6581:
klt is not available in Gentoo or Sabayon.

rtengine/CMakeLists.txt
```diff
@@ -20,12 +20,17 @@ set (RTENGINESOURCEFILES colortemp.cc curves.cc flatcurves.cc diagonalcurves.cc
     dirpyr_equalizer.cc
     calc_distort.cc lcp.cc dcp.cc ipretinex.cc
     cJSON.c camconst.cc
-    klt/convolve.cc klt/error.cc klt/klt.cc klt/klt_util.cc klt/pnmio.cc klt/pyramid.cc klt/selectGoodFeatures.cc
-    klt/storeFeatures.cc klt/trackFeatures.cc klt/writeFeatures.cc
     clutstore.cc
     ciecam02.cc
     )
 
+IF (NOT KLT_FOUND)
```
Floessie: @sharkcz Keywords/function calls in this file are lower case, so it should read `if (...)` instead of `IF (...)`. And while you're at it, you could also get rid of the pre-CMake-2.6 repetition of the condition in `else (...)` and `endif (...)`.

Best
Flössie

sharkcz:
Fedora provides the same version of KLT (1.3.4) as RT bundles. And I couldn't find KLT in Debian or Ubuntu or OpenSUSE.

sharkcz:
Because of a possible header clash between the bundled and system copy distro packagers should remove the rtengine/klt directory completely in their %prep phase before the actual build.

Proper solution would be to move the KLT library eg. to the top source directory and adapt the buildsystem for it.

Beep6581:
Factor into the equation that we've had no reports of problems related to KLT under any OS or distro, and it seems that most distros don't even provide that package, and often what does cause issues is touching something which works fine.

> Proper solution would be to move the KLT library eg. to the top source directory

Why?

heckflosse:
I also vote to leave it for 5.1

Hombre57:
5.1 is coming, what is the status of this PR ?

@sharkcz Can you add a line into the main `cmakefiles.txt` to find the KLT dependency ? If it works fine, I can update the cmakefiles to let dev and package maintainer use its `CUSTOM_KLT` or the bundled one (default). I'd vote to keep the bundled one into the rtengine. However the maintainer is responsible of providing a version that has a working KLT ;) (...if it builds correctly, it should works fine too).

Other possible solution : if the KLT library evolves, keep our version up to date.

Beep6581:
None of the distros I use ship KLT.

I can check if RT still compiles when the PR is ready.

sharkcz:
I plan to work on the PR this week again.

Beep6581:
@sharkcz what is the status?

Beep6581:
Ping @sharkcz 
This PR should be closed and a new one for the `dev` branch opened.

PR #3686, opened 2017-02-12
Decode compressed xtrans files. Thanks to libraw for this work!

heckflosse:


heckflosse:
Branch stays open for compressed bayer files from fuji cameras.

PR #3689, opened 2017-02-14
added "Auto White Balance temperature bias" feature

agriggio:
This new slider in the White Balance tools allows to alter the computation of
the "auto white balance" by biasing it towards warmer or cooler
temperatures. The bias is expressed as a percentage of the computed
temperature, so that the result is given by `computedTemp + computedTemp * bias`.


heckflosse:
@agriggio Thanks a lot for the pr. I tested it. It builds and works fine.
One suggestion: Wouldn't it be better instead of setting the bias to 0 when auto wb is not selected, to enable/disable it dependent on wb method using `tempBias->set_sensitive(bool);` ?

agriggio:
@heckflosse yes, I agree it would be better from a UI perspective. I simply couldn't follow the logic enough to do it with confidence... it is still unclear to me when exactly the various controls are updated, so I was concerned that I could accidentally miss some enable/disable call. I'll see what I can do however.

heckflosse:
@agriggio I can help you with that if you don't mind

agriggio:
@heckflosse I don't mind at all! :+1:

heckflosse:
@agriggio I already prepared a patch for your pr. But then I detected the bug #3690 which I want to solve first to avoid confusion

agriggio:
@heckflosse yes, I noticed that too while working on this. That's part of the reason why I was not confident enough to disable controls... great if you can fix it!

TooWaBoo:
> @heckflosse 
> One suggestion: Wouldn't it be better instead of setting the bias to 0 when auto wb is not selected, to enable/disable it dependent on wb method using tempBias->set_sensitive(bool); ?

Changing the AWB bias slider should not change the WB Method to "Custom" if "Auto" is not selected.
AWB bias is not stored in profiles. (I would like to set a default in my profiles)

heckflosse:
@TooWaBoo After my patch to the pr you won't be able to change awb bias when method is not "Auto"
Awb bias is stored in profile but something is wrong with restoring the values when method is "Auto"
See #3690

TooWaBoo:
> @heckflosse 
> After my patch to the pr you won't be able to change awb bias when method is not "Auto"

...and how can I save a default value for "AWB bias" to my **profile** when the Method is "Camera" ?

agriggio:
@TooWaBoo now I understand the problem (I think). I'll see what I can do. can you please check that everything works if the WB method is set to auto? I mean, open the pp3 and you should see a TemperatureBias entry under white balance. thanks!

heckflosse:
@TooWaBoo What sense does it make to set a value which is not used? AWB bias currently is only used when Method is "Auto".

TooWaBoo:
@agriggio 
The AWB bias is saved in profiles, picked the wrong file - sorry.
@heckflosse 
My normal workflow is:
1. Apply a profile (WB Method = Camera)
2. If WB is wrong, see what "Auto" will do and therefore I want a default value for AWB bias.

Question: How can a save a profile with WB Method = Camera and AWB bias = e.g -0.8
I'll hope this makes it clear.

heckflosse:
@TooWaBoo 
After patch:
Set Method = 'Auto'
Set AWB bias = -0.8
Set Method = 'Camera'
Save profile

TooWaBoo:
@heckflosse 
Thats OK for me. :-)

heckflosse:
@agriggio #3690 is really hard to fix. I need some time. Hope you don't mind the delay but it should be fixed before making changes to your pr.

agriggio:
@heckflosse no problem, I'm not in a hurry :-) just let me know when you want me to go on with this

heckflosse:
@agriggio I guess I solved #3690. Though it deserves some tests. After tests we can continue with your pr :+1:

agriggio:
@heckflosse I think I managed to implement your suggested change (after synchronizing with the `autowbfix` branch). Let me know when I can push the new changesets

heckflosse:
@agriggio please push :+1:

heckflosse:
@agriggio Works fine. May I suggest to use the same adjuster style (with blue and yellow dot) as for Temperature?

agriggio:
@heckflosse  sure, done!

heckflosse:
@agriggio `custom_tempBias` is never read, only written. Sure we need it?

agriggio:
@heckflosse I guess you are right. At some point I was using that variable for storing `tempBias` values when changing the AWB method (similar to what `custom_equal` does), but later I changed that but forgot to remove the variable. I'll fix that, thanks!

agriggio:
@heckflosse done. Let me know if there's anything else missing

heckflosse:
@agriggio One thing I found, though that's almost splitting hairs. The precision of the awb bias value in history should be the same as the precision of the adjuster. Currently the adjuster shows for example 0.14 while history shows 0.140

agriggio:
@heckflosse no need to justify yourself... it's your playground, we play by your rules :-) 
Also, ultimately we want RawTherapee to be perfect, don't we? So, should be fixed now.

heckflosse:
@agriggio One last suggestion, but that's completely up to you:
Feel free to insert a new entry [here](https://github.com/Beep6581/RawTherapee/blob/dev/AUTHORS.txt#L18) ;-)

agriggio:
@heckflosse thanks, done! Indeed, I hope I'll be able to contribute more in the future (mostly on the usability side, sadly I'm not knowledgeable enough on image processing to work on the core)

Floessie:
@agriggio 
> I hope I'll be able to contribute more in the future (mostly on the usability side...)

That's **highly** appreciated! :+1:

heckflosse:
Works fine on Win7/64
@Beep6581 @Floessie Any objections to merge this pr soon?

Floessie:
@heckflosse Of course not. :+1:

heckflosse:
@agriggio Alberto, thanks a lot for this new feature :+1:

Beep6581:
Thank you @agriggio !

PR #3692, opened 2017-02-15
Fix autowb issues, fixes #3690

### LINKED ISSUE #3690: White Balance shows wrong values when opening a file whichs wb method is 'Auto'
To reproduce:
Open amsterdam.pef
Set Auto WB
rt shows the calculated values for Temperature and Tint, fine
awb00
reopen amsterdam.pef
rt shows the hard coded default values for Temperature and Tint...
awb01

I will try to fix that.

heckflosse:


rtengine/rtengine.h
```diff
@@ -310,6 +310,14 @@ public :
 
 };
 
+class AutoWBListener
+{
+public :
+    virtual ~AutoWBListener() {}
+    virtual void WBChanged (double temp, double green) {}
+
+};
```
Floessie: @heckflosse I know, this is cloned from other listeners, but what's the sense in providing a single callback when that single callback isn't forced to be implemented? I'd suggest the following:

```c++
class AutoWBListener
{
public :
    virtual ~AutoWBListener() = default;
    virtual void WBChanged(double temp, double green) = 0;
};
```

agriggio:
I was trying to test this, and I got something weird. See attached screenshot
screenshot_20170215_092226
I noticed this in the terminal:
```
*** BUG ***
In pixman_region32_union: Malformed region reg1
Set a breakpoint on '_pixman_log_error' to debug
```

I'm on `54d1533a7dc47143b279a602692618abf41f429d`, here are some relevant (I hope) details:

```
Compiler: gcc-6 6.2.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.18.0
Build type: Debug
Build flags: -fdiagnostics-color=always -std=gnu++11 -march=native -Werror=unused-label -fopenmp -Werror=unknown-pragmas -g
Link flags:  -march=native
OpenMP support: ON
MMAP support: ON
```

What I did was to go back and forth a few times over 2-3 images, using the neutral profile and changing only the WB to "Auto". After a while, the above happens.

rtengine/rtengine.h
heckflosse: @Floessie As you wrote, I cloned it :) Thanks for the fix :+1:

heckflosse:
@agriggio Bug confirmed.

heckflosse:
@agriggio Should be fixed now

agriggio:
@heckflosse indeed, it seems to work fine now. Thanks!

heckflosse:
Because #3689 also has the fix I will close this pr now. Please test #3689 instead of this one.

PR #3700, opened 2017-02-17
re-enable dcraw code for reading lossy DNG files

agriggio:
The relevant dcraw code was #ifdef'd out, but reenabling it seems easy and working.

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -9,6 +9,7 @@
 /*RT*/#define NO_JASPER
 /*RT*/#define LOCALTIME
 /*RT*/#define DJGPP
+/*RT*/#include <jpeglib.h>
 
 #include "opthelper.h"
 
@@ -2604,7 +2605,7 @@ void CLASS kodak_radc_load_raw()
 
 #ifdef NO_JPEG
 void CLASS kodak_jpeg_load_raw() {}
-void CLASS lossy_dng_load_raw() {}
+// RT void CLASS lossy_dng_load_raw() {}
 #else
 
 METHODDEF(boolean)
@@ -2661,6 +2662,7 @@ void CLASS kodak_jpeg_load_raw()
 }
 
 void CLASS gamma_curve (double pwr, double ts, int mode, int imax);
+/*RT*/#endif
 
 void CLASS lossy_dng_load_raw()
 {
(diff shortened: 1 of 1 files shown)


agriggio:
One glitch that I have just seen: enabling the vignetting correction from one
of the lens correction profiles that comes with Adobe DNG Converter results in
a weird artifact on lossy DNG files. Basically, the picture is divided in two
zones of different brightness by a clearly visible vertical bar, similarly to
what a graduated filter (with a very steep transition) would do. I have no
clue why this happens though. The only thing that I noticed is that the lossy
DNG image has not the same resolution as the original one, but it is a few
pixels smaller (5992x3992 instead of 6016x4016): could this be a possible reason?
I can upload some samples somewhere if needed.

Beep6581:
@agriggio the preview is known to do crazy things when using LCP files, e.g. a 20% zoom could look like a 200% zoom, or it could be distorted. If the error you're seeing only happens with the LCP tool, it might be unrelated to the dcraw change.

agriggio:
@Beep6581 no sorry, the jpeg output is affected as well. But only for lossy DNG files.
I do believe that the two issues are unrelated though.
I'll upload a sample on my google drive and post a link.

heckflosse:
As @Beep6581 I also think, the lcp 'bug' is not related to your patch.
Maybe worth to note for users of `Lossy dng files` These files are not raw, they are already demosaiced, means using Raw auto-ca correction and other raw preprocessing tools won't work with that files.
We should mention that in rawpedia.

agriggio:
@heckflosse @Beep6581 ok, then if this is not a problem, I think the patch is ready for review.

Hombre57:
Should we hide the irrelevant tools for this kind of file ? How can we know that it is already demosaiced ?

Beep6581:
@Hombre57 I would greatly prefer disabling/greying-out, not hiding.

heckflosse:
No, We should not hide the tools for this kind of files imho. If we would do that we would also have to hide them for tiff and jpeg files.

Hombre57:
From a user point of view, it would be better to hide or greying out than leaving this out uselessly. If it's possible for lossy DNG file, it should be possible for standard image (png, jpg, ...). The Batch Editor would still have them all, as usual.

heckflosse:
@Hombre57 I agree. But isn't that a different issue?

agriggio:
maybe I just don't know what I'm talking about (quite possible), but why should vignetting correction work only on raw files? isn't it performed after demosaic? 
regarding turning off tools, you are thinking of the raw tab only, or is there anything else on other tabs? can't we simply grey out the whole raw tab for non-raw images?

heckflosse:
@agriggio Currently the lcp vignetting correction works on undemosaiced data:
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/rawimagesource.cc#L1779

Maybe we can change this. I will have a look...

Hombre57:
> I agree. But isn't that a different issue?

Probably. As @agriggio said, there are already unused tools still displayed for non raw images.

heckflosse:
@agriggio #3702

agriggio:
@heckflosse +1. so I was right, I really didn't know what I was talking about :-)

heckflosse:
@agriggio I had only one file to test with but it works fine. Code looks also good and does not change too much in dcraw.cc. :+1: for merge
@Floessie @Beep6581 @Hombre57 Objections?

heckflosse:
For grey-out of tools I opened #3704

Hombre57:
No objections.

agriggio:
@heckflosse thanks!

Issue #3702, opened 2017-02-18
LCP vignetting correction only works with undemosaiced raw files


heckflosse:
For some `raw` formats (canon mraw, sraw, lossy dng and foveon) current lcp vignetting correction code does not work correctly. I will post a fix soon.

agriggio:
seems to work fine with lossy DNGs! :+1:

heckflosse:
@agriggio Thank you very much for the test :)

PR #3703, opened 2017-02-18
LCP vignetting correction only works with undemosaiced raw files. fix…

heckflosse:
…es #3702


Beep6581:
RawPedia updated, would be good if you checked whether it's all valid:
http://rawpedia.rawtherapee.com/Lens/Geometry#Lens_Correction_Profile

heckflosse:
@Beep6581 
```
* Distortion, vignetting and chromatic aberration correction are all supported in raw files, but only distortion correction is supported in non-raw files.
```
is valid. It was not valid before my changes because lcp vignetting correction did not work on 3-channel raw files.

Beep6581:
RawPedia fixed.

Issue #3678, opened 2017-02-09
Update MacPorts instructions and macOS bundle scripts


ion12:
**Update macports-based compiling instructions:**

macports variants:
`+quartz -x11`
(+no_gnome +no_x11  are obsolete)

GTK3 dependencies:
`gtk3 gtkmm3 gtk-osx-application-gtk3 adwaita-icon-theme libsigcxx2 lcms2 libiptcdata fftw-3-single clang-3.9`
(openmp is now fully integrated into clang and doesn't need any explicit setting-up)

`sudo port install wget `
wget is a requirement in the current instructions for the libiconv patch, maybe it should be mentioned in the instructions

**Update `macosx_bundle.sh` for gtk3:**

I took the https://filebin.net/kdlrguptt56syza0 patch helmi03 linked to on issue #3339 and tweaked it a bit further to remove all obsolete references to RT4 and gtk2 from `macosx_bundle.sh` and `executable_loader.in` + some cosmetic changes that you can ignore if you don't like.
I also added a hard-coded fix to include the 16x16 Adwaita icons + index.theme 
I can't find any reason to include the full set of Adwaita icons - but If anyone has a more elegant solution to include Adwaita or if I'm missing any icons please improve/update.
I should mention that I don't have a Retina display, I don't know if that requires including higher resolution icons (?)

I would like to ask about partha's build: do we really need the entire (46MB !!!) folder of Adwaita icons? seems like a shame to bloat the size of the app bundle if they are not all necessary, the build I produce through the macports toolchain comes out significantly smaller and I think it's mostly to do with these icons.

attaching the patch and a build (I'm still on 10.11 so I don't know if it will work on 10.12 as well).

cheers

[osx.patch.zip](https://github.com/Beep6581/RawTherapee/files/765068/osx.patch.zip)
https://filebin.net/oauqratd8i2i56p9

heckflosse:
@ion12  '(openmp is now fully integrated into clang and doesn't need any explicit setting-up)'

Does that mean we should review this part of rt-code:

https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/opthelper.h#L72

Edit: Can you try a build without lines 72 and 74 of opthelper.h?

Ingo

heckflosse:
[...]

PR #3706, opened 2017-02-19
update macOS bundle scripts (MacPorts workflow)

ion12:
- critical: update macosx_bundle.sh and executable_loader.ini to correctly handle gtk3 and RT5
- important: update libiconv patch to correctly handle version 1.15
- macosx_bundle.sh (line 14): change script message color so it will be more legible on dark-themed terminal
- macosx_bundle.sh (lines 188, 213-215): change build output from disk image to .app bundle + distribution-ready zipped disk image


tools/osx/macosx_bundle.sh
```diff
@@ -122,32 +110,33 @@ CheckLink "${EXECUTABLE}"
...
+ditto {"${GTK_PREFIX}","${MACOS}"}/share/themes/Mac/gtk-3.0/gtk-keys.css
+ditto {"${GTK_PREFIX}","${MACOS}"}/share/themes/Default/gtk-3.0/gtk-keys.css
+ditto {"${GTK_PREFIX}","${MACOS}"}/share/themes/Adwaita/gtk-3.0/gtk.css
+# Adwaita icons
+iconfolders=("16x16/actions" "16x16/devices" "16x16/mimetypes" "16x16/places" "16x16/status" "48x48/devices")
+for f in "${iconfolders[@]}"; do ditto {"${GTK_PREFIX}","${MACOS}"}/share/icons/Adwaita/"$f"; done
+ditto {"${GTK_PREFIX}","${MACOS}"}/share/icons/Adwaita/index.theme
+"${GTK_PREFIX}"/bin/gtk-update-icon-cache-3.0 "${MACOS}"/share/icons/Adwaita
 # fontconfig files (X11 backend only)
 if otool -L "${EXECUTABLE}" | grep -sq 'libgtk-x11-2.0'; then
```
Beep6581: @ion12  Is this still needed? libgtk-x11-2.0 is from GTK2.

tools/osx/macosx_bundle.sh
```diff
@@ -11,7 +11,7 @@
 # - GTK_PREFIX
 
 function message {
-    printf '\e[34m-- %s\e[m\n' "$*"
+    printf '\e[35m-- %s\e[m\n' "$*"
```
Beep6581: Personally I would not use colors. I have used bold in the past, but I also ended up removing that, as when people copypasta terminal output into a bug report then the formatted text gets garbled.

tools/osx/macosx_bundle.sh
ion12: I don't know if a 'libgtk-x11-3.0' exists in port 'gtk3 +x11' but since we are demanding 'gtk3 -x11 +quartz' anyway I would assume it's safe to remove this check rather than update it. Left it as it was because I'm not sure why it was in there even in gtk2 times (maybe a leftover from a time building against a x11 backend was permitted ?)

tools/osx/macosx_bundle.sh
ion12: I prefer to leave in the colors if you don't mind, the script generates a lot of output and the colors do actually help.
This is a colored message copy-pasted from the terminal and it seems to have no formatting issues:
-- Removing disk image caches

Beep6581:
Hey @ion12 what is the status here?

ion12:
@Beep6581 - not sure if @Partha1b is opening a new fork to streamline all the changes?

Beep6581:
@ion12 ok if we merge this into #3678 ?

Beep6581:
@ion12 @Partha1b good to merge?

Partha1b:
@Beep6581 and @ion12 No issues from my side.

ion12:
@Beep6581 @Partha1b 
No further build-blocking issues on my side either.
1. Still bugs me we that are creating duplicate 'themes' dirs ('$DATADIR/themes' set for RT themes in '/CMakeLists.txt' and '$DATADIR/share/themes' set for GTK/Mac themes in 'macosx_bundle.sh')
2. Can we do without the '/tmp' symlink in the executable loader script?

@Beep6581 

3. Reminder that we changed manpage location for UNIX in '/CMakeLists.txt' but not for Windows (set separately in 'rtdata/CMakeLists.txt')
4. Reminder that we are using @Partha1b's fix to patch cairo [(link)](https://github.com/Beep6581/RawTherapee/issues/3678#issuecomment-289112570). To help other users build correctly we should add this patch to the repo as you did for libiconv, or at least defer this for a seperate issue.

ion12:
@Beep6581 to clarify - Don't wait for me to make any more changes without you and @Partha1b making the decision, as I'm not confident about what is "right" in the remaining few issues/notes that I mention.
If you two are happy leaving things as they are now please go ahead and merge.

@Beep6581 @Partha1b I did compare again how the app behaves in regards to fonts: 
On my machine @Partha1b's build shows 'Sans Regular' as the main/color picker in the preferences, while the current build (that doesn't explicitely install fonts/ sets up fontconfig path) shows 'None'. 
But in practice both builds use the sames exact font in the gui, and BOTH BUILDS CRASH as soon as you try to change the font in the preferences.
But this could be carried over to a seperate issue as to not further delay the merge? (And same for Cairo bug/patch?)

Cheers

Partha1b:
@ion12 @Beep6581 
I am very sorry that I have not responded yet as I was doing other stuff including just having built the dev version 2 days ago. 🙂 

@ion12 You said that you experienced a crash with my build? Can you try my new dev build? For your convenience, the URL is: https://www.partha.com/downloads/RawTherapee-dev.app.zip

I just tried it and experienced no crash. Here is a screenshot with the Jefferson handwriting font:
screen shot 2017-03-30 at 10 09 49 am

ion12:
@Partha1b sorry still crashing on my machine. I'm still on 10.11 btw.
Did remove config/cache from other versions before testing.
The crash:
[fonts_crash](https://github.com/Beep6581/RawTherapee/files/883546/fonts.txt)

Partha1b:
@ion12 

What does the log say? Can you post/send me the log? Thanks.

The log should be at /tmp/runRT-(stuff).log.

PS: Are you using my build with your dylibs?

ion12:
[runRT-log.zip](https://github.com/Beep6581/RawTherapee/files/883969/runRT-5.0-r1-gtk3-304-g209ea35-2017-03-30-23-40-19.log.zip)
I tried removing macports paths from PATH before launching your build just in case, but there is no difference, still crashes.
Do you have confirmation from other users that they can change the font in the preferences without issues?

Partha1b:
@ion12 OK, nothing in the log. 

Good point. I don't have confirmation from other users. I just posted on the discussion board. Let see what happens.

Partha1b:
@ion12 Just ran it on another Mac myself. It does not exhibit any crashes. That Mac is also running Sierra. But My builds are compatible with 10.9 and above. So, I am not sure what's happening with your machine. 

Let's wait for some other feedback.

ion12:
@Partha1b I got my hands on another machine, and both your build and my build seem to work fine with the font picker. Sorry for the false alarm. 
Since my build doesn't install/configure fonts and still accesses the font picker correctly could this not be It an indication that indeed the font configuration is only necessary for the X11 backend? It would be lovely if you could test this with your workflow.

I have run out of ideas for what is causing this crash on my main machine... :(

I made a (hopefully) final commit for this pull request on my end, after realizing that the path change I made in CMakeLists.txt for the 'Resources' dir was relying on absolute paths and that the bundle would not run without the cmake install prefix in place.
I think I got it right this time, although maybe there are prettier ways to do it.

@Beep6581 @Partha1b I won't have any time at all to work on this further for the next 2 weeks unfortunately, but I think it's as good as it will get on my end. I would really appreciate it @Partha1b if you could double check me and try to build from the scripts, and make corrections if you see fit.
I think that my bundle structure and config/cache paths should be correct according to your intentions of maintaining Apple-esque standards.
@Beep6581 you'd like to share this for users' testing here is a compiled bundle:
https://filebin.net/roixzk8ecgqmye8o


I still get some crashes upon exiting RT, but I really don't know if this is related to packaging so I can open a separate issue next month when I have time again.

cheers.

Beep6581:
@ion12 @Partha1b thank you for all the work and testing. As I can't commit to ion12:dev and you're away, I created a new branch [osxbuild](https://github.com/Beep6581/RawTherapee/tree/osxbuild) which I will merge into dev, handled in https://github.com/Beep6581/RawTherapee/issues/3678#issuecomment-292140396

PR #3708, opened 2017-02-20
improve readability of the partial profile paste dialog under the default theme

agriggio:
The partial profile paste dialog is a bit cluttered with RT's default GTK3 theme. This tries to make it a bit more readable. 
Current look:
before

After the patch:
after

Warning: tested only on gtkmm V3.18.0

### DIFF
--- a/rtdata/themes/RawTherapee-GTK3-20_.css
+++ b/rtdata/themes/RawTherapee-GTK3-20_.css
@@ -836,5 +836,5 @@ paned.vertical > separator {
 }
 
 #PartialPasteHeaderSep {
-    color: #D8D8D8;
+    background-color: #D8D8D8;
 }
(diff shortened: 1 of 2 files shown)


heckflosse:
@agriggio Looks good here on Win7/64. But as you I'm also on 3.18

TooWaBoo:
@agriggio 
The Gtk3.20 won't work. There are many of differences between Gtk <3.20 and Gtk>=3.20 in css styling

Change this:
```
#PartialPasteHeaderSep {
    color: #D8D8D8;
}
```

into this:
```
#PartialPasteHeaderSep {
    background-color: #D8D8D8;
}
```

agriggio:
@TooWaBoo thanks for the review! I've just pushed your suggested fix

PR #3717, opened 2017-02-25
Update themes, locallab & dynamic profiles

TooWaBoo:


TooWaBoo:
Update themes to reserve space in toolpaneltabs for locallab
Some changes in preferences for dynamic profiles

PR #3719, opened 2017-02-27
Improved DCB, less macroblocking on diagonals, much faster code

cuniek:
This is a collaboration between me (Jacek Gozdz), and heckflosse (Ingo Weyrich). Code compiles nicely.

changelog:
-A bug with tiles was detected and removed
-Code is about 2x faster (5 passes + enhancement)
-Enhancement code is new, gives less 2x2 macroblocking on diagonal lines (still not as good as AMAZE)

There are some big changes in the algorithm itself. It would be nice if someone else confirmed there are no visible disadvantages of new code (when enhancement is on). 

Currently there is code for benchmarking, and a lot of comments that should be removed when optimization is over.


rtengine/demosaic_algos.cc
```diff
@@ -3554,70 +3550,90 @@ void RawImageSource::dcb_pp(float (*image)[4], int x0, int y0)
...
 
+            g2 = h0 + h1 + h2;
+		
+            // new green value
             assert(indx >= 0 && indx < u * u);
-            image[indx][1] = (2.f + image[indx][c]) * (current * g1 + (16.f - current) * g2) * 0.0625f;
+            currPix *= (current * g1 + (16.f - current) * g2) / 48.f;
+
+            // get rid of the overshot pixels
+            float minVal = min(image[indx - 1][1], min(image[indx + 1][1], min(image[indx - u][1], image[indx + u][1])));
```
Floessie, line 3627: Tip: Make use of [`std::minmax()`](http://en.cppreference.com/w/cpp/algorithm/minmax) (3rd version) and [`std::tie()`](http://en.cppreference.com/w/cpp/utility/tuple/tie) here.

rtengine/demosaic_algos.cc
```diff
@@ -3717,27 +3739,53 @@ void RawImageSource::dcb_demosaic(int iterations, bool dcb_enhance)
...
+        dcb_correction(tile, map, x0, y0);
+        dcb_color(tile, x0, y0);
+        dcb_map(tile, map, x0, y0);
+        dcb_correction(tile, map, x0, y0);
+        dcb_map(tile, map, x0, y0);
+        dcb_correction(tile, map, x0, y0);
+        dcb_map(tile, map, x0, y0);
+        restore_from_buffer(tile, buffer);
+        
+		if (!dcb_enhance)
```
Floessie, line 3776: Whitespace trouble...

rtengine/demosaic_algos.cc
heckflosse: @Floessie I just tried std::minmax() and std::tie(). It slows down the whole dcb by about 3% (tested with 2 passes and dcb enhance) . Maybe the generated code has branches while, as we both know, the std::min() and std::max() code for floats has no branches ;-)

heckflosse: @Floessie Completely unrelated to this issue, but thinking about minmax-code I just detected that there is a better solution for [this code](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/sleefsseavx.c#L1389)

```
static inline void vswap( vmask condition, vfloat &a, vfloat &b) {
    // conditional swap the elements of two vfloats
    vfloat temp = vself(condition, a, b); // the values which fit to condition
    condition = vnotm(condition); // invert the condition
    a = vself(condition, a, b); // the values which fit to inverted condition
    b = temp;
}
```
Better solution would be:
```
static inline void vswap( vmask condition, vfloat &a, vfloat &b) {
    // conditional swap the elements of two vfloats
    vfloat temp = vself(condition, a, b); // the values which fit to condition
    a = vself(condition, b, a); // the values which fit to inverted condition
    b = temp;
}
```

heckflosse:
I vote for merging the pull request for the following reasons:

1. It's better and faster than current dcb
2. We get more tests of new dcb code
3. Though there is some whitespace trouble, that will be only for a short time, as I will make further changes to dcb code in a new branch once the pr is merged

Any objections?

Edit: As soon as the pr is merged I will remove the timing code and and push to dev. Then I will create a new branch

rtengine/demosaic_algos.cc
Floessie: @heckflosse As to the first point: It's a pity that often the more elegant code is the slower. But we have `min()` and `max()` with up to four parameters in [`rt_math.h`](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/rt_math.h#L82), which could be handy here.

As to the second point: Good catch. Also requires a facepalm when thinking about it.

heckflosse: @Floessie Yes, shame on me as I wrote this facepalm code ;-)

PR #3720, opened 2017-02-28
Variadic template version of `rtengine::(min|max)()`

Floessie:
This change allows for an arbitrary number of arguments to `min()` and `max()` by using recursion on variadic template functions. The disassembly of GCC 6.3 was carefully checked for regressions, but
nothing was found other than the flipping of arguments (recursion is now `(((a,b),c),d)` and was `(d,(c,(a,b)))` before).

I also unified the common type `_Tp` to to the even more common `T`.

### DIFF
--- a/rtengine/rt_math.h
+++ b/rtengine/rt_math.h
@@ -27,57 +27,68 @@ constexpr float RT_PI_F_2 = RT_PI_2;
 constexpr float RT_INFINITY_F = std::numeric_limits<float>::infinity();
 constexpr float RT_NAN_F = std::numeric_limits<float>::quiet_NaN();
 
-template <typename T>
-inline T SQR (T x)
+template<typename T>
+constexpr T SQR(T x)
 {
-//      return std::pow(x,2); Slower than:
     return x * x;
 }
 
 template<typename T>
-inline const T& min(const T& a, const T& b)
+constexpr const T& min(const T& a)
 {
-    return std::min(a, b);
+    return a;
+}
+
+template<typename T>
+constexpr const T& min(const T& a, const T& b)
+{
+    return a < b ? a : b;
 }
 
 template<typename T, typename... ARGS>
-inline const T& min(const T& a, const T& b, const ARGS&... args)
+constexpr const T& min(const T& a, const T& b, const ARGS&... args)
 {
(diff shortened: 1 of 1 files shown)


heckflosse:
@Floessie Would it be possible to get this kind of recursion
`((a,b),(c,d))`
to reduce dependency chain?

Floessie:
@heckflosse Have to think about it tomorrow. I'm too tired now. Might make trouble with an uneven number of arguments...

heckflosse:
@Floessie no need to hurry. Was just a thought

Floessie:
Sleeping on it usually leads to enlightment. This was also the case last night. Will test and try to push it tonight.

Floessie:
@heckflosse Is this what you were thinking about?

```c++
#include <iostream>
#include <string>

namespace
{

    template<typename T>
    std::string test(const T& a)
    {
        return "(" + a + ")";
    }

    template<typename T>
    std::string test(const T& a, const T& b)
    {
        return "(" + a + "," + b + ")";
    }

    template<typename T, typename... ARGS>
    std::string test(const T& a, const T& b, const ARGS&... args)
    {
        return test(test(a, b), test(args...));
    }

}

int main()
{
    std::cout << test(std::string("a")) << std::endl;
    std::cout << test(std::string("a"), std::string("b")) << std::endl;
    std::cout << test(std::string("a"), std::string("b"), std::string("c")) << std::endl;
    std::cout << test(std::string("a"), std::string("b"), std::string("c"), std::string("d")) << std::endl;
    std::cout << test(std::string("a"), std::string("b"), std::string("c"), std::string("d"), std::string("e")) << std::endl;
    std::cout << test(std::string("a"), std::string("b"), std::string("c"), std::string("d"), std::string("e"), std::string("f")) << std::endl;
    return 0;
}
```

Output:
```
(a)
(a,b)
((a,b),(c))
((a,b),(c,d))
((a,b),((c,d),(e)))
((a,b),((c,d),(e,f)))
```

Best,
Flössie

heckflosse:
@Floessie exactly, fantastic :+1:

heckflosse:
@Floessie builds and works fine. :+1: for merge

Issue #3691, opened 2017-02-14
cli version to speedup start time / rt-cli branch


Hombre57:
I've created `rt-cli` branch that build 2 executables : `rawtherapee.exe` and `rawtherapee-cli.exe`. The later one doesn't integrate the GUI code at all but includes all the command line options, while the first one now only support directory or files through the command line options. Use `-h` to see what is supported. You'll se a new `-q` option (for _Quick Start_ mode) that will not load the HaldCLUT files nore the color profiles. Users should use this flag if they start RT through command line for every new file to process (i.e. don't use wildcards).

To test the `-q` mode, just include film simulation, custom input profile and non standard output profile to the pp3. Works fine here.

Example of processing a 10MPix PEF file with lots of processing on Core-i7 / Windows7/64 :
- rawtherapee-cli.exe - w/o `-q` : 9.08s
- rawtherapee-cli.exe - w/ `-q` : 7.50s

The difference is (should be) the gain of the `-q`flag.

This will increase the install bundle by ~2Mb. Please tell me if it's worth it.

It's still possible to revert the usual `rawtherapee.exe` to its initial state (i.e. get back all the options flags) and reserve the cli version to people that wants performance and can build themselves...

Hombre57:
I've found one limitation : when using the `-q` mode, the `(automatic icc)` will not find any suitable intput profile even if it exist, just because we haven't loaded them all. It will still process the file but with a wrong result. How should we handle this problem ? I'd suggest to put a warning when `rawtherapee-cli` is started.

heckflosse:
> suggest to put a warning when  rawtherapee-cli  is started.

No, this has to be solved, not warned about!

Hombre57:
Ok. I'll scan the input profile directory for a suitable profile if this parameter is set and only load it, not load all of them.

heckflosse:
@Hombre57 iirc I made some changes to that part of code to only load profiles on demand, not in advance...

Hombre57:
Yes, but not for the auto selection mechanism (still have to look at the code, but now going to bed ;) ).

heckflosse:
@Hombre57 Oh, my bad :(

Beep6581:
@Hombre57 how is progress with this, should we move it to 5.2?

Hombre57:
Almost ready. Working on it tomorrow, should be ok this w.e.. It's up to you to move it to 5.2 if you want, I don't mind.

Hombre57:
[...]

Issue #3543, opened 2016-12-15
variable tracking size limit exceeded


Beep6581:
While compiling a `RelWithDebInfo` build of 1901f3a44031be79a026cba0fe6cf31ec67f8bc2 using GCC-4.9.3 I got this warning:
```
/home/morgan/programs/code-rawtherapee/rtexif/nikonattribs.cc: In constructor ‘rtexif::NALensDataInterpreter::NALensDataInterpreter()’:
/home/morgan/programs/code-rawtherapee/rtexif/nikonattribs.cc:327:5: note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without
     NALensDataInterpreter ()
     ^
```

```
/home/morgan/programs/code-rawtherapee/rtexif/sonyminoltaattribs.cc: In constructor ‘rtexif::SALensIDInterpreter::SALensIDInterpreter()’:
/home/morgan/programs/code-rawtherapee/rtexif/sonyminoltaattribs.cc:547:5: note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without
     SALensIDInterpreter ()
     ^
```

```
/home/morgan/programs/code-rawtherapee/rtgui/wavelet.cc: In constructor ‘Wavelet::Wavelet()’:
/home/morgan/programs/code-rawtherapee/rtgui/wavelet.cc:30:1: note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without
 Wavelet::Wavelet () : FoldableToolPanel(this, "wavelet", M("TP_WAVELET_LABEL"), true, true)
 ^
/home/morgan/programs/code-rawtherapee/rtgui/wavelet.cc: In constructor ‘Wavelet::Wavelet()’:
/home/morgan/programs/code-rawtherapee/rtgui/wavelet.cc:30:1: note: variable tracking size limit exceeded with -fvar-tracking-assignments, retrying without
```

Explanation here:
https://github.com/v8mips/v8mips/issues/213#issuecomment-158964300

I haven't seen it before (I haven't compiled RT on this laptop before), so I'm just raising awareness in case its new to you too.

Should we have some policy on what the `-fvar-tracking-assignments` value should be when providing debug builds to the public?

Floessie:
@Beep6581 That's a massive sequence of assignments. Maybe it would help the compiler, if we turned `lenses` into a `static const std::map<>` with a uniform initialization. Then there's no need to track assignments anymore, as they are impossible.

[...]

PR #3726, opened 2017-03-04
Dynamic processing profiles

agriggio:
support for dynamic processing profiles using rules to match against the image metadata. 
there are still some rough edges that need work, but I thought the PR would be the proper place for discussion.


TooWaBoo:
I've testet the latest version, no issues found. Like it. 👍

heckflosse:
I build it and it builds fine, but I didn't really test it because I have a question (maybe more a feature request) ;-)

Would it be possible to add other exif tags to the rules?
e.g. for pixelshift files it would be fantastic to be able selecting pixelshift 'demosaic' based on this tag in upcoming pixelshift branch:

ps_tag

Ingo

heckflosse:
I found a bug:

To reproduce:

Create a new dynamic profile and select 'Pop/Pop'
Press OK to close the dialog
Open the dynamic profile to edit it
It shows `Neutral` instead of `Pop`

Seems to happen only when the name of the folder of the profile is the same as the name of the profile

e.g. 
`Pop/Pop 1` works fine
`Portrait/Portrait` does not work
`Portrait/Portrait Lejto` works fine

agriggio:
thank you all for the feedback and testing! I'll have a look at the bug
asap. I am using (or trying to at least) the existing logic for finding the
profiles with the profileStore, I must be doing something wrong...

agriggio:
@heckflosse as I suspected, the bug seems to be upstream. To reproduce: select e.g. the bundled `Pop/Pop` as your default profile for RAW images, and open a RAW file that has no .pp3. You get this:
rt-error
I'll open a separate issue (**EDIT**: see #3727)

agriggio:
> @heckflosse: Would it be possible to add other exif tags to the rules?

See https://discuss.pixls.us/t/dynamic-processing-profiles-for-rawtherapee/3438/11
With the latest commit (f7f4575) it is now possible to combine the functionality of dynamic profiles and the custom profile builder. The latter is run after the former, and can be used to fine-tune the pp3.
For your use-case, the custom profile builder can be as simple as the following Python script (warning: don't take it literally as I haven't tested it on pixel-shift RAWs...):

```
#!/usr/bin/env python

import os, sys
import ConfigParser
import pyexiv2

class CaseConfigParser(ConfigParser.RawConfigParser):
    def optionxform(self, optstr):
        return optstr

def main():    
    p = CaseConfigParser()
    p.read(sys.argv[1])
    img = p.get('RT General', 'ImageFileName')
    out = p.get('RT General', 'OutputProfileFileName')
    pp3 = CaseConfigParser()
    if os.path.exists(out):
        pp3.read(out)
    meta = pyexiv2.ImageMetadata(img)
    meta.read()

    quality = meta.get('Exif.MakerNote.Quality')
    if quality and quality.value == 'RAW (pixel shift enabled)':
        pp3.set('RAW Bayer', 'Method', 'dcb') # or whatever...
        with open(out, 'w') as outf:
            pp3.write(outf)
    exit(0)

main()
```

heckflosse:
@agriggio `can be as simple as the following Python script` may be easy to use for you and me. But for the average rt user your new dynamic profile solution is way easier to use!!!

agriggio:
@heckflosse I agree, that's why I implemented it! :-) 
But IMHO your feature request is a "power user" one, so I was proposing a solution to that scenario.
Anyway, maybe I can consider the support for user-specified tags for version 2 of the editor.
For now, would the simple one that is in place be ok for inclusion in RT? I think the code is ready for review/consideration...

TooWaBoo:
> I think the code is ready for review/consideration...

Tested, no issues found.

heckflosse:
@agriggio Absolutely your code is ready for review (already done be @Floessie) and merge :+1: 
The feature I requested is not urgent because pixelshift branch is not merged yet ;-)

rtgui/profilestore.h
```diff
@@ -208,11 +208,16 @@ class ProfileStore :
         return internalDynamicEntry;
     }
 
-    std::vector<DynamicProfileRule> &getDynamicProfileRules()
+    const std::vector<DynamicProfileRule> &getDynamicProfileRules() const
     {
         return dynamicRules;
     }
 
+    void setDynamicProfileRules(const std::vector<DynamicProfileRule> &r)
```
Floessie: :+1: I was curious if you would go for `void set...(T)` or `void set...(const T&)`, and you went for the latter which I would have also done, though the former would be faster with an r-value-ref (which isn't how it's used here).

If I'm not mistaken, you could get rid of `#include "dynamicprofile.h"` if you PIMPL'ed the `ProfileStore` (that would be much appreciated) or if you turned `std::vector<DynamicProfileRule> dynamicRules` into `std::unique_ptr<std::vector<DynamicProfileRule>> dynamicRules`. There are already so many unnecessary header dependencies in RT, we shouldn't carelessly add more. :wink:

rtgui/thumbnail.cc
```diff
@@ -217,8 +217,16 @@ rtengine::procparams::ProcParams* Thumbnail::createProcParamsForUpdate(bool retu
 
     const CacheImageData* cfs = getCacheImageData();
     Glib::ustring defaultPparamsPath = options.findProfilePath(defProf);
+    bool create = (!hasProcParams() || forceCPB);
```
Floessie: Will that change later? If not, make it `const`. There are a lot of [philosophical reasons](http://www.phoronix.com/scan.php?page=news_item&px=MTI3NDQ) to that, but recently we also stumbled upon a [real issue](https://github.com/Beep6581/RawTherapee/issues/3543) worth considering `const` in most places. One exception I see are by-value parameters as they change the function signature without doing much of substance.

rtgui/thumbnail.cc
```diff
@@ -217,8 +217,16 @@ rtengine::procparams::ProcParams* Thumbnail::createProcParamsForUpdate(bool retu
...
 
-    if (defProf == DEFPROFILE_DYNAMIC && (!hasProcParams() || forceCPB) && cfs && cfs->exifValid) {
+    Glib::ustring outFName;
```
Floessie: Same here. `?:` is your friend. :wink:

rtgui/profilestore.h
agriggio: > though the former would be faster with an r-value-ref (which isn't how it's used here).

Since I'm the outsider here, I can go that way if you prefer... but if you
want to hear my opinion (and that's an if :-), this is totally immaterial in this context. I picked the const version for consistency with other parts of the code

> if you turned std::vector<DynamicProfileRule> dynamicRules into std::unique_ptr<std::vector<DynamicProfileRule>> dynamicRules. There are already so many unnecessary header dependencies in RT, we shouldn't carelessly add more

Sure, I'll do that! :+1:

rtgui/thumbnail.cc
agriggio: :+1:

rtgui/thumbnail.cc
agriggio: :+1: (we really have different taste it seems ;-)

Floessie:
@agriggio @heckflosse IMHO Alberto did a good job here and is nice to work with. :+1: for some final cleanups, testing, and merge.

agriggio:
@Floessie I've applied your suggested changes.
There's now one more thing that might be worth discussing, although it might be a separate issue.

I think the behaviour of the "Clear" and "Custom Profile Builder" actions under "Processing Profile Operations" is is a bit confusing:

- If you have a "static" processing profile as default, then "Clear" resets
  the pp3 to the default one, and "Custom Profile Builder" does nothing.

- If you have a dynamic profile, "Clear" resets to neutral, and "Custom
  Profile Builder" rebuilds the dynamic profile.

- If you have a custom profile builder tool specified, then "Clear" does not
  invoke it, but you need to call "Custom Profile Builder" instead.

Wouldn't it be simpler to have just one operation "Reset to default" that does whatever is needed to restore the default as specified in the user's preferences? Why are two separate methods needed? (Maybe I'm just missing something, of course)

(cc @heckflosse @Beep6581)

heckflosse:
@agriggio Thanks a lot for this work :+1: 
Works fine here.
I found one small issue:
The print in line 529 of dynamicprofilepanel.cc should also be only in verbose mode.

And a feature request for future:
In the `Filter` panel of filebrowser we have informations about the cameras and lenses of all images in current folder. Wouldn't it be cool to use this informations in `Dynamic Profile Rules` too?
This would make it much easier to specify especially lenses.

agriggio:
> @heckflosse The print in line 529 of dynamicprofilepanel.cc should also be only in verbose mode.

Thanks! Fixed now.

> Wouldn't it be cool to use this informations in Dynamic Profile Rules too?

Definitely. It is probably the first on my list of improvements (regarding dynamic profiles). I just have to think about the best UI for this. Suggestions are welcome! :-)

TooWaBoo:
> Suggestions are welcome

Wouldn't it be better to merge and offer it to the users to get feedback from them? Developer are thinking in a different way as a normal user. :-)

TooWaBoo:
Suggestion:

When creating a new entry, Camera and Lens are taken from the current picture in the editor as a default.

heckflosse:
@TooWaBoo 

> Wouldn't it be better to merge and offer it to the users to get feedback from them?

Absolutely. I already voted for merge.

agriggio:
thanks!

Il 06 Mar 2017 19:14, "Ingo Weyrich" <notifications@github.com> ha scritto:

agriggio:
@TooWaBoo: that's a good suggestion, I'll put it in the to-do list :-)

Il 06 Mar 2017 18:49, "TooWaBoo" <notifications@github.com> ha scritto:

TooWaBoo:
I'm doing the German translaten and I don't understand this part from the tooltip:
> ... Use the "re:" prefix to enter a regular expression.

agriggio:
@TooWaBoo sorry, it might be unclear indeed. Essentially, you can use regular expressions [https://en.wikipedia.org/wiki/Regular_expression](url) instead of just strings for matching. 
If you write e.g. `re:^SONY ILCE-6.00$` as camera model, the rule will match against all sony alpha a6x00 cameras, because the string after the `re:` prefix will be interpreted as a regular expression.
I hope it is a bit clearer now :-)

rtgui/profilestore.h
Floessie: > want to hear my opinion (and that's an if :-), this is totally immaterial in this context. I picked the const version for consistency with other parts of the code

That's why I wrote "which I would have also done". :wink:

agriggio:
:-) :+1:

Il 06 Mar 2017 21:28, "Floessie" <notifications@github.com> ha scritto:

heckflosse:
Another good feature would be some kind of `Show which profiles will be applied to this image` :) Should be not so difficult too...

rtgui/thumbnail.cc
Floessie: Haha. :laughing: There are other ways to assign a conditional `const` value, but `?:` sure is the fastest.

TooWaBoo:
> I hope it is a bit clearer now :-)

Thanks, it's clearer now.

Floessie:
@agriggio 
> `re:^SONY ILCE-6.00$`

Is the `line start` `^` and `line end` `$` mandatory / needed?

agriggio:
@Floessie no, I just wanted to add a bit more funny chars to make it clear it was a regexp :-)

Floessie:
@agriggio :rofl: You should have used a non-matching zero-width negative look-behind for it.

agriggio:
@Floessie I'll remember that for RawPedia ;-)

Floessie:
@agriggio Now that we have some trouble with #3691 (not your fault, relax :wink:), I must admit, I've overlooked something in my review: Would you please move [those](https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/dynamicprofile.h#L66) functions out off the global namespace. And there is no need to arbitrarily break lines at the ~80th column. Looks a bit ugly...

Best,
Flössie

agriggio:
@Floessie I think it is quicker if some of you do this directly on `dev` if that's not a problem. Otherwise I'd have to merge just for this trivial fix... thanks!

Floessie:
@agriggio Agreed. @Hombre57 is moving the code to `rtengine/` for `rt-cli`, maybe he can at least move those functions into the `rtengine` namespace.

PR #3740, opened 2017-03-08
*Fix crash* with dynamic profiles

agriggio:
we have to make sure we return the default Neutral profile if none of the dynamic rules match

I found this while trying to open the RAF file mentioned at
https://discuss.pixls.us/t/rawtherapee-for-raf-files-of-fuji-s5-pro/3458/8

Should I open an issue as well?


agriggio:
hold on, maybe a was too quick in my call... the bug might be somewhere else, and I just triggered it

agriggio:
Forget about this, the issue is #3741 
Closing

Issue #3742, opened 2017-03-08
Bad artifacts (white pixels) with the attached pp3 and ARW


agriggio:
RAW file from here:
https://pixls-discuss.s3.amazonaws.com/original/2X/f/fe651efcce1684b4f1d8f3cda2d65d821752f24d.ARW

PP3 in attach (adjust the path to the "FUJI Provia 100 Generic" film simulation, and take the lens correction profile from Adobe DNG converter). The result is in the attached jpg.

Bisect says the culprit is 6e7712831a5e146aa694165cc488a3b7230c669f

![dsc03922-4](https://cloud.githubusercontent.com/assets/11982086/23713598/bd557afa-0426-11e7-973b-2b9936f2385e.jpg)
[bad.pp3.txt](https://github.com/Beep6581/RawTherapee/files/828285/bad.pp3.txt)

Floessie:
@agriggio Alberto, as you are already at it, could you bisect it further?

1. Replace all new `constexpr` with `inline` again.
2. Make those functions depending on `min/max()` use `std::min/max()`.
3. Change `min/max()` to the [older](https://github.com/Beep6581/RawTherapee/commit/6e7712831a5e146aa694165cc488a3b7230c669f#diff-557116a6e0a898ee051eb7356bded351L58) recursive form.

You should find the real problem along the way. :+1:

heckflosse:
I guess the real cuprit is [here](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/curves.h#L998). For dark parts we get a division by zero.

Floessie:
That would also be a good place for the new `constexpr max()`. :wink:

heckflosse:
Using std::min and std::max in `LIM()` function fixes it...

Floessie:
@heckflosse That's strange. It's basically the [default](http://en.cppreference.com/w/cpp/algorithm/min) implementation without the braces. Precedence issue (although I'm pretty sure those braces are not needed)?

Floessie:
My GCC 6.3 libstdc++ has this:
```c++
  template<typename _Tp>
    _GLIBCXX14_CONSTEXPR
    inline const _Tp&
    min(const _Tp& __a, const _Tp& __b)
    {
      // concept requirements
      __glibcxx_function_requires(_LessThanComparableConcept<_Tp>)
      //return __b < __a ? __b : __a;
      if (__b < __a)
	return __b;
      return __a;
    }
```

[...]

PR #3753, opened 2017-03-11
Feedback request: a new processing pipeline for "Fast Export" mode

agriggio:
As the title says, this is not (yet) a real pull request, but more a request
for feedback.

In the last few days I've been playing with a new, maybe crazy, idea about the
"Fast Export" mode of RT.

A bit of context first. I've always liked the idea of a fast mode that trades
speed for quality, but I've never been very happy about the current fast
export mode. In my opinion, it gives up too much in features in order to
achieve a (reasonable) speed advantage compared to the normal mode.

So, inspired by the `--hq` flag of DarkTable (its description, I haven't
looked at the code), I started playing with the idea of a custom processing
pipeline to be used for the Fast Export mode. The idea is that of resizing the
image as early as possible, and run the rest of the processing steps on the
resized image. So, the more you scale down the output, the more you gain. This
of course means that you also lose quality, but that is part of the "contract"
of Fast Export: if you use it, you are ready to sacrifice some quality for
speed.

After some tweaking, I've reached a point in which I think the feature is
usable. Therefore, I'd like to see if there is somebody willing to give
his/her toughts about it (and maybe try it out). I've added a new check box in
the Fast Export tab (see screenshot), that enables the new pipeline. If you
leave it unchecked, you get the old behaviour of Fast Export. If you enable
it, the idea is that you can then turn of all the "Bypass" flags and still get
a reasonable performance boost (depending on how much you resize, of course)
while at the same time obtaining an output that is much closer to the "normal"
one. For instance, on my machine, resizing to 1920x1200 (from a 6000x4000
original), the new Fast Export is about 5x faster than the normal
pipeline. The quality is not too bad, especially if the ISO is not too high.

One final note. This required some refactoring of the processing pipeline code
in `simpleprocess.cc`. I *think* (or at least hope :-) I did not break
anything, but I'm not 100% sure...

fast-export-pipeline


heckflosse:
@agriggio That sounds very interesting! Can you share a raw and pp3 which gives the 5x speedup?

michaelezra:
This is a great idea. I don't have facility to compile and test, so I will only comment on the description.
In UI it would be more consistent to have radio buttons
Fast Export Pipeline
 1. Full processing on resized image (faster)
 2. Process with bypasses then resize (accurate)

Bypass flags should be active only when 2nd radio button is selected.
New pipeline can be set as default.

agriggio:
@heckflosse I'll post a pp3 asap, but anything with noise reduction, contrast by detail level and maybe a film simulation should get you in the ballpark...
@michaelezra thanks for the feedback! I like your idea :-)

heckflosse:
@agriggio Ok, thanks for the hints :+1: 
Btw: I asked for a raw file to get in line with your tests because it's much harder to get a 5x overall speedup for a Canon CR2 file compared to e.g. a Sony ARW file. Decoding (not demosaicing) of CR2 files is much slower than ARW files. That may have a reasonable impact on the relative speedup.

agriggio:
@heckflosse I see, thanks! never shot with a canon, so I didn't know :-).
then let me restate, and say for arw you get around 5x when resizing to
1920. I have some nef files, I'll try with those as well.

Il 11 Mar 2017 22:52, "Ingo Weyrich" <notifications@github.com> ha scritto:

heckflosse:
@agriggio Ok, I'll try with arw then. If you want to try with cr2, pick a raw from [here](http://www.photographyblog.com/reviews/canon_eos_5d_mark_iii_review/sample_images/)

heckflosse:
@agriggio 
I found a bug. To reproduce:

Apply neutral profile to an image. Set Resize max width and height in fast export to 1920
Put to queue for fast export.
Process queue.

This results in a correct 1920 x ... output on `dev`
But in a 900 x ... output on your branch

agriggio:
thanks for reporting. did you use the menu or the button in the fast export
tab?

Il 12 Mar 2017 3:28 PM, "Ingo Weyrich" <notifications@github.com> ha
scritto:

heckflosse:
@agriggio the button

agriggio:
hmm, that's weird. I'll investigate

Il 12 Mar 2017 17:31, "Ingo Weyrich" <notifications@github.com> ha scritto:

agriggio:
I measured a bit more carefully. for Sony arw resized to 1920, on average I get a speedup of 3.6-3.8x (I got a max of 6x in one case, but the quality loss was very visible on that file). I also tried on a canon file, with a speedup of about 2.5. so, my claimed 5x was definitely on the optimistic side :-) still, the speedup is visible, and for "non-critical" shots (a fuzzy concept, I admit) the quality is very acceptable IMHO. I'll try to post some RAWs soon

agriggio:
@heckflosse I've fixed the resizing bug, should work properly now. I've also
tweaked the implementation a little bit.

Regarding performance and examples, here are a few:

* RAW: https://pixls-discuss.s3.amazonaws.com/original/2X/5/5bb2bd21266a8d53f46666d2a748b2e7632d3895.ARW

  PP3: https://pixls-discuss.s3.amazonaws.com/original/2X/6/6382720ced6a82573cc189f45bde46cba8303eb0.pp3

  (be sure to adjust the path for the film simulation)

  Normal pipeline:
```
[alb@allblack test]$ time -p /opt/rawtherapee/bin/rawtherapee -s -o  2016-07-26-23-40-37-0-normal.jpg -c 2016-07-26-23-40-37-0.ARWRawTherapee, version Alberto
Processing: 2016-07-26-23-40-37-0.ARW
Merging sidecar procparams
real 63.19
user 105.59
sys 7.06
```

  Fast pipeline:
```
[alb@allblack test]$ time -p /opt/rawtherapee/bin/rawtherapee -s -o  2016-07-26-23-40-37-0-fast.jpg -f -c 2016-07-26-23-40-37-0.ARW
RawTherapee, version Alberto
Processing: 2016-07-26-23-40-37-0.ARW
  Merging sidecar procparams
real 11.54
user 18.64
sys 1.96
```

  That's about 6x, the fast output is obviously worse but I'd say not too bad.


* Using the ARW and pp3 from #3742, I get a 2.9x speedup only, because the
  Retinex tool is run on the full-size image. Again, the fast mode is worse,
  but still reasonable

* ARW from here: https://pixls-discuss.s3.amazonaws.com/original/2X/9/9f478a42b16a3964b9cebd48782fb98056bacf22.ARW

  pp3 in attachment (again, adjust the film sim path)

  The speedup is about 3.8x, and the output quality is quite close IMHO.


Finally, notice that the above are what I would consider challenging
examples. For "normal" shots at low ISO, the difference in quality is quite
minor IMHO.

[DSC04028.ARW.pp3.txt](https://github.com/Beep6581/RawTherapee/files/836887/DSC04028.ARW.pp3.txt)

heckflosse:
@agriggio I really like the new fast export option :+1: 
I think Defringe module (rtengine/PF_correct_RT.cc) needs some attention too. I applied defringe with max. settings for test (radius 5 and Threshold 0)

Here's a screenshot. Left is using 'special' fast export, right is using `old` fast export.
defringe

agriggio:
@heckflosse thanks for testing! I'll check defringe. l also think there might be other tools that need some tweaking for the new pipeline (e.g. some I never use like edges or shadows&highlights)

heckflosse:
@agriggio I would check all tools which have a `radius` adjuster ;-)

heckflosse:
@agriggio This may also be worth a look to speed up tonemapping ;-)

https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/EdgePreservingDecomposition.cc#L885

heckflosse:
@agriggio some additional ideas:

1. for xtrans files use 1-pass xtrans demosaic instead of 3-pass when using fast export. Maybe @sguyader (being an xtrans user) could comment on this.
2. I can make an option to amaze demosaic to bypass nyquist check. I guess there will be no visible difference between nyquist and non-nyquist amaze for downsized images. Speedup for amaze without nyquist check could be in 10% range (just a guess and also depends on image)

agriggio:
@heckflosse thanks, noted! 
If someone else is reading this, please comment :smile: As I said before, I'm not an expert at all (quite the contrary, actually) on image processing, so if there's something else I missed, let me know.

rtengine/simpleprocess.cc
```diff
@@ -1369,41 +1369,75 @@ class ImageProcessor {
...
         ipf.lab2rgb(*tmplab, *baseImg, params.icm.working);
         delete tmplab;
     }
 
+    void adjust_procparams(double scale_factor)
+    {
+        procparams::ProcParams& params = job->pparams;
+        procparams::ProcParams defaultparams;
+
+#define ADJUST_RADIUS_(n, f)                                              \
```
Floessie: Maybe you could turn this `#define` into a template function with three parameters or a lambda?

rtengine/simpleprocess.cc
```diff
@@ -1369,41 +1369,75 @@ class ImageProcessor {
...
-            params.dirpyrDenoise.luma *= tmpScale;
-            //params.dirpyrDenoise.smethod = "shal";
-
-            const double dirpyreq_scale = min(tmpScale * 1.5, 1.0);
-            for (int i = 0; i < 6; ++i) {
-                auto &n = params.dirpyrequalizer.mult[i];
-                n = 1.0 + (n - 1.0) * dirpyreq_scale;
-            }
+        // resize image
+        LabImage *resized = new LabImage(imw, imh);
```
Floessie: A `std::unique_ptr<>` would help maintainability. Also for `tmplab`.

agriggio: :+1:, I'll do that

rtengine/simpleprocess.cc
agriggio: I would be happy to, but I'm afraid I will need some help - my knowledge of C++1x is probably too low to come up with a concise and readable solution. Do you have any suggestions?

rtengine/simpleprocess.cc
```diff
@@ -1369,41 +1369,75 @@ class ImageProcessor {
...
+        } else {
+            ADJUST_RADIUS_(sharpening.radius, scale_factor);
+        }
+        params.impulseDenoise.thresh = 
+            int(float(params.impulseDenoise.thresh) * scale_factor);
+        if (scale_factor < 0.5) {
+            params.impulseDenoise.enabled = false;
+        }
+        params.wavelet.strength =
+            int(float(params.wavelet.strength) * scale_factor);
```
Floessie: `params.wavelet.strength` is an `int`, no need to cast to `int` on assignment. Additionally, [coercion](http://www.learncpp.com/cpp-tutorial/44-implicit-type-conversion-coercion/) will implicitly convert `params.wavelet.strength` to the type of `scale_factor` (which is `double`). So simply convert this line (and the same scheme above) simply to

```c++
params.wavelet.strength = params.wavelet.strength * scale_factor;
```

or even

```c++
params.wavelet.strength *= scale_factor;
```

If you need to cast, please prefer C++-style casts over C casts.

rtgui/filecatalog.cc
```diff
@@ -1095,41 +1095,41 @@ void FileCatalog::developRequested (std::vector<FileBrowserEntry*> tbe, bool fas
             // and also those which effect is not pronounced after reducing the image size
             // TODO!!! could expose selections below via preferences
             if (fastmode) {
-                if (options.fastexport_bypass_sharpening         ) {
-                    params.sharpening.enabled          = false;
-                }
+                if (!options.fastexport_use_fast_pipeline) {
+                    if (options.fastexport_bypass_sharpening         ) {
```
Floessie: If you stumble upon crazy whitespace, remove it! It might have made sense when first conceived, but tends to rot on every indenter run. Don't miss out to look above and below, if old code hurts the eyes. :wink:

rtengine/simpleprocess.cc
Floessie: Sure, but I've got to checkout your branch first...

rtgui/filecatalog.cc
agriggio: :+1:

rtengine/simpleprocess.cc
agriggio: Thanks for the link, I'll have a look at it!
I tend to do explicit type conversions because sometimes I have to work with the Microsoft C++ compiler and I like to avoid [this annoying warning](https://msdn.microsoft.com/en-us/library/2d7604yb.aspx).
But I agree that the implicit conversions are more readable
(and I suppose RT doesn't care about the MS compiler anyway...)

rtengine/simpleprocess.cc
Floessie: Okay, [generic lambdas](http://stackoverflow.com/a/17233649) (which would be the most obvious solution) are only available in C++14. We could create a function local class to emulate it, but that would lead to more code than before. So we have to use a plain old template function:

```patch
diff --git a/rtengine/simpleprocess.cc b/rtengine/simpleprocess.cc
index fd06df66..d741797c 100644
--- a/rtengine/simpleprocess.cc
+++ b/rtengine/simpleprocess.cc
@@ -1388,23 +1388,24 @@ private:
         delete tmplab;
     }
 
+    template<typename T>
+    void adjustRadius(const T& default_param, double scale_factor, T& param)
+    {
+        const double delta = (param - default_param) * scale_factor;
+        param = default_param + delta;
+    }
+
     void adjust_procparams(double scale_factor)
     {
         procparams::ProcParams& params = job->pparams;
         procparams::ProcParams defaultparams;
 
-#define ADJUST_RADIUS_(n, f)                                              \
-        do {                                                            \
-            auto delta = (params. n - defaultparams. n) * f;            \
-            params. n = defaultparams. n + delta;                       \
-        } while (false)
-        
         params.resize.enabled = false;
 
         if (params.prsharpening.enabled) {
             params.sharpening = params.prsharpening;
         } else {
-            ADJUST_RADIUS_(sharpening.radius, scale_factor);
+            adjustRadius(defaultparams.sharpening.radius, scale_factor, params.sharpening.radius);
         }
         params.impulseDenoise.thresh = 
             int(float(params.impulseDenoise.thresh) * scale_factor);
@@ -1427,13 +1428,13 @@ private:
 
         const double dirpyreq_scale = min(scale_factor * 1.5, 1.0);
         for (int i = 0; i < 6; ++i) {
-            ADJUST_RADIUS_(dirpyrequalizer.mult[i], dirpyreq_scale);
+            adjustRadius(defaultparams.dirpyrequalizer.mult[i], dirpyreq_scale, params.dirpyrequalizer.mult[i]);
             // auto &n = params.dirpyrequalizer.mult[i];
             // n = 1.0 + (n - 1.0) * dirpyreq_scale;
         }
 
-        ADJUST_RADIUS_(defringe.radius, scale_factor);
-        ADJUST_RADIUS_(sh.radius, scale_factor);
+        adjustRadius(defaultparams.defringe.radius, scale_factor, params.defringe.radius);
+        adjustRadius(defaultparams.sh.radius, scale_factor, params.sh.radius);
 
         if (params.raw.xtranssensor.method ==
             procparams::RAWParams::XTransSensor::methodstring[
@@ -1442,8 +1443,6 @@ private:
                 procparams::RAWParams::XTransSensor::methodstring[
                     procparams::RAWParams::XTransSensor::onePass];
         }
-
-#undef ADJUST_RADIUS_
     }
 
 private:
```

HTH,
Flössie

rtengine/simpleprocess.cc
Floessie: > I tend to do explicit type conversions because sometimes I have to work with the Microsoft C++ compiler and I like to avoid this annoying warning.

Yep, that can be a PITA. But in this case it's even suboptimal, as you first explicitly convert to `float` which is then converted to a `double` implicitly.

> I suppose RT doesn't care about the MS compiler anyway...

Me, too. :smile: 

But we do care about Clang.

rtengine/simpleprocess.cc
agriggio: :+1:
FWIW, I still think the macro is more readable in this context (less typing), but I won't argue. 
I'll incorporate your patch. Thanks for the feedback!

Floessie: :+1:

rtengine/simpleprocess.cc
```diff
@@ -1361,58 +1367,47 @@ class ImageProcessor {
...
 
         // resize image
-        LabImage *resized = new LabImage(imw, imh);
-        ipf.Lanczos(tmplab, resized, tmpScale);
-        delete tmplab;
-        tmplab = resized;
+        {
+            std::unique_ptr<LabImage> resized(new LabImage(imw, imh));
+            ipf.Lanczos(tmplab.get(), resized.get(), scale_factor);
+            tmplab.swap(resized);
```
Floessie: What you really want to do, is move-assign:
```c++
tmplab = std::move(resized);
```

Floessie:
:+1: (Especially the whitespace cleanup in `filecatalog.cc`.)

rtengine/simpleprocess.cc
agriggio: @Floessie can you elaborate on why? http://www.cplusplus.com/reference/memory/unique_ptr/swap/

Floessie: Sure. :smile: 

`swap()` is meant to exchange two things. For `std::unique_ptr<>` this means, both of them live on. You let the old instance die afterwards, I know, but that's not how it was implemented before:

```c++
delete tmplab;
tmplab = resized;
```

The equivalent with `std::unique_ptr<>` is:

```c++
tmplab = std::move(resize);
```

I know, this may seem like nitpicking to you, but consider that a destructor might have side-effects, or that the memory is released earlier. Both `swap` and `move` say what they do, and whenever I read `swap` I'll ask myself, why the writer preserved the other instance and what's next with it.

Best,
Flössie

agriggio: Fair enough :+1: I've changed it to using `std::move`

iliasg:
Nice feature !!

I was thinking long ago this downsampled fast export sceme .. in a more drastical way 

1. Use a really fast joint_demosaic_downsample technique like Dcraw -h (which calculates a 3 color sample at the center of 2X2 Bayer area by just averaging (G1+G2) / 2.0 and use R and B as are.
But with this we sometimes take a visible diaplacement in R and B because in fact the R / B samples are at different centers.
 A way to overcome this easily and fastly is to build the 3 color samples at the center of the R or B pixels ..  it is beneficial (less noise) to use the strongest of R or B (strongest is the one that has more exposure and is easily detected at the pre_demosaic autoWB step .. )
Then each R or B place will have 4 neighbours of the other two colors and the average of them should be the new calculated colors.
This way we have a 2X downscaled grid and the rest steps will be 4X faster ..

iliasg:
BTW this is same as "proxy" files used for fast video editing so maybe we should use this name .

agriggio:
Thanks for the feedback @iliasg! I don't know much about demosaicing, but I'll try to learn... :smile:

agriggio:
BTW, does anyone have any objections to announcing this on pixls, to get feedback from a wider audience?

heckflosse:
@agriggio No objections from me :+1:

Beep6581:
Hello

I'm happy to see such a feature being created, allowing one to work on a downscaled version of a raw image can be useful indeed.

I have a suggestion that could prove to be more intuitive and more usable:

* To downscale the raw image I would look in the "Raw" tab, while this option is currently hiding in the "Fast export" tab. Furthermore, if I zoom to 100% and apply Noise Reduction, Tone-mapping, or any of the size-dependent tools, I see an inaccurate preview, but it should be accurate at 100% zoom.
* By moving this option from "Fast export" to the "Raw" tab, this feature would be in a more intuitive place. More importantly, by enabling downscaling of the raw image, I should be able to zoom to 100% and get an accurate preview when using size-dependent tools. This is assuming that downscaling the raw image to an e.g. 800x800 bounding box means that a preview zoom of 100% shows an image with the widest dimension being 800px.

What do you think?

heckflosse:
@Beep6581 

> To downscale the raw image

There is no downscaling of the raw image atm

Beep6581:
Ok, but I think as long as the downscaling happens before any size-dependent tools then the argument is still valid (Retinex aside).

heckflosse:
@Beep6581 The argument is valid, but to implement that a lot of code needs to be changed...

agriggio:
@Beep6581 what @heckflosse said. what you ask is reasonable but IMHO would require a much more pervasive refactoring of the code. it also applies to the current pipeline as well: if you have resize on, there is currently no way of previewing the output accurately, you always see a preview of the original size. there are even workarounds described in rawpedia :-)
ideally, the fast export mode should produce a result that is as close as possible to the normal one when the output size is the same. of course it will not always be 100% accurate in general, but that's the goal. in fact, if you find some RAWs and pp3s for which the difference is significant, please let me know!

agriggio:
@Beep6581 thinking about this a bit more, what you ask could be done by working on the idea of @iliasg above: adding a new demosaicing algo doing what `dcraw -h` does. this could also be completely independent of the fast export mode. it probably will work only for Bayer (am i right @iliasg?) but might still be worth having

heckflosse:
@agriggio 

> this could also be completely independent of the fast export mode.

For that reason we should discuss this in a separate Issue ;)

Beep6581:
> what you ask is reasonable but IMHO would require a much more pervasive refactoring of the code.

It was just a suggestion. If it's not a feasible one then I'm happy if you ignore it.

> if you have resize on, there is currently no way of previewing the output accurately, you always see a preview of the original size.

Overlooking the size of the RT preview image, the differences between it and the saved image are not the result of using the resize tool but the result of using spatially varying tools. We spent many hours trying to tune those tools to mitigate these differences - it worked well for some, but was impossible for others.

> there are even workarounds described in rawpedia :-)

Yes, I wrote them ;)

> adding a new demosaicing algo doing what dcraw -h does

That, or pixel binning, would satisfy those who are satisfied by a halving in size of the resolution (for speed reasons), but the motivation for my suggestion was to be able to scale down the image at the beginning of the pipeline to any arbitrary size and retain an accurate 100% preview. That lies outside the scope of what this PR does so I'm happy leaving my fantasizing for now.

P.S. I wonder if I could decompose a raw image to RGB, extrapolate the pixels in each channel to fill the gaps, scale down the lot, cut out a new bayer pattern from the downscaled image and move over metadata to generate a downscaled raw image of arbitrary size... I will try that.

agriggio:
> It was just a suggestion. If it's not a feasible one then I'm happy if you ignore it.

I think it was a reasonable request, but yes, I have to ignore it for this PR.

> Yes, I wrote them ;)

That was my guess indeed :-)

> the motivation for my suggestion was to be able to scale down the image at the beginning of the pipeline to any arbitrary size and retain an accurate 100% preview. That lies outside the scope of what this PR does so I'm happy leaving my fantasizing for now.

I think what you describe is very interesting. Unfortunately, I know almost nothing about digital image processing, so I can't contribute to the discussion at the moment. But I find the topic fascinating, so I plan to study it (at least a bit). I hope that in the future I could be of some help :-)

Beep6581:
What is the status here?

agriggio:
Well, I use the feature regularly and without issues. Regarding the output quality, here are a couple of recent examples so you can judge:

https://pixls-discuss.s3.amazonaws.com/original/2X/9/9295c52765138aedb9f82d4fac753e3e249a6113.jpg
https://pixls-discuss.s3.amazonaws.com/original/2X/4/4a3a353e84272a49b866208b88bf99284f42df3e.jpg

That said, I also have to add that there are tools and/or options that I never use, so I don't know how it will behave there. If you want it for 5.1, then I'd need a bit of help in stressing it (and probably tweaking the adjustments made to the procparams for the fast pipeline). If you prefer to reconsider the merge after 5.1 is out, that's ok for me as well.

Beep6581:
The choice is yours, there is no rush or need for it to be in 5.1 from my end, but it seems to be stable so if you'd like it to be in 5.1 then it's better to commit soon for wider testing.

It seems like this tool will receive small tweaks over a long period of time. If it was an ordinary tool I would prefer not to merge until the tweaks are done, so that once it gets merged it stays unchanged, to not break PP3 compatibility. But since this is not a normal tool and since it comes with a disclaimer that its designed for speed and the results are not 100% accurate, I think the lack of tweaks is acceptable. If I understand the situation correctly and if you agree, then +1 from me to merge (to get wider testing).

heckflosse:
:+1: fro me too. [:)](https://www.youtube.com/watch?v=oJDGcxAf9D8)

agriggio:
OK, then I see no blocker. I'll try to merge tomorrow.

agriggio:
I'm ready to merge. One question though: should the new fast pipeline be enabled by default (when choosing fast export, of course)? If we want as much testing as possible, it probably makes sense... opinions?

heckflosse:
I vote for 'enabled' by default

agriggio:
ok, pushed!

Issue #3751, opened 2017-03-10
Slider mode no longer gives visual feedback


Beep6581:
GTK3 lets you put the slider into "normal" or "fine" adjustment mode using Shfit, and the height change if your indicator of that. Current dev does not give any feedback as to the slider's state. It used to work before. The slider should go thick when it's in fine adjustment mode.

TooWaBoo:
Insert this at line 246 in the RT theme to get back the feedback for the fine tune mode in Gtk>3.20

```
scale.fine-tune trough {
    margin: 5px;
    padding: 1px;
}
scale.fine-tune trough highlight {
    margin: -2px;
}

```

gaaned92:
@
Windows10 gtk3.22.9 : slider goes thick

TooWaBoo:
@gaaned92 

Where did you get Gtk3.22.**9** for Windows?
Did you test with the RT or the TWB theme?

gaaned92:
@TooWaBoo 
I just made an update of MSYS2 : pacman -Syu


```
-- Found PkgConfig: C:/msys64/mingw64/bin/pkg-config.exe (found version "0.29.1")
-- Checking for module 'gtk+-3.0>=3.16'
--   Found gtk+-3.0, version 3.22.9
-- Checking for module 'glib-2.0>=2.44'
--   Found glib-2.0, version 2.50.3
-- Checking for module 'glibmm-2.4>=2.44'
--   Found glibmm-2.4, version 2.50.0
-- Checking for module 'gtkmm-3.0>=3.16'
--   Found gtkmm-3.0, version 3.22.0
```
I tested  with the RT theme only. I dont know where to insert correction in TWB theme

TooWaBoo:
> ...where to insert correction...

Have you done the test with the modification above?
The TWB theme doesn't have this issue, 'cause I'm sizing everything.

` Found gtk+-3.0, version 3.22.9`
I get the same message, but "AboutThisBuild" shows Gtk3.22.0

gaaned92:
For the TWB you are right.

AboutThisBuild shows the gtkmm-3 version and not the gtk+-3 version.  gtkmm version is the last line of the pasted message above.
Perhaps AboutthisBuild should be modified as gtk+ version is more informative.
From Gnome: "gtkmm is the C++ API for GTK+, with bindings for other GNOME libraries too". This API seems to be stable.

TooWaBoo:
Ahh, thanks, now I get it.

[...]

PR #3757, opened 2017-03-12
Fix for issue https://github.com/Beep6581/RawTherapee/issues/3751

TooWaBoo:


heckflosse:
@Floessie @agriggio @Hombre57  I'm still on gtk 3.18 and can't test the fix. @Beep6581 is afk for about a week and can't test. So I have to rely on you to test and confirm the fix :+1:

Floessie:
@heckflosse Tested with GTK 3.22.8. Works. Thanks, @TooWaBoo! Ready for merge. :+1:

agriggio:
I'm also on 3.18, sorry

Beep6581:
@TooWaBoo thank you for the fix, I confirm it works.

In addition to mentioning the issue number please also describe the problem in the commit - otherwise browsing a commit log requires looking up each issue which makes it hell to e.g. write a changelog when dealing with hundreds or thousands of commits.

I will merge this into dev once the critical cli issue (I will open one in a moment) is fixed.

Issue #3687, opened 2017-02-12
Roadmap for 5.1


Beep6581:
This is a roadmap for RT5.1.

Feature freeze and new branch around April 15, followed by a release candidate, and a final release a week later.

- [x] Pixelshift #3489
- [x] Fujifilm compressed raw support #3229 
- [x] Sharpen only edges needs review and speedup #3656. I will fix an old bug for 5.1. review and speedup is for 5.2
- [x] White Balance shows wrong values when opening a file whichs wb method is 'Auto', #3690
- [x] Changed "Processing profile operations -> Custom Profile Builder" to "Reset to default" #3760
- [x] Adding more zoom levels to the image editor #3763
- [x] Incorrect zooming with lens correction profiles #3765 
- [x] segfault on RAF files from Fuji S5 pro #3741 
- [x] Compiling with -Wall shows a large amount of warnings which should be fixed #3790 #3798
- [x] Update MacPorts instructions and macOS bundle scripts #3678 
- [x] Update rtexif with ExifTool 10.49 #3802 
- [x] colour balance differences​ between thumbnail and main preview #3804 
- [x] CLI #3691
- [x] filter-by-default #3843 
- [x] Auto Levels completely wrong for monochrome raw files, sigma raw files and canon mraw/sraw files #3863

Hombre57:
May I suggest issue #3691 ? Just drop some comments there so I can finish it ASAP (99% done).

heckflosse:
@Hombre57 added

agriggio:
@Beep6581 how about #3760, #3763 and #3765? The first should be harmless (it just makes the action slightly more general), the second is (IMHO) ready, and the third fixes a long-standing and pretty annoying (for me) bug...

Beep6581:
@agriggio agreed for the reasons you stated. I added them. I will try to test them in the evening.

agriggio:
@Beep6581 thanks! :+1:

Beep6581:
@agriggio one other thing. Yesterday, in the dead of the night, we convened in the woods, set fire to an effigy, chanted some spells, photographed a goat, and agreed that we would be happy to give you commit rights as you are active, helpful and get along well. It comes with but one snag - please don't line-wrap the code at 80 chars ;) If you agree and are willing, I will raise the cat by its leg, spin it over my head and grant you commit rights before midnight.

agriggio:
[...]

PR #3760, opened 2017-03-14
Changed "Processing profile operations -> Custom Profile Builder" to "Reset to default"

agriggio:
The behaviour has changed slightly, so that clicking the menu item now resets
to the default processing profile specified in the preferences. If this
involves calling the custom profile builder, the behaviour is the same as
before. But this is a bit more general, in that it might also simply reapply
the static default profile, or regenerate the dynamic one (depending on the
user's settings)


TooWaBoo:
Tested and works fine.

Suggestion: Instead of a separate context menu item put it into the profile combobox like (Neutral) and (Last Saved). Then it's also available in the edit window.

agriggio:
@TooWaBoo sounds reasonable to me. Anyone else has an opinion? @Beep6581 ?

Beep6581:
"(Default)" is fine with me. @Hombre57?

Compiles and works fine.

TooWaBoo:
Bug:
If (Neutral) is set as default in the preferences (e.g. for jpg) "Reset to Default" doesn't switch to (Neutral). Custom profiles are working fine.

Beep6581:
Bug confirmed using (Neutral). (Dynamic) worked fine (I have no rules, so the result was neutral).

agriggio:
@TooWaBoo thanks, confirmed. I just pushed a fix.

TooWaBoo:
Tested and works fine. 👍

agriggio:
Adding this to the profile combo box is a bit more work than I thought... I propose to keep it in the context menu for now, and merge as is. Anyone against?

heckflosse:
:+1: for merge

Edit: and also :+1:  to mark it as fixed [here](https://github.com/Beep6581/RawTherapee/issues/3687#issue-207088602) :)

PR #3763, opened 2017-03-16
Adding more zoom levels to the image editor

agriggio:
see https://discuss.pixls.us/t/usability-of-crop-tool


heckflosse:
@agriggio Thanks for the pr :+1: 
Maybe in initZoomSteps use round() instead of int() for `lbl` because for example 10/11 now is shown as 90% though 91% is much closer.

agriggio:
@heckflosse I was trying to be consistent with this code in `zoompanel.cc`:

```
void ZoomPanel::refreshZoomLabel ()
{

    if (iarea->mainCropWindow) {
        int z = (int)(iarea->mainCropWindow->getZoom () * 100);

        if (z < 100) {
            zoomLabel->set_text (Glib::ustring::compose(" %1%%", z));
        } else {
            zoomLabel->set_text (Glib::ustring::compose("%1%%", z));
        }
    }
}
```

agriggio:
In fact, I see now that I could use `Glib::ustring::compose` instead of `sprintf`...

heckflosse:
@agriggio ah, ok.
I found a small issue. 
```
        sprintf(lbl, "%d00", s);
```

should be
```
        sprintf(lbl, "%d00%%", s);
```
To verify open a new detail window and it currently shows 100 instead of 100%

agriggio:
@heckflosse thanks, I'll fix it immediately. BTW, where do you see it? What window configuration do I need to use? I always use "single editor, vertical tabs" and the only zoom indicator I see is the one on the bottom of the screen, which is drawn by the code I pasted above...
zoom

heckflosse:
@agriggio press the button with the + right from 16% to get a new detail window. At top of that window then is 100% for example

agriggio:
:+1: thanks!
Should be fixed now.

heckflosse:
@agriggio I will look for a solution for #3764 now. It is not related to your pr but maybe we can include a fix for it in your pr because it's most likely the same part of code

agriggio:
:+1:

heckflosse:
@agriggio for #3764 can you please test this code for `CropWindow::zoomOut` ?
It may be not perfect but much better than current behaviour.
```
void CropWindow::zoomOut (bool toCursor, int cursorX, int cursorY)
{

    int x = -1;
    int y = -1;

    if (toCursor) {
        x = cursorX;
        y = cursorY;
    } else {
        screenCoordToImage(xpos + imgX + imgW / 2, ypos + imgY + imgH / 2, x, y);
    }

    zoomVersion = exposeVersion;
    changeZoom (cropZoom - 1, true, x, y);
    fitZoom = false;
}
```

agriggio:
@heckflosse zooming out seems to work. However, if you zoom in again, after a while (2nd or 3rd click, it seems to vary a bit) then the preview window gets centered... I suppose you need to tweak also `CropWindow::zoomIn`

heckflosse:
@agriggio and also zoom11, which has the same bug as zoomOut

agriggio:
@heckflosse  :+1: Just to make sure we don't do things twice: you confirm you will handle this, right?

heckflosse:
@agriggio the zoomIn bug seems to kind of a timing problem (maybe a missing lock). If I zoom in slowly all works fine. If I press the zoom in button like a maniac it centers at some point. Can you confirm?

heckflosse:
I don't know if I can handle the bug in zoomIn. If it's a missing lock somewhere, I'm the wrong person to solve it. @Hombre57 and @Floessie have better skills in that lock stuff than I.

Edit: But I guess I can handle the zoom11 bug

agriggio:
@heckflosse yes, that's what I see as well. I agree it might be a concurrency issue

heckflosse:
@agriggio Can you try this for zoom11 please

```
void CropWindow::zoom11 ()
{

    int x = -1;
    int y = -1;

    if (zoomSteps[cropZoom].zoom <= cropHandler.getFitZoom()) {
        if (cropHandler.cropParams.enabled) {
            x = cropHandler.cropParams.x + cropHandler.cropParams.w / 2;
            y = cropHandler.cropParams.y + cropHandler.cropParams.h / 2;
        } else {
            int fw, fh;
            cropHandler.getFullImageSize(fw, fh);
            x = fw / 2;
            y = fh / 2;
        }

        zoomVersion = exposeVersion;
    } else {
        screenCoordToImage(xpos + imgX + imgW / 2, ypos + imgY + imgH / 2, x, y);
    }

    changeZoom (zoom11index, true, x, y);
    fitZoom = false;
}
```

agriggio:
I'll try ASAP, probably tomorrow

heckflosse:
@agriggio Ok, no hurry :+1:

TooWaBoo:
I have done some tests with the additional zoom levels and the preview quality of the intermediate zoom levels is much worse than the standard zoom levels.

**Edit:** Not only the intermediate zoom levels. The zoom level of both pictures is set to 25%. The picture is taken with my pocket cam Nikon Coolpix P330.
RAW: https://filebin.net/0n1b0yf9ih7qkk3h

unbenannt
unbenannt2

agriggio:
@TooWaBoo that is expected. the code does a quick nearest neighbour resize,
which for the extra levels is even less accurate than the standard ones. I
don't think there's an easy way around, a proper solution would require a
much more complex patch. still, IMHO this is better than before, but I
understand that you might disagree :-)

Il 16 Mar 2017 7:51 PM, "TooWaBoo" <notifications@github.com> ha scritto:

I have done some tests with the additional zoom levels and the preview
quality of the intermediate zoom levels is much worse than the standard
zoom levels.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/Beep6581/RawTherapee/pull/3763#issuecomment-287156131>,
or mute the thread

TooWaBoo:
Did you read my "Edit"? Also the standard zoom level looks worse.

heckflosse:
I agree that for the standard zoom levels (especially visible on 25%) the loss in preview quality is not acceptable!

agriggio:
sorry, I read through the email and did not see your edit. I wasn't aware
of the loss of quality in standard levels, and I agree it's an issue. I'll
try to understand why it happens.

Il 16 Mar 2017 8:39 PM, "TooWaBoo" <notifications@github.com> ha scritto:

TooWaBoo:
I may also suggest to add a checkbox in the preferences to switch this feature on or off. For me it's really hard to work with lower quality.

agriggio:
I have to understand why it happens first. once I have a clear picture,
I'll post a summary. I think your request make sense though :+1:

Il 16 Mar 2017 20:58, "TooWaBoo" <notifications@github.com> ha scritto:

heckflosse:
Another point we should consider is zooming in/out using the scroll-wheel of the mouse. I'm used to scroll fast between the old zoom levels. With the new fine-grained zoom levels it takes much longer to change the zoom level e.g. between 50% and 100%

agriggio:
@heckflosse we could have the extra levels available only for the "fit to window" feature (assuming we can fix the quality loss). would that be ok?

heckflosse:
@agriggio That would be a very good solution!

agriggio:
@heckflosse @TooWaBoo my latest commit should fix both issues:

- restore the old quality for the previous zoom levels, and

- ignore the extra levels when zooming manually

Can you please test whether it works for you?

Thanks!

heckflosse:
@agriggio That looks quite good

@TooWaBoo completely off topic, but your example file suffers from wrong white point for your camera. Maybe @iliasg can improve that

twb_wl_001

twb_wl_002

Edit: Nice cars btw :)

TooWaBoo:
The standard zoom levels are OK. The quality of the extra levels are better than before but for me (only for me) still to bad. It would be nice if this feature can be enabled/disabled in the preferences. The size different "fit to screen" on my monitor is 25% to 28%. So it's not much I will get from this feature.

@heckflosse 
I'm using a DCP file from the Nikon P7800 which has the same sensor as the P330. With this color profile I don't get the color cast.
unbenannt

heckflosse:
@TooWaBoo I also would like to see the extended zoom levels selectable in preferences!

About the colour cast and the DCP. I would like to hear the expertise of @iliasg on this because I think that it is preferable to have a correct raw white point.

heckflosse:
@agriggio Let me explain why I vote for a setting in preferences to switch between standard zoom level and extended zoom level for `fit to screen`:

On my very old laptop (which I still use to process images on holidays) I would like to be able to use the extended zoom level for `fit to screen` because the laptop has a lowres display and `fit to screen` with extended zoom level would allow me a larger preview of the whole image (even if there is a loss of quality in rendering, it's still better then viewing a much smaller but better rendered preview)

On the other side, with my main machine using a larger display, I would like to use the `fit to screen` at standard zoom level because the advantage of larger preview when using `extended zoom level` doesn't knock out the worse rendering of preview.

agriggio:
@heckflosse @TooWaBoo :+1: makes sense! I'll add a setting in preferences. I also tried to change the rescale option of the pixbuf from nearest neighbour to bilinear. the performance is still ok (tested on a 8 year old laptop), but the picture becomes too soft for me. probably more pleasing to the eye, but not so useful for an editor IMHO.

rtgui/cropwindow.cc
```diff
@@ -93,31 +93,22 @@ CropWindow::~CropWindow ()
...
-    for (int s = 45; s >= 27; s -= 3) {
-        float z = 10./float(s);
-        sprintf(lbl, "% 2d%%", int(z * 100));
-        zoomSteps.push_back(ZoomStep(lbl, z, s));
-    }
-    for (int s = 25; s >= 11; --s) {
-        float z = 10./float(s);
-        sprintf(lbl, "% 2d%%", int(z * 100));
-        zoomSteps.push_back(ZoomStep(lbl, z, s));
+        bool is_major = (s == s/10 * 10);
```
Floessie: @agriggio Sure, this will work when the compiler optimizes? Shouldn't that be a modulo?

rtgui/cropwindow.cc
```diff
@@ -1928,7 +1919,11 @@ void CropWindow::zoomIn (bool toCursor, int cursorX, int cursorY)
         }
     }
 
-    changeZoom (cropZoom + 1, true, x, y);
+    int z = cropZoom + 1;
```
Floessie, line 1937: An index should be `unsigned`.

rtgui/cropwindow.cc
agriggio: why shouldn't it? `15/10 * 10` is well defined, and definitely different from `15 * (10/10)`. See e.g. section 5.6 of the C++ standard

agriggio:
I've added the option to the preferences, but should it be on or off by default? For now I picked off (i.e. the old behaviour), but maybe for new users enabling it by default could be better? I'm not sure...

rtgui/cropwindow.cc
Floessie: Yep, you're right. Being unfamiliar with this pattern, I tested it with both `-O0` and `-O3` and it yields nearly (`cmp`vs. `test`) the same assembly as a modulo with the former and exactly the same with the latter.

iliasg:
@agriggio @heckflosse 
I am not able to build and test for the moment .. but (as always) I have an idea ;)
We can have more steps at zoom out with the old algorithm by using two successive steps
100% to 200% then to 33.33 gives 66.67% .. 100%  to 300% then 25% = 75% ..

TooWaBoo:
> @iliasg 
> We can have more steps at zoom out with the old algorithm by using two successive steps
> 100% to 200% then to 33.33 gives 66.67%

Does this mean using two successive steps (100% to 200%) we can double the number of zoom steps without loosing quality?

iliasg:
@TooWaBoo 

> Does this mean using two successive steps (100% to 200%) we can double the number of zoom steps without loosing quality?

I am not sure .. but I hope so .. as I have no time to make samples .. you can try it by applying 200%, then capture the screen, load the capture in rt and apply 33%

TooWaBoo:
> you can try it by applying 200%, then capture the screen, load the capture in rt and apply 33%

I did this test with lower zoom steps and it looks good.

unbenannt2

agriggio:
@TooWaBoo can you try this patch and see whether you like the results better?

```
diff --git a/rtgui/crophandler.cc b/rtgui/crophandler.cc
--- a/rtgui/crophandler.cc
+++ b/rtgui/crophandler.cc
@@ -380,12 +380,12 @@
 
                     Glib::RefPtr<Gdk::Pixbuf> tmpPixbuf = Gdk::Pixbuf::create_from_data (ch->cropimg, Gdk::COLORSPACE_RGB, false, 8, ch->cropimg_width, 2 * ch->cropimg_height, 3 * ch->cropimg_width);
                     ch->cropPixbuf = Gdk::Pixbuf::create (Gdk::COLORSPACE_RGB, false, 8, imw, imh);
-                    tmpPixbuf->scale (ch->cropPixbuf, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_NEAREST);
+                    tmpPixbuf->scale (ch->cropPixbuf, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_BILINEAR);
                     tmpPixbuf.clear ();
 
                     Glib::RefPtr<Gdk::Pixbuf> tmpPixbuftrue = Gdk::Pixbuf::create_from_data (ch->cropimgtrue, Gdk::COLORSPACE_RGB, false, 8, ch->cropimg_width, 2 * ch->cropimg_height, 3 * ch->cropimg_width);
                     ch->cropPixbuftrue = Gdk::Pixbuf::create (Gdk::COLORSPACE_RGB, false, 8, imw, imh);
-                    tmpPixbuftrue->scale (ch->cropPixbuftrue, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_NEAREST);
+                    tmpPixbuftrue->scale (ch->cropPixbuftrue, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_BILINEAR);
                     tmpPixbuftrue.clear ();
                 }
 

```

TooWaBoo:
> can you try this patch

Sure, if you tell me how to apply a patch :-)

agriggio:
Oh, sorry :-) Be sure you are up to date with my branch. Then save the patch
to a file (say `bilinear.patch`), go to the RawTherapee top directory, open a terminal, and type:

```
patch -p1 < bilinear.patch
```

You need to have `patch` installed. If you are on Windows, you can get it from
http://gnuwin32.sourceforge.net/packages/patch.htm

Note: you can actually also use git for this, but I'll leave the instructions
for the git experts :-)

heckflosse:
@TooWaBoo 
`git apply < bilinear.patch` should also work

Floessie:
... as well as simply `git apply bilinear.patch`.

TooWaBoo:
I'm to stupid for that. :-)
I only get error messages.
```

$ git apply < bilinear.patch
<stdin>:9: trailing whitespace.
                    tmpPixbuf->scale (ch->cropPixbuf, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_BILINEAR);
<stdin>:15: trailing whitespace.
                    tmpPixbuftrue->scale (ch->cropPixbuftrue, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_BILINEAR);
error: Anwendung des Patches fehlgeschlagen: rtgui/crophandler.cc:380
error: rtgui/crophandler.cc: Patch konnte nicht angewendet werden
```

```
$ patch -p1 < bilinear.patch
patching file rtgui/crophandler.cc
Hunk #1 FAILED at 380.
1 out of 1 hunk FAILED -- saving rejects to file rtgui/crophandler.cc.rej

```

agriggio:
Hmmm... maybe a copy/paste issue. I'm attaching the file, try with this (after having cleaned your working directory -- `git checkout -f add-more-zoom-levels` should do)
[bilinear.patch.txt](https://github.com/Beep6581/RawTherapee/files/850629/bilinear.patch.txt)

heckflosse:
@TooWaBoo 
please post output of
```
git status
```

TooWaBoo:
It workes now.
```
$ git status
Auf Branch add-more-zoom-levels
Ihr Branch ist auf dem selben Stand wie 'origin/add-more-zoom-levels'.
Änderungen, die nicht zum Commit vorgemerkt sind:
  (benutzen Sie "git add <Datei>...", um die Änderungen zum Commit vorzumerken)
  (benutzen Sie "git checkout -- <Datei>...", um die Änderungen im Arbeitsverzeichnis zu verwerfen)

        geändert:       rtgui/crophandler.cc

Unversionierte Dateien:
  (benutzen Sie "git add <Datei>...", um die Änderungen zum Commit vorzumerken)

        bilinear.patch
        rtgui/crophandler.cc.rej

keine Änderungen zum Commit vorgemerkt (benutzen Sie "git add" und/oder "git commit -a")

```

heckflosse:
@agriggio I didn't apply your patch but changed manually in dev from 
```
INTERP_NEAREST
```
to
```
INTERP_BILINEAR
```

That segfaults as soon as I open an image in editor...

agriggio:
@heckflosse I noticed this as well, but only sometimes. I don't know why, but either it is a GTK bug, or I'd say that the fact that it was working previously is just pure luck... `INTERP_BILINEAR` should work there. Here's the documentation of `Gdk::Pixbuf::scale`:

```  
  /** Creates a transformation of the source image @a src by scaling by
   *  @a scale_x and @a scale_y then translating by @a offset_x and @a offset_y,
   * then renders the rectangle ( @a dest_x, @a dest_y, @a dest_width,
   *  @a dest_height) of the resulting image onto the destination image
   * replacing the previous contents.
   * 
   * Try to use scale_simple() first, this function is
   * the industrial-strength power tool you can fall back to if
   * scale_simple() isn't powerful enough.
   * 
   * If the source rectangle overlaps the destination rectangle on the
   * same pixbuf, it will be overwritten during the scaling which
   * results in rendering artifacts.
   * 
   * @param dest The Gdk::Pixbuf into which to render the results.
   * @param dest_x The left coordinate for region to render.
   * @param dest_y The top coordinate for region to render.
   * @param dest_width The width of the region to render.
   * @param dest_height The height of the region to render.
   * @param offset_x The offset in the X direction (currently rounded to an integer).
   * @param offset_y The offset in the Y direction (currently rounded to an integer).
   * @param scale_x The scale factor in the X direction.
   * @param scale_y The scale factor in the Y direction.
   * @param interp_type The interpolation type for the transformation.
   */
  void scale(const Glib::RefPtr<Gdk::Pixbuf>& dest, int dest_x, int dest_y, int dest_width, int dest_height, double offset_x, double offset_y, double scale_x, double scale_y, InterpType interp_type) const;
```

TooWaBoo:
The patch didn't work. So I did it the same way as @heckflosse and RT crash when I try to open a picture.

agriggio:
@TooWaBoo thanks for testing. I really don't know why `INTERP_BILINEAR` crashes. I guess I have to dig deeper...

heckflosse:
@agriggio This is from dev with INTERP_BILINEAR. Line numbers may be different to your branch...
```
Program received signal SIGSEGV, Segmentation fault.
0x00000000002e060f in ?? () from C:\msys64\mingw64\bin\libgdk_pixbuf-2.0-0.dll
(gdb) bt full
#0  0x00000000002e060f in ?? ()
   from C:\msys64\mingw64\bin\libgdk_pixbuf-2.0-0.dll
#1  0x00000000002e2019 in ?? ()
   from C:\msys64\mingw64\bin\libgdk_pixbuf-2.0-0.dll
#2  0x00000000002e2805 in ?? ()
   from C:\msys64\mingw64\bin\libgdk_pixbuf-2.0-0.dll
#3  0x00000000002d9f4c in ?? ()
   from C:\msys64\mingw64\bin\libgdk_pixbuf-2.0-0.dll
#4  0x0000000000567d50 in CropHandler::<lambda(gpointer)>::operator()(gpointer) const (__closure=0x0, data=0x1e8b1ab0)
    at H:/rt5master/rtgui/crophandler.cc:381
        imw = 1206
        tmpPixbuf = {pCppObject_ = 0x18fea0d0}
        tmpPixbuftrue = {pCppObject_ = 0x11261d50}
        czoom = 1000
        imh = 803
        idle_helper = 0x1e8b1ab0
        ch = 0x1eed1d38
#5  0x000000000056801a in CropHandler::<lambda(gpointer)>::_FUN(gpointer) ()
    at H:/rt5master/rtgui/crophandler.cc:410
No locals.
#6  0x00000000006839ed in IdleRegister::<lambda(gpointer)>::operator()(gpointer) const (__closure=0x0, data=0x121184d0) at H:/rt5master/rtgui/guiutils.cc:62
        data_wrapper = 0x121184d0
#7  0x0000000000683a72 in IdleRegister::<lambda(gpointer)>::_FUN(gpointer) ()
    at H:/rt5master/rtgui/guiutils.cc:72
No locals.
#8  0x0000000070ec17d7 in ?? () from C:\msys64\mingw64\bin\libgdk-3-0.dll
#9  0x00000000685f1c1e in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#10 0x00000000685f1f38 in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#11 0x00000000685f23c3 in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#12 0x000000000584a2ed in ?? () from C:\msys64\mingw64\bin\libgtk-3-0.dll
#13 0x0000000063ea4170 in ?? () from C:\msys64\mingw64\bin\libgtkmm-3.0-1.dll
#14 0x00000000005d469a in main (argc=1, argv=0x7472710)
    at H:/rt5master/rtgui/main.cc:381
        exname = "H:\\rt5master\\build3\\Debug\\rawtherapee.exe", '\000' <repeats 470 times>
        exePath = {static npos = 18446744073709551615, string_ = {
            static npos = <optimized out>,
            _M_dataplus = {<std::allocator<char>> = {<__gnu_cxx::new_allocator<char>> = {<No data fields>}, <No data fields>},
              _M_p = 0x7fd4fd0 "H:\\rt5master\\build3\\Debug"},
            _M_string_length = 25, {
              _M_local_buf = "\036\000\000\000\000\000\000\000▒bR\a\000\000\000", _M_allocated_capacity = 30}}}
        exnameU = L"H:\\rt5master\\build3\\Debug\\rawtherapee.exe", '\000' <repeats 470 times>
        consoleOpened = false
        m = <incomplete type>
        icon_path = {static npos = 18446744073709551615, string_ = {
            static npos = <optimized out>,
            _M_dataplus = {<std::allocator<char>> = {<__gnu_cxx::new_allocator<char>> = {<No data fields>}, <No data fields>},
              _M_p = 0x8ae5600 "H:\\rt5master\\build3\\Debug\\.\\images"},
            _M_string_length = 34, {
              _M_local_buf = "\"\000\000\000\000\000\000\000\002\000\000\000\000\000\000", _M_allocated_capacity = 34}}}
        defaultIconTheme = {pCppObject_ = 0x8abb030}
        screen = {pCppObject_ = 0x8b256d0}
        rtWindow = 0x8c43f00
(gdb)
```

Floessie:
If you ask me, the `IdleHelper` thing smells a bit. Maybe it's worth a closer look...

agriggio:
@Floessie I agree. It might be a concurrency issue...

TooWaBoo:
I've found a fix or a workaround for the crash.

```
>
> tmpPixbuf->scale (ch->cropPixbuf, 0, 0, imw, imh, 1, 1, czoom, czoom, Gdk::INTERP_TILES);
>                                                   ^  ^
> tmpPixbuftrue->scale (ch->cropPixbuftrue, 0, 0, imw, imh, 1, 1, czoom, czoom, Gdk::INTERP_TILES);
>                                                           ^  ^
```

I changed "INTERP_BILINEAR" into "INTERP_TILES" to show the pixel without antialiasing when upscaling.

Like it, looks good.  :smiley:

Floessie:
@agriggio I was on the wrong track: It was an OOB access, biting only sometimes. Fixed directly on `dev`.

TooWaBoo:
Maybe this is a better fix for the crash bug

```
            if (ch->cropimg) {
                if (ch->cix == ch->cropX && ch->ciy == ch->cropY && ch->ciw == ch->cropW && ch->cih == ch->cropH && ch->cis == (ch->zoom >= 1000 ? 1 : ch->zoom / 10)) {
                    // calculate final image size
                    float czoom = ch->zoom >= 1000 ?
                        ch->zoom / 1000.f :
                        float((ch->zoom/10) * 10) / float(ch->zoom);
-->                 int imw = ch->cropimg_width * czoom - 1;
-->                 int imh = ch->cropimg_height * czoom - 1;

                    if (imw > ch->ww) {
                        imw = ch->ww;
                    }

                    if (imh > ch->wh) {
                        imh = ch->wh;
                    }

                    Glib::RefPtr<Gdk::Pixbuf> tmpPixbuf = Gdk::Pixbuf::create_from_data (ch->cropimg, Gdk::COLORSPACE_RGB, false, 8, ch->cropimg_width, 2 * ch->cropimg_height, 3 * ch->cropimg_width);
                    ch->cropPixbuf = Gdk::Pixbuf::create (Gdk::COLORSPACE_RGB, false, 8, imw, imh);
                    tmpPixbuf->scale (ch->cropPixbuf, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_TILES);
                    tmpPixbuf.clear ();

                    Glib::RefPtr<Gdk::Pixbuf> tmpPixbuftrue = Gdk::Pixbuf::create_from_data (ch->cropimgtrue, Gdk::COLORSPACE_RGB, false, 8, ch->cropimg_width, 2 * ch->cropimg_height, 3 * ch->cropimg_width);
                    ch->cropPixbuftrue = Gdk::Pixbuf::create (Gdk::COLORSPACE_RGB, false, 8, imw, imh);
                    tmpPixbuftrue->scale (ch->cropPixbuftrue, 0, 0, imw, imh, 0, 0, czoom, czoom, Gdk::INTERP_TILES);
                    tmpPixbuftrue.clear ();
                }
```

Floessie:
@TooWaBoo Hm, as I read it, `Gdk::Pixbuf::create_from_data()` was provided with twice the height the source image really had. But maybe I misread it...

TooWaBoo:
> @TooWaBoo Hm, as I read it, Gdk::Pixbuf::create_from_data() was provided with twice the height

This was done by @agriggio 
My fix are only the two lines with the arrows in front (add - 1) and I changed "INTERP_BILINEAR" into "INTERP_TILES". The changes are done by female intuition = I don't know what I'm doing.  :grinning:

agriggio:
actually, that code is not mine :-) I only changed the zoom levels and the
interpolation mode.

Il 18 Mar 2017 11:04 PM, "TooWaBoo" <notifications@github.com> ha scritto:

TooWaBoo:
@agriggio 

Did you test my "female intuition fix 😀 " if it works on Linux too?
Works fine on Windows 10 x64 Gtk3.22.9

agriggio:
no time for testing yet. will do later today

Il 18 Mar 2017 11:21 PM, "TooWaBoo" <notifications@github.com> ha scritto:

agriggio:
@TooWaBoo I had a (quick) look at the code. I agree with @Floessie, the fact
that the pixbuf was created with twice the height of the source image seems to
be the culprit. I'm not sure I understand your fix instead... maybe you can
elaborate?

agriggio:
@TooWaBoo BTW, I agree that `Gdk::INTERP_TILES` is the best choice here :+1:

TooWaBoo:
> I'm not sure I understand your fix instead... maybe you can elaborate?

Sorry I can't. I only had the intuition that it has todo with the size or position of the image and then I played with some numbers. 😀

agriggio:
ok, then I hope you don't mind if I take @Floessie's fix :-)

Il 19 Mar 2017 12:25 PM, "TooWaBoo" <notifications@github.com> ha scritto:

TooWaBoo:
@agriggio 

I've compiled and tested the latest version (commit 7689502) without any issue. 👍

heckflosse:
@agriggio Builds and works fine here :+1: Did you include my patches for zoomOut and zoom11?

agriggio:
@heckflosse not yet. will do tonight!

Il 19 Mar 2017 3:39 PM, "Ingo Weyrich" <notifications@github.com> ha
scritto:

agriggio:
@heckflosse I've integrated your patches. For me, this is ready for consideration now. One last thing: what should be the default setting for the option to enable the extra zoom levels? Should it be on or off?

TooWaBoo:
**Bug:** JPGs without an embeded color profile shows wrong colors.

unbenannt

Floessie:
@TooWaBoo Looks like another bug I introduced to `ICCStore`. I'll have a look.

Floessie:
@TooWaBoo @agriggio Ah, no. It's just that 5632c787a3b44f4987dfff33c52d85670d410613 wasn't merged yet from `dev`. :sweat_smile:

agriggio:
@Floessie ok. I propose not to merge from `dev` however, but to simply wait for @Beep6581 testing it and then merging this to `dev` if nobody is against

TooWaBoo:
@all 
Sorry for the panic. 😀

Floessie:
@agriggio It won't hurt, if you merge from `dev`...

agriggio:
@Floessie ok, done !

heckflosse:
@agriggio 

> what should be the default setting for the option to enable the extra zoom levels? Should it be on or off?

I vote for 'off' and to document it on rawpedia and mention it in release notes of 5.1

Not that I won't use it. On my old laptop with small screen it's clearly an improvement :+1:

agriggio:
@heckflosse sounds good to me :+1:

TooWaBoo:
@agriggio 

Suggestion: The button "fit to crop" does nothing when no crop is set. It would be nice if this button would then do a "fit to screen".

agriggio:
@TooWaBoo sounds reasonable... done!

TooWaBoo:
> @TooWaBoo sounds reasonable... done!

Thanks, tested and works fine. 👍

Beep6581:
Love the new zoom levels!

Beep6581:
New zoom level: https://i.imgur.com/YBJGD3s.jpg
Old zoom level: https://i.imgur.com/fAB4VBV.jpg

Do we really need a Preferences checkbox for this? The quality difference is very minor. You get the same quality degradation if you use the Lens/Geometry tools.

My vote in order of preference (admittedly after limited testing):
1. No Preferences option.
2. Preferences option enabled by default.

agriggio:
I agree with @Beep6581

TooWaBoo:
Me too. Even on my huge monitor it's on. 😃

heckflosse:
Ok from me too

agriggio:
:+1: I'll remove the option then, and push

PR #3789, opened 2017-03-30
Ipwavelet heap buffers

agriggio:
I've incorporated @heckflosse and @Floessie feedback. I don't have the time right now to dig deeper into the real cause of the crash, but with this patch it goes away...

### DIFF
--- a/rtengine/alignedbuffer.h
+++ b/rtengine/alignedbuffer.h
@@ -33,7 +33,7 @@ private:
     void* real ;
     char alignment;
     size_t allocatedSize;
-    int unitSize;
+    unsigned int unitSize;
 
 public:
     T* data ;
@@ -69,7 +69,7 @@ public:
     * @param structSize if non null, will let you override the default struct's size (unit: byte)
     * @return True is everything went fine, including freeing memory when size==0, false if the allocation failed
     */
-    bool resize(size_t size, int structSize = 0)
+    bool resize(size_t size, unsigned int structSize = 0)
     {
         if (allocatedSize != size) {
             if (!size) {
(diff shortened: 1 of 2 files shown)


agriggio:
anyone against merging this?

heckflosse:
:+1:

Floessie:
Still we're missing a proof, that it really was a stack overflow.

agriggio:
you are right. in fact, I wonder why I am the only one seeing this, but
only in​ debug builds. I tried on different machines and with different
compilers, and I can reproduce it consistently (at least I could -- haven't
tried in the last few days). I suppose I could at least try to debug it --
it doesn't sound like fun but I'll see what I can do...

agriggio:
So, I dug a bit deeper. I think it was a stack corruption, but moving the buffers to the heap was not the proper fix, it was just hiding the problem. Here's a one-liner that **seems to** fix the problem:

```diff
diff --git a/rtengine/ipwavelet.cc b/rtengine/ipwavelet.cc
--- a/rtengine/ipwavelet.cc
+++ b/rtengine/ipwavelet.cc
@@ -537,7 +537,7 @@
     // begin tile processing of image
 
     //output buffer
-    int realtile;
+    int realtile = 0;
 
     if(params->wavelet.Tilesmethod == "big") {
         realtile = 22;
```

`realtile` is uninitialized when `params->wavelet.Tilesmethod == "full"`. Later on, we have:

```c++
    int tilesize = 128 * realtile;
    int overlap = (int) tilesize * 0.125f;
```

and further down our "infamous" stack-allocated buffers:

```c++
                    float Vmask[height + overlap] ALIGNED16;
                    float Hmask[width + overlap] ALIGNED16;
```

In one of my crashing runs from within gdb, here's what I saw:

```
(gdb) p overlap
$22 = -117191616
```

I'm going to push this on `dev` directly. I think it should solve the problem (I haven't seen crashes after applying it), but even if it doesn't and there's something else going on, it is still something that needs to be fixed. Interesting that neither `-Wall` nor cppcheck caught this...

Floessie:
@agriggio Alberto, that's it! :+1: :tada: Very good job!

PR #3791, opened 2017-03-30
Add a brighter TooWaBlue theme

TooWaBoo:
Bright
bright

Standard
normal

Dark
dark

### DIFF
--- a/rtdata/themes/TooWaBlue-Bright-GTK3-20_.css
+++ b/rtdata/themes/TooWaBlue-Bright-GTK3-20_.css
@@ -1,0 +1,69 @@
+/*
+  This file is part of RawTherapee.
+
+  Copyright (c) 2016-2017 TooWaBoo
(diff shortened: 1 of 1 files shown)


TooWaBoo:
Is there anything wrong with this theme?

agriggio:
I haven't tested it and I have nothing against it in principle, but aren't the three variants a bit too close to each other? The differences seem fairly minimal to me (but maybe they look more similar in the screenshots than in actual use?)

TooWaBoo:
The differences aren't big but big enough that some people like the brighter theme more. :smile:
In the old Gtk2 version there exists much more themes with less differences but it's up to you if you want to add the theme.

agriggio:
well, it's not up to me but more choices are good, if people find these
useful then :+1: for me

Beep6581:
Agreed that they are so similar one wonder's whether it's worth it, but then the file is only 3kB, so...

PR #3808, opened 2017-04-06
Update Deutsch locale

TooWaBoo:
Fast Export


Beep6581:
Hey @TooWaBoo, thank you for the update. Do you use `tools/generateTranslationDiffs` to sort these files, or something else? If you use `generateTranslationDiffs`, please paste the output of running the command `locale` in a terminal. I ask because the sorting method is incorrect - I have to re-sort your updates - so it would be good to nail this down.

TooWaBoo:
No, I've used the sort function of my editor. Next time I'll use `generateTranslationDiffs` with command `locale`.

Beep6581:
@TooWaBoo or don't sort at all - then the diff is readable, and the file will get sorted the next time I run generateTranslationDiffs.

Issue #3788, opened 2017-03-30
Bug with LCP profiles parsing


kznsq:
The XML parser do not accumulate the text while parsing with the 'LCPProfile::XmlTextHandler' and causes wrong reading numbers.
You can set the small (3 bytes, for example) BufferSize in the LCPProfile::LCPProfile and debug values readed, for example, here:
LCPProfile::XmlTextHandler(..)
{
..
pProf->pCurCommon->imgXCenter = atof(raw);
..
}

agriggio:
Fix in 633af0b (branch `lcp-parser-fix-xml-handle-text`). As usual, I went for the smallest patch (the LCP code would benefit from a general cleanup, but that's a separate issue). Please test, thanks!

agriggio:
@kznsq do you have an LCP profile that used to make the parser fail? I'd like to close this but I want to test your original scenario first. Thanks!

kznsq:
Hi!
Here the file.

Thanks for your response!

Also the LCPMapper needs to support fisheye mapping like this

void LCPMapper::correctDistortion(float& x, float& y) const
{
    if(!isFisheye) // flag needs to be added
    {
        double xd = (x - mc.x0) / mc.fx, yd = (y - mc.y0) / mc.fy;

        const float* aDist = mc.param;
        double rsqr = xd * xd + yd * yd;
        double xfac = aDist[swapXY ? 3 : 4], yfac = aDist[swapXY ? 4 : 3];

        double commonFac = (((aDist[2] * rsqr + aDist[1]) * rsqr +
aDist[0]) * rsqr + 1.)
                           + 2. * (yfac * yd + xfac * xd);

        double xnew = xd * commonFac + xfac * rsqr;
        double ynew = yd * commonFac + yfac * rsqr;

        x = xnew * mc.fx + mc.x0;
        y = ynew * mc.fy + mc.y0;
    }
    else
    {
        double u = x;
        double v = y;
        double u0 = mc.x0;
        double v0 = mc.y0;
        double du = u - u0;
        double dv = v - v0;
        double fx = mc.fx;
        double fy = mc.fy;
        double k1 = mc.param[0];
        double k2 = mc.param[1];
        double r = std::sqrt(du * du + dv * dv);
        double f = std::sqrt(fx*fy);
        double th = std::atan(r/f);
        double thCommon = (1.0/r)*(th + k1*th*th*th + k2*th*th*th*th*th);
        float ud = thCommon*fx*du + u0;
[...]

PR #3809, opened 2017-04-08
Correctly apply LCP "distortion" correction for Fisheye lenses

agriggio:
Note: as discussed e.g. at http://lensfun.sourceforge.net/manual/corrections.html, this is really a combination of distortion correction and change of projection (from fisheye to rectilinear), but "distortion correction" is how the Adobe camera model calls it, and this is how it appears in the RT gui

### DIFF
--- a/rtengine/dcrop.cc
+++ b/rtengine/dcrop.cc
@@ -1145,15 +1145,33 @@ bool Crop::setCropSizes (int rcx, int rcy, int rcw, int rch, int skip, bool inte
     parent->ipf.transCoord (parent->fw, parent->fh, bx1, by1, bw, bh, orx, ory, orw, orh);
     
     if (check_need_larger_crop_for_lcp_distortion(parent->fw, parent->fh, orx, ory, orw, orh, parent->params)) {
-        double dW = double(parent->fw) * 0.15 / skip; // TODO  - this is hardcoded ATM!
-        double dH = double(parent->fh) * 0.15 / skip; // this is an estimate of the max
-                                                      // distortion relative to the image
-                                                      // size. BUT IS 15% REALLY ENOUGH?
-                                                      // In fact, is there a better way??
-        orx = max(int(orx - dW/2.0), 0);
-        ory = max(int(ory - dH/2.0), 0);
-        orw = min(int(orw + dW), parent->fw - orx);
-        orh = min(int(orh + dH), parent->fh - ory);
+        // TODO - this is an estimate of the max distortion relative to the image size. ATM it is hardcoded to be 15%, which seems enough. If not, need to revise
+        int dW = int(double(parent->fw) * 0.15 / (2 * skip));
+        int dH = int(double(parent->fh) * 0.15 / (2 * skip));
+        int x1 = orx - dW;
+        int x2 = orx + orw + dW;
+        int y1 = ory - dH;
+        int y2 = ory + orh + dH;
+        if (x1 < 0) {
+            x2 += -x1;
+            x1 = 0;
+        }
+        if (x2 > parent->fw) {
+            x1 -= x2 - parent->fw;
+            x2 = parent->fw;
+        }
+        if (y1 < 0) {
+            y2 += -y1;
+            y1 = 0;
+        }
+        if (y2 > parent->fh) {
+            y1 -= y2 - parent->fh;
+            y2 = parent->fh;
+        }
+        orx = max(x1, 0);
+        ory = max(y1, 0);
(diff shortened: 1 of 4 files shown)


agriggio:
See also #3788

Beep6581:
+1 for merge.

agriggio:
I've fixed a corner case when you would get a wrong view for detail windows placed close to the corners. Now it should be ready for merging. If there are no objections, I'll do it either later tonight or tomorrow

morosakubek:
Hi, @agriggio,  there is still a problem with Canon's compact cameras. For example, with the Canon PowerShot G1 X model (RAW+lcp: https://drive.google.com/drive/folders/0BzMK6gsaNS8wdmh2eG1fcXZtdmM?usp=sharing)

agriggio:
thanks for reporting, I'll have a look ASAP

agriggio:
Hi @morosakubek, after some investigation, the distortion correction code is fine. The problem with your camera is that there is no entry for it in `camconst.json`, so RawTherapee can't determine the correct raw crop. You can fix it by using the following `camconst.json` (to put in your user configuration directory for RawTherapee -- see the comment at the beginning of the global `camconst.json` for more info):

```json
{"camera_constants": [
    {
        "make_model": "Canon PowerShot G1 X",
        "raw_crop": [ 200, 144, 4168, 3128 ], // Optional official 4/3 frame 4160x3120, 4pix  borders, Left Border 204-4, Top Border 148-4
        "masked_areas": [ 148, 2, 3340, 78 ],
        "ranges": { "white": 16300 }
    }    
]}
```

agriggio:
@Beep6581 might make sense to include the above in RT's `camconst.json`? (after polishing probably...). I just copied it from the entry for the G1 X MKII, but it seems to work

Beep6581:
@agriggio sure, are all of the values correct?

agriggio:
you can probably ignore the white value, it was a leftover from cut&paste. As for the rest, I'm not 100% sure but the result looks the same as the embedded JPEG of the camera and that created by the Adobe DNG converter.

morosakubek:
This value is bad. Too much crop compared to CameraRaw (D1 X and D1 X Mark II are two different cameras). Another issue is that I still have distorsia. I recently checked the Canon G7 Mark II on RawTherapee, and the same was on the narrower focal length. The difference occurs between CameraRaw and RT before applying LCP to CameraRaw, so it may be a problem reading the .CR2 files themselves in RawTherapee?

agriggio:
@morosakubek you are right -- I think yesterday I messed up the pp3 and came to a wrong conclusion. The LCP doesn't correct the distorsion properly, but I have no idea why. As a workaround, the automatic distortion correction does a good job here.

heckflosse:
@agriggio Alberto, maybe the correction is a bit off because RT passes 0 as focusDistance to the lcp code?

I also compared the FocusDistance values in the lcp file with the values in other lcp files. I don't know which unit FocusDistance is, but if it's `m` then the lcp covers only the range >= 64.4 m while for example
`PENTAX_smc_PENTAX-DA_18-55mm_F3.5-5.6_AL_-_RAW.lcp` covers the range >= 0.25 m

agriggio:
@heckflosse I just tried with the git master of LensFun, using [this script](https://github.com/lensfun/lensfun/blob/master/apps/lensfun-convert-lcp) to import the LCP profile, and then the example application `lenstool` that comes with lensfun, like this:

```
./lenstool --database=/home/alb/.local/share/lensfun/_lcps.xml -C"Canon PowerShot G1X" -L"fixed lens" -d /home/alb/Downloads/IMG_4077.png -o /home/alb/Downloads/out.png
```

I get the exact same result as RT. So, I'd say the LCP file is not precise for some reason, *or maybe* ACR is applying some transformation that is not documented in the Adobe lens model...

morosakubek:
I do not know if this info will help but Darktable has the ability to correct for canon g1x.

agriggio:
yes, darktable uses lensfun, which does a good job. eventually, also RT
will be able to use lensfun, hopefully in a not too distant future :-)
in the meantime, you can try auto distortion correction, it seemed ok to me

heckflosse:
@agriggio 

> eventually, also RT will be able to use lensfun, hopefully in a not too distant future :-)

Alberto, I started that long ago but stopped it for reasons. Maybe now is a good point to restart lensfun integration. Working together at lensfun integration would surely speedup the task :)

Ingo

agriggio:
:+1: yes, let's do that!

PR #3810, opened 2017-04-08
Added support for autodetecting the system monitor color profile also on Linux

agriggio:
Code borrowed from Geeqie.

Works fine here, but please test in other environments, particularly on Windows since I don't have that (to make sure I did not break anything there)

### DIFF
--- a/rtgui/editorpanel.cc
+++ b/rtgui/editorpanel.cc
@@ -105,7 +105,7 @@ bool find_default_monitor_profile(GdkWindow *rootwin, Glib::ustring &defprof, Gl
     // * Copyright (C) 2006 John Ellis
     // * Copyright (C) 2008 - 2016 The Geeqie Team
     // 
-    guchar *prof;
+    guchar *prof = NULL;
     gint proflen;
     GdkAtom type = GDK_NONE;
     gint format = 0;
--- a/rtgui/editorpanel.cc
+++ b/rtgui/editorpanel.cc
@@ -105,7 +105,7 @@ bool find_default_monitor_profile(GdkWindow *rootwin, Glib::ustring &defprof, Gl
     // * Copyright (C) 2006 John Ellis
     // * Copyright (C) 2008 - 2016 The Geeqie Team
     // 
-    guchar *prof = NULL;
+    guchar *prof = nullptr;
     gint proflen;
     GdkAtom type = GDK_NONE;
     gint format = 0;
--- a/rtgui/editorpanel.cc
+++ b/rtgui/editorpanel.cc
@@ -71,6 +71,67 @@ int setprogressStrUI ( void *p )
     return FALSE;
 }
 
+
+bool find_default_monitor_profile(GdkWindow *rootwin, Glib::ustring &defprof, Glib::ustring &defprofname)
+{
+#ifdef WIN32
+    HDC hDC = GetDC(nullptr);
+
+    if (hDC != nullptr) {
+        if (SetICMMode(hDC, ICM_ON)) {
+            char profileName[MAX_PATH + 1];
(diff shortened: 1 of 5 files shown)


Beep6581:
@agriggio I just wanted to make sure that the code is not from sourceforge but from http://geeqie.org/cgi-bin/gitweb.cgi?p=geeqie.git Sourceforge is obsolete but they can't remove the project for some reason, the git link above should be current. Their official website is http://geeqie.org/ (I updated it for them). I can confirm that monitor profile detection works in Geeqie-git in Linux - haven't tried this RT branch yet.

agriggio:
I've actually taken this from the latest release, not the git master. I'll
check if anything has changed and update the patch if needed. it works here
fwiw

agriggio:
@Beep6581 I confirm that the current git master uses the exact same code as the latest release for detecting the monitor color profile

gaaned92:
@agriggio no modification  for profile selection compared to dev branch on W10.
But I think there is a bug.
if  in the "option file, color management part" you have for instance
```
ICCDirectory=C:\\Windows\\System32\\spool\\drivers\\color
MonitorProfile=Rec2020
AutoMonitorProfile=true
```
instead of selecting the default system profile, the indicated profile is selected.

I observe that the selection at the bottom of editor panel is reset to the option file selection at RT start. It is logical but should be documented in rawpedia.

agriggio:
@gaaned92 thanks for testing. the change in behaviour is not intentional,
I'll see what is going on.

gaaned92:
@agriggio I  tested with dev branch and behaviour is the same, so it is certainly an old bug.

rtgui/editorpanel.cc
```diff
@@ -105,7 +105,7 @@ bool find_default_monitor_profile(GdkWindow *rootwin, Glib::ustring &defprof, Gl
     // * Copyright (C) 2006 John Ellis
     // * Copyright (C) 2008 - 2016 The Geeqie Team
     // 
-    guchar *prof;
+    guchar *prof = NULL;
```
Floessie: @agriggio Please use `nullptr` instead. I converted all places with `NULL` prior to the 5.0 release and it should stay consistent from now on.

agriggio:
> so it is certainly an old bug.

@gaaned92 can you open a separate issue then? Thanks!

rtgui/editorpanel.cc
agriggio: @Floessie sure :+1:

agriggio:
@Floessie I replaced `NULL` with `nullptr`. But FYI, I still see several uses of `NULL` in the code (unrelated to this patch). Should an issue be opened to fix them as well?

Floessie:
@agriggio Thanks for the fix. I'll care about the other places (and `0`) as soon as I'm back from my holydays.

agriggio:
any objections to merging this? According to @gaaned92 the behaviour on windows is unchanged, and on Linux it seems to work fine (tested on two different machines with two different desktop environments)

Beep6581:
Auto-detection works, but if I set the monitor default color profile to Foo, then enable "Use operating system's main monitor color profile" and open an image, the monitor profile selected by default under the preview is still Foo. Even if I change it to "System default (from GDK)" and restart RT, it still defaults to Foo. Since the Preferences options is called "monitor **default** color profile", "System default (from GDK)" should be on by default regardless of what is selected in the disabled combo.

agriggio:
@Beep6581 that's the same behaviour described by @gaaned92 I think. It seems this is a bug that is already in dev. I can try fixing that, but I propose we merge first. What do you think?

Beep6581:
@agriggio :+1:

Beep6581:
I opened a new issue #3821

PR #3817, opened 2017-04-10
Pentax Pixel Shift: Added per channel brightness equalization

heckflosse:
I added an improvement for some special cases of Pixel Shift shots where the light changed from frame to frame. That happens for example when shooting at sunset/sundown or when shooting while for example a cloud hides the sun for one shot, but not all.

Here's an example which shows the improvement which is already in dev:
Left is without equalization of frame brightness, right with the current global equalization of brightness
As you can see, there are still some vertical stripes in the orange part of the right.

ps_stripes_00

I added an option to equalize the brightness per colour-channel.
Left is with global (channel independent) equalization of frame brightness, right with the new per-channel equalization.
As you can see, the stripes are almost gone.

ps_stripes_01

If there are no objections, I would like to merge that into dev for 5.1

Ingo

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -720,6 +720,7 @@ HISTORY_MSG_471;PS Motion correction
 HISTORY_MSG_472;PS Smooth transitions
 HISTORY_MSG_473;PS Use lmmse
 HISTORY_MSG_474;PS Equalize
+HISTORY_MSG_475;PS Equalize channel
 HISTORY_NEWSNAPSHOT;Add
 HISTORY_NEWSNAPSHOT_TOOLTIP;Shortcut: <b>Alt-s</b>
 HISTORY_SNAPSHOT;Snapshot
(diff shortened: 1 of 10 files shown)


da-phil:
@heckflosse  I really appreciate your work you put in the pixelshift de-mosaicer! :+1: 
can't wait to try it out with some highres shots of my olympus E-M1mkII, I really like the idea that some shots which didn't work out in the in-camera JPG engine might still work out using RT without getting noticeable artifacts ;)

heckflosse:
@da-phil Pixel Shift is only for Pentax Pixel Shift raw files which have 4 frames with one pixel offset each.
Afaik the Olympus files already are combined in raw by the camera and don't contain multiple frames in one raw file.

da-phil:
ah, damnit olympus! I thought I could also benefit from this feature :(
in the "RAW spirit" it would have also made more sense to have access to all independent 8 frames in the olympus pixel shift system, because post-processing on a PC is always more powerful than in the camera.

Issue #3765, opened 2017-03-16
Incorrect zooming with lens correction profiles


agriggio:
It seems lens correction profiles interact badly with zooming. Using an image
with an active lens profile makes RT compute wrong zooms when the level is >=
33%: basically the image becomes huge (see screenshot).

* zoom at 25%
![zoom-25](https://cloud.githubusercontent.com/assets/11982086/24021791/fa296180-0aa2-11e7-8788-4f169793f8a9.png)

* zoom at 33%
![zoom-33](https://cloud.githubusercontent.com/assets/11982086/24021795/016669ca-0aa3-11e7-9b34-172769071418.png)

heckflosse:
@agriggio Yes, that's caused by auto-fill. There already is a very old issue about that somewhere

agriggio:
Ok. I tried to search but didn't see anything "obviously" related...

heckflosse:
@agriggio #3400 for example

agriggio:
@heckflosse ah, I was searching for open issues. this is so annoying I'd like to fix it. any suggestions on where to start?

heckflosse:
@agriggio Maybe [start here](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/iptransform.cc#L1067) ?

agriggio:
@heckflosse :+1: I'll see what I can do

agriggio:
@heckflosse this seems to work for me. Can you give it a try? (It seems that
LCP distortion correction should be applied to the full-size image to be
accurate)

```
diff --git a/rtengine/iptransform.cc b/rtengine/iptransform.cc
--- a/rtengine/iptransform.cc
+++ b/rtengine/iptransform.cc
@@ -948,7 +948,7 @@
     double hpteta  = fabs(hpalpha - rtengine::RT_PI / 2) < 3e-4 ? 0.0 : acos ((hpdeg > 0 ? 1.0 : -1.0) * sqrt((-oH * oH * tan(hpalpha) * tan(hpalpha) + (hpdeg > 0 ? 1.0 : -1.0) * oH * tan(hpalpha) * sqrt(16 * maxRadius * maxRadius + oH * oH * tan(hpalpha) * tan(hpalpha))) / (maxRadius * maxRadius * 8)));
     double hpcospt = (hpdeg >= 0 ? 1.0 : -1.0) * cos (hpteta), hptanpt = tan (hpteta);
 
[...]

PR #3318, opened 2016-05-30
Automatically disable Auto-Fill when enabling LCP Disortion Correction, #1791

Beep6581:
https://github.com/Beep6581/RawTherapee/issues/1791

Issue #1791, opened 2015-08-11
Preview image distorted when zoom to 33% or 50% using a lens correction profile in combination with 'Auto fill'


Beep6581:
Originally reported on Google Code with ID 1807
```
When using a lens correction profile, the preview image at 50% zoom is broken. With
some Canon profiles, it not zoom to 50%, but to maximum depth. Other profiles (like
the Pentax for the Amsterdam photo sample) produces a extremly distortion at 50%.

You can download a sample image with the lense profile at http://www.terminal.io/tmp/4623/example.zip

Steps to reproduce:

- Open the photo
- zoom to 50%
- load the lens correction profile

```

Reported by `cytrinox` on 2013-03-26 20:29:29

Beep6581:
Options:
1- Don't show the LCP's effect in preview at all,
2- Keep the buggy behavior as is,
3- Automatically disable Auto-fill from 33% upwards (either just from the preview or completely)
4- Automatically disable Auto-fill when enabling LCP
5- Fix the code so LCP works well with Auto-fill at all zoom levels.


1, 2 and 3 are unacceptable and from the discussion it seems 5 is not to be expected. This leaves us with 4. Any objections?

Beep6581:
Let's implement point 4 - Automatically disable Auto-fill when enabling LCP. The fact of Auto-fill being disabled should be reflected in history so that the user gets some feedback that this was done.

Starting point:
```
diff --git a/rtgui/lensprofile.cc b/rtgui/lensprofile.cc
index a7a6344..0362dc1 100644
--- a/rtgui/lensprofile.cc
+++ b/rtgui/lensprofile.cc
@@ -167,6 +167,7 @@ void LensProfilePanel::onUseDistChanged()
     useDistChanged = true;
 
     if (listener) {
+        listener->panelChanged (EvTransAutoFill, M("GENERAL_DISABLED"));
         listener->panelChanged (EvLCPUseDist, ckbUseDist->get_active() ? M("GENERAL_ENABLED") : M("GENERAL_DISABLED"));
     }
 }
```
It sends a message to the history panel, but doesn't actually disable Auto-fill yet.

Beep6581:
The commit above, in branch lcpautofill, disables Auto-Fill when loading an LCP file in non-batch mode and when enabling LCP Distortion Correction in non-batch mode. In both cases you get notified about this via the History panel.

[...]

PR #3822, opened 2017-04-12
LCP tweaks

agriggio:
A couple of tweaks to the UI for Lens Correction Profiles.
The most significant one is the removal of the workaround for #1791 (#3318), since now that bug has been fixed (see #3765)

### DIFF
--- a/rtengine/lcp.cc
+++ b/rtengine/lcp.cc
@@ -28,10 +28,17 @@
 #include <shlobj.h>
 #endif
 
+#include "settings.h"
 
 using namespace std;
 using namespace rtengine;
 
+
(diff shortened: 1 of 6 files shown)


Beep6581:
Builds and works fine.

There are some quirks with Auto-Crop when I tested using a fisheye LCP (strange crop position, couldn't resize along the right edge, left edge was missing) but these are probably unrelated to this patch.

agriggio:
@Beep6581 yes, that's probably unrelated. I'll merge then, thanks for testing

Beep6581:
Thanks for patching :)

PR #3880, opened 2017-05-18
Added support for using RawTherapee as a GIMP plugin for opening RAWs

agriggio:
plugin code taken from Darktable

quick instructions for the plugin:

```
$ cd tools/gimp-plugin
$ gimptool-2.0 --build file-rawtherapee.c
$ sudo cp file-rawtherapee /usr/lib/gimp/2.0/plug-ins
```

Assumes `rawtherapee` and `rawtherapee-cli` are on `$PATH`

Tested on linux only


gaaned92:
@agriggio 
Windows 10
The branch builds fine.
I am now struggling to generate the plugin.  That doesn't seem as simple as in your post above.  Surely I am missing something. But I progress.

I think we should compile once the plugin and provide this compiled plugin with the builds. Is it possible?

agriggio:
Hi @gaaned92,
sorry, as I wrote above I only tested on Linux. I have no idea how to build on windows unfortunately :-(

agriggio:
> I think we should compile once the plugin and provide this compiled plugin with the builds. Is it possible?

It should be possible, yes. The plugin has no build dependency on RT at all.

gaaned92:
@agriggio  Ah! no luck
I get those errors:

```
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpui-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpwidgets-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpmodule-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimp-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpmath-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpconfig-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpcolor-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgimpbase-2.0
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgegl-0.3
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lgegl-npd-0.3
C:/msys64/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/6.3.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot find -lbabl-0.1
```
I cannot find in my gimp installation something with those names. Have you an idea?

agriggio:
I think you need the equivalent of the `libgimp2.0-dev` package for windows... perhaps this can be useful

  https://wiki.gimp.org/wiki/Hacking:Building/Windows

gaaned92:
@agriggio 
1-  You wrote: the plugin has no build dependency on RT at all.
But perhaps, it can have dependencies with Gimp and thus work with a version and not with an other?

2- regarding the build error perhaps @heckflosse  or @Partha1b could help?

I have the Gimp-2.9.5-std built par Partha installed in C:/Gimp-2.9.5-std
I build under mingw64 shell:
```
#!/usr/bin/env bash
cd /y/rtsource/rawtherapee/tools/gimp-plugin
export set PATH=/c/GIMP-2.9.5-std/frameworks:/c/GIMP-2.9.5-std/gimp/bin:$PATH
echo $PATH
export set PKG_CONFIG_PATH=$PKG_CONGIG_PATH:'/c/GIMP-2.9.5-std/gimp/lib/pkgconfig'
gimptool-2.0 --build file-rawtherapee.c
cd /y/rtsource
```

In the Gimp  *.pc pkg files, I replaced the linux style 
`prefix=/opt/gimp/std`
by
`prefix=/c/GIMP-2.9.5-std/gimp`

but I suppose it is not the wright path as the linker complains.

Thanks for help.

Partha1b:
@gaaned92 In fact, it is "as easy as" building on Linux once the PKG_CONFIG_PATH is set correctly. This is handled by gimptool-2.0.exe on Windows. 

I don't have this branch. Where is the code for me to test?

gaaned92:
@Partha1b 
The github branch is gimp-plugin.
You will find here [https://drive.google.com/open?id=0B2q9OrgyDEfPS2FpdDAtMVI1RG8](https://drive.google.com/open?id=0B2q9OrgyDEfPS2FpdDAtMVI1RG8)
- the RT installer "RawTherapee_gimp-plugin_5.0-r1-gtk3-514-g54b66b77_WinVista_64.zip"
- the plugin source in "gimp-plugin.zip"

Apparently it is not so easy for me. Thank you in advance

Partha1b:
Sorry @gaaned92 . I was not meant as any dig at you. I meant that it was similar using Windows as in Linux.

It's possible that your shell is getting tripped (I've seen that happen with Msys2). 

So, do the following:
```gcc -o file-rawtherapee file-rawtherapee.c -I. `pkg-config --cflags --libs gimpui-2.0` ``` 

This assumes that you have set your PKG_CONFIG_PATH correctly. At a minimum, it should be like the following (this is my pkg path):
```
$ set |grep PKG_CONFIG_PATH
PKG_CONFIG_PATH=/opt/lib/pkgconfig:/usr/python/Lib/pkgconfig:/opt/share/pkgconfig:/opt/gimp/std/lib/pkgconfig:/opt/gimp/libmypaint/lib/pkgconfig
```
If it's not clear, please let me know.

File: rtgui/editorpanel.cc [hunk 1]
```diff
@@ -1885,6 +1893,30 @@ void EditorPanel::sendToGimpPressed ()
...
 
 
+bool EditorPanel::saveImmediately(const Glib::ustring &filename, SaveFormat sf)
+{
+    rtengine::procparams::ProcParams pparams;
+    ipc->getParams (&pparams);
+    std::unique_ptr<rtengine::ProcessingJob> job(rtengine::ProcessingJob::create (ipc->getInitialImage(), pparams));
+
+    // save immediately
+    rtengine::IImage16 *img = rtengine::processImage(job.get(), err, nullptr, options.tunnelMetaData, false);
```
Floessie, line 1907: @agriggio I like your use of a `std::unique_ptr` here. :+1: There's a handful of other places calling `rtengine::processImage()` where the `job` instance isn't freed after use. Maybe you could sanitize them, too?

rtgui/editorpanel.h
```diff
@@ -131,6 +131,8 @@ class EditorPanel final :
 
     void defaultMonitorProfileChanged(const Glib::ustring &profile_name, bool auto_monitor_profile);
 
+    bool saveImmediately(const Glib::ustring &filename, SaveFormat sf);
```
Floessie: `SaveFormat` is a rather heavy `struct`. Better pass it by `const&`.

rtgui/filepanel.cc
```diff
@@ -190,8 +190,13 @@ void FilePanel::init ()
     dirBrowser->fillDirTree ();
     placesBrowser->refreshPlacesList ();
 
-    if (argv1 != "" && Glib::file_test (argv1, Glib::FILE_TEST_IS_DIR)) {
-        dirBrowser->open (argv1);
+    //if (argv1 != "" && Glib::file_test (argv1, Glib::FILE_TEST_IS_DIR)) {
```
Floessie: Better nuke that line immediately. :wink:

rtgui/main.cc
```diff
@@ -292,12 +294,23 @@ int main(int argc, char **argv)
 
 #endif
 
-    simpleEditor = false;
+    //simpleEditor = false;
```
Floessie: Dito.

rtgui/main.cc
```diff
@@ -292,12 +294,23 @@ int main(int argc, char **argv)
...
 
-    if( !argv1.empty() )
+    if (!argv1.empty()) {
```
Floessie: :heart:

rtgui/rtwindow.cc
```diff
@@ -282,6 +282,14 @@ RTWindow::RTWindow ()
         show_all ();
 
         bpanel->init (this);
+
+        if (!argv1.empty()) {
+            Thumbnail* thm = cacheMgr->getEntry(argv1);
```
Floessie, line 288: Shouldn't this work, too?
```c++
            Thumbnail* const thm = cacheMgr->getEntry(argv1);
            if (thm) {
                fpanel->fileCatalog->openRequested({thm});
            }
```

tools/gimp-plugin/file-formats.h
```diff
@@ -0,0 +1,322 @@
+/* GIMP - The GNU Image Manipulation Program
+ * Copyright (C) 1995 Spencer Kimball and Peter Mattis
+ *
+ * file-rawtherapee.c -- raw file format plug-in that uses rawtherapee
+ * Copyright (C) 2016 Tobias Ellinghaus <me@houz.org>
```
Floessie, line 5: Didn't you edit this, too?

Partha1b:
@Floessie I am at a loss. What are your edits about? I don't see them related to the GIMP plugin which was the issue at hand?
@Beep6581 ?

Beep6581:
@Partha1b ?

Partha1b:
@Beep6581 I saw a whole lot of edits to rtgui above and was not sure it it belonged to this issue or not.

Beep6581:
@Partha1b you can see them in context when you click on the "Files changed" tab at the top.

Partha1b:
@Beep6581 Yes, I see. Thanks.

Partha1b:
@agriggio Suggest that you synchronize with the latest darktable code which includes trying to locate the app on a mac. 

Similar service does exist on Windows if you want. You can use "where" which is available on all Windows versions beyond  Windows XP 32bit and so safe to use. Also suggest that you use shellexecute on Windows which is a native API and supported by Microsoft.

agriggio:
@Partha1b sorry, I have no way of testing on windows. feel free to send me a patch (or commit on this branch directly), but I'm a bit reluctant to committing code that I can't test...

rtgui/editorpanel.h
agriggio: :+1: (although I wouldn't call that "heavy", but you are right anyway :-)

rtgui/filepanel.cc
agriggio: :+1:

rtgui/rtwindow.cc
agriggio: perhaps. But explicit is better than implicit in this case. The first version really says "if there is a file to open, let's try to do that". The second says "we expect that there is a file to open, and if we fail to load the thumbnail, we don't open anything"... I prefer version 1 :-)

tools/gimp-plugin/file-formats.h
agriggio: fixed, thanks

rtgui/rtwindow.cc
Floessie: My comment was about the lines below not above, sorry. It's more about saving the `std::vector<Thumbnail *> entries(1, thm);` not the control flow (which I hope is fine). :smile:

Beep6581:
@agriggio I merged dev into this branch, hope you don't mind.

I tried to test it, but it fails to compile:
```
gimptool-2.0 --build file-rawtherapee.c
x86_64-pc-linux-gnu-gcc  -pthread -I/usr/include/gimp-2.0 -I/usr/include/gegl-0.3 -I/usr/include/json-glib-1.0 -I/usr/include/gio-unix-2.0/ -I/usr/include/babl-0.1 -I/usr/include/gtk-2.0 -I/usr/lib64/gtk-2.0/include -I/usr/include/pango-1.0 -I/usr/include/atk-1.0 -I/usr/include/cairo -I/usr/include/pixman-1 -I/usr/include/libdrm -I/usr/include/gdk-pixbuf-2.0 -I/usr/include/libpng16 -I/usr/include/pango-1.0 -I/usr/include/harfbuzz -I/usr/include/pango-1.0 -I/usr/include/glib-2.0 -I/usr/lib64/glib-2.0/include -I/usr/include/freetype2 -o 'file-rawtherapee' file-rawtherapee.c  -lgimpui-2.0 -lgimpwidgets-2.0 -lgimpmodule-2.0 -lgimp-2.0 -lgimpmath-2.0 -lgimpconfig-2.0 -lgimpcolor-2.0 -lgimpbase-2.0 -lgegl-0.3 -lgegl-npd-0.3 -lm -Wl,--export-dynamic -lgmodule-2.0 -pthread -ljson-glib-1.0 -lgio-2.0 -lbabl-0.1 -lgtk-x11-2.0 -lgdk-x11-2.0 -lpangocairo-1.0 -latk-1.0 -lcairo -lgdk_pixbuf-2.0 -lgio-2.0 -lpangoft2-1.0 -lpango-1.0 -lgobject-2.0 -lglib-2.0 -lfontconfig -lfreetype 
/tmp/cc5zbOsI.o: In function `query':
file-rawtherapee.c:(.text+0x24d): undefined reference to `gimp_register_file_handler_raw'
collect2: error: ld returned 1 exit status
```

agriggio:
@beep6581 no problem. you need a recent git master of gimp to compile (I actually use a ppa called something like gimp-unstable). otherwise comment out that line

agriggio:
any objections to merging?

agriggio:
On my side, this is ready for merging. Objections?

heckflosse:
@agriggio Alberto, please wait 30 minutes. I will try to test now.

agriggio:
@heckflosse no hurry -- I didn't mean to merge *right now*. I can wait a few days without problems!

Beep6581:
I did the three steps from the first comment, added the RT path to $PATH, but now what?
Would be best to explain this in RawPedia.

agriggio:
@Beep6581 you're right, I should document this. Unfortunately, I'm a bit busy
at work these days, so for now I'm adding some comments here.

Now what you do is open Gimp (a recent 2.9.x version), go to `Edit -> Preferences -> Image import` and make sure that "Raw Importer" is set to the `file-rawtherapee` plug-in.

Then, just open a RAW file with Gimp: a RT editor window should open automatically, which you can use to tweak your RAW file. When you close the window, the image gets imported into Gimp.

If anything goes wrong, please let me know.

heckflosse:
@agriggio Alberto, I tested on Windows. It builds and works fine and does not break anything in rt. I didn't test the plugin because I don't have a recent GIMP 2.9 installed on windows. 
I tested starting rt without argument, with a folder as argument and with amsterdam.pef as argument.
All three worked fine. :+1: from me for merge.

Beep6581:
@agriggio I will help with the documentation, once I get it working. I don't have an "Edit -> Preferences -> Image import" option, and I'm using GIMP-2.9.4. Is it not new enough?
imgur-2017_06_19-19 39 53

Partha1b:
@Beep6581  and @gaaned92 No, it's not. You have use 2.9.5 from one of the new branches. Perhaps you can use my build. What you are looking for is this:

gimp-2 9-image-importer

Beep6581:
Ok, so don't wait for me to test, as it will likely take a few months before 2.9.5/6 is out and available on my platform (I stopped compiling GIMP from git).
+1 for merge.

Partha1b:
@agriggio & @Beep6581 One issue I would point out is that you are not going to get Windows or Mac users any benefit with this. In particular, unless RT is actually in the PATH environment variable, the gspawn command will fail and you won't get the app registered as a RAW plugin. There are really no PATHS for apps on a Mac.

At least for Windows, the recommended procedure is to use ShellExecuteEx. From the Microsoft recommendation [https://msdn.microsoft.com/en-us/library/windows/desktop/ee872121(v=vs.85).aspx](url)

> Finding an Application Executable
> 
> When the ShellExecuteEx function is called with the name of an executable file in its lpFile parameter, there are several places where the function looks for the file. We recommend registering your application in the App Paths registry subkey. Doing so avoids the need for applications to modify the system PATH environment variable.
> The file is sought in the following locations:
> The current working directory.
> The Windows directory only (no subdirectories are searched).
> The Windows\System32 directory.
> Directories listed in the PATH environment variable.
> Recommended: HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths
> 

For example, my GIMP builds are registered with App Paths. Also, we should have Windows RT build registered here as well if it is not.

agriggio:
@Beep6581 it might still work with 2.9.4: if you don't have other raw import plugins installed, gimp might just pick the RT one. at least you can try...

agriggio:
@Partha1b I'm happy to accept patches for win and Mac, as I can't test on them by myself...

Beep6581:
@agriggio I wrote the RawPedia article, but I don't know whether it's accurate as I can't test it yet (my GIMP is not new-enough): http://rawpedia.rawtherapee.com/GIMP_Plugin
Would be cool is GIMP included the plugin in their code, as those instructions require too much effort from a photographer/user.

agriggio:
@Beep6581 thanks for the documentation! Regarding putting RT in `$PATH`, though, there's no need to sudo-edit `/etc/environment`. You can just set `PATH` on your shell profile file. If you use `bash` (that's the default on most linux distributions), just add this to `~/.profile`:

```
export PATH=$PATH:$HOME/programs/rt
```

About distributing with GIMP: no problem for me, I just don't know how to do
it. I suppose we should release a new version of RT first?

Beep6581:
Hello

I don't have statistics, but I am "under the impression" that `/etc/environment` is more commonly available across distros than are `~/.profile` or `~/.bash_profile`. Another concern is that environment variables in `~/.*profile` might only be available to programs started from a terminal and not from a start menu or when RT is started from GIMP or otherwise. It would need to be tested to confirm or refute, but I can't test this myself as my GIMP is not new enough.

> I just don't know how to do it. I suppose we should release a new version of RT first?

Same here. I hope RT5.2 and GIMP-2.9.5 (or will it be called 2.9.6?) are both released soon, and it would be neat if they cooperated with minimal user effort.

Beep6581:
I asked GIMP to bundle the plugin and they're positive about that as long as we keep it up to date:
https://bugzilla.gnome.org/show_bug.cgi?id=784228

PR #3898, opened 2017-05-30
Cleaned pixelshift code and made a speedup for pixelshift without mot…

heckflosse:
…ion correction

About the speedup: PS combine of a K-1 file without motion detection was ~600 ms. Now it's ~230 ms


heckflosse:
Any objections to merge this pr?

Beep6581:
Compiles and runs fine, and no differences in output compared to 5.1.

PR #3912, opened 2017-06-05
Use system klt library - new PR

mattiaverga:
I've reworked @sharkcz PR #3573 by adding a cmake build option for choosing to use system's KLT library (default is to use the bundled KLT).
I'm opening this PR against `dev` branch as requested from @Beep6581 

@Hombre57 what did you mean with 

> add a line into the main cmakefiles.txt

What is `cmakefiles.txt`?


Hombre57:
@mattiaverga Since you've edited it, I guess you've found that by yourself :wink: ?

Please put all the `options` statement altogether around line 109-119 of `cmakelists.txt`.

mattiaverga:
@Hombre57 you wrote `cmakefiles.txt` instead of `cmakelists.txt` and that was confusing me 😉 

However, I've moved the option to the right line and renamed to be more like the other already used.

CMakeLists.txt
```diff
@@ -293,6 +294,9 @@ find_package(JPEG REQUIRED)
 find_package(PNG REQUIRED)
 find_package(TIFF REQUIRED)
 find_package(ZLIB REQUIRED)
+if (WITH_SYSTEM_KLT)
+  find_package (KLT REQUIRED)
+endif(WITH_SYSTEM_KLT)
```
Beep6581: Please don't repeat the condition in `else`s and `endif`s - use the same style (and capitalization) as in the rest of the file. The same goes for the other CMake files.

Beep6581:
Shall we merge?

mattiaverga:
I've done some local tests and it seems to be ok for me.

Beep6581:
Merged, thanks!

Floessie:
When sorting the files in `rtengine/CMakeLists.txt` I stumbled upon [this](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/CMakeLists.txt#L101). This doesn't seem right to me, as the KLT files are conditionally included later.

Beep6581:
@Floessie looks like an oversight. I tested removing them and RT compiled successfully when using `WITH_SYSTEM_KLT=OFF`.
Cleanup accidentally included in PR #5483

PR #3946, opened 2017-06-29
new feature: open images in an already-running instance of RT

agriggio:
This PR adds a new command-line flag, `-R`(standing for "remote"), which can be used to open images in an already-running instance of RawTherapee, instead of opening a new "simple editor" instance.
The first instance that is executed with the `-R` flag will act as a server for the successive client instances.
`-R` takes as optional argument the image to open. If not given, its effect is simply to raise the server window (or start a new one if none is running).

Currently, this is *not* implemented for Mac OS X. The reason is that there seems to be already some partial support for a "single-instance app" on the Mac (at least judging from some code in `rtwindow.cc`), and, not having a Mac on which to test, I didn't want to touch that part of the code.

Final note: tested on Linux only.

### DIFF
--- a/rtgui/main.cc
+++ b/rtgui/main.cc
@@ -481,8 +481,9 @@ int main(int argc, char **argv)
     // suppression of annoying error boxes
     SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX | SEM_NOOPENFILEERRORBOX);
 
-    if (argc > 1 || options.rtSettings.verbose) {
-        if (options.rtSettings.verbose || ( !Glib::file_test (fname_to_utf8 (argv[1]), Glib::FILE_TEST_EXISTS ) && !Glib::file_test (fname_to_utf8 (argv[1]), Glib::FILE_TEST_IS_DIR))) {
+    if(argc > 1) {
+        int ret = processLineParams( argc, argv);
+        if (options.rtSettings.verbose || (!remote && !Glib::file_test (argv1, Glib::FILE_TEST_EXISTS ) && !Glib::file_test (argv1, Glib::FILE_TEST_IS_DIR))) {
             bool stdoutRedirectedtoFile = (GetFileType(GetStdHandle(STD_OUTPUT_HANDLE)) == 0x0001);
             bool stderrRedirectedtoFile = (GetFileType(GetStdHandle(STD_ERROR_HANDLE)) == 0x0001);
 
@@ -536,18 +537,14 @@ int main(int argc, char **argv)
             }
         }
 
-        if(argc > 1) {
-            int ret = processLineParams( argc, argv);
-
-            if( ret <= 0 ) {
-                if(consoleOpened) {
-                    printf("Press any key to exit RawTherapee\n");
-                    FlushConsoleInputBuffer(GetStdHandle(STD_INPUT_HANDLE));
-                    getch();
-                }
-
-                return ret;
+        if( ret <= 0 ) {
+            if(consoleOpened) {
+                printf("Press any key to exit RawTherapee\n");
+                FlushConsoleInputBuffer(GetStdHandle(STD_INPUT_HANDLE));
+                getch();
             }
+
+            return ret;
         }
     }
 
--- a/rtgui/main.cc
+++ b/rtgui/main.cc
@@ -63,6 +63,7 @@ Glib::ustring argv1;
 Glib::ustring argv2;
 bool simpleEditor = false;
 bool gimpPlugin = false;
+bool remote = false;
 Glib::RefPtr<Gtk::CssProvider> cssForced;
 Glib::RefPtr<Gtk::CssProvider> cssRT;
 //Glib::Threads::Thread* mainThread;
@@ -89,8 +90,6 @@ Glib::ustring fname_to_utf8 (const char* fname)
 #endif
 }
 
-}
-
 // This recursive mutex will be used by gdk_threads_enter/leave instead of a simple mutex
 static Glib::Threads::RecMutex myGdkRecMutex;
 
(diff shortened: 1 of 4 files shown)


rtgui/main.cc
```diff
@@ -308,109 +580,51 @@ int main(int argc, char **argv)
...
-        try {
-            cssRT->load_from_path (filename);
-            Gtk::StyleContext::add_provider_for_screen(screen, cssRT, GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
-        } catch (Glib::Error &err) {
-            printf("Error: Can't load css file \"%s\"\nMessage: %s\n", filename.c_str(), err.what().c_str());
-        } catch (...) {
-            printf("Error: Can't load css file \"%s\"\n", filename.c_str());
+    int ret = 0;
+    if (remote) {
+        char *app_argv[2] = { nullptr, nullptr };
```
Floessie: This initialization makes no sense, as `app_argv[0]` is set immediately afterwards. Is there a reason for the `const_cast`? Suggestion:
```c++
const char* app_argv[2] = {argv0.c_str()};
```

rtgui/main.cc
```diff
@@ -308,109 +580,51 @@ int main(int argc, char **argv)
...
-                printf("Error: \"%s\"\n", err.what().c_str());
-            } catch (...) {
-                printf("Error: Can't find the font named \"%s\"\n", options.fontFamily.c_str());
+        RTApplication app;
+        ret = app.run(app_argc, app_argv);
+    } else {
+        if (init_rt()) {
+            Gtk::Main m(&argc, &argv);
+            gdk_threads_enter();
+            RTWindow *rtWindow = create_rt_window();
```
Floessie: Suggestion:
```c++
const std::unique_ptr<RTWindow> rtWindow(create_rt_window());
```

rtgui/main.cc
agriggio: https://developer.gnome.org/gtkmm/stable/classGtk_1_1Application.html#a41ede16109c06bbefce5194f4b601780

So, `char **`, not `const char **`

heckflosse:
@agriggio Alberto, I found a bug.
My rt is set to METM
I start rt with this command
`rawtherapee -R /h/2files/1.pef`
Rt starts and opens `1.pef` twice

samefiletwice

heckflosse:
@agriggio Alberto, in SETM it's also opened twice. It's just that you don't see it. But this small patch clearly shows it:

```patch
index 3c16163e..19393c85 100644
@@ -1519,7 +1519,7 @@ void RawImageSource::vflip (Imagefloat* image)

 int RawImageSource::load (const Glib::ustring &fname, int imageNum, bool batch)
 {
-
+std::cout << fname << std::endl;
     MyTime t1, t2;
     t1.set();
     fileName = fname;

```
Here's the console output (on Windows):
```
H:/2files/1.pef
H:\2files\1.pef
```
the first one has slashes while the second one has backslashes, strange...

heckflosse:
@agriggio
On Windows, using -R opens a console
Here's a patch which fixes that:
```patch
diff --git a/rtgui/main.cc b/rtgui/main.cc
index c9b5f2b7..4e4aaae1 100644
--- a/rtgui/main.cc
+++ b/rtgui/main.cc
@@ -481,8 +481,18 @@ int main(int argc, char **argv)
     // suppression of annoying error boxes
     SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX | SEM_NOOPENFILEERRORBOX);

+    bool remoteMode = false;
+    if(argc == 3) { // no console in remote mode
+        for(int i = 0; i < argc; i++) {
+            if(!strcmp(argv[i], "-R")) {
+                remoteMode = true;
+                break;
+            }
+        }
+    }
+
     if (argc > 1 || options.rtSettings.verbose) {
-        if (options.rtSettings.verbose || ( !Glib::file_test (fname_to_utf8 (argv[1]), Glib::FILE_TEST_EXISTS ) && !Glib::file_test (fname_to_utf8 (argv[1]), Glib::FILE_TEST_IS_DIR))) {
+        if (options.rtSettings.verbose || (!remoteMode && !Glib::file_test (fname_to_utf8 (argv[1]), Glib::FILE_TEST_EXISTS ) && !Glib::file_test (fname_to_utf8 (argv[1]), Glib::FILE_TEST_IS_DIR))) {
             bool stdoutRedirectedtoFile = (GetFileType(GetStdHandle(STD_OUTPUT_HANDLE)) == 0x0001);
             bool stderrRedirectedtoFile = (GetFileType(GetStdHandle(STD_ERROR_HANDLE)) == 0x0001);

```

agriggio:
@heckflosse Ingo, thanks for testing! I'm currently on vacation. I'll fix the issues next week when I'm back

heckflosse:
@agriggio Alberto, I will continue testing when the issues are fixed.

agriggio:
great, thanks!

heckflosse:
@agriggio Alberto, I will try to fix the 'file opening twice' bug, because it happens on Windows only

heckflosse:
@agriggio Alberto, may I push to your branch?

agriggio:
@heckflosse dire!

agriggio:
@heckflosse sorry, damn autocorrect. I meant: sure!

heckflosse:
@agriggio Alberto, I fixed the bugs I found. Works fine now on Windows. I tested SETM, METM and MEOW

Hombre57:
@agriggio @heckflosse Just tested now on Win7, it works great ! Ready to be committed for me.

Hombre57:
What's up with this PR @agriggio ? Is there something to complete or fix ? @heckflosse @Beep6581 Ok to commit ?

heckflosse:
@Hombre57 Look 9 posts above your post. Alberto is on vacation. We should wait until he is back!

agriggio:
Hi all, I'm back now. Ok for me to merge -- if there are no objections, I'll do it later today

PR #3947, opened 2017-07-02
Benchmark fixes

mattiaverga:
The tool benchmarkRT is not working.
I made some fixes:

- it now needs to use "rawtherapee-cli" instead of "rawtherapee" as exe name
- the URL from where sidecar files can be downloaded is no more working, changing to github
- the Neutral.pp3 sidecar is no more existent

However there's another problem: redirecting the output to /dev/null make rawtherapee-cli try to save to /dev/null.tif and this prevent the benchmark to work.
```
$ /usr/bin/rawtherapee-cli -o /dev/null -p "/tmp/rawtherapee-benchmark/Default.pp3" -t -c "/tmp/rawtherapee-benchmark/colorspace_flowers.pef" 
RawTherapee, version , command line
Processing: /tmp/rawtherapee-benchmark/colorspace_flowers.pef
  Merging procparams #0
Error saving to: /dev/null.tif
```
I think this needs to be addressed in the main program code.


Beep6581:
Commit 3fb5c7d39e886242788d70b86ed1668def4b050f makes this obsolete.

mattiaverga:
Please take a look at my commit 78082b0, if the computer locale is set to use comma as decimal separator, the line `t="$( printf "%s\n" "scale=3; $secOnly+$minOnly*60" | bc )"` will throw an error.
In my commit I've added `secOnly="$(tr , . <<<$secOnly)"` to make sure to use point as decimal separator for seconds.

Beep6581:
I wondered what that was for. Will look into it later.

PR #3958, opened 2017-07-07
Fix filtered horizontal thumbs

Floessie:
Take two images, the first in portrait, the second in landscape
orientation. Mark the second with a red dot and filter for red.
Now open that image in SETM. The top tumb is too slim.

### DIFF
--- a/rtgui/thumbbrowserbase.cc
+++ b/rtgui/thumbbrowserbase.cc
@@ -523,63 +523,49 @@ void ThumbBrowserBase::arrangeFiles ()
     // We could lock it one more time, there's no harm excepted (negligible) speed penalty
     //GThreadLock lock;
 
-    int N = fd.size ();
-
-    // apply filter
-    for (int i = 0; i < N; i++) {
-        fd[i]->filtered = !checkFilter (fd[i]);
-    }
+    const int N = fd.size();
 
     int rowHeight = 0;
 
-    // compute size of the items
-    for (int i = 0; i < N; i++)
+    for (int i = 0; i < N; i++) {
+        // apply filter
+        fd[i]->filtered = !checkFilter (fd[i]);
+
+        // compute size of the items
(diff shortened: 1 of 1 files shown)


heckflosse:
@Floessie May I push a change to `ThumbBrowserBase::arrangeFiles()` ?

heckflosse:
@Floessie

Here's a patch instead of a push ;-)

```
diff --git a/rtgui/thumbbrowserbase.cc b/rtgui/thumbbrowserbase.cc
index 5983f1c3..abb8ec68 100644
--- a/rtgui/thumbbrowserbase.cc
+++ b/rtgui/thumbbrowserbase.cc
@@ -575,6 +575,9 @@ void ThumbBrowserBase::arrangeFiles()
             if (!fd[i]->filtered && colsWidth + fd[i]->getMinimalWidth() <= availWidth) {
                 colsWidth += fd[numOfCols]->getMinimalWidth();
                 ++numOfCols;
+                if(colsWidth > availWidth) {
+                    break;
+                }
             }
         }
 
```

Floessie:
@heckflosse Yes, please push directly. 👍 

Was that a bug or is it an optimization?

heckflosse:
@Floessie It is an optimization which reduces another small O(n²) part to O(n)

Edit: Though it's still O(n²) when a filter is used which filters out a lot of files. But when no filter is used, breaking the loop when the max width of the row is reached should lead to O(n)

heckflosse:
@Floessie
I just pushed it

heckflosse:
@Floessie I must miss something. I tried to reproduce the bug, but could not...

Beep6581:
Without the patch, I get this:
imgur-2017_07_08-00 31 20

Beep6581:
And the patch fixes it.

heckflosse:
Bug and fix confirmed. :+1: from me for merge

Floessie:
@Beep6581 @heckflosse Fine. I'll merge.

Issue #3975, opened 2017-07-19
Always show verbose warning when failure writing options file


Beep6581:
When RT fails writing the options file, the user should always get a verbose popup warning, not only a printf in the console when running in verbose mode as is now.

Beep6581:
PR https://github.com/Beep6581/RawTherapee/pull/3976 shows a popup when RT fails to write the options file. It also shows a warning in the console:
> Gtk-Message: GtkDialog mapped without a transient parent. This is discouraged.

Ideally it should use the `writeFailed()` function, but I don't know how to get the parent. Ingo points out that the same warning is generated when emptying the trash.

I would like to commit this ASAP so that we can find and fix some issues on macOS, but it would be nice if someone knew how to get that parent parameter.

heckflosse:
go for merge from me

Beep6581:
Merged into dev, but could be done better - see https://github.com/Beep6581/RawTherapee/pull/3976#discussion_r128290755

PR #3976, opened 2017-07-19
Always show console warning and popup when saving options file fails, #3975

Beep6581:


rtgui/options.cc
```diff
@@ -2229,10 +2229,13 @@ int Options::saveToFile (Glib::ustring fname)
     FILE *f = g_fopen (fname.c_str (), "wt");
 
     if (f == nullptr) {
-        if (options.rtSettings.verbose) {
-            printf ("Options::saveToFile / Error: unable to open file \"%s\" with write access!\n", fname.c_str());
-        }
-
+        std::cout << "Warning! Unable to save your preferences to: " << fname << std::endl;
+#ifndef RAWTHERAPEE_CLI
+        Glib::ustring msg_ = Glib::ustring::compose(M("MAIN_MSG_WRITEFAILED"), fname.c_str());
```
agriggio, line 2234: I think it would be better to check the return code of `Options::saveToFile` and let the callee decide what to do (i.e. show a dialog message) instead of unconditionally show a dialog in here

agriggio: More specifically, we should change `Options::save` to return a `bool` status flag, and then check this in the 3 places where it is called (`rtgui/exportpanel.cc`, `rtgui/preferences.cc` and `rtgui/rtwindow.cc`). Then it's easy to retrieve a suitable parent for the dialog

heckflosse: @agriggio I agree, but I would prefer to do that for 5.3 To find some MacOS related bugs the quick and dirty solution should be good enough.

Beep6581: @agriggio good advice, I'll leave the issue open, but merging this for now to get feedback for macOS.

Beep6581: I tried doing as you suggested, but it turns out that getting the failed file path then is not possible, at least not without a bit of coding overkill which is beyond my skill. As such, I will close the issue.

agriggio:
Please don't close. We can easily implement it by having the function raise an exception instead of returning an error code. I'm fine with postponing this to 5.3 though.

Beep6581:
Ok, I'll re-open it. Looking forward to seeing how it's done :)

agriggio:
@Beep6581 my proposed solution in the new `options-save-raise-exception-on-failure` branch

heckflosse:
@agriggio Alberto, in a next step we could eliminate the code to save the options file from rth-cli target, because it does not make sense to save the options file from there. But then we would need the 
`-DRAWTHERAPEE_CLI` again

agriggio:
ok, but why would we need the define? rt-cli has its own main, it would be
enough to just not call `Options::save`, no?

heckflosse:
I thought about reducing size of the rth-cli executable

Edit: and about the possibility to use the define to join main.cc and main-cli.cc into one file to avoid doubled code

PR #3995, opened 2017-07-31
Update appstream metadata to current format

innir:
Hi,

I updated the appstream metadata the current format, following:

https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html

Best,
Philip


Beep6581:
Thank you.

Issue #3980, opened 2017-07-22
Including Data Type Information in Sidecar Files


lumpiluk:
Including information about the data type of any given key-value pair in PP3 files would be useful for interoperability with other programs as described in this thread: [https://discuss.pixls.us/t/parsing-pp3-sidecar-files/4690](https://discuss.pixls.us/t/parsing-pp3-sidecar-files/4690)

One way this could be done is to require a decimal point in floating point literals, to require quotes around strings, and to e.g. enclose double lists like so: `ClCurve=double_list(3;0;0;0.35;0.65;1;1;)`, as compared to the current format `ClCurve=3;0;0;0.35;0.65;1;1;` where the value of ClCurve might as well be a string.

P.S.: I'd be happy to try making these changes myself.

PR #3996, opened 2017-07-31
Load & save PP3 files with data type information (Issue #3980).

lumpiluk:
I changed the PP3 format to include information about the data type for each key/value pair as outlined in issue #3980.

Specifics from the new KeyFileWrapper class documentation:

Key files saved with the original Glib::KeyFile class do not allow foreign
applications to determine the data type of a given key-value pair.
For example, saving an integer with value 42 would yield the same result
as saving the string "42".
This wrapper class makes the following changes to the syntax of key files:

```
# Strings must be enclosed in quotation marks.
# (Inner quotation marks need not be escaped, for now.)
StringKey="StringValue"
# Regular integers are unchanged.
IntKey=42
# Booleans are unchanged.
BoolKey=true
# Double values must contain a decimal point. The decimal point need not be
# followed by another digit.
DoubleKey=42.
# 64-bit integers are marked as follows:
Int64Key=int64(42)
# Similarly, all remaining data types are clearly marked:
Uint64Key=uint64(42)
BoolListKey=bools(true;false;true;)
IntListKey=ints(4;2;42;)
DoubleListKey=doubles(4.0;2.0;42.0;)
StringListKey=strings("a";"b";)
```

### DIFF
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -29,6 +29,7 @@ Developement contributors, in last name alphabetical order:
   Jan Rinze
   Ben S.
   Andrey Skvortsov
+  Lukas Stratmann
   Fabio Suprani
   Anders Torger
   Ingo Weyrich
--- a/rtengine/CMakeLists.txt
+++ b/rtengine/CMakeLists.txt
@@ -80,6 +80,7 @@ set(RTENGINESOURCEFILES
     ipwavelet.cc
     jdatasrc.cc
     jpeg_ijg/jpeg_memsrc.cc
+    keyfile_wrapper.cc
     klt/convolve.cc
     klt/error.cc
     klt/klt.cc
--- a/rtengine/keyfile_wrapper.cc
+++ b/rtengine/keyfile_wrapper.cc
@@ -1,0 +1,199 @@
+#include "keyfile_wrapper.h"
+#include <string>
+#include <sstream>
+
+#include <iostream>
+
+namespace rtengine {
+    
+    const Glib::ustring KeyFileWrapper::INT64_PREFIX("int64");
+    
+    const Glib::ustring KeyFileWrapper::UINT64_PREFIX("uint64");
+    
+    const Glib::ustring KeyFileWrapper::BOOL_LIST_PREFIX("bools");
+    
+    const Glib::ustring KeyFileWrapper::INT_LIST_PREFIX("ints");
+    
+    const Glib::ustring KeyFileWrapper::DOUBLE_LIST_PREFIX("doubles");
+    
+    const Glib::ustring KeyFileWrapper::STRING_LIST_PREFIX("strings");
+    
+    KeyFileWrapper::KeyFileWrapper() {
+        // nothing to do here yet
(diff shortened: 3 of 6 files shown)


Beep6581:
@heckflosse @Hombre57 @Floessie @agriggio reminder that this needs review.

agriggio:
I'm neutral about this, but let me just point out that it will break the workflow of some people (i.e. those using a custom profile builder)

Beep6581:
I'm neither for nor against the change, but I am for avoiding breaking things if possible. If this PR breaks things for people using CPB, what would un-breaking it involve? Do you mean the CPB code itself would need to be updated, or that their "recipes" would need to be re-written? I've never used the CPB myself.

As those data types don't change, wouldn't describing them in RawPedia be enough?

agriggio:
The CPB is called to generate pp3 files. If the format changes, the CPB code needs to be updated to comply with the new format. So, people using a custom builder will have to tweak their code.

I personally don't think this is a deal breaker, using a CPB is an advanced feature so I expect people to be able to patch their code to comply with the new syntax, but I wanted to make sure we were all aware of this.

Hombre57:
I'll have a look tonight, me might find a compatibility solution by making it optional.

Floessie:
@Beep6581 
> As those data types don't change, wouldn't describing them in RawPedia be enough?

That's my impression, too.

I think, the proposed solution (to a problem, that's maybe not there, see above) is too invasive. If there really is a need for type annotations, then do it backward compatible. One thing that comes to my mind are comments in the INI. I don't know, if the parser we use supports them, but if so, one could add the type information as a specially decorated comment above or behind the key/value pair.

Another thing would be implicit type deduction in lapsestudio. So a test for `true` or `false` yields a `bool`, a number a `double`, a list of numbers a `std::vector<double>`, and everything else a `std::string`.

I think, we should first determine, if type annotations are needed at all.

Best,
Flössie

lumpiluk:
@Floessie 

> One thing that comes to my mind are comments in the INI. I don't know, if the parser we use supports them, but if so, one could add the type information as a specially decorated comment above or behind the key/value pair.

That's a good idea, I didn't think of that. The [parser](https://developer.gnome.org/glib/stable/glib-Key-value-file-parser.html) should support comments.

To be honest, in hindsight it doesn't seem like that big of a deal anymore to update LapseStudio whenever the PP3 format changes. The wall of code in LapseStudio for parsing these files still seems unnecessary, but the PP3 format doesn't seem to change so frequently that it would be a problem.

> Another thing would be implicit type deduction in lapsestudio. So a test for true or false yields a bool, a number a double, a list of numbers a std::vector<double>, and everything else a std::string.

I'm not sure this would work well for integers.

Floessie:
@lumpiluk That's good to hear. 👍 

Even if this PR is deemed unnecessary, we welcome you to contribute your ideas and code. Don't hesitate to start other PRs even if this one won't make it.

Best,
Flössie

Beep6581:
Shall we close this PR?

PR #4007, opened 2017-08-08
Update TooWaBlue themes

TooWaBoo:


Beep6581:
@TooWaBoo since the purpose of the new theme is not just to look nice, as is the case with the other themes, but to actually fulfill a technical need, maybe we should use a more descriptive name?

Maybe @Desmis and Elle have a better idea, here are some of mine:
* Middle Gray
* Average Surround

Desmis:
I think "average surround" is a good idea...

We should modify Rawpedia for Ciecam, to mention it, as a technical optimization of "viewing conditions"

jacques

TooWaBoo:
@Beep6581 

You can change the name to everything you want. It's up to you to decide.

Beep6581:
Theme renamed and RawPedia updated:
http://rawpedia.rawtherapee.com/Preferences#Theme

PR #4011, opened 2017-08-10
post-astyle cleanup

Beep6581:


Beep6581:
This cleanup mainly does two things:
- Broken table-type code blocks have been manually formatted, and I can confirm that re-running the files which contain those code blocks through astyle 3.0 does not destroy the manual table-like formatting, e.g. see https://github.com/Beep6581/RawTherapee/blob/astyle-fix/rtengine/procparams.cc#L73
- A long time ago some lines looked like this:
```c++
        if (!pedited || pedited->retinex.slope      )     keyFile.set_double    ("Retinex",    "Slope",      retinex.slope);
        if (!pedited || pedited->retinex.medianmap  )     keyFile.set_boolean   ("Retinex",    "Median",     retinex.medianmap);
```
After we started using astyle, they were changed to this:
```c++
        if (!pedited || pedited->retinex.slope      ) {
            keyFile.set_double    ("Retinex",    "Slope",      retinex.slope);
        }

        if (!pedited || pedited->retinex.medianmap  ) {
            keyFile.set_boolean   ("Retinex",    "Median",     retinex.medianmap);
        }
```
Note that the whitespace from the old style is still there. While these two example statements still line up to each other, they don't line up to the hundreds of other statements, which looks messy and does not improve readability. This cleanup removes that whitespace:
```c++
        if (!pedited || pedited->retinex.slope) {
            keyFile.set_double ("Retinex", "Slope", retinex.slope);
        }

        if (!pedited || pedited->retinex.medianmap) {
            keyFile.set_boolean ("Retinex", "Median", retinex.medianmap);
        }
```

It's not perfect, there are still overlooked gaps, but I think removing the whitespace relic from the past improves readability, and re-running astyle does not cause any issues.

@Floessie ok to merge?

Floessie:
@Beep6581 I like your changes very much. Though I must admit, I'm not fond of table-like formatting, so I'm biased. 😉 

👍 for merge.

Beep6581:
Thanks for reviewing @Floessie 

I like things which are maintainable, and table-like formatting of large sections is not maintainable - if someone adds a new line to the 'table' where some parameter is longer than any previous parameter in the table, it's up to that person to re-adjust the rest of the table. Experience shows this doesn't happen correctly. But in some cases, a table structure does improve readability, like with matrices. So I'm happy to fix the table structure where it helps readability, but tabulating the parameters of something like `ProcParams::save` would be a lot of work which would quickly lead to the same state my cat endeavours to lead to - chaos.

PR #4029, opened 2017-08-20
New middle gray preview background color and vertical togglebuttons

Beep6581:

PR #4030, opened 2017-08-21
Update TooWaBlue theme to v 2.54/2.55, vertical togglebuttons

TooWaBoo:
Reason: "New middle gray preview background color and vertical togglebuttons" #4029 
@Beep6581 
Merge this together with above PR, please.

### DIFF
--- a/rtdata/themes/TooWaBlue-GTK3-20_.css
+++ b/rtdata/themes/TooWaBlue-GTK3-20_.css
@@ -2,7 +2,7 @@
   This file is part of RawTherapee.
 
   Copyright (c) 2016-2017 TooWaBoo
-  Version 2.54 - requires RT 5.0 (Gtk+ >= 3.20)
+  Version 2.55 - requires RT 5.0 (Gtk+ >= 3.20)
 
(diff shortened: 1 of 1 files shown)


Beep6581:
@TooWaBoo good that you noticed a problem yourself, I was going to bring this up once a patch was decided on, but... what problem does this address? There were two that I noticed:
1. All of the crop mask and background color buttons take up too much space,
2. The "R" button specifically becomes much wider than the other G, B, L and F buttons when it directly follows the separator in the `bgcolor-h` branch, as you can see in this screenshot: https://user-images.githubusercontent.com/5844619/29497373-ea77030e-85e6-11e7-89aa-37f6192879d2.png

TooWaBoo:
I've fixed the "R" and "Focus" button. The G,B,L... buttons have the same width like in the current theme. I know, they are a bit big, but this is a compromize for higher dpi. They would become too small on those monitors, 'cause the images won't grow.

TooWaBoo:
I was wrong, the width of the G,B,L... buttons have changed, 'cause the images became bigger. I'll change it.

TooWaBoo:
Now it looks like this: 
grafik

PR #4043, opened 2017-08-25
Green equilibration speedup

heckflosse:


heckflosse:
Any objections to merge?

Issue #4044, opened 2017-08-26
Honor network address scheme in dynamic folder generation


johenning:
When editing files on a network volume, dynamic folder generation does not work. Instead of creating a converted folder under \\\192.168.1.2\\...\converted\ it creates a folder under C:\192.168.1.2\\...\converted\\.
![Screenshots for clarity](https://user-images.githubusercontent.com/3907965/29739962-d8b33c32-8a4b-11e7-9487-e1b655ac0e02.png)

System: Windows 10; i7-6600U; 16 GB RAM

Version: 5.2
Branch: release-5.2
Commit: 78e921cb
Commit date: 2017-07-23
Compiler: gcc 6.3.0
Processor: generic x86
System: Windows
Bit depth: 64 bits
Gtkmm: V3.22.0
Build type: release
Build flags: -mwin32 -m64 -mthreads -msse2 -std=c++11 -mtune=generic -Werror=unused-label -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -mwindows -Wno-aggressive-loop-optimizations -DNDEBUG -O3
Link flags: -m64 -mthreads -static-libgcc -mtune=generic -mwindows -s -O3
OpenMP support: ON
MMAP support: ON

PR #4046, opened 2017-08-26
Check for double backslash to enable output to network paths

johenning:
Fixes Issue #4044

### DIFF
--- a/rtgui/batchqueue.cc
+++ b/rtgui/batchqueue.cc
@@ -746,7 +746,7 @@ Glib::ustring BatchQueue::calcAutoFileNameBase (const Glib::ustring& origFileNam
     if (origFileName[0] == '/') {
         pa.push_back ("/" + da[0]);
 	} else if (origFileName[0] == '\\') {
-		if (origFileName[1] == '\\') {
+		if (origFileName.size() > 1 && origFileName[1] == '\\') {
 			pa.push_back ("\\\\" + da[0]);
 		}
 		else {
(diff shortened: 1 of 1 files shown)


Hombre57:
I don't manage to use a network path here ("\\127.0.0.1\Mes vidéos" on Win7), while it work when I type this path in a windows' explorer. How do you manage to use this kind of path ?

Beside that, could you update your identation please ? We uses spaces only, 4 spaces per tab.

johenning:
I'm not sure I follow. As far as I know, "\127.0.0.1\..." is not a network path. \ is equivalent to C:\ and thus refers to C:\127.0.0.1
In any case, I am only modifying the behavior if the path starts with a double backslash. Any path like yours is treated just like it was before.

I will update the indentation, didn't pay attention to that.

Hombre57:
@johenning Sorry, it was a typo, I meant "\\127.0.0.1\Mes vidéos" which is the kind of network path you're using. RT doesn't **browse** to that path at first, which I find strange since it works for you... So I can't test your patch for now.

johenning:
@Hombre57 If you type \\\\\ it shows up correctly ;)

Could you try "\\\127.0.0.1\Mes vidéos\\"? It doesn't seem to work without the trailing backslash. Also make sure the path contains at least one recognized file, as folders aren't displayed.

Hombre57:
@johenning I've tried and it works fine, minus an error message : "(rawtherapee.exe:5584): Gtk-CRITICAL **: gtk_tree_row_reference_new_proxy: assertion 'path->depth > 0' failed"

However that shouldn't prevent merging this PR.

PR #4051, opened 2017-09-01
Fix crash on empty command line arguments

dlichtenberger:
Command-line arguments can be empty, as in `rawtherapee-cli ''`.

Without this patch, RT crashes when parsing the command line above:

> terminate called after throwing an instance of 'std::out_of_range'                   
> what():  basic_string::at: __n (which is 0) >= this->size() (which is 0)           

The issue is very noticeable when scripting RawTherapee (at least with Bash). For example, if you have a call like `rawtherapee-cli $RT_PARAMS $FILENAME` in your script, an empty string is added to the argument list if $RT_PARAMS is empty.

### DIFF
--- a/rtgui/main-cli.cc
+++ b/rtgui/main-cli.cc
@@ -285,8 +285,7 @@ bool dontLoadCache ( int argc, char **argv )
 #if ECLIPSE_ARGS
         currParam = currParam.substr (1, currParam.length() - 2);
 #endif
-
-        if ( currParam.at (0) == '-' && currParam.at (1) == 'q' ) {
+        if ( currParam.length() > 1 && currParam.at(0) == '-' && currParam.at(1) == 'q' ) {
             return true;
         }
     }
@@ -317,6 +316,9 @@ int processLineParams ( int argc, char **argv )
 
     for ( int iArg = 1; iArg < argc; iArg++) {
         Glib::ustring currParam (argv[iArg]);
+        if ( currParam.empty() ) {
+            continue;
+        }
 #if ECLIPSE_ARGS
         currParam = currParam.substr (1, currParam.length() - 2);
 #endif
--- a/rtgui/main.cc
+++ b/rtgui/main.cc
@@ -122,6 +122,9 @@ int processLineParams ( int argc, char **argv )
 {
     for ( int iArg = 1; iArg < argc; iArg++) {
         Glib::ustring currParam (argv[iArg]);
+        if ( currParam.empty() ) {
+            continue;
+        }
 #if ECLIPSE_ARGS
         currParam = currParam.substr (1, currParam.length() - 2);
 #endif
@@ -697,4 +700,3 @@ int main (int argc, char **argv)
 
     return ret;
 }
-


heckflosse:
@dlichtenberger Thanks for reporting.
Bug confirmed and fix confirmed. Though it lacks the `Terminating without anything to do.` line you get when you start rawtherapee-cli without arguments. Nevertheless 👍  from me to merge.

Ingo

Beep6581:
I confirm a crash when running `./rawtherapee-cli ''`

Beep6581:
And I confirm that the branch fixes the crash, though without the message as @heckflosse pointed out.

Beep6581:
Wouldn't something like this be better?
```diff
diff --git a/rtgui/main-cli.cc b/rtgui/main-cli.cc
index 7908b913..07c836bd 100644
--- a/rtgui/main-cli.cc
+++ b/rtgui/main-cli.cc
@@ -100,6 +100,26 @@ int processLineParams ( int argc, char **argv );
 
 bool dontLoadCache ( int argc, char **argv );
 
+// Trim from start, in-place.
+static inline void ltrim(Glib::ustring &s) {
+    s.erase(s.begin(), std::find_if(s.begin(), s.end(), [](int ch) {
+        return !std::isspace(ch);
+    }));
+}
+
+// Trim from end, in-place.
+static inline void rtrim(Glib::ustring &s) {
+    s.erase(std::find_if(s.rbegin(), s.rend(), [](int ch) {
+        return !std::isspace(ch);
+    }).base(), s.end());
+}
+
+// Trim both ends, in-place.
+static inline void trim(Glib::ustring &s) {
+    ltrim(s);
+    rtrim(s);
+}
+
 int main (int argc, char **argv)
 {
     setlocale (LC_ALL, "");
@@ -243,11 +263,20 @@ int main (int argc, char **argv)
 #endif
 
     int ret = 0;
+    bool doTerminate = true;
 
     // printing RT's version in all case, particularly useful for the 'verbose' mode, but also for the batch processing
     std::cout << "RawTherapee, version " << RTVERSION << ", command line" << std::endl;
 
     if (argc > 1) {
+        argv1 = argv[1];
+        trim(argv1);
+        if (!argv1.empty()) {
+            doTerminate = false;
+        }
+    }
+
+    if (!doTerminate) {
         ret = processLineParams (argc, argv);
     } else {
         std::cout << "Terminating without anything to do." << std::endl;
@@ -286,7 +315,7 @@ bool dontLoadCache ( int argc, char **argv )
         currParam = currParam.substr (1, currParam.length() - 2);
 #endif
 
-        if ( currParam.at (0) == '-' && currParam.at (1) == 'q' ) {
+        if ( currParam.length() > 1 && currParam.at (0) == '-' && currParam.at (1) == 'q' ) {
             return true;
         }
     }
@@ -317,6 +346,9 @@ int processLineParams ( int argc, char **argv )
 
     for ( int iArg = 1; iArg < argc; iArg++) {
         Glib::ustring currParam (argv[iArg]);
+        if ( currParam.empty() ) {
+            continue;
+        }
 #if ECLIPSE_ARGS
         currParam = currParam.substr (1, currParam.length() - 2);
 #endif
diff --git a/rtgui/main.cc b/rtgui/main.cc
index f133c1e0..b94a9493 100644
--- a/rtgui/main.cc
+++ b/rtgui/main.cc
@@ -122,6 +122,9 @@ int processLineParams ( int argc, char **argv )
 {
     for ( int iArg = 1; iArg < argc; iArg++) {
         Glib::ustring currParam (argv[iArg]);
+        if ( currParam.empty() ) {
+            continue;
+        }
 #if ECLIPSE_ARGS
         currParam = currParam.substr (1, currParam.length() - 2);
 #endif
```

Then we see the "Terminating with nothing to do" message even for `./rawtherapee-cli '   '`

dlichtenberger:
I guess you could still use two empty arguments and not get the "Terminating..." message, but as far as I'm concerned that's not a realistic issue.

PR #4079, opened 2017-09-13
Show clipped shadow indicator only when all channels are clipped

### LINKED ISSUE #1926: The "clipshadow indicator" seems not to work as expected
Originally reported on Google Code with ID 1942

```
Hi,
I put all relevant files in http://filebin.net/9cqi31lyuw
When I load the raw file and hit ">" some parts of the picture are highlighted even
if I do not see why, 
They do not look that dark (the timber frame for example).
May be I do not know how to set up that feature?
Best regards
```

Reported by `dgaudry@condorcet93.fr` on 2013-08-06 15:50:05

midluk:
fixes #1926

### DIFF
--- a/rtgui/cropwindow.cc
+++ b/rtgui/cropwindow.cc
@@ -1623,7 +1623,7 @@ void CropWindow::expose (Cairo::RefPtr<Cairo::Context> cr)
                     const int shThreshold = options.shadowThreshold;
                     const float ShawdowFac = 64.f / (options.shadowThreshold + 1);
                     const float HighlightFac = 64.f / (256 - options.highlightThreshold);
-                    const bool showclippedAny = (!showR && !showG && !showB && !showL); // will show clipping if any of RGB chanels is clipped
+                    const bool showclippedAny = (!showR && !showG && !showB && !showL); // will show clipping if any (all) of RGB chanels is (shadow) clipped
 
 #ifdef _OPENMP
                     #pragma omp parallel for schedule(dynamic,16)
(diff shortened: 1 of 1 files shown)


rtgui/cropwindow.cc
```diff
@@ -1623,7 +1623,7 @@ void CropWindow::expose (Cairo::RefPtr<Cairo::Context> cr)
                     const int shThreshold = options.shadowThreshold;
                     const float ShawdowFac = 64.f / (options.shadowThreshold + 1);
                     const float HighlightFac = 64.f / (256 - options.highlightThreshold);
-                    const bool showclippedAny = (!showR && !showG && !showB && !showL); // will show clipping if any of RGB chanels is clipped
+                    const bool showclippedAny = (!showR && !showG && !showB && !showL); // will show clipping if any (all) of RGB chanels is (shadow) clipped
```
Beep6581, line 1626: It shouldn't be called `showclippedAny` anymore, but `showClippedAll`, and "any (all)" in the comment doesn't really make sense, more like
`// will show clipping if all RGB channels are below the shadow threshold`

midluk: showclippedAny is also used for highlight clipping, where it still does what was in the original comment

Beep6581: So `showClipped` with the comment `if all RGB channels are below the shadow threshold or either channel is above the highlights threshold`

midluk: `showClippedNormal` or what the bool value really reflects: `displayAllColors`

Issue #4100, opened 2017-09-23
Dev sets most batch options to ADD


Beep6581:
I compiled into an empty build folder and deleted the cache and options file. The Batch Processing options in Preferences are mostly all set to ADD, except for:
- sharpening radius
- sharpening damping
- sharpening iterations
- resize scale

They should all be SET.

I cancelled and examined the options file:
```
AdjusterBehavior=1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;
```

I re-ran RT, went to Preferences again and hit OK,
```
AdjusterBehavior=1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;0;0;0;0;0;0;
```

Six 0s but only 4 show up as "SET". Why?

This is a recurring issue: #2710, #2220, etc.

Beep6581:
This is also a quite serious problem as far as a usability is concerned since a clean installation starts "broken", so a v5.3 stopper.

Beep6581:
Another issue: the color of the radiobutton in the highlighted (selected) row is not visible as long as its selected. Theme needs fixing.
![imgur-2017_09_23-15 58 22](https://user-images.githubusercontent.com/5844619/30773784-1613161e-a078-11e7-85ea-fc6a5b2b687e.png)

Beep6581:
6 new 0s:
1. sharpening radius
2. sharpening damping
3. sharpening iterations
4. ?
5. ?
6. resize scale

Beep6581:
```diff
[...]

PR #4101, opened 2017-09-23
RT-Theme: radio and check button fix

TooWaBoo:
Issue: https://github.com/Beep6581/RawTherapee/issues/4100


heckflosse:
@TooWaBoo Thanks a lot! Works fine here.

PR #4107, opened 2017-09-26
Copy the Lensfun database into the app bundle

Benitoite:
This can be refined to not specify which version db folder (i.e. when we use something other than `version_1`) probably by someone.


tools/osx/macosx_bundle.sh
```diff
@@ -154,6 +154,10 @@ ditto {"${GTK_PREFIX}","${RESOURCES}"}/share/icons/Adwaita/index.theme
 #     cp -RL "${GTK_PREFIX}/etc/fonts" "${ETC}"
 # fi
 
+# Copy Lensfun Database Into App Bundle
+mkdir ~/repo-rt/build/RawTherapee.app/Contents/Resources/share/lensfun
```
agriggio, line 158: Maybe you want to use `${RESOURCES}` instead of a hard-coded path?

Benitoite: Well that makes sense.  I just used these commands because they are assumed values that hold true when compiling following the instructions.  I just don't know enough about the code to make it more universal if necessary.

Beep6581:
```diff
diff --git a/tools/osx/macosx_bundle.sh b/tools/osx/macosx_bundle.sh
index b216b313..250bc23e 100644
--- a/tools/osx/macosx_bundle.sh
+++ b/tools/osx/macosx_bundle.sh
@@ -154,9 +154,9 @@ ditto {"${GTK_PREFIX}","${RESOURCES}"}/share/icons/Adwaita/index.theme
 #     cp -RL "${GTK_PREFIX}/etc/fonts" "${ETC}"
 # fi
 
-# Copy Lensfun Database Into App Bundle
-mkdir ~/repo-rt/build/RawTherapee.app/Contents/Resources/share/lensfun
-cp /opt/local/share/lensfun/version_1/* ~/repo-rt/build/RawTherapee.app/Contents/Resources/share/lensfun/.
+# Copy the Lensfun database into the app bundle
+mkdir -p "${RESOURCES}/share/lensfun"
+cp /opt/local/share/lensfun/version_1/* "${RESOURCES}/share/lensfun"
 
 # Install names
 find -E "${CONTENTS}" -type f -regex '.*/(rawtherapee-cli|rawtherapee|.*\.(dylib|so))' | while read -r x; do
```

PR #4131, opened 2017-10-04
Fix missing filename after closing save-as dialog

PkmX:
Currently, if you open the save as dialog, press ESC to close the dialog and then re-open it, the filename is cleared to `"."`.

This patch sets `SaveAsDlg::fname` so RT always gets the original filename back even if the dialog is dismissed.


Floessie:
Problem and fix confirmed.

Beep6581:
Thanks! Something that has irritated me for a long time :]

PR #4158, opened 2017-10-25
Custom file chooser button

agriggio:
any objections to merge this?

### DIFF
--- a/rtdata/themes/RawTherapee-GTK3-20_.css
+++ b/rtdata/themes/RawTherapee-GTK3-20_.css
@@ -903,3 +903,13 @@ paned > separator {
 #PartialPasteHeaderSep {
     background-color: #D8D8D8;
 }
+
+/* All MyFileChooserButtons */
+button#MyFileChooserButton {
+    padding: 2px;
+    margin: 2px;
+}
+
+#ToolPanelNotebook button {
+	margin: 0px;
+}
--- a/rtdata/themes/RawTherapee-GTK3-_19.css
+++ b/rtdata/themes/RawTherapee-GTK3-_19.css
@@ -500,3 +500,9 @@ GtkNotebook {
 #PartialPasteHeaderSep {
     color: #D8D8D8;
 }
+
+
+#MyFileChooserButton {
+    padding-left: 3px;
+    padding-right: 3px;
+}
--- a/rtgui/darkframe.cc
+++ b/rtgui/darkframe.cc
@@ -46,7 +46,7 @@ DarkFrame::DarkFrame () : FoldableToolPanel(this, "darkframe", M("TP_DARKFRAME_L
     pack_start( *dfInfo, Gtk::PACK_SHRINK, 0);
 
     dfautoconn = dfAuto->signal_toggled().connect ( sigc::mem_fun(*this, &DarkFrame::dfAutoChanged), true);
-    dfFile = darkFrameFile->signal_file_set().connect ( sigc::mem_fun(*this, &DarkFrame::darkFrameChanged), true);
+    dfFile = darkFrameFile->signal_file_set().connect ( sigc::mem_fun(*this, &DarkFrame::darkFrameChanged)); //, true);
     btnReset->signal_clicked().connect( sigc::mem_fun(*this, &DarkFrame::darkFrameReset), true );
 
     // Set filename filters
--- a/rtgui/flatfield.cc
+++ b/rtgui/flatfield.cc
@@ -78,7 +78,7 @@ FlatField::FlatField () : FoldableToolPanel(this, "flatfield", M("TP_FLATFIELD_L
     pack_start( *flatFieldBlurRadius, Gtk::PACK_SHRINK, 0);
     pack_start( *flatFieldClipControl, Gtk::PACK_SHRINK, 0);
 
-    flatFieldFileconn = flatFieldFile->signal_file_set().connect ( sigc::mem_fun(*this, &FlatField::flatFieldFileChanged), true);
+    flatFieldFileconn = flatFieldFile->signal_file_set().connect ( sigc::mem_fun(*this, &FlatField::flatFieldFileChanged)); //, true);
     flatFieldFileReset->signal_clicked().connect( sigc::mem_fun(*this, &FlatField::flatFieldFile_Reset), true );
     flatFieldAutoSelectconn = flatFieldAutoSelect->signal_toggled().connect ( sigc::mem_fun(*this, &FlatField::flatFieldAutoSelectChanged), true);
     flatFieldBlurTypeconn = flatFieldBlurType->signal_changed().connect( sigc::mem_fun(*this, &FlatField::flatFieldBlurTypeChanged) );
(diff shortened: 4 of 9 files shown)


Floessie:
🤣 You two made my day!

agriggio:
:-) I think there was some miscommunication.

Here's a longer summary of what is happening:

1. @Beep6581 asked for a border for the new file chooser buttons, to better match the old look (and also because they look nicer indeed)

2. I first tried to do this with CSS, but failed. So, I added a hardcoded border to the code, but I also set a custom name for the widget, with the hope that this would be enough to override the default border in a CSS (for example in a TooWaBoo theme)

3. Apparently, this however would not work, as @TooWaBoo pointed out

4. So, my request can be expanded as follows: I have no problem removing the hardcoded border from the code, but someone must edit the CSS files of the themes shipped by RT to make sure that the border is there. *I would not know how to do it*, unfortunately...

5. So, it is true that currently there is nothing the in default CSS theme, but that's exactly the problem :-)


I hope it is clearer now...

TooWaBoo:
Argh, I've found an issue in the Preferences panel.
unbenannt
This can be fixed with css like this:

```
#PrefNotebook #MyFileChooserButton {
    margin: 2px;
}
```
... but only if you remove the objekt IDs from the subitems of the FileChooserButton. :-)
If necessary subitems can be referenced by eg. "#MyFileChooserButton image" and "#MyFileChooserButton label"

```
lbl_.set_name("MyFileChooserButton");
box_.set_name("MyFileChooserButton");
img->set_name("MyFileChooserButton");
```

The "hard coded" borders of items are not set by the item itself as a global setting, they are set individual for each item. But this isn't done consistent. Every stack in the Preferences panel has its own style. :-)

Beep6581:
I grepped through `rtgui/*` in this branch and confirm that not hard-coding padding is the good solution.
Patch for `RawTherapee-GTK3-20_.css` follows later after beef steak, potatos and coleslaw :)

TooWaBoo:
To only set the space for the 2. stack, the css rule looks like this:

```
#PrefNotebook stack:nth-child(2) #MyFileChooserButton {
    margin: 2px 6px;
}
```

Beep6581:
Something like so:
```diff
diff --git a/rtdata/themes/RawTherapee-GTK3-20_.css b/rtdata/themes/RawTherapee-GTK3-20_.css
index 94886d6f..e2b552cf 100644
--- a/rtdata/themes/RawTherapee-GTK3-20_.css
+++ b/rtdata/themes/RawTherapee-GTK3-20_.css
@@ -903,3 +903,9 @@ paned > separator {
 #PartialPasteHeaderSep {
     background-color: #D8D8D8;
 }
+
+/* All MyFileChooserButtons */
+button#MyFileChooserButton {
+    padding: 2px;
+    margin: 2px;
+}
diff --git a/rtgui/guiutils.cc b/rtgui/guiutils.cc
index 91e223a5..b1df584d 100644
--- a/rtgui/guiutils.cc
+++ b/rtgui/guiutils.cc
@@ -1177,16 +1177,15 @@ MyFileChooserButton::MyFileChooserButton(const Glib::ustring &title, Gtk::FileCh
     lbl_.set_ellipsize(Pango::ELLIPSIZE_MIDDLE);
     lbl_.set_justify(Gtk::JUSTIFY_LEFT);
     set_none();
-    box_.pack_start(lbl_, true, true, 3);
+    box_.pack_start(lbl_, true, true);
     Gtk::Image *img = Gtk::manage(new Gtk::Image());
     img->set_from_icon_name("document-open", Gtk::ICON_SIZE_BUTTON);
     box_.pack_start(*Gtk::manage(new Gtk::VSeparator()), false, false, 5);
-    box_.pack_start(*img, false, false, 3);
+    box_.pack_start(*img, false, false);
     box_.show_all_children();
     add(box_);
     signal_clicked().connect(sigc::mem_fun(*this, &MyFileChooserButton::show_chooser));
 
-    set_border_width(2);
     set_name("MyFileChooserButton");
     lbl_.set_name("MyFileChooserButton");
     box_.set_name("MyFileChooserButton");
```

There is still an issue with "Reset" button visible in the the Flat-Field/Dark-Frame panels. It's fixable for example with this:
```css
#ToolPanelNotebook button {
	margin: 0px;
}
```

TooWaBoo:
Please remove:
```
lbl_.set_name("MyFileChooserButton");
box_.set_name("MyFileChooserButton");
img->set_name("MyFileChooserButton");
```
...this makes no sense and makes the ID #MyFileChooserButton for css useless.

TooWaBoo:
Instead of:
```
#ToolPanelNotebook button {
	margin: 0px;
}
```

use:
```
#ToolPanelNotebook #MyFileChooserButton + button.image-button {
    margin: what ever you want.
}
```
This affects only the reset button

agriggio:
@TooWaBoo ok, wll do. fwiw, I copied the pattern from somewhere else in RT (don't remember where)

TooWaBoo:
At the moment, all things become a bit confused. :-)
I want to sort it:

1. @agriggio removes following lines from his code.
```
set_border_width(2);
...
lbl_.set_name("MyFileChooserButton");
box_.set_name("MyFileChooserButton");
img->set_name("MyFileChooserButton");
```
2. To fix the space for the myfileschooserbutton on stack 2 of the preferences dialog
32009432-6791d494-b9af-11e7-8bb1-2f7d0c8a98dd
add this to the RT theme:
```
#PrefNotebook > stack > :nth-child(2) #MyFileChooserButton {
    margin: 3px 6px;
}
```
... and everything is fixed. This is my suggestion.

unbenannt2

Beep6581:
> To fix the space for the myfileschooserbutton on stack 2 of the preferences dialog
add this to the RT theme:
#PrefNotebook > stack > :nth-child(2) #MyFileChooserButton {
    margin: 3px 6px;
}

Sorry, that is precisely what one should avoid! There is no reason to use a CSS selector which targets a specific instance of a widget to fix this issue. It just makes maintenance a nightmare.

TooWaBoo:
@Beep6581 
I agree with you but there should not be any style in the **guiutils.cc**. A clean solution is to clean the whole preferences panel (not only the FileChooserButton) which means removing all spaces and set it as a css style like it's done everywhere else. I think for that we need @Hombre57 .

Beep6581:
Agreed, but it's not just Preferences, it's the whole RT. `RawTherapee-GTK3-20_.css` is as it is because it went through many GTK+ versions and many changes. Now that things are stable, I think that CSS file should be rewritten from scratch, with the selectors being as broad as possible, and deviating from the default as little as necessary.

I am capable of doing the required rtgui work, both C++ and CSS, but I don't have time for that yet (it's been on my mind for over a year, but I had 3 concurrent jobs at one point - not much time for RT or anything), and having a new website is higher priority.

Hombre57:
@TooWaBoo Maybe you can answer better than me, but I think some spacing can't be set through CSS, like those set by `set_column_spacing` and `set_row_spacing` from `Gtk::Grid` ? That's the reason why I had to reintroduce some hardcoded style, for Gtk2 at least.

TooWaBoo:
@Hombre57 
Column spacing is not possible. In Gtk <3.20 padding and margin also didn't work for children.
In Gtk>=3.20 padding and margin can be used for children.

TooWaBoo:
I think it's ready to merge.

agriggio:
if someone wants to go ahead, please do. I am away until tomorrow night

Issue #4155, opened 2017-10-21
-Wextra


heckflosse:
I created a new branch [extraclean](https://github.com/Beep6581/RawTherapee/compare/extraclean) with the intention to get rt clean when built using `-Wextra`.
First commit 9e812bb cleans FTblockDN.cc for `-Wextra`

To check a compilation unit you can simply add 
`#pragma GCC diagnostic warning "-Wextra"`
after the last `#include` line.

gaaned92:
Using the -Wextra flag generates huge amount of warnings. Thus you will have a hard work to clean all the code.
If I can help don't hesitate to ask.

heckflosse:
@gaaned92 

> Using the -Wextra flag generates huge amount of warnings.

That's the reason I use 

`#pragma GCC diagnostic warning "-Wextra"`

after the last `#include` line per compilation unit.

Here's the list of rtengine compilation units:

- [x] CA_correct_RT.cc
- [x] EdgePreservingDecomposition.cc
- [x] FTblockDN.cc
- [x] PF_correct_RT.cc 54a07be
- [x] amaze_demosaic_RT.cc
- [x] cJSON.c 87c109a 7b3c36c
- [x] calc_distort.cc
- [x] camconst.cc
- [x] cfa_linedn_RT.cc
- [x] ciecam02.cc 15108af
- [x] cieimage.cc
- [x] clutstore.cc
- [x] color.cc a750585
- [x] colortemp.cc 7c9d428
- [x] coord.cc
- [x] cplx_wavelet_dec.cc
- [x] curves.cc 123c8b4
- [x] dcp.cc
- [ ] dcraw.cc (heckflosse: won't fix)
- [x] dcrop.cc
- [x] demosaic_algos.cc c8ac936
- [x] dfmanager.cc
- [x] diagonalcurves.cc
- [x] dirpyr_equalizer.cc 31c7586
- [x] dynamicprofile.cc
- [x] expo_before_b.cc
- [x] fast_demo.cc beb6347
- [x] ffmanager.cc
- [x] flatcurves.cc
- [x] gauss.cc
- [x] green_equil_RT.cc
- [x] hilite_recon.cc
- [x] iccjpeg.cc
- [x] iccstore.cc
- [x] icons.cc 7542cfc
- [x] iimage.cc
- [ ] image16.cc
- [ ] image8.cc
- [x] imagedata.cc 80b23e7
- [x] imagedimensions.cc
- [ ] imagefloat.cc
[...]

PR #4176, opened 2017-11-11
get rt cleaner when built using  -Wextra #4155

heckflosse:


heckflosse:
Branch stays open for further cleanup

PR #4177, opened 2017-11-11
Speedup for loading of Phase One files

heckflosse:


Floessie:
@heckflosse Ingo, I get a (new?) warning with Clang 5 on current `dev`:

```
/home/user/src/rawtherapee/rtengine/dcraw.h:319:70: warning: binding reference member 'ifp' to stack allocated parameter 'i' [-Wdangling-field]
   ph1_bithuff_t(DCraw *p,IMFILE *i,short &o):parent(p),order(o),ifp(i),bitbuf(0),vbits(0){}
                                                                     ^
/home/user/src/rawtherapee/rtengine/dcraw.h:354:13: note: reference member declared here
   IMFILE *&ifp;
            ^
```

Is that a problem?

Best,
Flössie

heckflosse:
@Floessie In the current use case `ph1_bithuff_t` is also stack allocated (in the same scope), so it's not a problem. But I'm open for improvements.

Floessie:
@heckflosse But you pass in a `IMFILE* i` and bind a reference to that pointer, i.e. the pointer is the temp value you bind a reference to. That makes no sense.

Some lines above is ` getbithuff_t(DCraw *p,IMFILE *&i, unsigned &z)`, which makes sense, if you want to reset the pointer from within that class. Could it be, you simply forgot the `&`?

heckflosse:
@Floessie That's confusing me as well (though it works fine). Let's start from scratch.

I need:

One instance of IMFILE per thread and one instance of ph1_bithuff_t per thread.
The per thread IMFILE instance has to be a copy of the global IMFILE instance.
The ph1_bithuff_t instance needs a reference to the IMFILE instance and the IMFILE instance is also needed outside the ph1_bithuff_t instance.
Additionally for all except the IMFILE instance of the master thread the attribute plistener has to be a nullptr.

Floessie:
@heckflosse 
> The ph1_bithuff_t instance needs a reference to the IMFILE instance and the IMFILE instance is also needed outside the ph1_bithuff_t instance.

But even if you pass an `IMFILE*` into `ph1_bithuff_t` that's a reference (here in the form of a pointer, to put it more precisely), and I don't see a place in `ph1_bithuff_t` where the value of `ifp` is changed.

To me, it all boils down to two possibilities:
1. Pass in `IMFILE*&` and store `IMFILE*&`. This way the value of `ifp` can be changed from inside `ph1_bithuff_t`. `ifp` de facto is a (partly compile time checked) `IMFILE**`.
2. Pass in `IMFILE*` and store `IMFILE*`. Now you can't change the value of the outside `ifp` from within `ph1_bithuff_t`. It would be even clearer, if the member was `IMFILE* const ifp;`.

Don't know, if this is helpful for your rules above or not. I just have the feeling, that passing in an `IMFILE*` and storing an `IMFILE*&` is wrong.

Best,
Flössie

heckflosse:
@Floessie Thank you! I pushed a fix :)

Floessie:
@heckflosse I'm more than glad, that you even made it `const`. 😉 🎉

PR #4181, opened 2017-11-13
Enable detection of non-Apple clang Clang features

Benitoite:
Ups macOS cmake requirement to 3.3 and sets policy 25, to differentiate compiler features between AppleClang and plain ole' clang.


Beep6581:
CMP0025: https://cmake.org/cmake/help/v3.3/policy/CMP0025.html

PR #4192, opened 2017-11-20
Procparams cleanup

Floessie:
Dear fellow RawTherapists,

When Alberto copied the lines for loading and saving the (Fattal) parameters the nth time, I decided that enough is enough. Copying code once might be okay, but having variations of the same ten or so lines over and over again is against my understanding of structured code. 😁 

So I took the challenge to tidy `procparams.*` by first introducing functions and templates for loading and saving the individual settings. The conversions were mostly done by regex.

Next was my take on the header: As all those parameters are about data and without accessors, I decided to turn them into `struct`s with constructors. That way I could get rid of the halfhearted `setDefaults()`. I scraped all those defaults up from `ProcParams::setDefaults()` and was surprised to find some differences in the already established constructors. `ProcParams::setDefaults()` took precedence over the old values. I hope, that fits in every place.

Writing about hope: There were some decisions to make and code outside `procparams.*` to convert, so chances are high that I broke something. I know @Beep6581 is good at testing settings and hope he'll find all those places where I was off the mark, maybe you all can join him. Apart from that I hope you like it, even though it will make merging on different branches more laborious once again.

The work isn't quite finished yet: I still plan to add `operator ==()` to every `struct`. But that's something, that won't affect other files anymore.

Best,
Flössie


agriggio:
yay!

Beep6581:
Great!

Beep6581:
@Desmis please pause changes to your branches until we can commit this and merge dev into your branches.

heckflosse:
@Floessie Maybe a good opportunity to fix this cppcheck warning too:

```
procparams.h line 1322:
The class 'PartialProfile' has 'operator=' but lack of 'copy constructor'.
```

heckflosse:
@Floessie Damn gui off cppcheck didn't let me copy these as text, but a png should also do the job:

procparams

Desmis:
@Beep6581 

OK :)

Floessie:
@heckflosse Great, I'll have a look at those cppcheck warnings. 👍

Floessie:
@heckflosse The first warning was easy:

```patch
diff --git a/rtengine/procparams.h b/rtengine/procparams.h
index 44f0268c..8e6bd48a 100644
--- a/rtengine/procparams.h
+++ b/rtengine/procparams.h
@@ -28,6 +28,7 @@
 
 #include "coord.h"
 #include "LUT.h"
+#include "noncopyable.h"
 
 class ParamsEdited;
 
@@ -1319,17 +1320,12 @@ private:
   * and hence is not responsible of their destructions. The function that instanciate
   * PartialProfile object has to handle all this itself.
   */
-class PartialProfile
+class PartialProfile :
+    public NonCopyable
 {
 public:
     rtengine::procparams::ProcParams* pparams;
     ParamsEdited* pedited;
-    PartialProfile& operator =(const PartialProfile& rhs)
-    {
-        pparams = rhs.pparams;
-        pedited = rhs.pedited;
-        return *this;
-    };
 
     PartialProfile      (bool createInstance = false, bool paramsEditedValue = false);
     PartialProfile      (ProcParams* pp, ParamsEdited* pe = nullptr, bool fullCopy = false);
diff --git a/rtgui/filebrowser.cc b/rtgui/filebrowser.cc
index cda73728..b308911a 100644
--- a/rtgui/filebrowser.cc
+++ b/rtgui/filebrowser.cc
@@ -1021,7 +1021,7 @@ void FileBrowser::pasteProfile ()
 
         for (unsigned int i = 0; i < mselected.size(); i++) {
             // copying read only clipboard PartialProfile to a temporary one
-            rtengine::procparams::PartialProfile cbPartProf = clipboard.getPartialProfile();
+            const rtengine::procparams::PartialProfile& cbPartProf = clipboard.getPartialProfile();
             rtengine::procparams::PartialProfile pastedPartProf(cbPartProf.pparams, cbPartProf.pedited, true);
 
             // applying the PartialProfile to the thumb's ProcParams
@@ -1069,7 +1069,7 @@ void FileBrowser::partPasteProfile ()
             for (unsigned int i = 0; i < mselected.size(); i++) {
                 // copying read only clipboard PartialProfile to a temporary one, initialized to the thumb's ProcParams
                 mselected[i]->thumbnail->createProcParamsForUpdate(false, false); // this can execute customprofilebuilder to generate param file
-                rtengine::procparams::PartialProfile cbPartProf = clipboard.getPartialProfile();
+                const rtengine::procparams::PartialProfile& cbPartProf = clipboard.getPartialProfile();
                 rtengine::procparams::PartialProfile pastedPartProf(&mselected[i]->thumbnail->getProcParams (), nullptr);
 
                 // pushing the selected values of the clipboard PartialProfile to the temporary PartialProfile
```

The others from the screenshot seem false positives to me, but I'll try to silence them with the latest cppcheck.

Beep6581:
0fc6d2f9
When launching RT I get:
imgur-2017_11_21-11 13 56
and a bunch of these in the console:
`-->Key file contains key “FlatFieldClipControl” in group “RAW” which has a value that cannot be interpreted.`

I built in a clean `build` folder. The last RT I ran was probably d47e7f67.

I can't find the string "Key file contains key" anywhere in the code. I'm guessing the key file is the `options` file: https://filebin.net/kbz98ofws8ve1xrw/options

`/home/morgan/.config/RawTherapee/profiles/Sony A7II/Sony A7II.pp3` exists.
My `profiles` folder is a symlink:
`/home/morgan/.config/RawTherapee/profiles -> /home/morgan/RTstuff/profiles/`

Floessie:
@Beep6581 Confirmed. Looking...

Floessie:
@Beep6581 Here's a patch:

```patch
index 150dbbb7..6695c557 100644
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -1703,7 +1703,7 @@ RAWParams::RAWParams() :
     ff_BlurRadius(32),
     ff_BlurType(getFlatFieldBlurTypeString(FlatFieldBlurType::AREA)),
     ff_AutoClipControl(false),
-    ff_clipControl(0),
+    ff_clipControl(false),
     ca_autocorrect(false),
     cared(0.0),
     cablue(0.0),
@@ -3668,7 +3668,9 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
             // For compatibility to elder pp3 versions
             assignFromKeyfile(keyFile, "RAW", "HotDeadPixels", pedited, raw.hotPixelFilter, pedited->raw.hotPixelFilter);
             raw.deadPixelFilter = raw.hotPixelFilter;
-            pedited->raw.deadPixelFilter = pedited->raw.hotPixelFilter;
+            if (pedited) {
+                pedited->raw.deadPixelFilter = pedited->raw.hotPixelFilter;
+            }
             assignFromKeyfile(keyFile, "RAW", "HotPixelFilter", pedited, raw.hotPixelFilter, pedited->raw.hotPixelFilter);
             assignFromKeyfile(keyFile, "RAW", "DeadPixelFilter", pedited, raw.deadPixelFilter, pedited->raw.deadPixelFilter);
             assignFromKeyfile(keyFile, "RAW", "HotDeadPixelThresh", pedited, raw.hotdeadpix_thresh, pedited->raw.hotdeadpix_thresh);
diff --git a/rtengine/procparams.h b/rtengine/procparams.h
index 44f0268c..80b2784e 100644
--- a/rtengine/procparams.h
+++ b/rtengine/procparams.h
@@ -1185,7 +1185,7 @@ struct RAWParams {
     int ff_BlurRadius;
     Glib::ustring ff_BlurType;
     bool ff_AutoClipControl;
-    int ff_clipControl;
+    bool ff_clipControl;
 
     bool ca_autocorrect;
     double cared;
diff --git a/rtgui/flatfield.cc b/rtgui/flatfield.cc
index 2b57aa47..9facac74 100644
--- a/rtgui/flatfield.cc
+++ b/rtgui/flatfield.cc
@@ -288,7 +288,9 @@ void FlatField::setAdjusterBehavior (bool clipctrladd)
 
 void FlatField::trimValues (rtengine::procparams::ProcParams* pp)
 {
-    flatFieldClipControl->trimValue(pp->raw.ff_clipControl);
+    int value = pp->raw.ff_clipControl;
+    flatFieldClipControl->trimValue(value);
+    pp->raw.ff_clipControl = value;
 }
 
 void FlatField::setDefaults(const rtengine::procparams::ProcParams* defParams, const ParamsEdited* pedited)
```

But there's one thing, I don't understand: The parameter `ff_clipControl` was originally [saved as a boolean](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/procparams.cc#L3445) but internally treated as an integer. It belongs to the `flatFieldClipControl` adjuster, that goes from 0 to 100. Is the way we used to save it wrong, or the UI?

Beep6581:
Patch applied and the default raw profile now gets applied correctly.

Btw. when I open RT I get one of these for every image in the startup folder which uses Fattal: `ToneMapFattal02 took 9 ms`
I get these timings every time I click on the Editor tab or on the File Browser tab. Does that mean that RT recomputes Fattal every time? Shouldn't it just show the cached thumb without recomputing anything?

Beep6581:
Ticked means tested, it has no bearing on the result of the test.

# Editor:
- [x] Exposure
- [x] Shadows/Highlights
- [x] Tone Mapping
- [x] HDR Tone Mapping
- [x] Retinex
- [x] Vignette Filter
- [x] Graduated Filter
- [x] L\*a\*b Adjustments
- [x] CIECAM02
- [x] Sharpening
- [x] Edges
- [x] Microcontrast
- [x] Impulse** Noise Reduction
- [x] Noise Reduction
- [x] Defringe
- [x] Contrast by Detail Levels
- [x] White Balance
- [x] Vibrance
- [x] Channel Mixer
- [x] Black-and-White
- [x] HSV Equalizer
- [x] Film Simulation
- [x] RGB Curves
- [x] Color Toning
- [x] Color Management
- [x] Wavelets
- [x] Crop
- [x] Resize
- [x]  Lens / Geometry
- [x] Sensor with Bayer Matrix
- [x] Sensor with X-Trans Matrix
- [x] Raw White Points
- [x] Preprocessing
- [x] Dark-Frame
- [x] Flat-Field

# Batch:
When testing in batch mode, select two or more images and make sure changes propagate to all selected images obeying your Add/Set settings in Preferences > Batch Processing.
- [x] Exposure
- [x] Shadows/Highlights
- [x] Tone Mapping
- [x] HDR Tone Mapping
- [x] Retinex
- [x] Vignette Filter
- [x] Graduated Filter
- [x] L\*a\*b Adjustments
- [x] CIECAM02
- [ ] Sharpening
- [ ] Edges
- [ ] Microcontrast
- [ ] Impulse** Noise Reduction
- [ ] Noise Reduction
- [ ] Defringe
- [ ] Contrast by Detail Levels
- [ ] White Balance
- [ ] Vibrance
- [ ] Channel Mixer
- [ ] Black-and-White
- [ ] HSV Equalizer
- [ ] Film Simulation
- [ ] RGB Curves
- [ ] Color Toning
- [ ] Color Management
- [ ] Wavelets
- [ ] Crop
- [ ] Resize
- [ ]  Lens / Geometry
- [ ] Sensor with Bayer Matrix
- [ ] Sensor with X-Trans Matrix
- [ ] Raw White Points
- [ ] Preprocessing
- [ ] Dark-Frame
- [ ] Flat-Field

heckflosse:
@Beep6581 

> Shouldn't it just show the cached thumb without recomputing anything?

The thumb is cached without all the tools. rt has to recompute everything for the thumbs when you open a folder. But it should not recompute when switching between editor and file browser if you didn't make changes imho

Floessie:
@heckflosse About those false positives: As soon as there is an `if (pedited)` cppcheck will warn when calling `assignFromKeyfile()`, disregarding the fact, that we pass a reference that is checked within that function. I tried to work around it by adding special `assignFromKeyfile()` for enums etc. but cannot exclude `if (pedited)` everywhere. Maybe you have an idea on how to keep `assignFromKeyfile()` a one-liner and passing in `pedited` in a different way.

The "member not initialized in constructor" warnings are absolutely dumb. Don't know how to work around these.

heckflosse:
@Floessie 

> The parameter ff_clipControl was originally saved as a boolean but internally treated as an integer. It belongs to the flatFieldClipControl adjuster, that goes from 0 to 100. Is the way we used to save it wrong, or the UI?

That's clearly a copy/paste bug from me :( 
ff_clipControl has to be an int and saved as int

heckflosse:
@Floessie 

> As soon as there is an if (pedited) cppcheck will warn when calling assignFromKeyfile(), disregarding the fact, that we pass a reference that is checked within that function.

But wouldn't for example `pedited->defringe.radius` in
```
assignFromKeyfile(keyFile, "Defringing", "Radius", pedited, defringe.radius, pedited->defringe.radius);
```

dereference a null pointer in case pedited == nullptr?

Edit: I just saw that assignFromKeyfile() is a template...

Floessie:
@heckflosse 
```c++
template<typename T>
bool assignFromKeyfile(
    const Glib::KeyFile& keyfile,
    const Glib::ustring& group_name,
    const Glib::ustring& key,
    bool has_params_edited, // <-- protection variable
    T& value,
    bool& params_edited_value // <-- passed in as a reference
)
{
    if (keyfile.has_key(group_name, key)) {
        getFromKeyfile(keyfile, group_name, key, value);

        if (has_params_edited) { // <-- protection variable
            params_edited_value = true; // <-- dereferenced here
        }

        return true;
    }
    return false;
}
```

Floessie:
@Beep6581 @heckflosse 
> That's clearly a copy/paste bug from me :(
ff_clipControl has to be an int and saved as int

Okay, so I only fixed the SEGV. Older PP3s will yield the `which has a value that cannot be interpreted` warning until they are overwritten.

@Beep6581 I don't see that popup about the missing default profile. Do I have to remove the `.config/RawTherapee5-dev` directory?

heckflosse:
@Floessie 

> Older PP3s will yield the which has a value that cannot be interpreted warning until they are overwritten.

I will try to fix that.

heckflosse:
@Floessie Here's a patch

```patch
index e2c16a35..59a6330c 100644
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -3661,7 +3661,15 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
             assignFromKeyfile(keyFile, "RAW", "FlatFieldBlurRadius", pedited, raw.ff_BlurRadius, pedited->raw.ff_BlurRadius);
             assignFromKeyfile(keyFile, "RAW", "FlatFieldBlurType", pedited, raw.ff_BlurType, pedited->raw.ff_BlurType);
             assignFromKeyfile(keyFile, "RAW", "FlatFieldAutoClipControl", pedited, raw.ff_AutoClipControl, pedited->raw.ff_AutoClipControl);
-            assignFromKeyfile(keyFile, "RAW", "FlatFieldClipControl", pedited, raw.ff_clipControl, pedited->raw.ff_clipControl);
+            if (ppVersion < 328) {
+                // with ppversion < 328 this value was stored as a boolean, which is nonsense
+                // To avoid annoying warnings we read this value as boolean and then convert to int
+                bool temp = false;
+                assignFromKeyfile(keyFile, "RAW", "FlatFieldClipControl", pedited, temp, pedited->raw.ff_clipControl);
+                raw.ff_clipControl = temp;
+            } else {
+                assignFromKeyfile(keyFile, "RAW", "FlatFieldClipControl", pedited, raw.ff_clipControl, pedited->raw.ff_clipControl);
+            }
             assignFromKeyfile(keyFile, "RAW", "CA", pedited, raw.ca_autocorrect, pedited->raw.ca_autocorrect);
             assignFromKeyfile(keyFile, "RAW", "CARed", pedited, raw.cared, pedited->raw.cared);
             assignFromKeyfile(keyFile, "RAW", "CABlue", pedited, raw.cablue, pedited->raw.cablue);
diff --git a/rtgui/ppversion.h b/rtgui/ppversion.h
index d2e9be09..d3eca590 100644
--- a/rtgui/ppversion.h
+++ b/rtgui/ppversion.h
@@ -2,11 +2,13 @@
 #define _PPVERSION_
 
 // This number has to be incremented whenever the PP3 file format is modified or the behaviour of a tool changes
-#define PPVERSION 327
+#define PPVERSION 328
 #define PPVERSION_AEXP 301 //value of PPVERSION when auto exposure algorithm was modified
 
 /*
   Log of version changes
+   328  2017-11-21
+        Fix wrong type of ff_clipControl
    327  2017-09-15
         [Profiles Lens Correction] Added Lensfun
    326  2015-07-26

```

Floessie:
@heckflosse Ingo, thanks for the patch. But honestly, what's the difference between 0 and 1 in a range from 0 to 100? Wouldn't it be better to assume 0 and skip the reading completely?

agriggio:
@Floessie, how is this going? I'd like to perform a couple of tasks that require to manipulate the procparams, but I obviously don't want to go before this is merged...

Floessie:
@agriggio I'm finished, as far as I can see. I'm not at home ATM, so I won't be able to fix things until next week. Maybe someone can use that time to check for regressions...

Best,
Flössie

agriggio:
I think this should not stay open for long. The more we wait for merging, the harder it will be to work on other branches. So, I'm for merging right now and deal with the bugs (if any) when the pop up.

Hombre57:
Very nice work @Floessie ! Could you tell us what to check in particular (i.e. what could possibly be wrong for you), or do we have to check as much things as we can ?

Floessie:
@Hombre57 Thanks for the flowers! 😃 

> Could you tell us what to check in particular (i.e. what could possibly be wrong for you), or do we have to check as much things as we can ?

Well, there were a few type mismatches (one was a real cracker and @heckflosse fixed it), so maybe take a look at the first or second commit (I can't remember anymore and have a slow connection here) where I adjusted some types (`colorrank` and some others). Those are the things that could be critical. Other than that, take some PP3s developed with the latest `dev` and save them with this branch again. Then you can `diff` them. That's maybe the fastest and most automated approach...

Thanks for your testing,
Flössie

Beep6581:
I tested all tools, though not in all possible modes. Processing profile fill mode set to "fill". I opened `amsterdam.pef` in dev and in this branch ("ppc"), then either tool by tool or tab by tab (including Exif and IPTC) I tested like this: in dev I enable the tool/all tools in a tab, change all parameters from their default values, saved a PP3. In ppc I open that PP3, changed some parameter then changed it back, e.g. if slider foo was 19 I'd change it to 18 then back to 19, then save the new PP3, then diff.

I found the following parameter differences between a PP3 saved using 4f9d272c (dev) and 95614c7e (procparams-cleanup):
- BeforeCurveMode
- Gammafree
- ThresholdHighlight
- SaturatedOpacity (from the color toning tool, dev `1` ppc `0`)
- FlatFieldClipControl (`false` to `0` is not a bug, but it always saves `0` when the slider is set to non-zero which is a bug)

I can't guarantee that those are bugs, the differences might be invalid for some reason or other, but they should be examined.

+1 for merging. It will get more testing in dev.

Floessie:
@Beep6581 Thanks for your review. I'll take a look ASAP. I'll be back from my trip by tomorrow evening and might take the time then to look for the differences. Otherwise on Tuesday.

Best,
Flössie

Desmis:
I did various tests and I did not find any dysfunctions

Jacques

Floessie:
@Beep6581 

### Investigations

> BeforeCurveMode

No idea why there are differences, the code looks sane and I'll have to dig deeper.

> Gammafree

There's an oddity already in the old code: When saving it uses `pedited->icm.gamma` while when loading it uses `pedited->icm.gamfree`. The latter seems to be the right one.

> ThresholdHighlight

This is only saved in `dev` already. Easy to fix.

> SaturatedOpacity

No idea, code was regex-converted and no type change.

> FlatFieldClipControl

@heckflosse and I already fixed that long-standing bug here. We adjust the value to `0`, if `ppVersion < 328`. Could that just be the case in your test?

I'll fix those ASAP and prepare the branch for merge.

Best,
Flössie

Beep6581:
@Floessie I cleared the PP3 and assigned neutral when testing `FlatFieldClipControl`. Does your procparams branch use <328?

Floessie:
@Beep6581 No, `procparams-cleanup` has v328. I just tested it, and it loads and saves `FlatFieldClipControl` just fine. It's `dev`which has v327, so going back and forth will reset `FlatFieldClipControl` to `0`.

Just to prevent misunderstandings: I took `procparams-cleanup`, applied `Neutral`, changed the said field to `42` in the GUI, quit, checked the PP3, opened it in RT again, and the GUI still showed `42`.

Desmis:
@Floessie 
I could summarize all this work by "bravo", very good job of simplification and optimization
Jacques

Floessie:
@Desmis Thanks for your kind words, Jacques. I only did some *travail appliqué*, but I'm glad you like it and we discovered some real and old bugs on the way (the more you test, the more it breaks). 😄 

Cordialement,
Flössie

Desmis:
@Floessie 

Merci, mais c'est du très bon "travail appliqué" :)
Cordialement
Jacques

Floessie:
@Beep6581 

> BeforeCurveMode

Found it:
```patch
index 6629589a..65028340 100644
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -174,7 +174,7 @@ bool assignFromKeyfile(
     const Glib::ustring& group_name,
     const Glib::ustring& key,
     bool has_params_edited,
-    const std::map<const char*, T>& mapping,
+    const std::map<std::string, T>& mapping,
     T& value,
     bool& params_edited_value
 )
@@ -183,7 +183,7 @@ bool assignFromKeyfile(
         Glib::ustring v;
         getFromKeyfile(keyfile, group_name, key, v);
 
-        const typename std::map<const char*, T>::const_iterator m = mapping.find(v.c_str());
+        const typename std::map<std::string, T>::const_iterator m = mapping.find(v);
 
         if (m != mapping.end()) {
             value = m->second;
@@ -3442,7 +3442,7 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
                 toneCurve.shcompr = 100; // older pp3 files can have values above 100.
             }
 
-            const std::map<const char*, ToneCurveParams::TcMode> tc_mapping = {
+            const std::map<std::string, ToneCurveParams::TcMode> tc_mapping = {
                 {"Standard", ToneCurveParams::TcMode::STD},
                 {"FilmLike", ToneCurveParams::TcMode::FILMLIKE},
                 {"SatAndValueBlending", ToneCurveParams::TcMode::SATANDVALBLENDING},
@@ -3774,7 +3774,7 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
             assignFromKeyfile(keyFile, "Color appearance", "Datacie", pedited, colorappearance.datacie, pedited->colorappearance.datacie);
             assignFromKeyfile(keyFile, "Color appearance", "Tonecie", pedited, colorappearance.tonecie, pedited->colorappearance.tonecie);
 
-            const std::map<const char*, ColorAppearanceParams::TcMode> tc_mapping = {
+            const std::map<std::string, ColorAppearanceParams::TcMode> tc_mapping = {
                 {"Lightness", ColorAppearanceParams::TcMode::LIGHT},
                 {"Brightness", ColorAppearanceParams::TcMode::BRIGHT}
             };
```
Notes to myself: 
1. You never learned C, so don't use it!
2. Don't try to write [faster code](https://en.wikipedia.org/wiki/Program_optimization#When_to_optimize) than Ingo!

Floessie:
@Beep6581 

> SaturatedOpacity (from the color toning tool, dev 1 ppc 0)

Works as desired here with `procparams-cleanup` (while `ppc` might be shorter, it's normally short for  the platform I'm usually working with, i.e. *PowerPC*) . With `Neutral` it defaults to `80`, so where does your `0` come from?

Desmis:
__"You never learned C, so don't use it!
    Don't try to write faster code than Ingo!"__

I strictly apply these two principles **!__**

Floessie:
@Beep6581 It's ready for merge now. If it's okay with you, go ahead and merge it. Thanks.

Beep6581:
@Floessie 
> so where does your 0 come from?

I was actually wondering where the "1" comes from, as I applied "Neutral" in "Fill mode". I couldn't reproduce it intentionally, but unintentionally I did, so I posted it with the "might be invalid for some reason or other" disclaimer.

Thank you @Floessie !

Merged.
Builds and runs fine.

I will leave the branch for you to deleted.

PR #4203, opened 2017-11-30
Update french translation

Stefal:
Correction of a wrong file path in the tooltip:
was `/home/tom/image/02-09-2006/dsc0012.nef`
now `/home/tom/photos/2010-10-31/dsc0042.nef`


Hombre57:
Merci pour la correction.

Issue #3413, opened 2016-09-02
Cannot scroll toolbox with mousewheel when HideTPVScrollbar=true


Beep6581:
In `gtk3` using gtk+-3.18.9, when the vertical scrollbar is hidden, one cannot scroll the toolbox with the mousewheel. Perhaps we should disable that option in gtk3, since the vertical scrollbar auto-hides and so is never in the way.

TooWaBoo:
I can scroll with the mousewheel when the scrollbar is hidden in Windows 10 RT 4.2.1174 GTK3.
Maybe it's only a Linux issue

Beep6581:
Which Gtk3 version do you use?

TooWaBoo:
Branch: gtk3
Version: 4.2.1174
Changeset: a6179ef9697c8d042e686c3a70d5da0243b7cc5f
Compiler: gcc 5.3.0
Processor: generic x86
System: Windows
Bit depth: 64 bits
**Gtkmm: V3.18.0**
Build type: release
Build flags: -m64 -mwin32 -mthreads -std=gnu++11 -mtune=generic -Werror=unused-label -fopenmp -Werror=unknown-pragmas -msse -msse2 -mwindows -fopenmp -Wno-aggressive-loop-optimizations -DNDEBUG -O3
Link flags: -m64 -mwin32 -mthreads -static-libgcc -mtune=generic -mwindows -s -O3
OpenMP support: ON
MMAP support: ON

Beep6581:
I found that with the vertical scrollbar set to hidden using Gtk+-3.18.9, scrolling with the mouse scrollwheel only happens if the cursor is over a combobox, for example the highlight reconstruction method combobox. When it's not over the combobox, I cannot scroll down.

Strangely, if I did scroll the toolbox down by hovering the cursor over a combobox, scrolling down, moving it over the combobox again, scrolling down, and now I move it away from the combobox and scroll down one click, the toolbox scrolls up one click!

TooWaBoo:
With GTK+ 3.18.0 no problems at all.

heckflosse:
I also tried using Win7 gtk3 3.18.0. No Problems here

Beep6581:
Summary: gtk+-3.18.9 and possibly other versions seem to have a bug regarding the mousewheel and scrollbar. Regardless, in Gtk3 the vertical scrollbar auto-disappears. Is there any use keeping the option?

Beep6581:
[...]

PR #4218, opened 2017-12-08
Tool panel vertical scrollbar bug in Linux #3413

Beep6581:
When the vertical scrollbar is set to hidden, one cannot scroll the
toolbox vertically using the mouse scrollwheel in Linux. This patch
works around the problem by forcing the scrollbar to always be visible
in Linux.


Beep6581:
Any objections?

agriggio:
No objections, but FWIW I scrolling works here (gtk+ 3.18.0)

Beep6581:
@agriggio which Linux distro is that?
Please answer in #3413

sguyader:
I see the same issue as described by @Beep6581 when hiding the vertical scrollbar in Manjaro (Gtk+ 3.22.2).
By the way, is really useful to have the option of hiding the scrollbar? Who finds it essential?

heckflosse:
@sguyader It's not essential to hide the scrollbar, but it's useful to hide it if scrolling works without it (as it does on windows builds) ;-)

PR #4228, opened 2017-12-14
include string header for colortemp.h

Benitoite:
To fix mac clang compilation ```error: implicit instantiation of undefined template 'std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >'
    std::string method;```


Benitoite:
This file previously included a larger header which contained the <string> info.

Floessie:
Thanks, @Benitoite!

PR #4237, opened 2017-12-18
Sony pixel shift

agriggio:
Any reason to not merge this?


Beep6581:
@agriggio 
This sneaked in: `rtengine/procparams.cc.save-failed`

agriggio:
thanks, deleted

Beep6581:
Compiles and runs fine using GCC-5.4.0.

heckflosse:
@agriggio Shall we keep `pixelshift: Optionally use one green instead of averaging two greens` or not?

PR #4238, opened 2017-12-19
A new local contrast tool

agriggio:
Ported from G'MIC. The code is extremely simple, and it works quite well for me.
It might be that the same effect can be achieved with a combination of existing RT tools, but that's not obvious. This one is very simple to use...


agriggio:
ah, missing still: partial paste and add/set behaviour preferences in batch mode.
I'll work on that asap

heckflosse:
Left is neutral profile, right is neutral profile + local contrast default settings:

grafik

Local contrast increases the black border of the clock far too much at default settings.

agriggio:
@heckflosse Ingo, can you try with G'MIC inside GIMP (in L-channel mode) and compare the results?

agriggio:
Note, I lowered the amount value from its default in G'MIC, and it gives reasonably good results for me here. But I am totally open to better defaults!

heckflosse:
@agriggio Alberto, I don't have G'MIC installed on my Windows machine. But you can try yourself. [Here's](http://rawtherapee.com/shared/test_images/hdrmerge_045.dng) the file

heckflosse:
@agriggio Alberto, the amount is not the problem here. The large radius (80) is the culprit

agriggio:
radius is the same as the G'MIC default. In fact, G'MIC by default produces even more extreme results:

* G'MIC
hdrmerge_045-gmic

* RT
hdrmerge_045

Beep6581:
Compiles and runs fine.

Simple way of adding a little punch to an image.

imgur-2017_12_19-21 12 23
imgur-2017_12_19-21 12 31

gaaned92:
Windows 10: builds and runs ok
Simple and provides some punch
Build available here:
https://keybase.pub/gaaned92/RTW64NightlyBuilds/RawTherapee_localcontrast_5.3-322-g59da3c5a_WinVista_64.zip

heckflosse:
👍 for merge

heckflosse:
@agriggio Alberto, I would like to push this patch before you merge. Objections?

```
diff --git a/rtengine/iplocalcontrast.cc b/rtengine/iplocalcontrast.cc
index 0445746a..24739fb6 100644
--- a/rtengine/iplocalcontrast.cc
+++ b/rtengine/iplocalcontrast.cc
@@ -27,18 +27,13 @@
 #endif

 #include "improcfun.h"
-#include "settings.h"
 #include "gauss.h"
-#include "boxblur.h"
 #include "array2D.h"
-#define BENCHMARK
-#include "StopWatch.h"

 namespace rtengine {

 void ImProcFunctions::localContrast(LabImage *lab)
 {
-    BENCHFUN
     if (!params->localContrast.enabled) {
         return;
     }
@@ -52,12 +47,12 @@ void ImProcFunctions::localContrast(LabImage *lab)
     const float sigma = params->localContrast.radius / scale;

 #ifdef _OPENMP
-    #pragma omp parallel
+    #pragma omp parallel if(multiThread)
 #endif
     gaussianBlur(lab->L, buf, width, height, sigma);

 #ifdef _OPENMP
-    #pragma omp parallel for
+    #pragma omp parallel for if(multiThread)
 #endif
     for (int y = 0; y < height; ++y) {
         for (int x = 0; x < width; ++x) {

```

agriggio:
Sure, go ahead

heckflosse:
Ok, pushed

Issue #1559, opened 2015-08-11
Support for file type 32 bit floating point TIFF


Beep6581:
Originally reported on Google Code with ID 1575

```
Version: 4.0.9.133

Many HDR programs can save 32 bit floating point TIFFs. The convention of these seems
to be to have values in the range 0 - 1 and if icc profile is embedded the gamma in
that is set to 1.0, i e the expected gamma is 1.

Recent Lightroom release now can load 32 bit tiffs, and it would be nice if also RawTherapee
had this feature. This is usually used when HDR merging is made in a HDR software,
photomatix for example, but tonemapping is left to the raw program, which often is
a good idea if you strive for a natural-looking result.

Loading a 32 bit float into a Image16 is a 10 minute hack (I've done that on my local
build), but then the precision is lost. There is an Imagefloat type too, but it does
not seem to be used for loading. Maybe this is a quick hack for someone who understand
RT image loading.

There is no need to be able to save 32 bit float, just load.
```

Reported by `torger@ludd.ltu.se` on 2012-10-03 21:42:18

Issue #2920, opened 2015-09-26
Amaze Demosaicing Tile Size Optimization.


TimZaman:
Hi.

Browsing through the code i found the tile size in amaze demosaicing to be quite small. 
The tiles are used to cut up the work in tiles of this certain size. I found that when the size of the tiles are set small (160) and the image is large (80 megapixel), the speed improvement of making the tile of size 512 increases the speed by 20%. If i set it at 1024, it only increases by 10%.
If i use a 40 megapixel file, and set the tile size to 512 it increases the speed by 10%.

I would suggest increasing the tile size at least 2x, but maybe someone else first confirm such preliminary findings, or someone can elaborate why the value of 160 was chosen.

https://github.com/Beep6581/RawTherapee/blob/master/rtengine/amaze_demosaic_RT.cc#L51-L52

heckflosse:
@TimZaman maybe this is of interest for you: https://github.com/Beep6581/RawTherapee/issues/1819#issuecomment-129995754

TimZaman:
Ah that's spot on. I wonder what other parameters are of importance,
like image size or maybe even image saliency.
Anyone up for trying this again?

heckflosse:
I'm not against to make a new try to speedup Amaze. The last time I took a look at Amaze it ended with the assumption (only in my brain) that a further speedup of about 25% could be possible (without changing tile size)

TimZaman:
I was planning on looking into porting it to opencl or cuda, probably the
former. Been on my list for ages. Duo to the complexity i have some trouble
slicing and chunking down the process though.

heckflosse:
@TimZaman I know about the complexity of Amaze code. I made all the optimizations (including SSE) for it.

heckflosse:
@TimZaman 
The tile size should be chosen to fit 'almost' all tile buffers into L2 cpu cache for each core. At the time I made the opts for amaze that was the case. Don't know if that's still the case for newer cpus. Maybe we find a better way to (dynamically) adjust tile size (determine size of L2 cache and adjust tile size).

PR #4250, opened 2017-12-23
Add BENCHFUN to more places

gatoatigrado:
I was interested in improving performance, and I think getting better benchmarks might be helpful. I tried `WITH_PROF=ON`, but the output I got wasn't very useful -- a large chunk of time is missing, and calls like `sony_arw2_load_raw()` are quite off (`BENCHFUN` and print statements reveal it taking a few hundred milliseconds, not 4.78 seconds). So I added BENCHFUN to a few more functions which look like they'll take time. One downside of this, however, is that sometimes timers are nested ... I could add some more code to display more user-friendly output of that if desired.

(NOTE: gprof invocation is here, LMK if I'm doing something wrong --
 https://gist.github.com/gatoatigrado/6fc0cc7dcac58ef759aa8d2cce17b3e9)

An example of the new output is here (https://gist.github.com/gatoatigrado/c0009ec97b64ecbc8def720c28a0de3e), and here are some pie charts (just from google sheets)

flickr06
flickr06
flickr06

Overall ExecCMSTransform is taking the most time ... if you all want to point me to a thread on that, I'd love to discuss. I see a note here (https://github.com/Beep6581/RawTherapee/issues/1559) that things got faster in version 2.4, but I seem to have version 2.6 installed so probably I'm already seeing that benefit.

Finally, I noticed that multi-core operation is not very much faster; my timings with OpenMP were like 23.4s user time on the default RawTherapee profile (sum over all cores, using a Haswell i7-4790K), whereas single-threaded operation is around 8.2s user time on the default RawTherapee profile. Since my use case can parallelize across images, I think I might do that & just use the single-threaded version. But maybe multi-threading can be made more efficient (coarser tasks?).

Cheers!

### DIFF
--- a/rtengine/StopWatch.h
+++ b/rtengine/StopWatch.h
@@ -19,6 +19,8 @@
  *  Author: reine
  */
 
+#define BENCHMARK
+
 #ifndef STOPWATCH_H
 #define STOPWATCH_H
 #include <iostream>
--- a/rtengine/StopWatch.h
+++ b/rtengine/StopWatch.h
@@ -19,8 +19,6 @@
  *  Author: reine
  */
 
-#define BENCHMARK
-
 #ifndef STOPWATCH_H
 #define STOPWATCH_H
 #include <iostream>
--- a/rtengine/amaze_demosaic_RT.cc
+++ b/rtengine/amaze_demosaic_RT.cc
@@ -26,7 +26,6 @@
 //
 ////////////////////////////////////////////////////////////////
 
-#define BENCHMARK
 #include "rtengine.h"
 #include "rawimagesource.h"
 #include "rt_math.h"
--- a/rtengine/amaze_demosaic_RT.cc
+++ b/rtengine/amaze_demosaic_RT.cc
@@ -26,6 +26,7 @@
 //
 ////////////////////////////////////////////////////////////////
 
+#define BENCHMARK
 #include "rtengine.h"
 #include "rawimagesource.h"
 #include "rt_math.h"
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -122,6 +122,8 @@ typedef unsigned char uchar;
 typedef unsigned short ushort;
 
 #include "dcraw.h"
+#include "StopWatch.h"
+
 /*
    RT All global variables are defined here, and all functions that
    access them are prefixed with "CLASS".  Note that a thread-safe
@@ -3208,6 +3210,7 @@ void CLASS sony_load_raw()
 
 void CLASS sony_arw_load_raw()
 {
+  BENCHFUN
   ushort huff[32770];
   static const ushort tab[18] =
   { 0xf11,0xf10,0xe0f,0xd0e,0xc0d,0xb0c,0xa0b,0x90a,0x809,
@@ -3228,6 +3231,7 @@ void CLASS sony_arw_load_raw()
 
 void CLASS sony_arw2_load_raw()
 {
+  BENCHFUN
 
 #if defined( _OPENMP ) && defined( MYFILE_MMAP )
 #pragma omp parallel
--- a/rtengine/image16.cc
+++ b/rtengine/image16.cc
@@ -21,6 +21,7 @@
 #include "image8.h"
 #include <cstdio>
 #include "rtengine.h"
+#include "StopWatch.h"
 
 namespace
 {
@@ -333,6 +334,7 @@ Image16::tofloat()
 // Parallized transformation; create transform with cmsFLAGS_NOCACHE!
 void Image16::ExecCMSTransform(cmsHTRANSFORM hTransform, const LabImage &labImage, int cx, int cy)
 {
+    BENCHFUN
     // LittleCMS cannot parallelize planar Lab float images
     // so build temporary buffers to allow multi processor execution
 #ifdef _OPENMP
--- a/rtengine/iplab2rgb.cc
+++ b/rtengine/iplab2rgb.cc
@@ -26,6 +26,7 @@
 #include "curves.h"
 #include "alignedbuffer.h"
 #include "color.h"
+#include "StopWatch.h"
 
 namespace rtengine
 {
@@ -264,6 +265,7 @@ Image8* ImProcFunctions::lab2rgb (LabImage* lab, int cx, int cy, int cw, int ch,
  */
 Image16* ImProcFunctions::lab2rgb16 (LabImage* lab, int cx, int cy, int cw, int ch, const procparams::ColorManagementParams &icm, GammaValues *ga)
 {
+    BENCHFUN
 
     if (cx < 0) {
         cx = 0;
--- a/rtengine/ipsharpen.cc
+++ b/rtengine/ipsharpen.cc
@@ -23,6 +23,7 @@
 #include "rt_math.h"
 #include "sleef.c"
 #include "opthelper.h"
+#include "StopWatch.h"
 
 using namespace std;
 
@@ -36,6 +37,7 @@ namespace rtengine
 extern const Settings* settings;
 SSEFUNCTION void ImProcFunctions::dcdamping (float** aI, float** aO, float damping, int W, int H)
 {
+    BENCHFUN
 
     const float dampingFac = -2.0 / (damping * damping);
 
@@ -92,6 +94,7 @@ SSEFUNCTION void ImProcFunctions::dcdamping (float** aI, float** aO, float dampi
 
 void ImProcFunctions::deconvsharpening (float** luminance, float** tmp, int W, int H, const SharpeningParams &sharpenParam)
 {
+    BENCHFUN
     if (sharpenParam.deconvamount < 1) {
         return;
     }
@@ -151,6 +154,7 @@ void ImProcFunctions::deconvsharpening (float** luminance, float** tmp, int W, i
 
 void ImProcFunctions::sharpening (LabImage* lab, float** b2, SharpeningParams &sharpenParam)
 {
+    BENCHFUN
 
     if (!sharpenParam.enabled) {
         return;
@@ -256,6 +260,7 @@ void ImProcFunctions::sharpening (LabImage* lab, float** b2, SharpeningParams &s
 
 void ImProcFunctions::sharpenHaloCtrl (float** luminance, float** blurmap, float** base, int W, int H, const SharpeningParams &sharpenParam)
 {
+    BENCHFUN
 
     float scale = (100.f - sharpenParam.halocontrol_amount) * 0.01f;
     float sharpFac = sharpenParam.amount * 0.01f;
@@ -324,6 +329,7 @@ void ImProcFunctions::sharpenHaloCtrl (float** luminance, float** blurmap, float
 // Thanks to Manuel for this excellent job (Jacques Desmis JDC or frej83)
 void ImProcFunctions::MLsharpen (LabImage* lab)
 {
+    BENCHFUN
     // JD: this algorithm maximize clarity of images; it does not play on accutance. It can remove (partialy) the effects of the AA filter)
     // I think we can use this algorithm alone in most cases, or first to clarify image and if you want a very little USM (unsharp mask sharpening) after...
     if (!params->sharpenEdge.enabled) {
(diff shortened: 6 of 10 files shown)


heckflosse:
@gatoatigrado 
lcms (ExecCMSTransform) is indeed a bottleneck.
sony_arw2_load_raw() scales well on my machine (if the arw is loaded from SSD or file system cache).
If the arw is loaded from network or HD, it does not scale well

Can you elaborate which functions do not scale well on your machine?

heckflosse:
@gatoatigrado 

> amaze_demosaic_RT took 1446 ms

that sounds too much for amaze demosaicing an arw file.
I just tried on my rusty AMD FX-8350 amaze demosaicing a 42.3 MP arw from Sony 7RM3.
That takes 700 ms

heckflosse:
@gatoatigrado I just saw that you used rawtherapee-cli for your tests. That of course has some overhead you won't have when you process the images in rt queue. Can you verify your measurements in rt queue?

heckflosse:
@gatoatigrado About amaze demosaic: Long time ago we optimized amaze demosaic to perform well on my AMD FX8350 and also on an I7 6 cores/ 12 threads. One critical part here is the tile size, #defined by AMAZETS in amaze_demosaic_RT.cc 
[For reference](https://github.com/Beep6581/RawTherapee/issues/2920)

gatoatigrado:
Hi @heckflosse, thanks for all of the comments. Since I am relatively new, I wouldn't be surprised if my setup is just wrong. FWIW I am building with,

```
cmake -DCMAKE_BUILD_TYPE="relwithdebinfo" \
      -DPROC_TARGET_NUMBER="2" \
      -DBUILD_BUNDLE="ON" \
      -DCACHE_NAME_SUFFIX="5-dev" \
      -DWITH_LTO="OFF" -DWITH_PROF=ON -DOPTION_OMP=OFF ..
```

Just to clarify, the 700ms you mention before is from running AMaZE on a single thread?

I can try just `release` instead of `relwithdebinfo` if you think that'd make a significant difference? IIUC the debug symbols are just stored separately and shouldn't affect performance though.

Please point me to docs or let me know how to use rt queue, I am not familiar. I don't see why the CLI version would be slower if I am getting benchmark results for the particular function though?

Alternately, if you wish to re-run my benchmark, my input is from an A7R (jelly of your mk 3 ;) i totally want one), a test file is here: https://drive.google.com/file/d/1CTX2np2yCVGxNHZohhC1ZkJ09Qe753W3/view?usp=sharing

the command line is

```
/tmp> which rawtherapee-cli
.../RawTherapee/build/relwithdebinfo/rawtherapee-cli
/tmp> time rawtherapee-cli -Y -t -c /tmp/test.arw                                       
```

Some kind of "not retained" trigger was being hit so I had to hack `main-cli.cc`, please LMK what a better way would be (https://gist.github.com/gatoatigrado/168b7f7d820c2ef9522f2d9f3a677578)

As far as `ExecCMSTransform`, I tried making just 1 call on a whim, but it didn't help. Is that just doing a colorspace transform? I don't understand why that'd be expensive...

As far as multi-threaded scaling, I can look more soon; gprof seemed to suggest `rtengine::wavelet_level<float>::AnalysisFilterHaarHorizontal` was taking an amount of time, so perhaps the task granularity for that is wrong or something.

heckflosse:
@gatoatigrado 

> calls like sony_arw2_load_raw() are quite off (BENCHFUN and print statements reveal it taking a few hundred milliseconds, not 4.78 seconds)

In your gprof output, look at the `self seconds` column, not on the `cumulative seconds`

heckflosse:
> Just to clarify, the 700ms you mention before is from running AMaZE on a single thread?

No, 8 threads

Edit: single thread is > 3000 ms, but my machine is limited by memory bandwidth. For that reason I don't get better scaling most of the time (except for really expensive computations (i.e. exp or log) where memory bandwidth does not matter)

gatoatigrado:
OK, then I think the 1400ms I was getting on a single thread is pretty good. I saw the "self time" column, but still can't make sense of that (the LCMS call is too small, I'm guessing its library isn't included -- https://gist.github.com/gatoatigrado/f6870f1ba7ac6eb39a7b559d776e0593). Here is a gprof2dot output, but it also looks wonky,

graph

I'm going to try and get Google's periodic interrupt-based stacktrace profiler (https://github.com/gperftools/gperftools) to work ...

heckflosse:
Of course, for a single function the cli version is not slower, but your time command (which you used to compare single vs. multithreaded execution) also included the start time of rt (where for example lookup tables are build an so on)

heckflosse:
> my input is from an A7R (jelly of your mk 3 ;)
> OK, then I think the 1400ms I was getting on a single thread is pretty good.


I don't own a mk3. I used a arw from such a camera to test ;-)
Your 1400ms for a single thread are pretty good. What's the processing time if you enable all cores?

Your arw needs ~2600 ms using a single thread and ~530 ms using 8 threads on my machine (in amaze demosaic function)

About the queue: Just start rawtherapee (without the -cli), add your pictures to the queue and press the start button in queue tab.

heckflosse:
addtoqueue

startqueue

heckflosse:
Also I would not use -DWITH_PROF=ON at all

heckflosse:
> As far as ExecCMSTransform, I tried making just 1 call on a whim, but it didn't help. Is that just doing a colorspace transform? I don't understand why that'd be expensive...

It's a while ago I checked why lcms is slow. At that time it turned out to be slow because it uses function pointers and expensive (exp, log, whatever) functions, which (using standard c++ libs) won't be vectorized.
But it's more than 2 years ago, that I checked that. Maybe lcms code has changed meanwhile, don't know.

heckflosse:
@gatoatigrado btw: here's my configuration

```
cmake -DCMAKE_BUILD_TYPE="Release" -DPROC_TARGET_NUMBER="2" -DBUILD_BUNDLE="ON" -DCACHE_NAME_SUFFIX=5-dev ..
```

gatoatigrado:
Thanks! For rt queue, I think I am running into the same issue I had to hack around with the "not retained" issue. I did get the Google profiler to work though, here's the commands (no binary modification necessary apparently),

```
> aptitude install libgoogle-perftools-dev
> export CPUPROFILE=/tmp/prof.out CPUPROFILE_FREQUENCY=1000
> rm /tmp/prof.out; LD_PRELOAD=/usr/lib/libprofiler.so rawtherapee-cli -Y -t -c /tmp/test.arw && ls /tmp/prof.out
> google-pprof --svg $(which rawtherapee-cli) /tmp/prof.out > profile.svg && google-chrome profile.svg
```

Here is a graph for single-threaded performance,

rawtherapee_profile svg

Here is a graph for multi-threaded performance,

openmp-profile svg

It looks like the LittleCMS calls are proportionally smaller in multi-threaded, but AMaZE and rgbProc are larger, so those probably scale worse. LittleCMS is still the biggest though. I don't know what `cmsGetProfileInfoASCII` is, but anything with ASCII in the name doesn't sound super fast ....

heckflosse:
@gatoatigrado That's very interesting. I knew that lcms was a bottleneck, but that cmsGetProfileInfoASCII bottleneck new to me. Though I don't know how to improve that atm, as it's not in RT code. I will investigate nevertheless...

heckflosse:
> It looks like the LittleCMS calls are proportionally smaller in multi-threaded, but AMaZE and rgbProc are larger, so those probably scale worse.

Yes, that's imho caused by LittleCMS calls using expensive functions (exp, log) which are not so much influenced by memory bandwidth as the simple operations in demosaic and rgbproc.

heckflosse:
Anyway, I forgot to mention, that I really appreciate your efforts to speed up RT and that I'm not against adding more timing code (BENCHFUN)

👍

gatoatigrado:
Thanks :). I compiled lcms from source and got a new profile, which doesn't have the "ASCII", so maybe that's a misnomer (but the time is still high).

openmp-profile svg

(this is with OpenMP).

as a general comment (I haven't dug into lcms yet), vectorization can still help for expensive functions; a coworker wrote some fast approximations for neural networks, cf https://github.com/google/sling/blob/master/sling/myelin/kernel/avx-math.cc#L245 . maybe we can do something similar.

heckflosse:
> vectorization can still help for expensive functions

But not for external libraries which are used by RT (like lcms), or?

In rt code we we already use vectorized math functions (exp, log, sincos...) but we can't change lcms code. That's a different project...

heckflosse:
@gatoatigrado Your graph is very interesting!

gatoatigrado:
Yeah, I'm not exactly sure if I'll be able to make an intervention. I'm a little tempted to train a replacement in TensorFlow, but that seems like not very convenient since it'd be fixed to a particular ICC profile.

It seems the existing code calls a lot of function pointers, with non-vectorized arguments,

```
static
void _LUTeval16(register const cmsUInt16Number In[], register cmsUInt16Number Out[],  register const void* D)
{
    cmsStage *mpe;
...
             mpe ->EvalPtr(&Storage[Phase][0], &Storage[NextPhase][0], mpe);
...
}
```

iiuc this `EvalPtr` is a function pointer,

```
    _cmsStageEvalFn     EvalPtr;        // Points to fn that evaluates the stage (always in floating point)
```

```
typedef void (* _cmsStageEvalFn)     (const cmsFloat32Number In[], cmsFloat32Number Out[], const cmsStage* mpe);
```

In my experience this style is pretty much guaranteed to be slow. If it were C++ it might be possible to template in the function being called ... I don't know how receptive the lcms maintainer is to that kind of change.

gatoatigrado:
OK, I manually inlined a few functions in LCMS and it dropped the time from ~740ms to ~670ms.

New profiling script: https://gist.github.com/gatoatigrado/158c52a8ed5f15b1b12e151d3a23ea41
LCMS diff that inlines a few functions: https://gist.github.com/gatoatigrado/5ed4f608f06a5e68586ad3e850478b10 (I think CLion changed the encoding, so ignore those weird changes to "Cache" or something)

Old traces look like this:

flickr06

New traces look like this:

flickr06

Runs:

```
# Old
oliver:/tmp> for i in {1..3}; do python ~/bin/alias/rt_profile.py 2>&1 | grep ExecCMSTransform; done
ExecCMSTransform took 740 ms
ExecCMSTransform took 748 ms
ExecCMSTransform took 758 ms
# New
oliver:/tmp> for i in {1..3}; do python ~/bin/alias/rt_profile.py 2>&1 | grep ExecCMSTransform; done
ExecCMSTransform took 677 ms
ExecCMSTransform took 668 ms
ExecCMSTransform took 666 ms
```

I guess this partly demonstrates that there is a lot of low-hanging fruit in LCMS performance-wise, but doesn't exactly give us a quick solution. I will try to do some TensorFlow modeling as well ... perhaps if we can find a convex optimization problem, running something like libsvm inline would be reasonable.

Also I think we should probably migrate this discussion to an issue ... since the Google profiler seems to provide useful graphs, perhaps we could just close/discard this pull request? Either way is fine ... I guess BENCHFUN means people don't have to install the Google profiler (I dunno how Mac/Windows compatibility is too).

heckflosse:
@gatoatigrado I just prepared a commit which allows to set `-DWITH_BENCHMARK=ON` in cmake command (means BENCHMARK is #defined then). I'll push it soon.

heckflosse:
@gatoatigrado I think we have two different use cases here:

1. people who know what they are doing (certainly you, maybe also me ;), Maybe also other rt developers. I have to dig into the google profiler a bit more, I have to admit.

2. normal users, which experience performance issues. For them we can provide a build with `-DWITH_BENCHMARK=ON` to detect bottlenecks, if we, as you suggested, plant some more `BENCHFUN` into the code.

Beep6581:
When this gets merged, the compilation docs need to be updated to reflect the possibility of the
`-DWITH_BENCHMARK=ON` flag.

heckflosse:
@gatoatigrado I committed https://github.com/Beep6581/RawTherapee/commit/11ad1dace8458a4373f74f3e37d98f2b8e56e255 a while ago

Beep6581:
@gatoatigrado thank you for the PR! The documentation and the build scripts have been updated:
http://rawpedia.rawtherapee.com/Linux#CMake
09d020cca3020e4051cc2329d60d598af26a9806

The placement of some of these definitions can be improved, so we will now close this PR and add them in optimal places in other commits.

PR #4253, opened 2017-12-25
Use AlignedBuffer helper class in rgbProc, use SSE in standard tone c…

gatoatigrado:
For my Sony profile, this seemed to be a mild hot spot. Obviously I'm a newish candidate contributor, so please let me know what the best way to test this would be. I checked that it seems to be getting called, and seems to use fewer cycles according to the google profiler than the previous method (15 vs. 43 samples, using the method in https://github.com/Beep6581/RawTherapee/pull/4250). On my test image, this seems to produce the same result as before (and I checked that disabling the tone curve code entirely yielded the wrong image).

flickr06

Previously, `rgbProc()` seemed to give the last channel more padding than necessary.  I switched this so that all channels get the same padding (currently `perChannelSizeBytes`). Please review carefully to make sure it's OK.

### DIFF
--- a/rtengine/LUT.h
+++ b/rtengine/LUT.h
@@ -95,6 +95,8 @@ protected:
     // list of variables ordered to improve cache speed
     unsigned int maxs;
     float maxsf;
+    // possibly-more-correct value for sse routine (see unit test for details)
+    float maxIndexFloat;
     T * data;
     unsigned int clip;
     unsigned int size;
@@ -129,6 +131,7 @@ public:
         upperBound = size - 1;
         maxs = size - 2;
         maxsf = (float)maxs;
+        maxIndexFloat = ((float)upperBound) - 1e-5;
 #if defined( __SSE2__ ) && defined( __x86_64__ )
         maxsv =  F2V( maxs );
         sizeiv =  _mm_set1_epi32( (int)(size - 1) );
@@ -158,6 +161,7 @@ public:
         upperBound = size - 1;
         maxs = size - 2;
         maxsf = (float)maxs;
+        maxIndexFloat = ((float)upperBound) - 1e-5;
 #if defined( __SSE2__ ) && defined( __x86_64__ )
         maxsv =  F2V( maxs );
         sizeiv =  _mm_set1_epi32( (int)(size - 1) );
@@ -228,6 +232,7 @@ public:
             this->upperBound = rhs.upperBound;
             this->maxs = this->size - 2;
             this->maxsf = (float)this->maxs;
+            this->maxIndexFloat = ((float)this->upperBound) - 1e-5;
 #if defined( __SSE2__ ) && defined( __x86_64__ )
             this->maxsv =  F2V( this->size - 2);
             this->sizeiv =  _mm_set1_epi32( (int)(this->size - 1) );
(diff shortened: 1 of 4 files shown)


heckflosse:
@gatoatigrado 

Accessing the lut with integer inidices instead of float indicies should give a different result because the interpolation is lost.

Your vectorized code uses [this operator](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/LUT.h#L362), which does not interpolate

The scalar access with a float uses [this one](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/LUT.h#L447), which interpolates.

A proper solution would be to make a correct version of [this one](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/LUT.h#L297) and access it with a vector of floats.

Edit: I added a sreenshot which shows that your version gives different output. I used `compare image1.tif image2.tif.gif  compare.png` from imagemagick to check for differences.

grafik

gatoatigrado:
Thanks! Yes this isn't a difference that's easily visible. I'm working on fixing up the current version. I also noticed there is a slight "bug", if you have a 4-element LUT with values

```
[10, 11, 12, 13]
```

then looking up 2.5 yields 13.0, not 12.5. I think the solution would be to replace `maxsf` with `upperBound - epsilon` (i.e. `size - (1+epsilon)`) instead of `upperBound - 1` (or `size - 2`). I hope it's OK if I replace this in my SSE implementation.

Also how do you feel about adding gunit (https://github.com/google/googletest/blob/master/googletest/README.md)? I got it integrated in cmake, but I'm not sure how to make it a conditional dependency. Essentially Ubuntu users would have to `apt-get install libgtest-dev`, which seems easy enough, but LMK.

gatoatigrado:
Oh, sorry, I hit the wrong button, didn't mean to close.

gatoatigrado:
@heckflosse I wrote a new interpolating routine, and it seems to work. I wrote a unit test here (https://gist.github.com/gatoatigrado/b769582c36ca1f0bc566d78acb292f71) but I think my integration of gunit into the CMakeLists resulted in rebuilding the rawtherapee target. At least it seemed to be rebuilding more files than I recall it was before. Maybe I can open an issue to discuss integrating a unit test framework?

rtengine/LUT.h
```diff
@@ -293,72 +298,37 @@ class LUT :
...
+        vint indexes = _mm_cvttps_epi32(clampedIndexes);
+        int indexArray[4];
+        _mm_storeu_si128(reinterpret_cast<__m128i*>(&indexArray[0]), indexes);
+
+        // Load data from the table. This reads more than necessary, but there don't seem
+        // to exist more granular operations (though we could try non-SSE).
+        // Cast to int for convenience in the next operation (partial transpose).
+        vint values[4];
+        for (int i = 0; i < 4; ++i) {
+            values[i] = _mm_castps_si128(LVFU(data[indexArray[i]]));
```
heckflosse, line 330: This potentially reads out of bound. Maybe better to increase the size of the new[] in line 128 by 3?

rtengine/LUT.h
```diff
@@ -293,72 +298,37 @@ class LUT :
...
-        tempv = _mm_shuffle_ps(tempv, tempv, _MM_SHUFFLE(1, 1, 1, 1));
-        p2v = _mm_move_ss( p2v, tempv);
-        // now p2v is 3 2 1 0
-
-        vfloat diffv = indexv - _mm_cvtepi32_ps ( idxv );
-        diffv = vself(vorm(maxmask, minmask), _mm_setzero_ps(), diffv);
-        resultv = p1v + p2v * diffv;
-        return resultv  ;
+        vfloat diff = clampedIndexes - _mm_cvtepi32_ps(indexes);
+        return (_mm_set1_ps(1.0f) - diff) * lower + (diff * upper);
```
heckflosse: Maybe using `vintpf` from [here](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/sleefsseavx.c#L1408) is faster. I didn't test.

rtengine/LUT.h
```diff
@@ -456,6 +426,8 @@ class LUT :
             }
 
             idx = 0;
+            // Note: Maybe this should be 'idx > maxsf'? See unit test where a LUT with
+            // values [10, 11, 12, 13] gets looked up at 2.5 and returns 12.5.
         } else if (index > maxsf) {
```
heckflosse: Yes, this should be `idx > maxsf` or `idx > maxs`. Good catch 👍

rtengine/LUT.h
```diff
@@ -456,6 +426,8 @@ class LUT :
             }
 
             idx = 0;
+            // Note: Maybe this should be 'idx > maxsf'? See unit test where a LUT with
```
heckflosse: Yes, this should be `idx > maxsf` or `idx > maxs`. Good catch 👍

Edit: I have to think about that again...
Edit: `idx > maxs` should be correct and fastest (not tested)

heckflosse:
@gatoatigrado 

> Also how do you feel about adding gunit (https://github.com/google/googletest/blob/master/googletest/README.md)?

Please open an issue to discuss that. 👍

heckflosse:
@gatoatigrado That looks really good! But I think there is still one issue (though it's not obvious):
The Standard Tone Curve uses a LUT which does not clip (LUT_CLIP_BELOW/LUT_CLIP_ABOVE are not set), means for your example `[10, 11, 12, 13]` it should return `13.5` if you access it with `3.5`. As your use case is the first one we can do that by changing

```
        vfloat diff = clampedIndexes - _mm_cvtepi32_ps(indexes);
```

to

```
        vfloat diff = indexv - _mm_cvtepi32_ps(indexes);
```
Please correct me if I'm wrong


Edit: scratch that. I was wrong, LUT_CLIP_BELOW and LUT_CLIP_ABOVE are defaults. It's too late in the evening...

heckflosse:
I have no objections to merge this pr. @Floessie @agriggio ?

Edit: I checked for differences. Found none (no red pixels using compare) 👍 

grafik

Edit: I also measured processing time of rgbproc for a 36MPixel file using neutral profile + Tone Curve 1 Standard. As usual the median of 7 measures:

Before: 434 ms
After: 404 ms

Good job!

heckflosse:
@gatoatigrado May I point you to some other possible use cases of your vectorized LUT access in `rgbProc()`?

I will open an issue in case you don't mind ;-)

heckflosse:
@gatoatigrado Some suggestions to change (though I'm willing to merge your pr even without this):

We have some macros (though I don't like macros and @Floessie also does not like them ;-) and also some inlines for better readability (imho at least).

These are:

ZEROV for _mm_set_zero_ps()
F2V(x) for _mm_set1_ps(x) // F2V means Float To Vector
vminf(x) and vmaxf(x) for _mm_min_ps(x) and _mm_max_ps(x)

Additionally, could you please add a comment to `operator[](vfloat)` that the current implementation is restricted to LUTs which clip at lower and upper bound? I know, it's my fault that I didn't add a comment in my crappy try which I marked `not ready for prodcution`, but I cannot push the comment to your fork ;-)

gatoatigrado:
@heckflosse Thanks for all of the comments. I hope the last version addresses everything, but LMK if I missed something. Please feel free to vectorize other call sites; I will continue to investigate based on CPU profiling, but that'll only lead me to look at whatever is currently in my [RawTherapee / Sony] profile. Cheers!

gatoatigrado:
Also, I added you to my fork contributors, maybe that will let you directly edit PRs

Floessie:
> I have no objections to merge this pr. @Floessie @agriggio ?

Me neither, though I didn't have the time to dig as deep as you did (it's Christmas after all).

@gatoatigrado If you come across C-style casts like in ebc92e1c350b028ae16a113051b8c3e46d4da0f1, don't hesitate to convert them to proper `static_cast<>`s or whatever fits best.

Good job! 👍 

> We have some macros (though I don't like macros and @Floessie also does not like them ;-) and also some inlines for better readability (imho at least).

😋 

Best,
Flössie

agriggio:
no objections from me either, thanks for your work! :+1:

I'm actually very interested in learning how to properly vectorize code and make the most out of SSE, so this is a good excuse to dig deeper into this :-)

Beep6581:
> but that'll only lead me to look at whatever is currently in my [RawTherapee / Sony] profile

Speaking of which, if you need images to test on, there's a whole bunch of them from various cameras here (`crash_kodak_pb645m.dcr`, `panasonic_dmc-fz28_crashes.rw2` and possibly `nikon_d1x_lightning.dng` are corrupt to catch now-handled crashes):
http://rawtherapee.com/shared/test_images/

heckflosse:
@gatoatigrado 

> Please feel free to vectorize other call sites

I'm already working on that. First pr (which improves processing time of rgbProc independent on the settings) will follow soon.

PR #4273, opened 2018-01-03
Crop tweaks

agriggio:
added a few tweaks to improve the crop tool (or at least to allow a bit more customization). You can now select whether to perform "auto zoom to fit" when creating/editing a crop, and whether to hide the guides when not editing it. Also, if you use a solid background, it will become transparent while you are modifying the crop. Finally, I've added a new crop ratio "As Image"


heckflosse:
@agriggio Suggestion:

The Combo box right from crop ratio box could be hidden or disabled when crop ratio as image is selected

heckflosse:
@agriggio No objections to merge from my side

PR #4297, opened 2018-01-12
Speedup for Colour Toning Method 'Saturation 2 colours'

heckflosse:
Measured in rgbproc() with a 36Mpixel file using neutral profile + Color Toning Method `Saturation 2 colours` 

```
neutral                                           :  300 ms
neutral + ColorToning Saturation 2 colours before : 1940 ms
neutral + ColorToning Saturation 2 colours after  :  700 ms
```


Beep6581:
No issues found.

PR #4301, opened 2018-01-13
Bugfix in LabGrid class + code refactoring

Hombre57:
- suppressed trailing `_` char for class' parameters
- renamed function name (those not related to Gtk) to lowerCamelCase
- changed the mecanism of dragging point, to avoid throwing unecessary `panelChange` event
- added a `reset` function, with support of `toInitial` to get back the values at load time (CTRL+click on reset button)
- `on_draw` optimisation: now LabGrid IS A BackBuffer

### DIFF
--- a/rtdata/languages/Francais
+++ b/rtdata/languages/Francais
@@ -1381,6 +1381,7 @@ TP_COLORTONING_HIGHLIGHT;Hautes lumières
 TP_COLORTONING_HUE;Teinte
 TP_COLORTONING_LAB;Mixage Lab
 TP_COLORTONING_LABEL;Virage Partiel
+TP_COLORTONING_LABGRID_VALUES;HL: a=%1 b=%2\nO: a=%3 b=%4
 TP_COLORTONING_LUMA;Luminance
 TP_COLORTONING_LUMAMODE;Préserver la luminance
 TP_COLORTONING_LUMAMODE_TOOLTIP;Si activé, lorsque vous changez la couleur (rouge, vert, cyan, bleu, etc.), la luminance de chaque pixel est préservé
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -1407,6 +1407,7 @@ TP_COLORTONING_HUE;Hue
 TP_COLORTONING_LAB;L*a*b* blending
 TP_COLORTONING_LABEL;Color Toning
 TP_COLORTONING_LABGRID;L*a*b* color correction grid
+TP_COLORTONING_LABGRID_VALUES;HL: a=%1 b=%2\nS: a=%3 b=%4
 TP_COLORTONING_LUMA;Luminance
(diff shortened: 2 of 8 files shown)


agriggio:
> - suppressed trailing _ char for class' parameters

I hate you :-)

> - renamed function name (those not related to Gtk) to lowerCamelCase

now even more! ;-)

seriously though, sorry for messing with the style -- I was using my usual one without thinking that RT has different conventions...

Hombre57:
@agriggio 
>I hate you :-)

I feared that :-/

In first place, I thought it was the naming convention from DT that you copied the code from.

RT is >90% consistent with regard to conventions. I'm not the one who will say that it has to be enforced, that's not _my_ project so if we agree to have no conventions, I have no problem to revert to your coding style.

agriggio:
@Hombre57 I was kidding :-) 
I do like my conventions (a lot) as I find them helpful, but I can totally live without them. I used them simply because I sort of have them "hardcoded" and I was not paying enough attention. I will be more careful in the future and (try to) stick with the commonly-used ones.

Hombre57:
Ok, I was a bite unsure... 😅 

Just tell me if you have objection for merging ? Same for @Beep6581

Beep6581:
Tested:
* Compiles fine, no warnings.
* Seems to work fine.
* Saved image matches preview.
* Partial paste seems to work fine.
* Batch editing mostly works, but I found an odd issue:
0. Clear the processing profiles from a bunch of images.
1. Click a thumbnail in the File Browser, in Batch Edit enable Color Toning, set method to CCGrid, and make some tone changes.
2. Shift+select one more thumb, so now two are selected. The "Color toning" power button is in inconsistent state, good. Click it once, it turns CT off. Click it again, now both images have CT on, good. The second thumb now has the same color toning as the first, good.
3. Shift+select a third image. The "Color toning" power button is in inconsistent state, good. Click it once, it turns CT off. Click it again. You'd expect all three images to have CT on, but in my first two tests, only the first two images had CT on, the third image had it off, even though CT was enabled and the CCGrid method was selected and I modified the two nodes. However, the third time I tried to reproduce this I couldn't - all images worked as they should, even when I selected up to 5 images...

Beep6581:
+1 for merging. I'm not even sure the above thing is a bug - it's late.

agriggio:
no objections to merging!

Hombre57:
@Beep6581 @agriggio Thanks.

I'll try to reproduce your issue.

Beep6581:
@Hombre57 I deleted `~/.cache/RawTherapee*` and `~/photos/test_images/*.pp3` and re-tested. Reproduced it again, on the second thumb. The second PP3 had color toning enabled and method grid, but:
```
LabGridALow=0
LabGridBLow=0
LabGridAHigh=0
LabGridBHigh=0
```

PR #4312, opened 2018-01-17
Histogram matching

agriggio:
By popular demand, here you go:
https://filebin.net/7knvtwuauiddowy7/Peek_2018-01-17_22-00.mp4

more or less works, but it needs to be tested with as many images from as many cameras as possible.
Also batch mode is a bit weird in that it will not update the thumbnail (but the setting will be recorded in the pp3). This is not so much for performance (which might be an issue though) but rather to avoid introducing more mess in an already messy part of RT (rtthumbnail)

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -724,6 +724,7 @@ HISTORY_MSG_491;White Balance
 HISTORY_MSG_492;RGB Curves
 HISTORY_MSG_493;L*a*b* Adjustments
 HISTORY_MSG_COLORTONING_LABGRID_VALUE;CT - Color correction
+HISTORY_MSG_HISTMATCHING;Auto-matched Tone Curve
 HISTORY_MSG_LOCALCONTRAST_AMOUNT;Local Contrast - Amount
 HISTORY_MSG_LOCALCONTRAST_DARKNESS;Local Contrast - Darkness
 HISTORY_MSG_LOCALCONTRAST_ENABLED;Local Contrast
@@ -1552,6 +1553,7 @@ TP_EXPOSURE_CURVEEDITOR1;Tone curve 1
 TP_EXPOSURE_CURVEEDITOR2;Tone curve 2
 TP_EXPOSURE_CURVEEDITOR2_TOOLTIP;Please refer to the "Exposure > Tone Curves" RawPedia article to learn how to achieve the best results by using two tone curves.
 TP_EXPOSURE_EXPCOMP;Exposure compensation
+TP_EXPOSURE_HISTMATCHING;Auto-matched Tone Curve
 TP_EXPOSURE_LABEL;Exposure
 TP_EXPOSURE_SATURATION;Saturation
 TP_EXPOSURE_TCMODE_FILMLIKE;Film-like
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -1554,6 +1554,7 @@ TP_EXPOSURE_CURVEEDITOR2;Tone curve 2
 TP_EXPOSURE_CURVEEDITOR2_TOOLTIP;Please refer to the "Exposure > Tone Curves" RawPedia article to learn how to achieve the best results by using two tone curves.
 TP_EXPOSURE_EXPCOMP;Exposure compensation
 TP_EXPOSURE_HISTMATCHING;Auto-matched Tone Curve
+TP_EXPOSURE_HISTMATCHING_TOOLTIP;Automatically adjust sliders and curves (except exposure compensation) to match the look of the embedded JPEG thumbnail.
 TP_EXPOSURE_LABEL;Exposure
 TP_EXPOSURE_SATURATION;Saturation
 TP_EXPOSURE_TCMODE_FILMLIKE;Film-like
--- a/rtengine/CMakeLists.txt
+++ b/rtengine/CMakeLists.txt
@@ -116,6 +116,7 @@ set(RTENGINESOURCEFILES
     rtlensfun.cc
     tmo_fattal02.cc
     iplocalcontrast.cc
+    histmatching.cc
     )
 
 if(LENSFUN_HAS_LOAD_DIRECTORY)
--- a/rtengine/histmatching.cc
+++ b/rtengine/histmatching.cc
@@ -25,6 +25,7 @@
 #include "rt_math.h"
 #include "iccstore.h"
 #include "../rtgui/mydiagonalcurve.h"
+#include "improcfun.h"
 
 
 namespace rtengine {
@@ -54,7 +55,7 @@ std::vector<int> getCdf(const IImage8 &img)
 int findMatch(int val, const std::vector<int> &cdf, int j)
 {
     if (cdf[j] <= val) {
-        for (; j < cdf.size(); ++j) {
+        for (; j < int(cdf.size()); ++j) {
             if (cdf[j] == val) {
                 return j;
             } else if (cdf[j] > val) {
@@ -123,27 +124,39 @@ void mappingToCurve(const std::vector<int> &mapping, std::vector<double> &curve)
 void RawImageSource::getAutoMatchedToneCurve(std::vector<double> &outCurve)
 {
     const int rheight = 200;
-    RawMetaDataLocation rml;
-    eSensorType sensor_type;
-    int w, h;
     ProcParams neutral;
+    std::unique_ptr<IImage8> target;
+    { 
+        int tr = TR_NONE;
+        int fw, fh;
+        getFullSize(fw, fh, tr);
+        int skip = fh / rheight;
+        PreviewProps pp(0, 0, fw, fh, skip);
+        ColorTemp currWB = getWB();
+        std::unique_ptr<Imagefloat> image(new Imagefloat(int(fw / skip), int(fh / skip)));
+        neutral.raw.bayersensor.method = procparams::RAWParams::BayerSensor::getMethodString(procparams::RAWParams::BayerSensor::Method::FAST);
+        neutral.raw.xtranssensor.method = procparams::RAWParams::XTransSensor::getMethodString(procparams::RAWParams::XTransSensor::Method::FAST);
+        getImage(currWB, tr, image.get(), pp, neutral.toneCurve, neutral.raw);
+
+        // this could probably be made faster -- ideally we would need to just
+        // perform the transformation from camera space to the output space
+        // (taking gamma into account), but I couldn't find anything
+        // ready-made, so for now this will do. Remember the famous quote:
+        // "premature optimization is the root of all evil" :-)
+        convertColorSpace(image.get(), neutral.icm, currWB);
+        ImProcFunctions ipf(&neutral);
+        LabImage tmplab(image->getWidth(), image->getHeight());
+        ipf.rgb2lab(*image, tmplab, neutral.icm.working);
+        image.reset(ipf.lab2rgbOut(&tmplab, 0, 0, tmplab.W, tmplab.H, neutral.icm));
+        target.reset(image->to8());
+    }
     std::unique_ptr<IImage8> source;
     {
+        RawMetaDataLocation rml;
+        eSensorType sensor_type;
+        int w, h;
         std::unique_ptr<Thumbnail> thumb(Thumbnail::loadQuickFromRaw(getFileName(), rml, sensor_type, w, h, 1, false, true));
-        source.reset(thumb->quickProcessImage(neutral, rheight, TI_Nearest));
-    }
-    std::unique_ptr<IImage8> target;
-    if (true) {
-        neutral.icm.working = "RT_sRGB";
-        // faster, but has problems likely due to color space transformations that I do not properly understand...
-        double scale;
-        std::unique_ptr<Thumbnail> thumb(Thumbnail::loadFromRaw(getFileName(), rml, sensor_type, w, h, 1, 0.0, false));
-        target.reset(thumb->processImage(neutral, sensor_type, rheight, TI_Nearest, getMetaData(), scale, false));
-    } else {
-        ProcessingJob *job = ProcessingJob::create(this, neutral, true);
-        int err = 0;
-        std::unique_ptr<IImagefloat> tmp(processImage(job, err, nullptr, false));
-        target.reset(static_cast<Imagefloat *>(tmp.get())->to8());
+        source.reset(thumb->quickProcessImage(neutral, target->getHeight(), TI_Nearest));
     }
     if (target->getWidth() != source->getWidth() || target->getHeight() != source->getHeight()) {
         Image8 *tmp = new Image8(source->getWidth(), source->getHeight());
--- a/rtengine/histmatching.cc
+++ b/rtengine/histmatching.cc
@@ -96,31 +96,35 @@ void mappingToCurve(const std::vector<int> &mapping, std::vector<double> &curve)
 
     auto coord = [](int v) -> double { return double(v)/255.0; };
     auto doit =
-        [&](int start, int stop, int step) -> void
+        [&](int start, int stop, int step, bool addstart) -> void
         {
             int prev = start;
+            if (addstart) {
+                curve.push_back(coord(start));
+                curve.push_back(coord(mapping[start]));
+            }
             for (int i = start; i < stop; ++i) {
                 int v = mapping[i];
                 bool change = i > 0 && v != mapping[i-1];
                 int diff = i - prev;
                 if (change && std::abs(diff - step) <= 1) {
-                    curve.emplace_back(coord(i));
-                    curve.emplace_back(coord(v));
+                    curve.push_back(coord(i));
+                    curve.push_back(coord(v));
                     prev = i;
                 }
             }
         };
-    doit(0, idx, idx > step ? step : idx / 2);
-    doit(idx, int(mapping.size()), step);
-    if (curve[1] > 0.01) {
-        curve.insert(curve.begin(), 0.0);
-        curve.insert(curve.begin(), 0.0);
-    }
-    if (curve.back() < 0.99 || (1 - curve[curve.size()-2] > step / 512.0 && curve.back() < coord(mapping.back()))) {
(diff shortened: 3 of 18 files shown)


Beep6581:
The Rosetta Stone for anyone wanting to learn how to convert Python to C++ while learning RT internals at the same time.

Floessie:
Very nice, Alberto! 👍 I only had a look at the diff yet. Here are my remarks:

- Shouldn't the `ppVersion` be incremented?
- There's an [unchecked pointer](https://github.com/Beep6581/RawTherapee/compare/histogram-matching#diff-846023f8f7705bc76e1d3bb073932543R170). Can't say if it's relevant, but makes me nervous. 😉 
- `std::container<T>::emplace*()` is used with already constructed objects, which is okay but silently decays to an `insert()`/`push_back()`. I'd rather like the latter in those places, so that it obvious, what's going on.
- During optimization the fixed size `std::vector<>`s could be converted to `std::array<>`s.

HTH,
Flössie

agriggio:
Hi @Floessie, thanks for the review!

> Shouldn't the ppVersion be incremented?

I don't think so, this is fully backwards compatible. If the entry is not found in the pp3, no histogram matching will be done.

> There's an unchecked pointer.

Thanks, I'll fix that.

> std::container<T>::emplace*() is used with already constructed objects, which is okay but silently decays to an insert()/push_back(). I'd rather like the latter in those places, so that it obvious, what's going on.

> During optimization the fixed size std::vector<>s could be converted to std::array<>s.

Ok, those are sort of immaterial though.

Floessie:
> I don't think so, this is fully backwards compatible. If the entry is not found in the pp3, no histogram matching will be done.

Okay, I didn't really understand [*modified*](https://github.com/Beep6581/RawTherapee/blob/fb8148acaba28de27521b8480e7fe8f4e494e1e4/rtgui/ppversion.h#L3) in the comment in `ppversion.h`. You're right, I think.

> Ok, those are sort of immaterial though.

Yep. The first one helps review, the latter lowers the allocation overhead a bit.

agriggio:
Here are a couple of TODOs (besides testing and speedups where possible):

1. make sure that the thumbnails get converted to sRGB. This is typically the case, but some cameras allow to use also AdobeRGB I think. The matching will be obviously wrong in these cases

2. deal with thumbnails that have a different aspect ratio than the full raw image. It should be sufficient to simply crop the target before computing its histogram

2 is easy, but suggestions are welcome for 1

heckflosse:
Histogram matching on http://www.rawtherapee.com/shared/test_images/crash_kodak_pb645m.dcr segfaults. Most likely because of the unchecked pointer mentioned by @Floessie

agriggio:
@heckflosse thanks, should be fixed now (and also the other bug you reported in private)

heckflosse:
@agriggio Fixes confirmed. Thank you. One minor issue: When there is no emedded jpeg, no curve is generated, good. But then the curve type should also stay as before. Currently it's set to film-like in that case.

agriggio:
@heckflosse I'm not really sure what the correct behaviour would be, I suppose you can argue both ways...  the current way is simpler, that's why I picked it. But if you prefer I can change it.

heckflosse:
@agriggio Let's wait for comments of other contributors

agriggio:
quick update: I've implemented point 2 above. For point 1, I've performed a quick test and it seems that even if the thumbnail is AdobeRGB the match is pretty good. Of course more is needed, but just in case the raw is here (I've also set the thumbnail to be 16:9 so you can test that this is handled properly as well):
https://filebin.net/dry5zsoahejy464s/_DSC9477.ARW

sguyader:
I tested with files form Fujifilm X-T1 and X-T2, no issues. Great job!

heckflosse:
@agriggio I just found one small issue. To reproduce:

1) clear processing profile from a raw in file browser
2) batch apply `Auto-matched Tone Curve` => works fine (correctly updates the thumb)
3) set processing profile to neutral in file browser
4) batch apply `Auto-matched Tone Curve` => does not work (does not update the thumb)

Imho it should either work or not work for both cases.

agriggio:
@heckflosse histogram matching should never update the thumb in batch mode... I'll see what is going on

agriggio:
@heckflosse either I don't understand what you mean (quite possible), or I cannot reproduce...

agriggio:
@Beep6581 @heckflosse @Hombre57 @Desmis @Floessie do we want this for 5.4 or do you prefer to wait? I know it's a bit late, so if we don't reach 100% consensus I have no problem postponing it...

heckflosse:
@agriggio 👍 for 5.4

When a raw file has no processing profile, clicking the auto matching button in batch updates the thumb, while it does not when a profile already is assigned

agriggio:
@heckflosse the thumbnail gets updated because a pp3 is created, but the thumbnail does not reflect the effect of histogram matching -- in other words, it should be the same as when you go from "no pp3" to "neutral". Can you confirm this? If so, this is the expected behaviour (and that's part of why I wrote at the beginning that batch editing is a bit weird... :-)

heckflosse:
@agriggio I was wrong. It's unrelated to histogram matching, which makes it even worse.
I will open a seperate issue.

heckflosse:
@agriggio It is related to histogram matching

1) set your default profile to a profile which has `Auto levels` enabled
2) clear processing profile from a raw in file browser
3) batch apply `Auto-matched Tone Curve` => as you can see, `Auto levels` in batch edit is still enabled and the thumb is updated (looks like `Auto levels` is applied to the thumb).
4) open the file in editor => Auto matched tone curve is applied, but `Auto levels` is not applied, which in fact is good, but confused me a lot and also may confuse other users.

Edit:

5) close the editor and open the file again in editor. Now auto levels is applied.

There clearly is something wrong

Beep6581:
:+1: for 5.4

Desmis:
👍 for 5.4

heckflosse:
This patch fixes the issue I mentioned above:

```diff
diff --git a/rtengine/improccoordinator.cc b/rtengine/improccoordinator.cc
index 75fc81e9..e470aa48 100644
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -450,6 +450,7 @@ void ImProcCoordinator::updatePreviewImage (int todo, Crop* cropCall)
             imgsrc->getAutoMatchedToneCurve(params.toneCurve.curve);
 
             params.toneCurve.histmatching = false;
+            params.toneCurve.autoexp = false;
             params.toneCurve.curveMode = ToneCurveParams::TcMode::FILMLIKE;
             params.toneCurve.curve2 = { 0 };
             params.toneCurve.expcomp = 0.0;

```

Floessie:
👍 for 5.4, as this is something really useful. I didn't have the time to check it yet (and it doesn't look as if time will permit on the weekend) but maybe next week.

Hombre57:
@agriggio 👍 for 5.4

Hombre57:
@agriggio 

> > Shouldn't the ppVersion be incremented?

> I don't think so, this is fully backwards compatible. If the entry is not found in the pp3, no histogram matching will be done.

I don't think we should bump version only if necessary, but also when something is updated in its possible content, i.e. adding new stuff. So I'd bump it.

agriggio:
@Floessie @Hombre57 ok, no problem to bump the version (I still think it's not necessary, but it's definitely not a problem :-)

Desmis:
@agriggio 
This morning, I tested with numerous files
* generaly result is good
* for some files as : amsterdam_moving_boat_1.pef, or london_bridge_moving_1.pef, the result is underexposed - histogram is shift to the left

Jacques

kazah7:
Hi, what about this kind of behaviour? 
chorwacja
 
https://filebin.net/7sxqpfow0sbp23nj

heckflosse:
@kazah7 Please compare with inspector view which shows the embedded jpeg. Does it look similar?

kazah7:
@heckflosse Yes, looks the same..

heckflosse:
@kazah7 Then it works as designed :)

heckflosse:
@Desmis Jacques, I had a look at amsterdam_moving_boat_1.pef
The embedded jpeg looks the same as the raw with auto matched tone curve.

Desmis:
@heckflosse 

OK, I see :)
But in this case(s) what's the goal of such tool ?

Beep6581:
To make the image in RT look similar to the out-of-camera JPEG.

Desmis:
I must have missed a step, but what is the interest - if not intellectual, if it leads to a very unworkable image.

Probably, In Rawprocess, by multipliing by "gain", or applying a special gamma, we can have a "normal" image :)
But I admit, that these cases are rare.

Beep6581:
Many people new to raw shooting want to be able to use an image similar to what their camera produced as a starting point. This feature makes that possible.

Desmis:
OK, but _a minima_, we shall write something to  indicate the principle of this operation

Beep6581:
Yes, all tools must be documented.

agriggio:
Since we have consensus (thanks to all!) I'll merge later today. Here a couple of answers:

> what's the goal of such tool ?
> what is the interest - if not intellectual, if it leads to a very unworkable image.

As @Beep6581 said:

> Many people new to raw shooting want to be able to use an image similar to what their camera produced as a starting point.

But also I find this to be true:

> I tested with numerous files
> - generaly result is good

IOW, jpeg engines of modern cameras are good IMHO, so since we have the information we should exploit it! (Of course they can still get the WB off, they might use poor demosaicing, too aggressive noise reduction, and so on... and that's why we shoot raw :-)

> OK, but a minima, we shall write something to indicate the principle of this operation

Definitely agree! In fact, we should probably revise this:

http://rawpedia.rawtherapee.com/The_Image_Editor_Tab#Eek.21_My_Raw_Photo_Looks_Different_than_the_Camera_JPEG

Desmis:
ok, but the title and content of this "pull request" do not predict such a goal. 
Where then you have to be diviner  :)

But I approved.

jacques

agriggio:
sorry Jacques for the confusion, but I think the title is accurate,  and when I said "popular demand" in the description I was (ironically) referring to this thread:

https://discuss.pixls.us/t/profile-application-after-opening-file/6394/8?u=agriggio

if you don't like this and want to take back your vote, please let me know and I will not merge... I won't do anything before hearing back from you :-)

Desmis:
Alberto
No problem you can merge :)

It is a good purpose.
But probably my english is so bad that I dont understand.
Next time in Latin or in french
Jacques

Beep6581:
Alea jacta est!

heckflosse:
grafik
 😁

Desmis:
@heckflosse 

I am an old man, but for now, I have not a wooden leg :)

Desmis:
@agriggio 

I have not seen this thread...I don't look all the threads....for all "Software, Processing, Digital painting, etc."
For those who participated in the thread, the title is obvious, for others ...

It the tittle was: "Histogram matching with settings coming from embedded JPG", it would be clearer

And probably, If for "Auto-matched Tone Curve" button, a tooltip would make the use clearer. Which tone  Curve?, to do what ?

agriggio:
Jacques, I already apologized for the confusion. I have no problem doing it again, unless you want me to go through the caudine forks (yes, I'm in a classical mood ;-)

you are absolutely right about the tooltip, I'll add one! :+1:

Desmis:
@agriggio 

Alberto
No problem at all :)

Floessie:
@agriggio I'm having a session with this branch right now. **THIS IS JUST GREAT! YAY!** I'm having a new camera with a supported DCP. And now this makes developing a breeze. Just some slight tweaks to the matched curve, occasionally TC2, DCP look table, HLR, defringe, sharpening, and I'm done. 🎉 

This was the missing piece!!!

Enthusiastic,
Flössie

Desmis:
I have made some test with my (old) files in Maldives, with my Sony NEX-6 (jpg)

The 2 images seems "identical".
Very good job :)

Jacques

heckflosse:
I would like to suggest a change.
For underexposed images I always get a curve like this (example is from a -1 ev shot):

grafik

That's fine if I don't change exposure. But when I change exposure to e.g. +1 I also have to move the last curve point up to 1 to avoid losing (compressing) highlights. That makes using the tool in batch almost impossible for underexposed shots
Why don't we always set the last curve point to 1/1?

Edit: [Here's a file](https://filebin.net/f3mg4q2ooa117sbn/D700_20150725_6496.NEF) to reproduce

Edit 2: It does not make using the tool in batch mode in combination with exposure settings unusable, because that doesn't work per design. Why is exposure set back to zero when histogram matching is enabled?

heckflosse:
@agriggio How about this patch to fix the issues I mentioned above?

```diff
diff --git a/rtengine/histmatching.cc b/rtengine/histmatching.cc
index 5b7647f4..be618d4d 100644
--- a/rtengine/histmatching.cc
+++ b/rtengine/histmatching.cc
@@ -119,7 +119,7 @@ void mappingToCurve(const std::vector<int> &mapping, std::vector<double> &curve)
     doit(idx, int(mapping.size()), step, idx - step > step / 2);
     if (curve.size() <= 2 || curve.back() < 0.99 || (1 - curve[curve.size()-2] > step / 512.0 && curve.back() < coord(mapping.back()))) {
         curve.emplace_back(1.0);
-        curve.emplace_back(coord(mapping.back()));
+        curve.emplace_back(1.0);
     }
     if (curve.size() < 4) {
         curve = { DCT_Linear }; // not enough points, fall back to linear
diff --git a/rtengine/improccoordinator.cc b/rtengine/improccoordinator.cc
index d390d94c..d3a32140 100644
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -449,10 +449,13 @@ void ImProcCoordinator::updatePreviewImage (int todo, Crop* cropCall)
         if (params.toneCurve.histmatching) {
             imgsrc->getAutoMatchedToneCurve(params.toneCurve.curve);
 
+            if (params.toneCurve.autoexp) {
+                params.toneCurve.expcomp = 0.0;
+            }
+
             params.toneCurve.autoexp = false;
             params.toneCurve.curveMode = ToneCurveParams::TcMode::FILMLIKE;
             params.toneCurve.curve2 = { 0 };
-            params.toneCurve.expcomp = 0.0;
             params.toneCurve.brightness = 0;
             params.toneCurve.contrast = 0;
             params.toneCurve.black = 0;
diff --git a/rtengine/simpleprocess.cc b/rtengine/simpleprocess.cc
index 0cc223da..45956f26 100644
--- a/rtengine/simpleprocess.cc
+++ b/rtengine/simpleprocess.cc
@@ -743,14 +743,18 @@ private:
         if (params.toneCurve.histmatching) {
             imgsrc->getAutoMatchedToneCurve(params.toneCurve.curve);
 
+            if (params.toneCurve.autoexp) {
+                params.toneCurve.expcomp = 0.0;
+            }
+
             params.toneCurve.autoexp = false;
             params.toneCurve.curveMode = ToneCurveParams::TcMode::FILMLIKE;
             params.toneCurve.curve2 = { 0 };
-            params.toneCurve.expcomp = 0.0;
             params.toneCurve.brightness = 0;
             params.toneCurve.contrast = 0;
             params.toneCurve.black = 0;
             params.toneCurve.hlcompr = 0;
+
         }        
 
         // at this stage, we can flush the raw data to free up quite an important amount of memory
diff --git a/rtgui/tonecurve.cc b/rtgui/tonecurve.cc
index 4a25794f..fd4c0827 100644
--- a/rtgui/tonecurve.cc
+++ b/rtgui/tonecurve.cc
@@ -936,7 +936,6 @@ bool ToneCurve::histmatchingComputed()
     GThreadLock lock;
     disableListener();
     enableAll();
-    expcomp->setValue(0);
     brightness->setValue(0);
     contrast->setValue(0);
     black->setValue(0);
@@ -947,6 +946,7 @@ bool ToneCurve::histmatchingComputed()
     }
 
     if (autolevels->get_active() ) {
+        expcomp->setValue(0);
         autoconn.block(true);
         autolevels->set_active(false);
         autoconn.block(false);

```

agriggio:
ok for the expcomp part, but setting the last point to be 1,1 doesn't always work in my tests...

heckflosse:
@agriggio Can you share a file where 1,1 for last point does not work? Maybe we find a better solution

agriggio:
@heckflosse I'll have to redo some tests, unfortunately I didn't keep a record of this. btw one solution (that I would very much like) could be to have our curves behave like those of GIMP. they are easier to edit IMHO, and less sensitive to outliers. I haven't checked the math yet but would like to...

agriggio:
@heckflosse actually, I just realized that the bad results that I mentioned were with my python script, not with the RT implementation. I have since improved the curve generation code, so it might be that now the issue is gone. So, :+1: for me to go with your proposal, and if we find some problematic shots, we can then start digging deeper. What do you think?

Desmis:
@agrigio @heckflosse 
I have made numerous test, and I think (for non big underexposed picture), that a good solution is to use the second curve "Tone curve 2". with "Control cage" and act on low light.

For example with "amsterdam_moving_boat_1.pef", I obtain an image I think better than RT by default
Same things for other under exposed images.

Do you think, or no,  to "guide" users, in Rawpedia documentation.

heckflosse:
@agriggio Ok, I will push my changes now

Desmis:
I just tested your changes on a very underexposed images, and now we can used correctly "Expcomp"

No differences for others under exposed images

I think, take into account "auto level", leeds in some cases at over exposed  for high values of luminance

:)

Beep6581:
@agriggio https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/histmatching.cc#L136

heckflosse:
@agriggio No objection from me to keep the BENCHFUN, but #define BENCHMARK should be removed as it [now](https://github.com/Beep6581/RawTherapee/commit/11ad1dace8458a4373f74f3e37d98f2b8e56e255) can be set in cmake command line

agriggio:
sorry, I just forgot to remove this. please go ahead

PR #4324, opened 2018-01-23
Removed most of the __x86_64__ dependencies. Also removed _RT_NESTED_…

heckflosse:
…OPENMP

### DIFF
--- a/rtengine/EdgePreservingDecomposition.cc
+++ b/rtengine/EdgePreservingDecomposition.cc
@@ -291,7 +291,7 @@ bool MultiDiagonalSymmetricMatrix::LazySetEntry(float value, int row, int column
     return true;
 }
 
-SSEFUNCTION void MultiDiagonalSymmetricMatrix::VectorProduct(float* RESTRICT Product, float* RESTRICT x)
+void MultiDiagonalSymmetricMatrix::VectorProduct(float* RESTRICT Product, float* RESTRICT x)
 {
 
     int srm = StartRows[m - 1];
@@ -693,7 +693,7 @@ EdgePreservingDecomposition::~EdgePreservingDecomposition()
     delete A;
 }
 
-SSEFUNCTION float *EdgePreservingDecomposition::CreateBlur(float *Source, float Scale, float EdgeStopping, int Iterates, float *Blur, bool UseBlurForEdgeStop)
+float *EdgePreservingDecomposition::CreateBlur(float *Source, float Scale, float EdgeStopping, int Iterates, float *Blur, bool UseBlurForEdgeStop)
 {
 
     if(Blur == nullptr)
@@ -884,7 +884,7 @@ float *EdgePreservingDecomposition::CreateIteratedBlur(float *Source, float Scal
     return Blur;
 }
 
-SSEFUNCTION void EdgePreservingDecomposition::CompressDynamicRange(float *Source, float Scale, float EdgeStopping, float CompressionExponent, float DetailBoost, int Iterates, int Reweightings)
+void EdgePreservingDecomposition::CompressDynamicRange(float *Source, float Scale, float EdgeStopping, float CompressionExponent, float DetailBoost, int Iterates, int Reweightings)
 {
     if(w < 300 && h < 300) { // set number of Reweightings to zero for small images (thumbnails). We could try to find a better solution here.
         Reweightings = 0;
--- a/rtengine/FTblockDN.cc
+++ b/rtengine/FTblockDN.cc
@@ -474,7 +474,7 @@ void ImProcFunctions::Tile_calc(int tilesize, int overlap, int kall, int imwidth
 int denoiseNestedLevels = 1;
 enum nrquality {QUALITY_STANDARD, QUALITY_HIGH};
 
-SSEFUNCTION void ImProcFunctions::RGB_denoise(int kall, Imagefloat * src, Imagefloat * dst, Imagefloat * calclum, float * ch_M, float *max_r, float *max_b, bool isRAW, const procparams::DirPyrDenoiseParams & dnparams, const double expcomp, const NoiseCurve & noiseLCurve, const NoiseCurve & noiseCCurve, float &nresi, float &highresi)
+void ImProcFunctions::RGB_denoise(int kall, Imagefloat * src, Imagefloat * dst, Imagefloat * calclum, float * ch_M, float *max_r, float *max_b, bool isRAW, const procparams::DirPyrDenoiseParams & dnparams, const double expcomp, const NoiseCurve & noiseLCurve, const NoiseCurve & noiseCCurve, float &nresi, float &highresi)
 {
 BENCHFUN
 //#ifdef _DEBUG
@@ -783,7 +783,7 @@ BENCHFUN
                 numthreads = MIN(numthreads, options.rgbDenoiseThreadLimit);
             }
 
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
             denoiseNestedLevels = omp_get_max_threads() / numthreads;
             bool oldNested = omp_get_nested();
 
@@ -916,7 +916,7 @@ BENCHFUN
 
                             if (!denoiseMethodRgb) { //lab mode
                                 //modification Jacques feb 2013 and july 2014
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 #pragma omp parallel for num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
 
@@ -964,7 +964,7 @@ BENCHFUN
                                     }
                                 }
                             } else {//RGB mode
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 #pragma omp parallel for num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
 
@@ -999,7 +999,7 @@ BENCHFUN
                                 }
                             }
                         } else {//image is not raw; use Lab parametrization
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                             #pragma omp parallel for num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
 
@@ -1159,7 +1159,7 @@ BENCHFUN
                             if (!memoryAllocationFailed) {
                                 // precalculate madL, because it's used in adecomp and bdecomp
                                 int maxlvl = Ldecomp->maxlevel();
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 #pragma omp parallel for schedule(dynamic) collapse(2) num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
 
@@ -1284,7 +1284,7 @@ BENCHFUN
                                         if (!memoryAllocationFailed) {
                                             // copy labdn->L to Lin before it gets modified by reconstruction
                                             Lin = new array2D<float>(width, height);
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                             #pragma omp parallel for num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
 
@@ -1337,14 +1337,14 @@ BENCHFUN
                                     }
                                 }
 
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 int masterThread = omp_get_thread_num();
 #endif
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 #pragma omp parallel num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
                                 {
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                     int subThread = masterThread * denoiseNestedLevels + omp_get_thread_num();
 #else
                                     int subThread = 0;
@@ -1354,7 +1354,7 @@ BENCHFUN
                                     float *fLblox = fLbloxArray[subThread];
                                     float pBuf[width + TS + 2 * blkrad * offset] ALIGNED16;
                                     float nbrwt[TS * TS] ALIGNED64;
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                     #pragma omp for
 #endif
 
@@ -1454,7 +1454,7 @@ BENCHFUN
                                 }
                                 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 #pragma omp parallel for num_threads(denoiseNestedLevels) if (denoiseNestedLevels>1)
 #endif
 
@@ -1591,7 +1591,7 @@ BENCHFUN
                                     realred /= 100.f;
                                     realblue /= 100.f;
 
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                     #pragma omp parallel for schedule(dynamic,16) num_threads(denoiseNestedLevels)
 #endif
 
@@ -1641,7 +1641,7 @@ BENCHFUN
                                         }
                                     }
                                 } else {//RGB mode
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                     #pragma omp parallel for num_threads(denoiseNestedLevels)
 #endif
 
@@ -1681,7 +1681,7 @@ BENCHFUN
 
                                 }
                             } else {
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
                                 #pragma omp parallel for num_threads(denoiseNestedLevels)
 #endif
 
@@ -1748,7 +1748,7 @@ BENCHFUN
                 }
             }
 
-#ifdef _RT_NESTED_OPENMP
+#ifdef _OPENMP
             omp_set_nested(oldNested);
 #endif
 
@@ -2043,7 +2043,7 @@ BENCHFUN
 //%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 
 
-SSEFUNCTION void ImProcFunctions::RGBtile_denoise(float * fLblox, int hblproc, float noisevar_Ldetail, float * nbrwt, float * blurbuffer)  //for DCT
+void ImProcFunctions::RGBtile_denoise(float * fLblox, int hblproc, float noisevar_Ldetail, float * nbrwt, float * blurbuffer)  //for DCT
 {
     int blkstart = hblproc * TS * TS;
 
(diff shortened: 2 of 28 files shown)


Beep6581:
Compiles and runs fine on x86_64/GCC-5.4.0.

heckflosse:
@Beep6581 Thanks for testing 👍

Floessie:
Compiles and runs fine on openSUSE Leap 42.1 x86_64 (GCC 5.3.1).

Floessie:
Compiles and runs on Debian Testing i386 with SSE2 hardware (GCC 7.2.0).

Alas, the colors are [wrong](https://filebin.net/7pxrfbeavrkz8lwv).

Floessie:
Compiles and runs on Debian Testing i386 **without** SSE2 hardware (GCC 7.2.0).

The colors are only wrong when using the shipped Pentax K-1 DCP with "base table".

Floessie:
@heckflosse Scratch my comments above: I was testing `dev`. :facepalm:

Floessie:
Okay, compiles and runs fine on Debian Testing i386 with and without `__SSE2__` (GCC 7.2.0). Sadly, `qemu-system-i386` doesn't trap on SSE2 instructions even with `-M pentium3` and `/proc/cpuinfo` not listing `sse2`.

However, the problem with the DCP base table remains:
- Debian Testing AMD64 `dev` ✅ 
- Debian Testing i386 `dev` :x:
- Debian Testing i386 `remove_x86_64_dependencies` ❌ 

HTH,
Flössie

heckflosse:
@Floessie What does the X mean?

Floessie:
@heckflosse ❌ means wrong colors with DCP base table.

I've checked out the 5.3 tag, and there we don't have that problem. So definitely a regression. I'll try to bisect, though this VM is slow.

Best,
Flössie

heckflosse:
@Floessie with our without SSE2?

Floessie:
@heckflosse Doesn't matter.

Really slow machine. ~5 bisect steps left...

Floessie:
@heckflosse It's really unlikely, but bisect revealed 0498048ca05078cfe07636bb636629a80e3890e9. I'll give it a try with `dev` now...

Floessie:
@heckflosse Yep, it's the `LIKELY()`.

heckflosse:
@Floessie Thanks for bisecting 👍 
What dou you suggest? Remove the LIKELY() at that place or just disable LIKELY() in opthelper.h for 32bit builds? Maybe also report upstream?

heckflosse:
@Floessie does this work?

```diff
diff --git a/rtengine/dcp.cc b/rtengine/dcp.cc
index 82ea35f0..c5d86cf0 100644
--- a/rtengine/dcp.cc
+++ b/rtengine/dcp.cc
@@ -1074,7 +1074,8 @@ void DCPProfile::apply(
                 float s;
                 float v;
 
-                if (LIKELY(Color::rgb2hsvdcp(newr, newg, newb, h , s, v))) {
+                bool apply = Color::rgb2hsvdcp(newr, newg, newb, h , s, v);
+                if (LIKELY(apply)) {
 
                     hsdApply(delta_info, delta_base, h, s, v);
 
```

Floessie:
No, that was my idea, too, but it doesn't work. Next, I'll remove the `inline`...

Floessie:
@heckflosse 
> What dou you suggest? Remove the LIKELY() at that place or just disable LIKELY() in opthelper.h for 32bit builds? Maybe also report upstream?

I'd disable `LIKELY()` for 32bit builds. It's not immediately clear what triggers the bug in this place, so others might be affected, too. But it might as well be a qemu problem. Maybe we should ask someone with a real 32bit installation.

About upstreaming: I've not yet found a bug report that matches, so we should probably report it once we've reproduced it on real hardware.

Build finished: Removing `inline` did not help.

Next: Older compiler, if available...

heckflosse:
@Floessie Ok, then this patch for branch `remove_x86_64_dependencies` should fix it:

```diff
diff --git a/rtengine/opthelper.h b/rtengine/opthelper.h
index ce1f620e..ce69b03d 100644
--- a/rtengine/opthelper.h
+++ b/rtengine/opthelper.h
@@ -30,8 +30,13 @@
 
     #ifdef __GNUC__
         #define RESTRICT    __restrict__
-        #define LIKELY(x)   __builtin_expect (!!(x), 1)
-        #define UNLIKELY(x) __builtin_expect (!!(x), 0)
+        #if __SIZEOF_POINTER__ == 8 // there seems to be a bug with __builtin_expect on 32bit systems
+            #define LIKELY(x)   __builtin_expect (!!(x), 1)
+            #define UNLIKELY(x) __builtin_expect (!!(x), 0)
+        #else
+            #define LIKELY(x)    (x)
+            #define UNLIKELY(x)  (x)
+        #endif
         #define ALIGNED64 __attribute__ ((aligned (64)))
         #define ALIGNED16 __attribute__ ((aligned (16)))
     #else

```

Floessie:
@heckflosse 🎉 Doesn't happen with GCC 6.4! AFAIK GCC 7.3 is set for release this week. I could compile it to test for our problem then. It's often better to report against a vanilla GCC. In the meantime you should disable `LIKELY()` for GCC >= 7 on 32bit.

heckflosse:
@Floessie Done for `dev` and `remove_x86_64_dependencies`

Floessie:
Fix confirmed! Thanks! 👍

heckflosse:
Any objections to merge?

Floessie:
Not from me. At least under Linux with a recent (but not brand-new 😉) GCC no apparent problems with or without SSE2 on 32bit x86.

Floessie:
GCC 7.3.0 is out, so I built it:

gcc730

The error is still present.

@Beep6581 AFAIK, you have a GCC bugzilla account. Would you report the bug upstream?

Beep6581:
@Floessie I do not have a [GNU bugzilla](https://gcc.gnu.org/bugzilla/) account.

Floessie:
GCC bugzilla account creation is restricted ATM, I'm attending for an account via email...

Floessie:
[Bug filed.](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=84077)

Floessie:
Jakub Jelinek is looking at our problem ATM. Cross your fingers!

Floessie:
So, we have an analysis:

> The problem goes away even in the original unmodified dcp.ii with -msse2 -mfpmath=sse or with -ffloat-store, which means most likely the source just relies on excess precision not happening.
If so, this is a dup of PR323.  To make i?86 32-bit code use no excess precision in float and double, one needs to use -msse2 -mfpmath=sse, or to get rid of it in most cases while paying in performance -ffloat-store, otherwise some computations can be done in the 80-bit precision and rounding can happen whenever something needs to be forced out of the i387 floating point stack.
Instead of -ffloat-store, you could also insert optimization barriers in certain spots (e.g. using inline asm and for FLT_EVAL_METHOD non-zero only) where you really require rounding to float or double rather than allowing excess precision.
> In any case, most likely not a bug on the GCC side.

Shall we try to find the wrong spot, or shall we keep it as is?

Floessie:
A final hint by Jakub:

> Besides inline asm barriers another option is volatile {float,double} vars in certain spots (to ensure rounding to {float,double} where required).  And again, you don't need to do anything if FLT_EVAL_METHOD is 0, or for double computations also if FLT_EVAL_METHOD is 1.

Floessie:
@heckflosse As already suspected, this patch heals the problem:

```patch
diff --git a/rtengine/color.h b/rtengine/color.h
index 4b865f1d..54d3eb5e 100644
--- a/rtengine/color.h
+++ b/rtengine/color.h
@@ -376,7 +376,7 @@ public:
         if(var_Min < 0.f) {
             return false;
         } else {
-            float var_Max = max(r, g, b);
+            volatile float var_Max = max(r, g, b);
             float del_Max = var_Max - var_Min;
             v = var_Max / 65535.f;
 
```

But going the `-msse2 -mfpmath=sse` way would be much better.

HTH,
Flössie

Beep6581:
@Floessie should `-msse2 -mfpmath=sse` be a requirement for people building on i386?

Floessie:
@Beep6581 As far as I can see, yes. The `LIKELY()` wasn't the problem, but the i387 80 bit math.

Beep6581:
@Floessie and it applies to that hardware regardless of OS?

Floessie:
@Beep6581 Yes, as soon as i387 is used, some internal assumptions are no longer valid, if I understand it right. Ping @heckflosse.

Beep6581:
@Floessie at first I thought we should add this to RawPedia, but it seems better that this goes into the CMake file. Can i387 be targetted via CMake?

heckflosse:
@Floessie Using `-msse2 -mfpmath=sse` won't solve the bug for machines which don't have sse.

How about this instead?

```diff
diff --git a/rtengine/color.h b/rtengine/color.h
index 4b865f1d..660379c1 100644
--- a/rtengine/color.h
+++ b/rtengine/color.h
@@ -377,7 +377,11 @@ public:
             return false;
         } else {
             float var_Max = max(r, g, b);
+#ifndef __SSE__ // see https://github.com/Beep6581/RawTherapee/pull/4324#issuecomment-365214360
+            volatile float del_Max = var_Max - var_Min;
+#else
             float del_Max = var_Max - var_Min;
+#endif
             v = var_Max / 65535.f;
 
             if (fabsf(del_Max) < 0.00001f) {

```

Floessie:
@heckflosse 
> Using -msse2 -mfpmath=sse won't solve the bug for machines which don't have sse.

Maybe we're misunderstanding each other, but here are my thoughts on the problem:
- x87 is unique, as it predates IEEE 754 and internally uses 80 bit registers.
- Results between pure 64 bit and 80 bit representations differ.
- The compiler uses 64 bit FP for compile time constants. Thus there might be visible differences as in our case.
- `-msse2 -mfpmath=sse` is a x86 peculiarity. It should only be used when compiling for x86-32 (it's implicit for x86-64).
- Other current architectures are not affected by x87 80 bit mode. So compiling for anything other than x86-32 should be no problem.

> How about this instead?

The problem lies deeper: How shall we know, where else our code gets broken by this?

Floessie:
@Beep6581 
> Can i387 be targetted via CMake?

The rule should be "if x86 and 32 bit", IMHO.

heckflosse:
@Floessie But doesn't that mean that we don't support x86-32 non sse builds anymore?
Then I'm fine with a CMake fix.

Floessie:
@heckflosse 
> But doesn't that mean that we don't support x86-32 non sse builds anymore?

Yes. Quoting from the [GCC Wiki](https://gcc.gnu.org/wiki/FloatingPointMath#x86note):
> Whether or not intermediate results are rounded to double precision or extended precision depends on optimizations being able to keep values in floating-point registers. The option -ffloat-store prevents GCC from storing floating-point results in registers. While this avoids the indeterministic behavior just described (at great cost), it does not prevent accuracy loss due to double rounding.
> On more modern x86 processors that support SSE2, specifying the compiler options -mfpmath=sse -msse2 ensures all float and double operations are performed in SSE registers and correctly rounded. These options do not affect the ABI and should therefore be used whenever possible for predictable numerical results.

For **very** uncomplaining users, we could optionally use `-ffloat-store` (CMake option or auto-detected), which would enable x86-32-non-SSE builds. 🐌

agriggio:
@heckflosse SSE2 on x86 is from 2001. I'd say it's safe to assume that anyone that wants to use RT on a 32-bit machine has SSE2 these days...

heckflosse:
@agriggio I'm fine with dropping non SSE2 support for x86 machines 👍

Floessie:
Here's a patch for `dev`:

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index fa153855b..cb45c573f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,6 +92,17 @@ set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${PROC_FLAGS}")
 # Stop compilation on typos such as std:swap (missing colon will be detected as unused label):
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror=unused-label")
 
+# Special treatment for x87 and x86-32 SSE (see GitHub issue #4324)
+include(FindX87Math)
+if(HAVE_X87_MATH)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -ffloat-store")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -ffloat-store")
+endif()
+if(HAVE_X86_SSE_MATH)
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse2 -mfpmath=sse")
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -msse2 -mfpmath=sse")
+endif()
+
 if(WIN32)
     # Add additional paths. Look in the MinGW path first, then in the Gtkmm path.
     # If you wish to build some dependent libraries, you have to install them in MinGW to use them:
diff --git a/cmake/modules/FindX87Math.cmake b/cmake/modules/FindX87Math.cmake
new file mode 100644
index 000000000..4e136bac6
--- /dev/null
+++ b/cmake/modules/FindX87Math.cmake
@@ -0,0 +1,54 @@
+# This file is part of RawTherapee.
+#
+# Copyright (C) 2018 Flössie <floessie.mail@gmail.com>
+#
+# RawTherapee is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# RawTherapee is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with RawTherapee.  If not, see <http://www.gnu.org/licenses/>.
+
+include(CheckCXXSourceCompiles)
+
+set(TEST_SOURCE
+"
+#if !defined(__i386) && !defined(_M_IX86)
+#error
+#endif
+
+#if defined(__SSE2__)
+#error
+#endif
+
+int main()
+{
+}
+")
+
+CHECK_CXX_SOURCE_COMPILES("${TEST_SOURCE}" HAVE_X87_MATH)
+
+set(TEST_SOURCE
+"
+#if !defined(__i386) && !defined(_M_IX86)
+#error
+#endif
+
+#if !defined(__SSE2__)
+#error
+#endif
+
+int main()
+{
+}
+")
+
+CHECK_CXX_SOURCE_COMPILES("${TEST_SOURCE}" HAVE_X86_SSE_MATH)
+
+unset(TEST_SOURCE)
diff --git a/rtengine/opthelper.h b/rtengine/opthelper.h
index 711322f56..ce1f620e1 100644
--- a/rtengine/opthelper.h
+++ b/rtengine/opthelper.h
@@ -30,13 +30,8 @@
 
     #ifdef __GNUC__
         #define RESTRICT    __restrict__
-        #if __SIZEOF_POINTER__ == 4 &&  __GNUC__ >= 7 // there seems to be a bug with __builtin_expect on 32bit systems when using gcc >= 7
-            #define LIKELY(x)    (x)
-            #define UNLIKELY(x)  (x)
-        #else
-            #define LIKELY(x)   __builtin_expect (!!(x), 1)
-            #define UNLIKELY(x) __builtin_expect (!!(x), 0)
-        #endif
+        #define LIKELY(x)   __builtin_expect (!!(x), 1)
+        #define UNLIKELY(x) __builtin_expect (!!(x), 0)
         #define ALIGNED64 __attribute__ ((aligned (64)))
         #define ALIGNED16 __attribute__ ((aligned (16)))
     #else
```

HTH,
Flössie

heckflosse:
@Floessie When building on Windows7/64 I get the following, which is correct, but confusing:

```
-- Performing Test HAVE_X86_SSE_MATH
-- Performing Test HAVE_X86_SSE_MATH - Failed
```

You could use `if(CMAKE_SIZEOF_VOID_P EQUAL 4)` to don't check on 64bit

Floessie:
@heckflosse We had to raise `cmake_minimum_required()` then...

heckflosse:
@Floessie Why? We already use that [here](https://github.com/Beep6581/RawTherapee/blob/dev/CMakeLists.txt#L409)

Floessie:
@heckflosse I must correct myself: It's present in CMake 2.8. I tested its value when cross-compiling to PowerPC (not RT) and there it is **unset**.

Floessie:
@heckflosse We could use [`CMAKE_REQUIRED_QUIET`](https://gitlab.kitware.com/cmake/cmake/commit/4c0cc9ab9195d348ea5f1b004789ec334531bbb1) to silence the test output.

innir:
[Sorry, didn't read the complete thread, so my assumption might be wrong :-(]

From what I see you now check at _**compile time**_ for x87 and x86-SSE. I fear this does not work for Linux distributions that provide compiled code to their users (like Debian, Ubuntu, OpenSUSE, Fedora, ...).
They compile software on CPUs that might have different hardware features than the CPUs the user run the software on. I fear with the choice now rawtherapee could crash or give wrong results for users...

Beep6581:
None of those distributions ship such builds, nor would we recommend them doing so.

> It took 8h to compile and 20m to open the RAW

heckflosse:
@innir 

> They compile software on CPUs that might have different hardware features than the CPUs the user run the software on. I fear with the choice now rawtherapee could crash or give wrong results for users...

That would mean, we could also not use `#ifdef __SSE2__`. I doubt this would have remained undetected over the years...

innir:
@heckflosse Sorry for not being clear enough. My concerns are only about the changes of C / CXX flags in CMakeLists.txt.

heckflosse:
@innir Ok, then it's at @Floessie to answer. I'm a cmake noob :)

Floessie:
@innir 
> From what I see you now check at compile time for x87 and x86-SSE. I fear this does not work for Linux distributions that provide compiled code to their users (like Debian, Ubuntu, OpenSUSE, Fedora, ...).
They compile software on CPUs that might have different hardware features than the CPUs the user run the software on. I fear with the choice now rawtherapee could crash or give wrong results for users...

Hi Philip,

The decision whether to use x87 or SSE math on x86-32 can only be made at compile time. After all, it's a completely different register set and ISA. GCC has a feature called ["Function Multiversioning"](https://gcc.gnu.org/wiki/FunctionMultiVersioning) to deal with ISA extensions at run time, but we would have to identify every location in the sources that needs a special treatment. As the effects of x87 rounding and comparison are subtle and hard to grasp, we better cope with it globally at compile time.

As @heckflosse already wrote: The `#ifdef __SSE2__` decisions in the source files are also determined at compile time, and in fact the CMake module utilizes test compiles exactly in the same way as the main compilation, so this is no CMake feature.

I don't know which processor target you choose when building the Debian package of RT for i686, but I suspect it's one with SSE2 support. If so, this binary will SIGILL on a Pentium 3 or earlier, and this has always been the case. If you use `-DPROC_TARGET_NUMBER=3` (which is `-march=pentium`), you get a binary that will run on every i586, albeit slowly, and this, too, has always been the case. Using `-DPROC_TARGET_NUMBER=1` depends on the toolchain and would probably produce a i686 binary without SSE2.

What's new is that when SSE2 is available on the target architecture, it will now be used exclusively for FP math, whereas x87 was used as well before, leading to subtle bugs. As I guess you've always packaged a SSE2 binary, this is beneficial for you and the other packagers. When only x87 is available, we now employ `-ffloat-store` to force 64 bit FP math even with 80 bit internal registers. This, too, aids consistent behavior between the ISAs.

I hope that's coherent. If not, feel free to ask.

HTH,
Flössie

innir:
@Floessie Thanks for the detailed explanation! Things are clearer now. I'd expect nothing to change as we build the binaries for Debian with `-DPROC_FORCED_LABEL=generic`.

Issue #4224, opened 2017-12-12
Detail windows seem broken


Beep6581:
The detail windows show a greatly pixelized image when I zoom in past about 300-400%. In the screenshot you can see that some detail windows already break at 400% zoom while others can still show the preview at 400% correctly. 

Tested using a SONY SLT-A77V raw file from:
https://filebin.net/vko4wznvvaulyz40/DSC00989.ARW
Mirror: https://discuss.pixls.us/t/playraw-presolana-was-salvaging-a-raw/5769/4

Branch: colortab-tools-onoff
`15.3-283-g`11ca61e3

![imgur-2017_12_12-08 36 41](https://user-images.githubusercontent.com/5844619/33872558-9a172d8c-df17-11e7-8a9c-8127bef6bf26.png)
![imgur-2017_12_12-08 40 04](https://user-images.githubusercontent.com/5844619/33872667-16c0c6d6-df18-11e7-8027-c3d5ec508ba2.png)

heckflosse:
Confirmed

Hombre57:
I think that what we see is the coarse image instead of the real preview. Might be connected to the bug of the detail window frozen to the coarse image when it opens (I don't see any opened issue for this).

This might be a race condition somewhere.

Floessie:
@Hombre57 
> Might be connected to the bug of the detail window frozen to the coarse image when it opens (I don't see any opened issue for this).

It was mentioned on [pixls](https://discuss.pixls.us/t/hi-a-couple-questions-suggestions-but-first-therapee-is-awesome/5798/11) lately.

heckflosse:
This fixes the issue when zooming in:

```diff
diff --git a/rtgui/crophandler.cc b/rtgui/crophandler.cc
index f6ff83a7..0abff627 100644
--- a/rtgui/crophandler.cc
+++ b/rtgui/crophandler.cc
@@ -426,11 +426,11 @@ bool CropHandler::getWindow (int& cwx, int& cwy, int& cww, int& cwh, int& cskip)
     cwh = cropH;
 
     // hack: if called before first size allocation the size will be 0
-    if (cww < 10) {
+    if (cww == 0) {
         cww = 10;
     }
 
-    if (cwh < 32) {
+    if (cwh == 0) {
         cwh = 32;
     }
 
```

heckflosse:
@Hombre57 

> Might be connected to the bug of the detail window frozen to the coarse image when it opens

[...]

PR #4336, opened 2018-01-26
Patch up crop updates

Floessie:
There are currently two problems with crop updates (#4224):

1. No update and thus a coarse (internally also called *rough*) display.
2. A deadlock when quickly going back and forth per F3/F4.

These two problems are addressed so far. Another deadlock is pending, which is neither analyzed nor treated yet.

### DIFF
--- a/rtgui/crophandler.cc
+++ b/rtgui/crophandler.cc
@@ -29,19 +29,35 @@
 
 using namespace rtengine;
 
-CropHandler::CropHandler ()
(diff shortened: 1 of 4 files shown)


heckflosse:
Works fine here 👍

agriggio:
Works good here as well. @Floessie, I think you should go ahead and merge :+1:

Floessie:
Merged. Thanks for testing!

PR #4340, opened 2018-01-28
Added new processing profiles #2055

### LINKED ISSUE #2055: Ideas for a default profile to produce predictable "camera-like" output
Originally reported on Google Code with ID 2071

```
Autolevels are today mainly used to make the first open of an image to look good (it's
used in the default profile). I'd prefer if first open should would just look similar
to a standard in-camera jpeg, ie just a tonecurve to brighten and compress highlights,
and that's it.

I've attached how such a tonecurve looks. Not sure if it can be achieved with the Exposure
box sliders. Try open a file, select the neutral profile (make sure highlight recovery
is also at 0 if you don't have bleeding edge when this is fixed), and then apply this
type of curve -- you'll see how very similar it becomes to the camera's JPEG output.

I think autolevels should not be included in any of the bundled profiles, and maybe
not even be saved into the .pp3 when enabled. Instead just seen as a settings "assistant",
you press once and get some values as a starting point for more advanced processing.
I could be convinced otherwise though, but I certainly don't think it's good to have
in the Default profile.

Why do I want this? Because ideally I'd like that you could dump all your raw photos
from a shoot on RT and do nothing and get predictable output. Autolevels is not good
to have in such a situation, it will try to correct an underexposed image, expand contrast
in low contrast conditions etc. A fixed curve (or corresponding) will give you something
that looks like normal "unprocessed" camera output.

It's also what basically all other raw converters do, default profile is something
stable similar to default camera jpegs. A few converters do have an "autolevels" function
(Lightroom has it for example), but it's not something that gets enabled per default,
as the results generally require further tuning by a human operator.

What do you think?
```

Reported by `torger@ludd.ltu.se` on 2013-11-20 19:25:54

Beep6581:
Includes new default processing profile for raw photos.
Closes #2055


agriggio:
How about enabling auto-matched lens corrections (with Lensfun) in the new profiles?

Beep6581:
Yes, agreed.

heckflosse:
@agriggio Agreed for distortion correction and vignetting correction. For CA-correction only a few lensfun profiles include ca-correction, means for a default profile raw-auto-ca-correction is the better choice

PR #4345, opened 2018-01-30
Histmatching fixes and tweaks

agriggio:
histogram matching now gives always the same result regardless of the options and/or pp3 parameters.
the implementation is as fast as the old one
it should also behave a bit better with underexposed thumbnails.

in summary, I think this should go in.


Beep6581:
Compiling...

heckflosse:
@agriggio Alberto, for Sony arw files the additional decode by using `Thumbnail::loadFromRaw(...)` does not really matter, because decoding Sony arw files is one the fastest decoders in rt. For Canon CR2 the overhead by decoding the whole raw file again is significant.

Beep6581:
On my laptop, [Intel Core m3-6Y30](https://en.wikipedia.org/wiki/List_of_Intel_Core_M_microprocessors#"Skylake-Y"_(SoC,_dual-core,_14_nm)) 0.9GHz, the feature feels acceptably fast, but as Ingo pointed out, CR2 files are slower to decode, and when I tested a whole bunch of formats, the difference was most noticeable on a Canon EOS 6D MKII CR2 file, 26.1MP. Commit 2f1ab34ef1231688e6f0a6e247ccb68510fd16e2 before the patch took about 1s, and c04150be35bbbc4591944593d438ed39ea4bca0c after the patch needed about 2s for the same raw file.

rtengine/histmatching.cc
```diff
@@ -86,12 +86,19 @@ void mappingToCurve(const std::vector<int> &mapping, std::vector<double> &curve)
...
         }
     }
+    if (idx == int(mapping.size())) {
+        for (idx = 1; idx < int(mapping.size()); ++idx) {
+            if (mapping[idx] >= idx) {
+                break;
+            }
+        }
+    }
     int step = max(int(mapping.size())/npoints, 1);
```
Floessie: Which `max()` is that?

rtengine/histmatching.cc
```diff
@@ -185,50 +202,48 @@ void RawImageSource::getAutoMatchedToneCurve(std::vector<double> &outCurve)
...
     {
-        int tw = source->getWidth(), th = source->getHeight();
-        float thumb_ratio = float(std::max(tw, th)) / float(std::min(tw, th));
-        float target_ratio = float(std::max(fw, fh)) / float(std::min(fw, fh));
+        RawMetaDataLocation rml;
+        eSensorType sensor_type;
+        double scale;
+        int w = fw / skip, h = fh / skip;
+        std::unique_ptr<Thumbnail> thumb(Thumbnail::loadFromRaw(getFileName(), rml, sensor_type, w, h, 1, false, false));
+        target.reset(thumb->processImage(neutral, sensor_type, fh / skip, TI_Nearest, getMetaData(), scale, false));
```
Floessie: I might have a déjà-vu, or is `Thumbnail::loadFromRaw()` guaranteed to return a `Thumbnail`? If so, please bear with me.

rtengine/histmatching.cc
agriggio: max Von sydow -- I thought it was obvious ;-P

or are you saying we have different implementaions of max that give different results *over the integers*?

rtengine/histmatching.cc
agriggio: this is legit, thanks! I'll fix :+1:

rtengine/histmatching.cc
```diff
@@ -185,50 +202,48 @@ void RawImageSource::getAutoMatchedToneCurve(std::vector<double> &outCurve)
...
                 // crop the width
-                int cw = fw - (fh * float(tw) / float(th));
+                int cw = tw - (th * float(sw) / float(sh));
                 cx += cw / 2;
-                fw -= cw;
+                tw -= cw;
             }
             if (settings->verbose) {
-                std::cout << "histogram matching: cropping target to get an aspect ratio of " << std::fixed << std::setprecision(2) << thumb_ratio << ":1, new full size is " << fw << "x" << fh << std::endl;
+                std::cout << "histogram matching: cropping target to get an aspect ratio of " << std::fixed << std::setprecision(2) << thumb_ratio << ":1, new size is " << tw << "x" << th << std::endl;
```
Floessie: That's the drawback of streams: They aren't stateless. Coverity already warned about `std::cout` being changed without resetting...

agriggio:
@heckflosse sorry for the canon slowdown, but I saw no better way to fix the problems. I have faith in you though, I'm sure you can find a way to speed this up...

do you think we should wait to merge before it gets solved?

rtengine/histmatching.cc
Floessie: Aha, I thought it was `max_planck()`. 😄

> or are you saying we have different implementaions of max that give different results over the integers?

No, we have `std::max()` and `rtengine::max()` both yielding the same result, with the latter even being `constexpr` and taking more than two arguments. But without giving the namespace it's unclear which one is used. I bit further below you use `std::max()`, so it would be more consistent to add the namespace here, too (and do without `using namespace` for these very reasons).

agriggio:
I've pushed one more tweak to the curve matching. I think this is the best I got so far. Now the wonky curves should have gone -- at least on my tests.

heckflosse:
@agriggio Alberto, in rawimagesource.cc there is ri->data[] which is the raw data before preprocessing and demosaic. Using that you can save the decoding from file.

agriggio:
@heckflosse sorry Ingo, I'm not comfortable enough to do major changes at this point. If you want to go ahead, you're more than welcome though. Otherwise this is fine as is for me. And it's also fine if we want to keep this out of 5.4 if you think the performance hit is too much.

agriggio:
@heckflosse I took a (quick) look at the code. The problem is that `rtengine::Thumbnail::loadFromRaw` does its "sort-of" demosaicing using `ri->image`, not `ri->data`. So we would need to either:

- modify `Thumbnail::loadFromRaw` to work on `ri->data` instead of `ri->image`, or
- add some `RawImage::uncompress_image` method to reinitialize `image` from `data`.

In both cases, it seems it's too late now for a change that would need proper testing to make sure nothing gets broken... I'd rather avoid that and accept the (small, to me) performance hit for Canon users. After all, who shoots Canon anyway? :stuck_out_tongue_winking_eye: 

Jokes aside, what are your opinions? Ping also @Floessie @Beep6581 @Hombre57 @Desmis

heckflosse:
@agriggio I will take a look now

Floessie:
@agriggio I for one shoot Canon. But I have a fast PC (albeit Intel, but this is all userspace anyway). 😜 

@heckflosse Thanks on behalf of all Canon users! 👍

agriggio:
@Floessie given Canon's market share, I guess you are not alone... that's why my ":stuck_out_tongue_winking_eye: "

@heckflosse a quick-but-ugly hack could be to not free `image` in `RawImageSource::load`, or (even uglier) to not free it for Canon cameras. Then we could extend the signature of `Thumbnail::loadFromRaw` to take an optional `RawImage *` as input. If `nullptr`, the method would create a new `RawImage`, otherwise it would reuse the existing one.

But it's ugly an I vote against it :-)

heckflosse:
@agriggio My idea was to make a 
```
Thumbnail* Thumbnail::getFromRaw (int &w, int &h, int fixwh, bool rotate, RawImage *ri)
```

agriggio:
@heckflosse there's still the problem of `RawImage::data` vs `RawImage::image`. How would you solve that? Anyway, the stage is yours :-)

Beep6581:
Commit 34bd19f was mistakenly pushed to the wrong branch, should have gone into `dev`.

Desmis:
@agriggio 
I come back from a walk by the sea, and as a reminder I am not at all a computer scientist, but caution seems important to me here. 
In addition, I think we need to stabilize to allow assessments of these changes :-)

heckflosse:
@agriggio @Floessie A fast and clean solution is possible, but I will do that after 5.4 is released, because it requires some work and testing.
:+1: for merge

agriggio:
Thanks to all for the feedback. I will merge later tonight then!

heckflosse:
@agriggio There probably is an easy solution if we trade speed for memory. It would require width * height * 6 bytes additional memory per open editor.

rtengine/histmatching.cc
```diff
@@ -213,6 +213,7 @@ void RawImageSource::getAutoMatchedToneCurve(std::vector<double> &outCurve)
             histMatchingCache = outCurve;
             return;
         }
+        skip = LIM(5, skip * fh / h, 10); // adjust the skip factor -- the larger the thumbnail, the less we should skip to get a good match
```
heckflosse: for loadfromraw() skip % 2 should be zero, because in loadfromraw() it will be adjusted to a multiple of 2 to match the bayer pattern. https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/rtthumbnail.cc#L464.

Edit: Also LIM(5, x, 10) is equal to max(5, x), looks like a bug...

I just saw a possible bug in loadfromraw where skip can be set to 1. I will prepare a fix for that.

heckflosse:
@agriggio Can you please include this in your merge?

```
diff --git a/rtengine/rtthumbnail.cc b/rtengine/rtthumbnail.cc
index 2dcb6b9e..5318c5ea 100644
--- a/rtengine/rtthumbnail.cc
+++ b/rtengine/rtthumbnail.cc
@@ -466,8 +466,8 @@ Thumbnail* Thumbnail::loadFromRaw (const Glib::ustring& fname, RawMetaDataLocati
         skip--;
     }
 
-    if (skip < 1) {
-        skip = 1;
+    if (skip < 2) {
+        skip = 2;
     }
 
     int hskip = skip, vskip = skip;

```

PR #4352, opened 2018-02-01
Speedup for histogram matching

heckflosse:


heckflosse:
With last commit it's an about 25% speedup measured using a 24MP arw file

agriggio:
:+1:

heckflosse:
branch stays open for further speedups

PR #4354, opened 2018-02-02
Text toggle button CSS

Beep6581:


agriggio:
Almost works, but there's still a glitch with "hovering" over radio buttons:
screenshot from 2018-02-02 09 33 22

Beep6581:
Fixed.

PR #4356, opened 2018-02-02
made histogram matching depend on the input profile

agriggio:
This improves the accuracy of the matching when using non-default profiles.

IMHO this is the "right thing" to do, though I'm fine with postponing to after 5.4 if it's too late now.


heckflosse:
Compiling...

heckflosse:
Works fine here 👍

Beep6581:
Works well, cool!

Beep6581 (approved):

PR #4366, opened 2018-02-04
Softproofing output profile

agriggio:
As agreed on IRC, merge so that it gets tested for 5.4


agriggio:
keeping the branch in case we want to improve later (e.g. see @Desmis feedback)

PR #4373, opened 2018-02-06
Avoid unnecessary processing when zooming to >= 100% or opening detai…

heckflosse:
…l windows

This reduces (but not completely stops) the initial blocky opening of detail windows.


Beep6581:
I will review this evening.

Issue #4319, opened 2018-01-22
`setenv("LANG", ...)` too late


Floessie:
Alberto extended the `MultiLangMgr` to set the "LANG" environment variable in #4278. This is definitely the way to go, but doesn't fix the problem for me:

```
Version: 5.3-514-gac1238e7
Branch: dev
Commit: ac1238e7
Commit date: 2018-01-22
Compiler: cc 6.3.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.22.0
Lensfun: V0.3.2.0
Build type: release
Build flags:  -std=c++11 -march=native -Werror=unused-label -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -O3 -DNDEBUG
Link flags:  -march=native
OpenMP support: ON
MMAP support: ON
```

If I place the `setenv()` in front of [`gtk_init()`](https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/main.cc#L492) it's effective, but not after it.

Best,
Flössie

heckflosse:
@Floessie Great you found a solution. Why don't you just push it?

Floessie:
@heckflosse No, that's no real solution, thus I didn't push it. The problem is, that `MultiLangMgr` depends on the options being ready, while the options depend on things that happen after `gtk_init()`. Just taking the options loading from `init_rt()` to the line before `gtk_init()` made it crash and I didn't have the time to dig deeper yesternight. I'll give it another try this week...

heckflosse:
@Floessie Did you test whether it works using `putenv()` instead of `setenv()`?
I had to use `putenv()` for windows builds, because `setenv()` is not available on msys2.

Floessie:
@heckflosse I'll give it a try! 👍

Beep6581:
Could be related:
https://discuss.pixls.us/t/rawtherapee-5-3-524-and-gimp-2-9-9-on-ubuntu-studio-17-10/6546

Floessie:
@heckflosse I've tried it without success (as I already expected). I hope the msys2 supplied `putenv()` is implemented better than what [POSIX](https://stackoverflow.com/questions/5873029/questions-about-putenv-and-setenv) mandates.

Neither `setenv()` nor `putenv()` are thread-safe, so they have to be called as early as possible.

heckflosse:
@Floessie it's `putenv(const char *_EnvString)`

[...]

PR #4382, opened 2018-02-09
Mac Executable Loader: Set LC_ALL

Benitoite:
This prevents a crash when switching to queue with ö/ü in image path using en_US introduced in dev.
https://github.com/Beep6581/RawTherapee/issues/4319#issuecomment-364468235

### DIFF
--- a/tools/osx/executable_loader.in
+++ b/tools/osx/executable_loader.in
@@ -51,5 +51,6 @@ esac
 # Prevent crash when directory name contains special characters
 AppleLocale=`defaults read -g AppleLocale`
 export LANG=${AppleLocale%@*}.UTF-8
+export LC_ALL=${AppleLocale%@*}.UTF-8
 
 exec "${cwd}/rawtherapee-bin" "$@"


Beep6581:
@Benitoite can you test whether `LANG` is needed at all? In glibc, `LC_ALL` is inspected before `LANG`.
Also, for my info, what does `${AppleLocale}` output?

Benitoite:
@Beep6581 `defaults read -g AppleLocale` outputs `en_DE` for example.  That is the command used to load the AppleLocale value in the Mac executable loader (confusing style using same name thoug legit).   I will test omitting the LANG assignment when I get back home from an outing.

Benitoite:
@Beep6581 Without setting LANG in the executable loader, I crash instantly when launching:

```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libsystem_c.dylib             	0x00007fff9dde0b52 strlen + 18
1   rawtherapee-bin               	0x000000010c172c41 Options::load(bool) + 7457
2   rawtherapee-bin               	0x000000010c13c937 main + 4695
3   libdyld.dylib                 	0x00007fff9ddaa235 start + 1

```

Benitoite:
Closing as now not needed in latest dev pull.

heckflosse:
@Benitoite Does that mean my latest push fixed the issue completely? If so, I will merge into release5.4

Benitoite:
Your latest push fixed the Mac issue with LC_ALL not being set crashing the queue using en_US.  That was temporary. The problem the queue with en_DE remains, so the Mac performance remains as it is in `release5.4` branch.

heckflosse:
@Benitoite Ok, means the issue with en_DE is not fixed, but I should merge it into release5.4 because now it's not worse than before and covers more cases.

Btw: what does en_DE compared to de_DE or en_GB?

Benitoite:
@heckflosse Yes.

Queue crash when image path has ö/ü:
en_DE crashes
de_DE is OK
en_UK is OK
de_US crashes

heckflosse:
@Benitoite According to your `locale -a` there is no `en_DE` on your system, or am I blind?

Floessie:
@heckflosse I'm off to bed now, but as a tip: We could call `setlocale()` ourselves to check, if the locale is present, and learn from [DT](https://github.com/darktable-org/darktable/blob/85985f95bb18cc6d7555025f86824eddda6c15fa/src/common/l10n.c#L31) (I mean the `gtk_disable_setlocale()`). Or was the outcome, that we SEGV on macOS as soon as we call `setlocale()`? I doubt that from the latest outcomes...

@heckflosse No, neither `en_DE` nor `de_US`.

Benitoite:
@heckflosse Correct; it is possible to mix and match languages and countries in the macOS control panel & end up with a locale that is not among the `locale -a` list.

Beep6581:
I use:
```
LANG=en_SE.UTF-8
LC_CTYPE="en_SE.UTF-8"
LC_NUMERIC=C
LC_TIME="en_SE.UTF-8"
LC_COLLATE="en_SE.UTF-8"
LC_MONETARY="en_SE.UTF-8"
LC_MESSAGES="en_SE.UTF-8"
LC_PAPER="en_SE.UTF-8"
LC_NAME="en_SE.UTF-8"
LC_ADDRESS="en_SE.UTF-8"
LC_TELEPHONE="en_SE.UTF-8"
LC_MEASUREMENT="en_SE.UTF-8"
LC_IDENTIFICATION="en_SE.UTF-8"
LC_ALL=
```

PR #4406, opened 2018-02-19
review of 'rtengine/PF_correct_RT.cc'

heckflosse:
This pr is **not** for 5.4

I made some changes to `rtengine/PF_correct_RT.cc` which result in (hopefully) better code, speedups and reduced memory usage.
There are 4 functions in the file. Here are test profiles. 1.pp3 is for first function and so on.
[pp3.zip](https://github.com/Beep6581/RawTherapee/files/1736837/pp3.zip)
I tested using [this](http://www.rawtherapee.com/shared/test_images/pippi.arw) 24MPixel file.

Here are my results (in ms):

```
                                                      before           after

PF_correct_RT (defringe)                                524             398
PF_correct_RTcam (defringe when ciecam02 is enabled)    666             484
Badpixelscam (ciecam02 hot/bad pixel filter)           1068             795
BadpixelsLab (CbdL reduce artifacts enabled)           1574             796
```

memory usage (in byte):
```
                                                      before           after

PF_correct_RT (defringe)                              16*w*h          12*w*h
PF_correct_RTcam (defringe when ciecam02 is enabled)  20*w*h          12*w*h
Badpixelscam (ciecam02 hot/bad pixel filter)          24*w*h          20*w*h
BadpixelsLab (CbdL reduce artifacts enabled)          24*w*h          12*w*h
```

```
       before     after

loc     1866      1166
```

Edit: Because I could eliminate some epsilons which were used to avoid divisions by zero, there may be differences between before and after output.


Floessie:
@heckflosse I'll push some minor cleanups. BTW: The const-correctness of `JaggedArray<>` is broken.

Floessie:
@heckflosse `DefringeParams::threshold` is a `float`, but the functions take is as an `int`. Is that a problem?

rtengine/PF_correct_RT.cc
```diff
@@ -1059,8 +1059,8 @@ void ImProcFunctions::BadpixelsLab(LabImage * src, double radius, int thresh, fl
 
     // end luma badpixels
 
-    float ** tmaa = tmL; // reuse tmL buffer
-    const JaggedArray<float> tmbb(width, height);
+    float** const tmaa = tmL; // reuse tmL buffer
+    JaggedArray<float> tmbb(width, height);
```
heckflosse: @Floessie Why did you remove the const here and from other JaggedArrays? Just curious

Edit: Now I saw your comment about const-correctness of JaggedArray<>. Shouldn't we change that at other places in code too?

Floessie: @heckflosse Absolutely. But not tonight. I'll fix `JaggedArray<>` during this week (in which branch?), the compiler will tell us the places then...

heckflosse: @Floessie Should those fixes to JaggedArray go into 5.4? If not, just change in my branch

heckflosse:
@Floessie 

> DefringeParams::threshold is a float, but the functions take is as an int. Is that a problem?

Not really, because in gui it's an int too. But we should change it also to an int in procparams imho

btw: thanks for your review :+1:

rtengine/PF_correct_RT.cc
Floessie: @heckflosse No, not for 5.4 as it's not a functional bug. I'll push it to this branch then.

heckflosse:
@Floessie @Beep6581 What shall we do with the head comment of `rtengine/PF_correct_RT.cc`?

```
////////////////////////////////////////////////////////////////
//
//      Chromatic Aberration Auto-correction
//
//      copyright (c) 2008-2010  Emil Martinec <ejmartin@uchicago.edu>
```

Obviously it's just copied from [here](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/CA_correct_RT.cc#L1)

Beep6581:
I'm not sure what you're asking. From what I know, it's up to every dev to add their name to the list, since adding it to each line is not reasonable. Don't think dates are necessary.

heckflosse:
@Beep6581 The comment is copied from raw ca-correction code, which is not was PF_correct_RT.cc does. It is misleading

Beep6581:
So it needs to be fixed, but it's not obvious to me what "PF" means ;)

heckflosse:
@Beep6581 

> but it's not obvious to me what "PF" means

agreed ;)

heckflosse:
I would like to merge this pr into dev as soon as 5.4 is released.
Final reviews are welcome @Floessie @agriggio @Hombre57 @Desmis @Beep6581 @iliasg 

Ingo

Desmis:
@heckflosse

Hello Ingo
I just tested  this branch
With my image test and with default settings for Defringe result is good  :)

If I enabled CIECAM with the same image, If I enable Defringe, fringe "desappears"  but  strong artifacts red and green.
jacques

heckflosse:
@Desmis Thanks for testing. Bug confirmed. Looking...

heckflosse:
@Desmis Jacques, I pushed a fix

rtengine/PF_correct_RT.cc
```diff
@@ -9,6 +9,9 @@
 // optimized: September 2013, Ingo Weyrich
 // further optimized: February 2018, Ingo Weyrich
 //
+//  Ingo Weyrich March 1028: The above comment 'Chromatic Aberration Auto-correction' sounds wrong
```
Beep6581: `s/1028/2018/` ?

Floessie: Alternatively, add that to [Wikipedia](https://en.wikipedia.org/w/index.php?title=1028&action=edit), there isn't much information on that year...

heckflosse: @Floessie @Beep6581 I was faster :)

Desmis:
@heckflosse 

Ingo

I tested, works fine :)

jacques

PR #4411, opened 2018-02-26
Update Deutsch locale

TooWaBoo:


heckflosse:
@TooWaBoo Thank you :+1:

PR #4413, opened 2018-02-27
Speedup for rcd demosaic

heckflosse:
I made a first speedup for rcd demosaic.
Measured using a 100MP file and a 24MP file:

```
                     before        after

100 MP              3850 ms      1320 ms
24 MP                940 ms       340 ms
```
On my machine rcd demosaic now is as fast as amaze demosaic.
To compare with amaze demosaic I placed a stopwatch also in amaze demosaic

Memory usage is also reduced:

```
before:  width * height * 24 byte
after:   c * ts * ts * 24 byte, where ts = 228 and c is the number of cpu cores
```

I will add further commits to this pr.


heckflosse:
With last commit I reduced the number of calculations in `STEP 1: Find cardinal and diagonal interpolation directions`

```
                     before     first commit   last commit

100 MP              3850 ms        1320 ms       1140 ms
24 MP                940 ms         340 ms        295 ms
```

heckflosse:
Processing times with `another small speedup`

```
100 MP : 1090 ms
24 MP:    285 ms
```

heckflosse:
Processing times with last commit

```
100 MP : 1085 ms
24 MP:    270 ms
```

Issue #4438, opened 2018-03-12
New RT Out Of Gamut function in RT5.4 is too sensitive


RawConvert:
This new functionality allows out-of-gamut areas to be highlighted in relation to the RT Output profile or the Monitor profile.  You would surely not expect to see OOG areas if you opened an sRGB jpeg and set sRGB for output, however you do, as this photo shows -

![indira-jpeg-shows-oog](https://user-images.githubusercontent.com/16868433/37284509-af4ada1e-25f3-11e8-9260-3bc664608b38.png)

To reproduce -
. take the uploaded _6D_7984-Test-RT_sRGB.jpg.pp3 which has profile RT_sRGB
. open in RT with Neutral profile and with no printer profile set in Preferences and with RT_sRGB set as Output in Colour / Output Profile
. press both the soft-proofing buttons at the bottom of the preview window.

Note that the cyan areas do not disappear quickly if you reduce saturation, you can reduce Saturation to -52 and still see small OOG areas on the side of the black drum.

Another example - _6D_7667-Test-canal-RT_sRGB.jpg - this also has fairly persistent OOG areas.  I have not had to search around to find these examples, I imagine most sRGB jpegs will show the behaviour.
![_6d_7984-test-rt_srgb](https://user-images.githubusercontent.com/16868433/37284543-cd48a9e2-25f3-11e8-9a47-3d99949f6c14.jpg)
![_6d_7667-test-canal-rt_srgb](https://user-images.githubusercontent.com/16868433/37284575-f4b44ca2-25f3-11e8-95b4-edd6e391eeab.jpg)

uglybrian89:
Delete my account. Please. Thank you

On Mar 12, 2018 5:52 AM, "Andrew" <notifications@github.com> wrote:

This new functionality allows out-of-gamut areas to be highlighted in
relation to the RT Output profile or the Monitor profile. You would surely
not expect to see OOG areas if you opened an sRGB jpeg and set sRGB for
output, however you do, as this photo shows -

[image: indira-jpeg-shows-oog]
<https://user-images.githubusercontent.com/16868433/37284509-af4ada1e-25f3-11e8-9260-3bc664608b38.png>

To reproduce -
. take the uploaded _6D_7984-Test-RT_sRGB.jpg.pp3 which has profile RT_sRGB
. open in RT with Neutral profile and with no printer profile set in
Preferences and with RT_sRGB set as Output in Colour / Output Profile
[...]

PR #4452, opened 2018-03-20
Improved gamut warning

agriggio:
Based on PhotoFlow's approach. Seems to be more accurate than lcms's built-in support. See #4438


agriggio:
Any objections to merge?

PR #4453, opened 2018-03-20
Denoising filter for Sony PDAF stripes

agriggio:
As seen on various dpreview sample photos for the A7III (but occurring also on other Sony cameras)


heckflosse:
:+1: for merge

PR #4459, opened 2018-03-22
feature: added option to use a (fast) neutral RAW rendering in 'inspector mode'

agriggio:
This is useful e.g. if your camera doesn't have full-size embedded JPEGs, and you still want to quickly check if the shot is in focus, there is motion blur, etc...

I've added an option in preferences that lets you select between 3 modes for "Inspector mode":

1. show the embedded thumbnail (default, current behaviour)
2. show a quick rendering of the RAW
3. show the thumbnail if full size, otherwise render the RAW quickly

inspector

For the rendering, I tried to make it as fast as possible without touching the existing code. Performance is acceptable but I'm sure we can do better if needed (by "we", I mean @heckflosse of course ;-). The image goes to sRGB without any color management at all, to keep things fast. This is similar to how the inspector mode currently works btw.

Feedback welcome.


rtdata/languages/default
```diff
@@ -1157,6 +1157,10 @@ PREFERENCES_TAB_IMPROC;Image Processing
 PREFERENCES_TAB_PERFORMANCE;Performance & Quality
 PREFERENCES_TAB_SOUND;Sounds
 PREFERENCES_THEME;Theme
+PREFERENCES_THUMBNAIL_INSPECTOR_JPEG;Embedded JPEG preview
+PREFERENCES_THUMBNAIL_INSPECTOR_MODE;Image to show
+PREFERENCES_THUMBNAIL_INSPECTOR_RAW;Neutral RAW rendering
```
Beep6581: s/RAW/raw/

agriggio:
ok to merge?

PR #4460, opened 2018-03-22
Unbounded processing

agriggio:
This branch removes clips and clamps from the RT pipeline. I am not sure it is complete yet, but the best way to know (and to find bugs) is to test this more extensively. If we want this feature, I think we should merge now that 5.4 has just been released, and be ready to live with the consequences for a while, until all the major bugs have emerged. FWIW, I've been using this for a while without pain, but there are several tools that I almost never use in RT, so my testing is still quite limited...

Opinions?


agriggio:
any comment on this?

Beep6581:
@agriggio +1 for merge to get more testing.

agriggio:
Unless someone objects, I'll merge later today

PR #4461, opened 2018-03-22
added possibility to specify extra working spaces via a json file

agriggio:
The JSON file is called workingspaces.json, it can be either in the global iccprofiles directory, or in the user's ICC profiles dir (set in preferences).

The format is the following:

```json
{"working_spaces": [
    {
        "name" : "ACES",
        "file" : "/path/to/ACES.icc"
    },
    {
        "name" : "ACEScg",
        "matrix" : [0.7184354, 0.16578523, 0.09882643, 0.29728935, 0.66958117, 0.03571544, -0.00647622, 0.01469771, 0.66732561]
    }
]}
```

if "matrix" is present, "file" is ignored. If only "file" is present, the matrix is extracted from the ICC profile. For this, we look only at the R, G, and B matrix columns and the white point set in the profile. Bradford adaptation is used to convert the matrix to D50. Anything else (LUT, TRC, ...) in the profile is ignored.

It is the user's responsibility to ensure that the profile is suitable to be used as a working space.

### DIFF
--- a/rtengine/clutstore.cc
+++ b/rtengine/clutstore.cc
@@ -286,7 +286,7 @@ void rtengine::HaldCLUT::splitClutFilename(
     profile_name = "sRGB";
 
     if (!name.empty()) {
-        for (const auto& working_profile : rtengine::ICCStore::getWorkingProfiles()) {
+        for (const auto& working_profile : rtengine::ICCStore::getInstance()->getWorkingProfiles()) {
             if (
                 !working_profile.empty() // This isn't strictly needed, but an empty wp name should be skipped anyway
                 && std::search(name.rbegin(), name.rend(), working_profile.rbegin(), working_profile.rend()) == name.rbegin()
--- a/rtengine/dcp.cc
+++ b/rtengine/dcp.cc
@@ -37,30 +37,36 @@ namespace
 
 DCPProfile::Matrix invert3x3(const DCPProfile::Matrix& a)
 {
-    const double res00 = a[1][1] * a[2][2] - a[2][1] * a[1][2];
-    const double res10 = a[2][0] * a[1][2] - a[1][0] * a[2][2];
-    const double res20 = a[1][0] * a[2][1] - a[2][0] * a[1][1];
-
-    const double det = a[0][0] * res00 + a[0][1] * res10 + a[0][2] * res20;
-
-    if (std::fabs(det) < 1.0e-10) {
+    DCPProfile::Matrix res = a;
+    if (!invertMatrix(a, res)) {
         std::cerr << "DCP matrix cannot be inverted! Expect weird output." << std::endl;
-        return a;
     }
-
-    DCPProfile::Matrix res;
-
-    res[0][0] = res00 / det;
-    res[0][1] = (a[2][1] * a[0][2] - a[0][1] * a[2][2]) / det;
-    res[0][2] = (a[0][1] * a[1][2] - a[1][1] * a[0][2]) / det;
-    res[1][0] = res10 / det;
-    res[1][1] = (a[0][0] * a[2][2] - a[2][0] * a[0][2]) / det;
-    res[1][2] = (a[1][0] * a[0][2] - a[0][0] * a[1][2]) / det;
-    res[2][0] = res20 / det;
-    res[2][1] = (a[2][0] * a[0][1] - a[0][0] * a[2][1]) / det;
-    res[2][2] = (a[0][0] * a[1][1] - a[1][0] * a[0][1]) / det;
-
     return res;
+    
+    // const double res00 = a[1][1] * a[2][2] - a[2][1] * a[1][2];
+    // const double res10 = a[2][0] * a[1][2] - a[1][0] * a[2][2];
+    // const double res20 = a[1][0] * a[2][1] - a[2][0] * a[1][1];
+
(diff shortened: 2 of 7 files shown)


Desmis:
@agriggio 
Alberto

For me it is a good idea :)

jacques

Beep6581:
Sounds good!

agriggio:
ok to merge?

Desmis:
no problem :)

Beep6581:
GCC-6.4.0
Compilation fails:
```
/home/morgan/programs/code-rawtherapee/rtengine/iccstore.cc: In member function ‘bool rtengine::ICCStore::Implementation::loadWorkingSpaces(const Glib::ustring&)’:
/home/morgan/programs/code-rawtherapee/rtengine/iccstore.cc:808:18: error: ‘invertMatrix’ is not a member of ‘rtengine’
             if (!rtengine::invertMatrix(pMatrices.back()->toMatrix(), b)) {
                  ^~~~~~~~
/home/morgan/programs/code-rawtherapee/rtengine/iccstore.cc:808:18: note: suggested alternative:
In file included from /home/morgan/programs/code-rawtherapee/rtengine/color.h:24:0,
                 from /home/morgan/programs/code-rawtherapee/rtengine/iccstore.h:29,
                 from /home/morgan/programs/code-rawtherapee/rtengine/iccstore.cc:32:
/home/morgan/programs/code-rawtherapee/rtengine/rt_math.h:145:6: note:   ‘invertMatrix’
 bool invertMatrix(const std::array<std::array<T, 3>, 3> &in, std::array<std::array<T, 3>, 3> &out)
      ^~~~~~~~~~~~
```

agriggio:
@Beep6581 I might have messed up with rebasing, let me check

agriggio:
@Beep6581 should work now, please confirm when you have time

Beep6581:
@agriggio compiles and seems to work correctly.

Beep6581:
@agriggio has this been documented?

Beep6581:
It is now:
http://rawpedia.rawtherapee.com/Color_Management#Working_Profile

agriggio:
@Beep6581 thanks!

agriggio:
@Beep6581 I've slightly edited the page to fix the location of the global `workingspaces.json`. Please take a look when you can (I'm not sure what is the path on macs)

Beep6581:
Current text:
```text
Windows: %RT_DIR%\iccprofiles
Linux: $RT_INSTALL_PREFIX/share/rawtherapee/iccprofiles
```

When I tested, I used `/usr/share/color/icc/workingspaces.json` where `/usr` is not the `$RT_INSTALL_PREFIX`.
A bundle-build is forced-on for Windows and macOS, but optional in Linux. My bundle build `~/programs/rawtherapee` has a `share` folder but the `iccprofiles` folder is in the root of the bundle, i.e. in `~/programs/rawtherapee/iccprofiles`.

Maybe we should change the docs (code markup removed for readability) to something like:
```text
* Windows: %RT_DIR%\iccprofiles
* Linux:
** When `BUILD_BUNDLE="OFF"`: $RT_INSTALL_PREFIX/share/rawtherapee/iccprofiles
** When `BUILD_BUNDLE="ON"`: $RT_INSTALL_PREFIX/share/rawtherapee/iccprofiles
```

`$RT_INSTALL_PREFIX` implies that it's an environment variable (because it's uppercase), and that it gets set somewhere, but that is not the case. Shouldn't we use `$CMAKE_INSTALL_PREFIX` instead of `$RT_INSTALL_PREFIX`? Or replace `$RT_INSTALL_PREFIX` with `<rt install prefix>` so it's clear it's not an env var, but I think `$CMAKE_INSTALL_PREFIX` would be more clear, as the only people who would understand this part are people who compile RT themselves.

However, while that might be clear to a package maintainer, an end-user has no idea that `$RT_INSTALL_PREFIX` = `$CMAKE_INSTALL_PREFIX` = `/usr`
So maybe we should instead write:
```text
* Windows: %RT_DIR%\iccprofiles
* Linux:
** When installed using your package manager: /usr/share/rawtherapee/iccprofiles
** When self-compiled with `BUILD_BUNDLE="ON"`: $RT_INSTALL_PREFIX/share/rawtherapee/iccprofiles
```
While `/usr` might not be true for all Linux distributions, I think it's correct for the vast majority, or in fact all that I'm aware of.
What do you think?

agriggio:
@Beep6581 yes, the current text was essentially a draft, I agree with your proposal (especially in using `/usr` directly in the example)

Beep6581:
@agriggio ok, before I make the change I have one more question. For Windows is the only valid folder `%RT_DIR%\iccprofiles`, or it it also `C:\Windows\system32\spool\drivers\color`?

agriggio:
The first one is what is referred to as "the RT profiles dir". The second one is ok if that is the value of the user ICC folder in preferences (perhaps that's the default value?)

Beep6581:
Yes, that's usually the default:
https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/options.cc#L507

Beep6581:
What do you think?
http://rawpedia.rawtherapee.com/Color_Management#Adding_Custom_Working_Profiles

I didn't mention `/usr/share/color/icc` since that is what `Directory containing color profiles` refers to by default.

PR #4474, opened 2018-03-27
Bump cJSON to 1.7.5

Floessie:
Hi there,

with the lastest reuse of cJSON in RT, I thought the time has come to update that part. The API is stable, so that no other changes were necessary.

Best,
Flössie


agriggio:
@Floessie I think you should merge as soon as you have the chance, I see no reason to not do it

Floessie:
@agriggio @heckflosse @Beep6581 Thanks. Done.

PR #4492, opened 2018-04-06
Remove ciecam02 double precision processing

heckflosse:
Though in past the double precision ciecam02 code was the reference ciecam02 code, imho that's not the case anymore. For that reason I have removed (hopefully) all the double precision ciecam02 code and would like to push that into dev branch if there are no objections from other devs.

Especially ping @Desmis


Desmis:
@heckflosse 

Hello Ingo

No problem you can push :)

jacques

Issue #3801, opened 2017-04-05
Adopt libexiv2 or ExifTool for metadata handling


Beep6581:
Metadata handling in RawTherapee is far from optimal. I rip out and process chunks of metadata from ExifTool and manually merge them into the rtexif/* files. This is tedious and error prone, and in all the years that RT has been open-source I'm the only person to have done that. This issue is to adopt libexiv2 or ExifTool to outsource our metadata handling needs. Although this will initially take considerable effort, in the long run it should result in less work when it comes to having up to date camera and lens info, and adding support for reading and writing metadata to various file formats.

Robin of Exiv2 was willing to help in some capacity with libexiv2 adoption. *3

Furthermore, libiptcdata which we're using has been dead since 2009. IPTC is being phased out in favor of XMP. I believe both libexiv2 and ExifTool handle IPTC and XMP, solving that problem as well.

References:
http://owl.phy.queensu.ca/~phil/cpp_exiftool/
http://www.exiv2.org/getting-started.html#lib
https://github.com/Beep6581/RawTherapee/issues/3581#issuecomment-270509717
https://github.com/Beep6581/RawTherapee/issues/1314#issuecomment-276317545
https://github.com/Beep6581/RawTherapee/issues/2307#issuecomment-172172408

Hombre57:
Found this just today : http://dev.exiv2.org/projects/exiv2/wiki/How_does_Exiv2_compare_to_Exiftool

agriggio:
I can open a PR that replaces rtexif with exiv2. The basics should work, but I haven't tested it much and won't be able to maintain it and/or perform the needed cleanups for integration into dev (if any).
Therefore, I'm basically asking if someone would be willing to "adopt" the branch and do the remaining work to get this merged.

If there is interest, please let me know.

Floessie:
@agriggio Is there a branch we can have a look at?

agriggio:
@Floessie I have a branch I can push, but as I said I won't be able to work on it. My question was like "should I push it anyway?". 

If your answer is "yes", I'll push later today :-)

Beep6581:
Yes!

Floessie:
Yes, of course! You don't have to create a PR, though...

agriggio:
Pushed branch `metadata-exiv2`.

A couple of comments:

- some lenses have different names, this is just the way it is with exiv2 vs rtexif

- exif tags editing is limited only to editing the following tags:

  User Comment
  Artist
[...]

PR #4498, opened 2018-04-10
added support for user-specific DCP and input ICC profiles dirs

agriggio:
Look for direcories `dcpprofiles/` and `iccprofiles/input/` under `Options::rtdir`
(typically something like `$HOME/.config/RawTherapee`)

### DIFF
--- a/rtengine/dcp.cc
+++ b/rtengine/dcp.cc
@@ -1737,51 +1737,50 @@ void DCPStore::init(const Glib::ustring& rt_profile_dir, bool loadAll)
     file_std_profiles.clear();
 
     if (!loadAll) {
-        profileDir.assign (rt_profile_dir);
+        profileDir = { rt_profile_dir, Glib::build_filename(options.rtdir, "dcpprofiles") };
         return;
     }
 
-    if (!rt_profile_dir.empty()) {
-        std::deque<Glib::ustring> dirs = {
-            rt_profile_dir
-        };
+    std::deque<Glib::ustring> dirs = {
+        rt_profile_dir,
+        Glib::build_filename(options.rtdir, "dcpprofiles")        
+    };
 
-        while (!dirs.empty()) {
-            // Process directory
-            Glib::ustring dirname = dirs.back();
-            dirs.pop_back();
+    while (!dirs.empty()) {
+        // Process directory
+        Glib::ustring dirname = dirs.back();
+        dirs.pop_back();
 
-            std::unique_ptr<Glib::Dir> dir;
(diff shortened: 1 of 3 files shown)


Floessie:
Alberto, I've fiddled in exactly the same region for #4500, so we should coordinate the merging a bit. I'd suggest, @Beep6581 tests this first, so you can quickly merge and I'll resolve the conflicts in my branch then.

agriggio:
@Floessie thanks for the warning! As you wish though -- I don't mind merging later if you prefer

Floessie:
@agriggio No problem, you can go first. :+1:

iliasg:
Not sure which is the better solution but ..
There is a conflict with model names under different tags ..
In DNGs and DCPs
0xc614 | UniqueCameraModel
In DNGs 
0xc615 | LocalizedCameraModel
In Canon maker data each model is globally identified at "CanonModelID" tag 
0x0010 | CanonModelID | int32u
https://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/Canon.html#CanonModelID
(where we can take a full list for use ..)
Dcraw.c makes use of this list in a table which we (RT) do not maintain any more .. :( 

Same with Sony 0xb001 | SonyModelID | int16u
https://www.sno.phy.queensu.ca/~phil/exiftool/TagNames/Sony.html#SonyModelID

I failed to locate such a table for Panasonic ..

Beep6581:
Please keep #3801 in mind. RT's metadata support is slowly rusting away and IMHO movement in any direction other than adopting an actively maintained external metadata library should be considered a temporary workaround.

agriggio:
@Beep6581 ok, but this PR has nothing to do with metadata per se. It merely adds the option to replace an "auto matched" DCP/ICC profile with a custom one without overwriting the original, or to add custom DCP/ICC profiles that RT doesn't ship yet and have them play nicely with the existing "auto matching"

gaaned92:
```
Look for direcories dcpprofiles/ and iccprofiles/input/ under Options::rtdir
(typically something like $HOME/.config/RawTherapee)
```
According to above, where these dirs should be located on W10?
I think they should be subdirs of the "config folder"?

agriggio:
@gaaned92 I think so, yes, but I don't have a windows machine to test... the dirs are relative to the bundle dir I think

gaaned92:
@agriggio 
I tested and effectively on W10 it is %LOCALAPPDATA%/\<config folder\>/dcpprofiles

agriggio:
anyone against merging?

Beep6581:
@agriggio no objections.

PR #4499, opened 2018-04-10
added SoftLight tool

agriggio:
This is a simple tool that emulates the effect of blending an image with a copy of itself in "soft-light" mode in GIMP. The effect is to add a bit of contrast and saturation, resulting in a nice (to me) 'pop'.
You can get the exact same effect in the existing RT, by using a proper CLUT (open the identity clut in GIMP, duplicate in soft-light mode, and save), but I find this so nice and useful that I thought a dedicated tool was nicer.
A couple of screenshots:

- with default applied (Auto Matched ISO low):
default

- adding a bit of soft-light:
softlight


rtengine/ipsoftlight.cc
```diff
@@ -0,0 +1,73 @@
...
+{
+    if (!params->softlight.enabled || !params->softlight.strength) {
+        return;
+    }
+
+    const float blend = params->softlight.strength / 100.f;
+    const float orig = 1.f - blend;
+
+    const auto apply =
+        [=](float x) -> float
```
Floessie, line 39: This is a pleasure to read. :+1:

Only a nitpicking concerning line 39 above: Using `[=]` (capture all by copy) isn't recommended, better keep the scope minimal, i.e. `[blend, orig]` as far as I can see.

rtengine/procparams.cc
```diff
@@ -3306,6 +3327,10 @@ int ProcParams::save(const Glib::ustring& fname, const Glib::ustring& fname2, bo
         saveToKeyfile(!pedited || pedited->hsvequalizer.scurve, "HSV Equalizer", "SCurve", hsvequalizer.scurve, keyFile);
         saveToKeyfile(!pedited || pedited->hsvequalizer.vcurve, "HSV Equalizer", "VCurve", hsvequalizer.vcurve, keyFile);
 
+// Soft Light
+        saveToKeyfile(!pedited || pedited->softlight.enabled, "SoftLight", "Enabled", softlight.enabled, keyFile);
+        saveToKeyfile(!pedited || pedited->softlight.strength, "SoftLight", "Strength", softlight.strength, keyFile);
+
```
Floessie, line 3333: Here soft light is in front of film simulation. In most other places it's FS then SL. Would be nice and clean if you could streamline it.

agriggio: good catch. This is a leftover from when I was putting soft light before film simulation in the pipeline. I'll fix that, thanks for spotting!

sguyader:
@agriggio Nice, simple tool. At first I thought it wasn't really useful, as a somewhat similar effect can be obtained with contrast and saturation/chromaticity sliders. But in use, it is simpler gives a subtle effect.

Hombre57:
@agriggio I agree that it provide more punch to the image. I like this tool, and hope to see it in dev.

agriggio:
what to do about this?

heckflosse:
@agriggio no objection from me

Beep6581:
I tested it in April, found no issues. No objections.

heckflosse:
\o/

PR #4503, opened 2018-04-11
Add StartupWMClass to desktop file

tribut:
This allows desktop environments to better associate the application
with its windows and seems to be necessary for the appimage build at
least.


Beep6581:
Ref: https://standards.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html

Beep6581:
File validates.

PR #4515, opened 2018-04-20
Lens corrections: apply CA correction after distortion, and not before

agriggio:
As the title says. This seems to produce better results (and more consistent with darktable). I only tested the lensfun backend though. Example screenshots:

- current dev
old

- this branch (sorry for the wrong name, I just realized that ;-)
new

- darktable
dt

### DIFF
--- a/rtengine/iptransform.cc
+++ b/rtengine/iptransform.cc
@@ -327,25 +327,29 @@ void ImProcFunctions::transform (Imagefloat* original, Imagefloat* transformed, 
     } else {
         bool highQuality;
         std::unique_ptr<Imagefloat> tmpimg;
+        Imagefloat *dest = transformed;
         if (!needsCA() && scale != 1) {
             highQuality = false;
         } else {
             highQuality = true;
             // agriggio: CA correction via the lens profile has to be
-            // performed before all the other transformations (except for the
-            // coarse rotation/flipping). In order to not change the code too
-            // much, I simply introduced a new mode
-            // TRANSFORM_HIGH_QUALITY_CA, which applies *only*
-            // profile-based CA correction. So, the correction in this case
-            // occurs in two steps, using an intermediate temporary
-            // image. There's room for optimization of course...
+            // performed before separately from the the other transformations
+            // (except for the coarse rotation/flipping). In order to not
+            // change the code too much, I simply introduced a new mode
+            // TRANSFORM_HIGH_QUALITY_CA, which applies *only* profile-based
+            // CA correction. So, the correction in this case occurs in two
+            // steps, using an intermediate temporary image. There's room for
+            // optimization of course...
             if (pLCPMap && params->lensProf.useCA && pLCPMap->isCACorrectionAvailable()) {
                 tmpimg.reset(new Imagefloat(original->getWidth(), original->getHeight()));
-                transformLCPCAOnly(original, tmpimg.get(), cx, cy, pLCPMap.get());
-                original = tmpimg.get();
+                dest = tmpimg.get();
             }
         }
-        transformGeneral(highQuality, original, transformed, cx, cy, sx, sy, oW, oH, fW, fH, pLCPMap.get());
+        transformGeneral(highQuality, original, dest, cx, cy, sx, sy, oW, oH, fW, fH, pLCPMap.get());
+        
+        if (highQuality && dest != transformed) {
+            transformLCPCAOnly(dest, transformed, cx, cy, pLCPMap.get());
+        }
     }
 }
 


heckflosse:
@agriggio Alberto, based on your example I would disagree. Your branch shows more blue ca than current dev. But imho the example has not very much ca. We should try with some more example files.

Iirc, lensfun requires a specific order of the corrections. I'll have a look ...

agriggio:
@heckflosse yes, lensfun requires a specific order, which is the reason of this pr. the fact that it matches darktable is a further indication of that. as regards quality, neither of the two is superior in this particular pic, but I suspect in more extreme cases the difference will be more visible. it's just that I didn't have better examples ready...

heckflosse:
@agriggio I just tried using amsterdam.pef and can't see a difference between both branches :+1: .
It's a bit difficult to test on good examples because the example file should have visible ca and also the lensfun profile has to support distortion and ca correction.
I will try to find some more example files.

heckflosse:
Here's another example. I can't see a difference between your branch (middle) and current dev (left). For reference at the right is the result with raw auto ca-correction.
grafik

Edit: another screenshot which includes the uncorrected result
bottom right: no ca correction
top right: raw auto ca-correction
bottom left: lensfun ca-correction your branch
top left: lensfun ca-correction current dev
grafik

heckflosse:
@agriggio Here's another file to try
[D700_20131003_2549.zip](https://github.com/Beep6581/RawTherapee/files/1933725/D700_20131003_2549.zip)

agriggio:
@heckflosse I have no real concrete example where there is a big difference, but I saw someone complaining on dpreview and I thought this could be the issue. unfortunately, that person was totally unhelpful so I don't have much more information :-(

heckflosse:
@agriggio 

> I have no real concrete example where there is a big difference, but I saw someone complaining on dpreview and I thought this could be the issue. unfortunately, that person was totally unhelpful so I don't have much more information :-(

I know ;-)

heckflosse:
@agriggio Actually this unhelpful person cost a lot of your and my time without getting an improvement (don't know if there is something wrong at all as we have no good example). 
How about concentrating on proper reported issues?
Of course with exceptions, as your fix for pdaf stripes proved ;-) Though it wasn't reported here, there was still a lot of support and interest to fix this (great job btw :+1: )

agriggio:
@heckflosse I agree, let's move on. I'll keep the branch in case we get a proper bug report.
btw the main reason for the PDAF stripes fix was that I could reproduce on my a6000 (although it never occurred on normal shots), had it not happened there I would not have bothered :-)

Beep6581:
`PPVERSION` should be bumped.

Looking forward to the merge!

Beep6581:
If the decision is not to merge (at this point), @agriggio what do you think about closing the PR (but keeping the branch)?

agriggio:
fine for me. I'll reopen when (and if) I can find some evidence that this is the way to go

ff2000:
I just wanted to say that in quite contrasty areas my Lumix 12-60 Kit-lens can show quite some CA. I always struggle to get the best correction (I am red-green-blind and just can't define the color of the CA as well as the labels).
Here is an example RAW file:
https://www.dropbox.com/s/7thrtjjl76ukfqa/_1040857.RW2?dl=0
Look at the edges of the cave.

Unfortunately this lens has no lensfun file... But I hope it helps.

heckflosse:
@ff2000 For your example file:

grafik

and

grafik

Beep6581:
Closing PR as agreed.

agriggio:
Here is some more evidence, taken from the thread at https://discuss.pixls.us/t/ca-correction-issue

- CA correction off
no-ca

- RAW CA correction
raw-ca

- current lensfun CA correction
lensfun-old-ca

- new lensfun CA correction
lensfun-new-ca

To me, it seems the latter gives the best results... opinions?

heckflosse:
@agriggio Alberto, I will try to find some time tomorrow to have a look at the new examples

agriggio:
@heckflosse no hurry, enjoy your vacation! :-)

TooWaBoo:
@agriggio 
I've tested some of my images and the result is better than the RAW-CA-Correction. Great Job. 👍

rtengine/iptransform.cc
```diff
@@ -327,25 +327,29 @@ void ImProcFunctions::transform (Imagefloat* original, Imagefloat* transformed,
...
             highQuality = true;
             // agriggio: CA correction via the lens profile has to be
-            // performed before all the other transformations (except for the
-            // coarse rotation/flipping). In order to not change the code too
-            // much, I simply introduced a new mode
-            // TRANSFORM_HIGH_QUALITY_CA, which applies *only*
-            // profile-based CA correction. So, the correction in this case
-            // occurs in two steps, using an intermediate temporary
-            // image. There's room for optimization of course...
+            // performed before separately from the the other transformations
```
Beep6581, line 336: This comment came out squirrelly.

Beep6581:
RawTherapee offers raw CA correction, manual post-raw CA correction, and profiled post-raw CA correction. This branch affects only profiled post-raw CA correction (LCP + lensfun).

agriggio:
Hi, can we decide the fate of this before 5.5? Ping @Beep6581 @heckflosse

Thanatomanic:
Hi @agriggio, based on the provided examples I would say that changing the order improves the end result. And even if there is hardly any improvement, it doesn't seem to hurt to change the order. Also, I like it that it will match with the way lensfun intended ([see the very last paragraph](http://lensfun.sourceforge.net/manual/v0.3.95/corrections.html)).
+1 for merge

heckflosse:
:+1: from me as well

agriggio:
great, thanks! I'll merge tomorrow then

Beep6581:
It makes no visible difference to my Pentax K10D test images, so if it helps you, +1

Issue #4469, opened 2018-03-25
Icon cleanup


Beep6581:
Branch https://github.com/Beep6581/RawTherapee/tree/iconcleanup

1. Remove all obsolete icons from `rtdata/images`
2. Redraw all SVG icons using new Inkscape 0.92 format, with consistent template, grid, metadata, etc.
3. Create SVGs for icons which don't currently have an SVG plate.
4. Revise icon tree structure.
5. Write new scripts to make generating PNGs from SVGs clean and easy (Bash4 for now, Python in the future, not in this issue).

Beep6581:
PR #4556 deletes unused icons.

Beep6581:
Problematic files:
```text
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/unrated.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/ratednot.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/ratednotg.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/fullscreen-exit.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/Light/actions/histChro.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/Light/actions/histChrog.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/Dark/actions/histChro.png' @ warning/png.c/MagickPNGWarningHandler/1665.
identify: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50 `rtdata/images/Dark/actions/histChrog.png' @ warning/png.c/MagickPNGWarningHandler/1665.
[...]

PR #4526, opened 2018-04-28
Removed obsolete icons, #4469

Beep6581:
Unused icons were removed.
Any objections to merge?


Floessie:
I arbitrarily picked `cross.png`: This file is referenced in the sources but no longer installed. Now the hand tool shows a hand instead of a cross cursor. Is that wanted?

Beep6581:
@Floessie thanks for catching that, I'll revise later.

Beep6581:
I will make a new branch.

PR #4541, opened 2018-05-07
Update camera_model_aliases.json

yuki77:
Add Panasonic camera.


Beep6581:
Hi @yuki77 
I can confirm the GX85 and GX80 based on Panasonic's firmware page. Where did you find info on the remaining cameras and aliases?

yuki77:
Hi @Beep6581
I find in [Adobe support page](https://helpx.adobe.com/camera-raw/kb/camera-raw-plug-supported-cameras.html).

Beep6581:
Thanks @yuki77 

Note: we could blindly add other aliases, but there is no guarantee that RT will read the camera name the same way that page reports them. e.g. that page uses `ILCE-A7M2` for the Sony A7II, while the raw files identify themselves (in RT, and using Exiv2 and ExifTool) as `SONY ILCE-7M2`.

PR #4556, opened 2018-05-13
Deleted unused icons

Beep6581:
```bash
for f in rtdata/images/*.png; do
    c="$(grep -rim1 "$(basename $f)" rtgui/*)"
    if [[ -z $c ]]; then
        echo "$f"
    fi
done
```
Manually verified each hit.
Ignored `cglabel*.png`.


Floessie (approved):
I could not find a reference to any of those files erased in the sources.

Beep6581:
Thanks @Floessie

Issue #2467, opened 2015-08-11
Better sorting of padded numbers and fractions in File Browser


Beep6581:
Originally reported on Google Code with ID 2484

```
Padded number sorting could be improved:
http://i.imgur.com/ZNmKFBv.png
```

Reported by `entertheyoni` on 2014-08-28 07:06:04

Beep6581:
The file sort order is still human-unfriendly:
![screenshot_20180516_114827](https://user-images.githubusercontent.com/5844619/40109974-1841d854-58ff-11e8-885b-2b5475ac0466.png)

Expected:
```text
GX9whiteISO200.RW2
GX9whiteISO400.RW2
GX9whiteISO800.RW2
GX9whiteISO1600.RW2
GX9whiteISO3200.RW2
GX9whiteISO6400.RW2
GX9whiteISO12800.RW2
GX9whiteISO25600.RW2
```

Floessie:
@Beep6581 I already have an idea how to solve that...

Beep6581:
Maybe this helps, `sort` from `coreutils` has a `--human-numeric-sort` option:
https://github.com/coreutils/coreutils/blob/master/src/sort.c

Floessie:
@Beep6581 Thanks for the pointer, although I saw it too late. I did it my way. :notes: :wink:

Beep6581:
@Floessie great :)
Note that `case '.'` will fail on locales which use the comma as a decimal separator, and dragons beware for locales which use `.` as a thousands separator (i.e. 0 or more `.` are valid).

Beep6581:
Note that my note could be rubbish, since I didn't read the whole patch ;)

Floessie:
@Beep6581 
> Note that case '.' will fail on locales which use the comma as a decimal separator, and dragons beware for locales which use . as a thousands separator (i.e. 0 or more . are valid).

Yes, the code is "locale-unaware", but it's about filenames, so this shouldn't be a problem for the f-numbers you mentioned in 2014. The thousands separator per se shouldn't be a problem, but more than one `.` is treated as such and not condensed to one. Improvements welcome...

Floessie:
> The thousands separator per se shouldn't be a problem, but more than one `.` is treated as such and not condensed to one.

Thinking a night about it, I stand corrected: It would reverse the order.

Floessie:
[...]

PR #4559, opened 2018-05-16
Sort file numbers naturally (#2467)

Floessie:
Hi,

This is an implementation to sort `ThumbBrowserEntryBase`s naturally as proposed in #2467. It does so by padding number strings to a common length without converting them to actual numbers. ~~Fractions are cared for.~~ My first implementation also included negative numbers, but that was rubbish given ISO dates.

The code is sufficiently simple so that I don't mind, if you dislike it.

Best,
Flössie

### DIFF
--- a/rtgui/thumbbrowserentrybase.cc
+++ b/rtgui/thumbbrowserentrybase.cc
@@ -23,6 +23,153 @@
 
 #include "../rtengine/mytime.h"
 
+namespace
+{
+
+Glib::ustring getPaddedName(const Glib::ustring& name)
+{
+    enum class State {
+        OTHER,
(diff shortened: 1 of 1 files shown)


rtgui/thumbbrowserentrybase.cc
```diff
@@ -23,6 +23,153 @@
...
+
+                    default: {
+                        if (number.size() < pad_width) {
+                            res.append(pad_width - number.size(), '0');
+                        }
+                        res += number;
+                        res += c;
+                        number.clear();
+
+                        state =
```
Beep6581: What a nice clear way of writing this on multiple lines!

PR #4569, opened 2018-05-19
Artifacts when using auto ca-correction on pixelshift files

heckflosse:
When using auto ca-correction on pixelshift files sometimes there are artifacts because the auto ca-correction is called per frame. This pr changes the behaviour and uses the calculated parameters from first frame for all 4 frames.

Example: left is before, right is after
grafik

### DIFF
--- a/rtengine/CA_correct_RT.cc
+++ b/rtengine/CA_correct_RT.cc
@@ -27,7 +27,7 @@
 #include "rawimagesource.h"
 #include "rt_math.h"
 #include "median.h"
-
+#include "StopWatch.h"
 namespace {
 
 bool LinEqSolve(int nDim, double* pfMatr, double* pfVect, double* pfSolution)
@@ -111,7 +111,7 @@ bool LinEqSolve(int nDim, double* pfMatr, double* pfVect, double* pfSolution)
 using namespace std;
 using namespace rtengine;
 
-void RawImageSource::CA_correct_RT(const bool autoCA, const double cared, const double cablue, const double caautostrength, array2D<float> &rawData)
+void RawImageSource::CA_correct_RT(const bool autoCA, const double cared, const double cablue, const double caautostrength, array2D<float> &rawData, double *fitParamsTransfer, bool fitParamsIn, bool fitParamsOut, float *buffer, bool freeBuffer)
 {
 // multithreaded and vectorized by Ingo Weyrich
     constexpr int ts = 128;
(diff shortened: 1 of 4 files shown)


heckflosse:
As a side effect the processing time for auto ca-correction of a 42.3 MP pixelshift file went down from ~1080 ms to ~660 ms.

Floessie:
@heckflosse Ingo, I'm not at home during the next week, so I can only review this a week later. As I see allocated memory newly shared between functions, this calls for a review. :wink:

Floessie:
> As a side effect the processing time for auto ca-correction of a 42.3 MP pixelshift file went down from ~1080 ms to ~660 ms.

Great result, BTW! :+1:

heckflosse:
@Floessie 

> As I see allocated memory newly shared between functions, this calls for a review.

Exactly for that reason I requested a review from you and Alberto ;-)

Edit: no need to hurry

rtengine/rawimagesource.h
```diff
@@ -245,7 +245,7 @@ class RawImageSource : public ImageSource
     inline  void interpolate_row_rb     (float* ar, float* ab, float* pg, float* cg, float* ng, int i);
     inline  void interpolate_row_rb_mul_pp (float* ar, float* ab, float* pg, float* cg, float* ng, int i, float r_mul, float g_mul, float b_mul, int x1, int width, int skip);
 
-    void CA_correct_RT  (const bool autoCA, const double cared, const double cablue, const double caautostrength, array2D<float> &rawData);
+    float* CA_correct_RT  (const bool autoCA, const double cared, const double cablue, const double caautostrength, array2D<float> &rawData, double *fitParamsTransfer = nullptr, bool fitParamsIn = false, bool fitParamsOut = false, float * buffer = nullptr, bool freeBuffer = true);
```
Floessie: `CA_correct_RT` is only used in line 2020 et seq. of `rtengine/rawimagesource.cc`, so I don't think there's a need for all those defaults. It's harder to review and debug when default arguments are involved. Maybe consider approriate forwarding overloads.

rtengine/rawimagesource.cc
```diff
@@ -2015,9 +2015,13 @@ void RawImageSource::preprocess  (const RAWParams &raw, const LensProfParams &le
...
         if(numFrames == 4) {
-            for(int i=0; i<4; ++i) {
-                CA_correct_RT(raw.ca_autocorrect, raw.cared, raw.cablue, 8.0, *rawDataFrames[i]);
+            StopWatch Stop1("ps ca correction");
+            double fitParams[64];
+            float *buffer = CA_correct_RT(raw.ca_autocorrect, raw.cared, raw.cablue, 8.0, *rawDataFrames[0], fitParams, false, true, nullptr, false);
+            for(int i = 1; i < 3; ++i) {
+                CA_correct_RT(raw.ca_autocorrect, raw.cared, raw.cablue, 8.0, *rawDataFrames[i], fitParams, true, false, buffer, false);
             }
+            CA_correct_RT(raw.ca_autocorrect, raw.cared, raw.cablue, 8.0, *rawDataFrames[3], fitParams, true, false, buffer);
```
Floessie: This hasn't become nicer, but it's sane memory-wise. :wink:

rtengine/CA_correct_RT.cc
```diff
@@ -1153,14 +1184,27 @@ void RawImageSource::CA_correct_RT(const bool autoCA, const double cared, const
...
-    free(blockwt);
-    free(RawDataTmp);
+    if(freeBuffer) {
+        free(buffer);
+    }
 
     if(plistener) {
         plistener->setProgress(1.0);
     }
+    return buffer;
```
Floessie, line 1210: Better return `nullptr` if you freed the pointer.

Floessie (changes requested):
Moving a `std::unique_ptr<>` in and out could make things clearer. Memory handling is okay, though. The only thing I'd really like you to change is the return of an already freed pointer.

rtengine/CA_correct_RT.cc
heckflosse: Done

Issue #4587, opened 2018-06-05
Strange color artifacts in Jupiter


luisfl:
RT version: rawtherapee-5.4-5.7.x86_64 in fully updated Fedora 28 (distro RPM package).

When I open this ORF image of Jupiter with RawTherapee,

http://luis.impa.br/photo/lixo/Dated_JN-180601-B_21202.orf

RT shows lots of color artifacts, independently of the demosaicing algorithm.
The same RAW file processed with Darktable or Dcraw (through UFRAW) shows no artifacts (see attached jpg crops at 800%).
I have only observed all these artifacts with astrophotography of planets, and only recently. It seems old RT versions did not show this problem.

![dated_jn-180601-b_21202-dt](https://user-images.githubusercontent.com/15525268/40948396-a445b70a-683e-11e8-8965-974a239e86d6.jpg)
![dated_jn-180601-b_21202-rt](https://user-images.githubusercontent.com/15525268/40948397-a46631e2-683e-11e8-9223-68c505a1f1ce.jpg)
![dated_jn-180601-b_21202-ufraw](https://user-images.githubusercontent.com/15525268/40948398-a482c596-683e-11e8-8a1b-9739dd284ad7.jpg)

agriggio:
are you using auto-ca correction by any chance? (just guessing)

Thanatomanic:
That doesn't seem to matter. I tested this on the latest build and I can reproduce the weird behaviour with all demosaicing modes and with everything else off...

gaaned92:
VNG4 seems OK

Beep6581:
darktable uses PPG by default for .orf files.

Floessie:
Color aside: At least you can spot Ganymed on the RT rendition. :grin:

agriggio:
Try with green equilibration set to 100% and a couple of steps of false colour suppression:

- AMaZe:
![amaze](https://user-images.githubusercontent.com/11982086/40966741-9b0a9a02-68b1-11e8-9182-f548f2869d8c.jpeg)

- LMMSE:
![lmmse](https://user-images.githubusercontent.com/11982086/40966748-9fcc69bc-68b1-11e8-8165-629ea51c638e.jpeg)

Beep6581:
@luisfl could you upload some other shots from this camera - both astro and non-astro?

heckflosse:
Neutral profile + following patch gives this

[...]

PR #4593, opened 2018-06-07
Move green equilibration info to camconst.json

agriggio:
As the title says, this moves the info about whether global green equilibration is needed for the current camera from `rawimagesource.cc` to `camconst.json`. I've added a new key `"global_green_equilibration"`, taking `0` or `1` as value. Currently this is enabled for all Panasonic cameras and all Olympus cameras whose model name starts with E, as was the case before. This might need to be revised though (see #4587).

In a separate commit I've also fixed the case of the keys `"pdafPattern"` and `"pdafOffset"`, to make them consistent with the rest of the file, using `lowercase_and_unserscores`.

### DIFF
--- a/rtengine/camconst.cc
+++ b/rtengine/camconst.cc
@@ -311,17 +311,17 @@ CameraConst::parseEntry(void *cJSON_, const char *make_model)
         }
     }
 
-    ji = cJSON_GetObjectItem(js, "pdafPattern");
+    ji = cJSON_GetObjectItem(js, "pdaf_pattern");
 
     if (ji) {
         if (ji->type != cJSON_Array) {
-            fprintf(stderr, "\"pdafPattern\" must be an array\n");
+            fprintf(stderr, "\"pdaf_pattern\" must be an array\n");
             goto parse_error;
         }
 
         for (ji = ji->child; ji != nullptr; ji = ji->next) {
             if (ji->type != cJSON_Number) {
-                fprintf(stderr, "\"pdafPattern\" array must contain numbers\n");
+                fprintf(stderr, "\"pdaf_pattern\" array must contain numbers\n");
                 goto parse_error;
             }
 
@@ -329,11 +329,11 @@ CameraConst::parseEntry(void *cJSON_, const char *make_model)
         }
     }
 
-    ji = cJSON_GetObjectItem(js, "pdafOffset");
+    ji = cJSON_GetObjectItem(js, "pdaf_offset");
 
     if (ji) {
         if (ji->type != cJSON_Number) {
-            fprintf(stderr, "\"pdafOffset\" must contain a number\n");
+            fprintf(stderr, "\"pdaf_offset\" must contain a number\n");
             goto parse_error;
         }
 
--- a/rtengine/camconst.cc
+++ b/rtengine/camconst.cc
@@ -25,6 +25,7 @@ CameraConst::CameraConst() : pdafOffset(0)
     memset(raw_crop, 0, sizeof(raw_crop));
     memset(raw_mask, 0, sizeof(raw_mask));
     white_max = 0;
+    globalGreenEquilibration = -1;
 }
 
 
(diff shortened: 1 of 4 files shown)


rtengine/camconst.cc
```diff
@@ -339,6 +340,17 @@ CameraConst::parseEntry(void *cJSON_, const char *make_model)
         cc->pdafOffset = ji->valueint;
     }
 
+    ji = cJSON_GetObjectItem(js, "global_green_equilibration");
+
+    if (ji) {
+        if (ji->type != cJSON_Number) {
+            fprintf(stderr, "\"global_green_equilibration\" must be a number\n");
```
Floessie: @agriggio Are you planning for something different than `0` and `1` in the future? If not, why not use a JSON boolean here?

rtengine/rawimagesource.cc
```diff
@@ -1945,8 +1946,16 @@ void RawImageSource::preprocess  (const RAWParams &raw, const LensProfParams &le
         }
     }
 
-    // check if it is an olympus E camera or green equilibration is enabled. If yes, compute G channel pre-compensation factors
-    if ( ri->getSensorType() == ST_BAYER && (raw.bayersensor.greenthresh || (((idata->getMake().size() >= 7 && idata->getMake().substr(0, 7) == "OLYMPUS" && idata->getModel()[0] == 'E') || (idata->getMake().size() >= 9 && idata->getMake().substr(0, 9) == "Panasonic")) && raw.bayersensor.method != RAWParams::BayerSensor::getMethodString( RAWParams::BayerSensor::Method::VNG4))) ) {
+    // check if green equilibration is needed. If yes, compute G channel pre-compensation factors
+    const auto globalGreenEq =
+        [&]() -> bool
```
Floessie, line 1951: Maybe reduce the lambda capture to `[&ri]`. See Scott Meyers' "Effective modern C++, Item 31" for some, I admit, "constructed" examples. Nevertheless: The more precise the better to read and optimize.

rtengine/camconst.cc
```diff
@@ -606,6 +618,24 @@ CameraConst::get_WhiteLevel(const int idx, const int iso_speed, const float fnum
     return lvl.levels[idx];
 }
 
+bool
+CameraConst::has_globalGreenEquilibration()
+{
+    return globalGreenEquilibration >= 0;
```
Floessie, line 624: I know the `-1` "trick" is very common. A nicer solution will be [`std::optional<>`](http://en.cppreference.com/w/cpp/utility/optional) in C++17. In the meantime, we could add our own handy `Optional<>`:

```c++
template<class T>
class Optional
{
public:
    Optional() :
        present(false),
        value(T())
    {
    }

    Optional(const T& _value) :
        present(true),
        value(_value)
    {
    }

    ~Optional() = default;

    Optional(const Optional& other) = default;
    Optional& operator =(const Optional& other) = default;

    explicit operator bool() const
    {
        return present;
    }

    const T& operator *() const
    {
        return value;
    }

    T& operator *()
    {
        return value;
    }

    const T* operator ->() const
    {
        return &value;
    }

    Optional& operator =(const T& new_value)
    {
        present = true;
        value = new_value;
        return *this;
    }

    void reset()
    {
        present = false;
        value = T();
    }

private:
    bool present;
    T value;
};
```

**Edit:** I removed the `getValue()` function for a cleaner interface.

rtengine/camconst.cc
agriggio: @Floessie I looked for it in cJSON but could not find it... perhaps I missed it?

Floessie: @agriggio Yes, it seems a bit tricky, but googling revealed [this](https://sourceforge.net/p/cjson/discussion/998969/thread/3fd954a7/#cc36/4230) and [this](https://sourceforge.net/p/cjson/patches/3/#3c31) (BTW: Simon Budig is one of the Gimp developers).

PR #4594, opened 2018-06-07
add bayer Amaze+vng4 and xtrans 4pass demosaicers

heckflosse:


heckflosse:
Merged to get more tests.

PR #4605, opened 2018-06-13
reduce the range of the lab grid colour toning method

agriggio:
This allows for more fine-tuning where it matters most. 
I have never needed to use the extreme settings of the tool: moving the markers close to the edges of the grid gives a very strong colour cast to the images; this might have a use for artistic purposes, but for "colour corrections" it is not so useful IMHO. For the latter use case, the markers usually stay very close to the center. With this PR, I reduced the intensity of the correction, so that the full grid (well, almost) can be used effectively for subtle colour corrections.

If you want stronger effects, there are still the other colour toning methods available.

I had to bump the ppversion because the parameters need to be rescaled for this to work.

### DIFF
--- a/rtengine/improcfun.cc
+++ b/rtengine/improcfun.cc
@@ -5740,10 +5740,11 @@ void ImProcFunctions::lab2rgb (const LabImage &src, Imagefloat &dst, const Glib:
 void ImProcFunctions::colorToningLabGrid(LabImage *lab, int xstart, int xend, int ystart, int yend, bool MultiThread)
 {
     const float factor = ColorToningParams::LABGRID_CORR_MAX * 3.f;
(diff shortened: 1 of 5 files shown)


heckflosse:
@agriggio I really like the way it works now, but when opening a file where an extreme setting (near the border) was used, I'm not able to change it because I can't reach the node.

grafik

agriggio:
@heckflosse yeah, I was hoping nobody was actually using those extreme settings :stuck_out_tongue_winking_eye: 
One possibility would be to simply clamp in the GUI, what do you think?

heckflosse:
@agriggio those extreme settings were just to test. No objection to clamp here :)

heckflosse:
:+1: for merge

PR #4611, opened 2018-06-14
Update Deutsch locale

TooWaBoo:


rtdata/languages/Deutsch
```diff
@@ -749,21 +750,21 @@ HISTORY_MSG_441;(Retinex) - Einstellungen\nVerstärkung und Ausgleich\nTransmiss
...
-HISTORY_MSG_464;(Sensor-Matrix)\nFarbinterpolation\nUnschärfebewegungsmaske
-HISTORY_MSG_465;(Sensor-Matrix)\nFarbinterpolation\nUnschärferadius
-HISTORY_MSG_468;(Sensor-Matrix)\nFarbinterpolation\nLücken in der Bewegungs-\nmaske erkennen
-HISTORY_MSG_469;(Sensor-Matrix)\nFarbinterpolation\nMedian
-HISTORY_MSG_471;(Sensor-Matrix)\nFarbinterpolation\nBewegungskorrektur
-HISTORY_MSG_472;(Sensor-Matrix)\nFarbinterpolation\nWeicher Übergang
-HISTORY_MSG_473;(Sensor-Matrix)\nFarbinterpolation\nLMMSE für Bewegungs-\nteile verwenden
-HISTORY_MSG_474;(Sensor-Matrix)\nFarbinterpolation\nFrame-Helligkeit angleichen
-HISTORY_MSG_475;(Sensor-Matrix)\nFarbinterpolation)\nAusgleich pro Kanal
+HISTORY_MSG_445;(Sensor-Matrix)\nFarbinterpolation - PS\nUnterbild
```
heckflosse: There are also non PS files with more than one frame. For example dual-pixel files from canon EOS 5D Mark IV

TooWaBoo: Ok, I'll change it.

TooWaBoo:
Changed

heckflosse:
Thank you!

rtdata/languages/Deutsch
```diff
@@ -1000,7 +1002,7 @@ PARTIALPASTE_RAW_DMETHOD;Farbinterpolation: Methode
 PARTIALPASTE_RAW_FALSECOLOR;Farbinterpolation: Falschfarbenreduzierung
 PARTIALPASTE_RAW_IMAGENUM;Farbinterpolation: Unterbild
 PARTIALPASTE_RAW_LMMSEITERATIONS;Farbinterpolation: LMMSE-Verbesserungsstufen
-PARTIALPASTE_RAW_PIXELSHIFT;Farbinterpolation: Pixel Shift
+PARTIALPASTE_RAW_PIXELSHIFT;Farbinterpolation: Pixel-Shift
```
Beep6581, line 1005: @TooWaBoo why have you hyphenated it? All camera manufacturers who use this term use it without hyphenation (even though the practice of pixel shifting is over 10 year old).

TooWaBoo:
In German connected substantives are written together like **Pixelshift** .
In some cases e,g more than 2 substantives, they can be connected with a hyphen like **Pixel-Shift-Kamera** .
To split substantives with a space (like in English) is uncommon in German even for terms.
Terms from other languages can be written together or combined with a hyphen, even if they consist of only two substantives. This is the normal spelling in German.
**Pixelshift** or **Pixel-Shift** both is correct. **Pixel Shift** isn't wrong 'cause it's not a German word but it's not common in German.

Beep6581:
@TooWaBoo The way we use "Pixel Shift" in `default` and the way camera manufacturers use it is not as a noun phrase or noun+verb, but as a trade name, ergo the capitalization. I wasn't aware trade names were allowed to be changed in German, i.e. the same should apply to General Motors, 21st Century Fox, Rolls-Royce and Taco Bell.

Floessie:
@Beep6581 Counter-example from [Sony](https://support.d-imaging.sony.co.jp/support/ilc/psms/de/index.html). :smiley:

Beep6581:
Fair enough
https://support.d-imaging.sony.co.jp/support/ilc/psms/en/index.html

Beep6581:
c3

c3 1

TooWaBoo:
The German grammer has as many rules as exceptions. 😁
What should I use:
1. Pixelshift
2. Pixel-Shift
3. Pixel Shift
4. PixelShift (this is wrong)

Beep6581:
Up to you :) Seeing as even the manufacturer translations aren't consistent. I will be using "Pixel Shift" in the English version, but use whatever feels correct in German.

TooWaBoo:
Ok, I take 2. :-)

PR #4613, opened 2018-06-15
increased precision of the channel mixer (allow 1 decimal figure)

agriggio:
I'd like to have a bit more precision in the channel mixer. This PR adds one significant figure. 
I hope there are no objections...

### DIFF
--- a/rtengine/improcfun.cc
+++ b/rtengine/improcfun.cc
@@ -2255,15 +2255,15 @@ void ImProcFunctions::rgbProc (Imagefloat* working, LabImage* lab, PipetteBuffer
 
     float Balan = float (params->colorToning.balance);
 
-    float chMixRR = float (params->chmixer.red[0]);
-    float chMixRG = float (params->chmixer.red[1]);
-    float chMixRB = float (params->chmixer.red[2]);
-    float chMixGR = float (params->chmixer.green[0]);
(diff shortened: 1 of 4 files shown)


heckflosse:
:+1: from me

iliasg:
@agriggio Alberto, thanks for the precision increase
I would also like an increase in the range from -200..200 to a much bigger (say -500..500) because the current range is many times restrictive when we want to apply a 3X3 color correction matrix via the channel mixer ..

agriggio:
@iliasg that's exactly my use case :-) though for me -200,200 has been enough so far. but I can definitely increase the range if needed :+1:

iliasg:
@agriggio Alberto, I've met a lot of cases where -200,200 is not enough
To summarize my experience .. maybe less than 80% of cases are fine with -200,200 .. -300,300 will cover more than 95% .. some extreme cases (say underwater) may need more .. 

BTW I always wondered why this base 100 is used instead of 1.000 .. will changing it at the present situation arise compatibility issues ??.

agriggio:
@iliasg understood! changing the base to 1000 is exactly what I did, except that the gui shows 100.0 instead :-) I think this is less confusing for people used to the old way. I'll extend the range to -500,500 when I get back tomorrow night

Issue #3547, opened 2016-12-18
Scalable graphics - HiDPI


Beep6581:
The aim of this issue is to get RawTherapee to look well when scaled up, something which is becoming increasingly common. The change should not cause a performance penalty.

Reading SVG icons is slow, but once they are read they will be drawn just as fast as PNG icons. As startup time does matter to us, we will use PNG icons. The icons will be stored as SVG in the repository, but will be converted to PNG for builds for performance reasons.

Steps:
- [x] Clean up icon situation, #4469
- [x] Use a script to render PNGs of various sizes, optimize, maybe create iconsets.
- [x] Remove `rtdata/images/*` and `tools/source_icons`
- [x] Keep SVGs separate from PNGs.
- [x] Do not ship SVGs with builds.
- [x] Update CSS themes as needed.
- [x] Implement HiDPI support in RT code - the UI including the preview and thumbnails should scale correctly.

https://wiki.gnome.org/HowDoI/HiDpi

Hombre57:
How long do you expect to support Gtk2 ?

Hombre57:
I definitely vote for create a single iconset file (one for Dark, one for Light... do we still want to support the Light theme btw ?)

Beep6581:
I discussed this with Ingo the other day. RT5 should have a GTK2 version, but after that I'll only invest time in merging as long as the time required is short - there are better things in life than supporting GTK2.

Basically we will decide at some point in the near future to commit all changes to the `gtk3` branch first (we can rename `master` to `gtk2` for clarity, we could also rename `gtk3` to `master` though that is not a Git requirement and it could be confusing when looking at Git history), and then I could merge changes from `gtk3` to `gtk2` as long as the merge conflicts do not require much time to solve. Once we hit a merge conflict which would require much time to solve, I will cease merging. Of course anyone is free to take over.

Beep6581:
Considering all RT themes are dark and for good reason, I don't see a need to complicate things by shipping a light theme.

reagle:
I updated to Rawtherapee 5.2 and the browser thumbnails are no longer 1/4 size as mentioned in #3936 . Thanks! Everything is much larger now, so I then configured the font sizes down and it looks great.

samdoshi:
I've done a little bit of investigation on this.

For example the following diff applied to `thumbbrowserentrybase.cc`:

```diff
[...]

Issue #1598, opened 2015-08-11
RAW histogram, clipping indicators, and RAW white point correction


Beep6581:
Originally reported on Google Code with ID 1614

```
There are no clipping indicators in RAW histogram, even for images where all RAW data
are certainly saturated.

There is also, I think, a related issue: the the RAW Tab/White Point linear correction
slider:  My understanding is that it adjust the value at which the raw data is considered
fully maxed-out.  The RAW histogram doesn't seem to react to it - shouldn't it? (it
reacts to Black Point corrections).

I think it would be useful to explicitly indicate (via RAW histogram warning)if and
to what degree the original raw data is maxed out.

BTW: is the default saturation level in raw histogram taken as the default for a camera
hard-coded in dcraw and is this the RHS y-axis in RAW histogram?
```

Reported by `mcjtom` on 2012-11-03 12:57:30

RawConvert:
I was looking at #4667 and thought it would be good to help the user keep an eye on raw values in relation to white point.  I thought this was the appropriate issue to update.  I think it's useful to see actual raw values straight as they've come from the camera.  This is to a degree possible by selecting None as the demosaic method and cursoring around the image.  However -
1. if the RGB values are set to % or [0-1], then I am getting silly values like 5230.3%.
2. if set to [0-255] I am getting e.g. 13338.  13338 is probably correct as it's an overexposed area and from memory 13338 is around the white point.
3. The "13338" depends on the white point correction slider.  Is it truly showing the camera value if the slider is at 1.00?  Should it depend on the slider when you've selected the raw view of the histogram?

Issue #1596, opened 2015-08-11
Histogram X-axis scale in EV stops


Beep6581:
Originally reported on Google Code with ID 1612

```
An enhancement proposal: to scal histogram grid/x-axis divisions somewhere closer to
EV stops to make it more relevant photographically (in practice, I guess the histogram
shows gamma-corrected pixel values, so perhaps equally spaced, but with 10-12 divisions,
which should be close enough?)

BTW: how is the x-axis scaled in RAW histogram?  And where the minimum and maximum
clipping ends of the histogram come from (dark level? and saturation? for the camera?
Can that be adjusted by hand?
```

Reported by `mcjtom` on 2012-11-02 10:44:57

Thanatomanic:
This was implemented more or less as described above by #4618

PR #4618, opened 2018-06-15
Histogram update: different scaling modes and a new look

Thanatomanic:
Hi all, as hinted and encouraged [here](https://discuss.pixls.us/t/technical-explanation-histogramarea-histogramrgbarea/7896) I did some work on the look and feel of the current histogram.
A short summary of the changes:
- Changed the aspect ratio to 1:1 so that more details can be made visible when resizing the sidebar. This should pose little interface problems, even on smaller screens.
- Completely removed fullMode, and replaced this with three scaling modes that can be toggled by a button (icons still needed!) or by double clicking the histogram. Available modes: normal, logarithmic and a magic sauce (complicated cubic function...).
- Disabled the luma curve by default. This is a personal preference - things seemed cluttered enough.
- Revamped the look of the graphs itself, inspired by the histogram in darktable. The graphs are now drawn filled with additive coloring. I am not entirely sure what benefit this may have, but I kinda like it better than the single lines. Disagreement is anticipated ;-)

Let me know what you think. Comments and advice are more than welcome!

Edit: a preview image
image
Edit2: this graphical change has already been reverted


Thanatomanic:
I already see that I need to tell my text editor to get rid of tabs and replace them by spaces... Things appear messy now, right?

agriggio:
hi,

> Disagreement is anticipated ;-)

good guess ;-) I find this harder to read.
but thanks a lot for the other mods! (except I'd put the luma curve back...)

TooWaBoo:
The histogram looks good but its too big. It takes to much space. I prefer the old size.

grafik

TooWaBoo:
The TooWaBlue theme is using the name: #fullButton
If the histogram height is not a multiple of the button height I use that for the compensation.
Please add this back
...->set_name("fullButton");
Add this name to the "bar" button.

There is an issue with the "bar" button icon and "mode/full" button icon too.
The "bar" and "mode/full" button should be swapped.

Thanatomanic:
@TooWaBoo I hadn't considered placement on the right. It _does_ seem a bit too much there.
I see three alternatives then, and I would like hear some opinions:

1. Revert the change in aspect ratio completely.
2. Only make the change when the histogram is on the left.
3. Invest some effort into making the histogram vertically scalable (like the snapshots panel is currently).

I would not like option 1, because on high resolution / ultrawide monitors the histogram is just too tiny to read. Option 2 seems okay. Option 3 is the most flexible, but I would need somebody to point me in the right direction on how to make that work. Maybe there is an option 4?

> There is an issue with the "bar" button icon and "mode/full" button icon too.
The "bar" and "mode/full" button should be swapped.

Done

> The TooWaBlue theme is using the name: #fullButton
If the histogram height is not a multiple of the button height I use that for the compensation.
Please add this back
...->set_name("fullButton");
Add this name to the "bar" button.

I looked at your CSS, and I see some extra padding for #fullButton, but I see no visual differences between the old situation (with fullButton intact) and my new situation. Where should this padding be?
image

TooWaBoo:
Without padding it looks like this:
unbenannt
The one pixel missing at the bottom.

Beep6581:
1. >Changed the aspect ratio to 1:1

    It's nice having the possibility of a larger histogram, especially considering HiDPI support is on its way #3547 #4469. Having a scalable histogram could be nice, or we could have a setting in Preferences for the histogram's aspect ratio: 2:1 or 1:1, or toggle the aspect ratio via double-click, or something.
2. >Completely removed fullMode, and replaced this with three scaling modes

    :+1: 
3. >Disabled the luma curve by default

    Need to give this more thought, but I'm currently leaning toward agreeing. See for example http://rawtherapee.com/shared/test_images/tracteur.dng
screenshot_20180616_094548
The luminance histogram looks very different to the others, but I can't say that it gives me any vital info in most cases, as in most cases I just care about not clipping the colors and having the histogram touch both ends.
This would not be an issue at all is RT remembered the last-used histogram settings between sessions.
4. > Revamped the look of the graphs itself

    They might look nicer in screenshots but are less clear, so I disagree here, the not-filled look is easier to work with.
5. >I already see that I need to tell my text editor to get rid of tabs and replace them by spaces... Things appear messy now, right?

    No tabs, 4 spaces. Refer to this file for the style: 
[rawtherapee.astylerc](https://github.com/Beep6581/RawTherapee/blob/dev/rawtherapee.astylerc)
[astyle documentation](http://webcache.googleusercontent.com/search?q=cache:mHMuDTdZnI4J:astyle.sourceforge.net/astyle.html+&cd=1&hl=en&ct=clnk&gl=se&lr=lang_de%7Clang_en)
If you do run the file through astyle, be careful that you don't commit unnecessary changes, so that this doesn't cause merge problems for the [many open branches](https://github.com/Beep6581/RawTherapee/branches/all).
5. > Please add this back
...->set_name("fullButton");
Add this name to the "bar" button.

    It needs a name, but "fullButton" is no longer descriptive of what it does, so it needs a better name.

Please don't touch any icon files in this branch, as I'm plowing the whole icon-field in #4469.

iliasg:
Nice to see work progressing on histogram .. thanks.
Is the log scaling a log10 or log2 ??. I think the later is more usefull (I mean horizontal scaling mostly .. but vertical would also be usefull)

TooWaBoo:
I've modified my theme.
unbenannt
[TooWaBlue-HistMod-GTK3-20_.css.txt](https://github.com/Beep6581/RawTherapee/files/2108025/TooWaBlue-HistMod-GTK3-20_.css.txt)

The name #fullbutton is'nt needed anymore.

I prefer: 3. Invest some effort into making the histogram vertically scalable (like the snapshots panel is currently).

Thanatomanic:
@Beep6581
3. I'll await the consensus of the devs on disabling luma by default. So far it's 1 vs 1.
4. The graphical changes are not getting enough love so far, so I'd better remove them again. Was worth a try though and fun delving into GTK.
And understood about the icons. I'm following the development closely :-)

@iliasg This is log10, I can try log2. And are you suggesting to also apply the scaling horizontally? Certainly possible, but what benefit do you see from that?

iliasg:
@Thanatomanic 
- For horizontal axis .. log2 (i.e. stops) is a natural way of photographic thinking and calculations .. 
- for vertical .. just allows representation of bigger populations than linear (but less than log10) but with less compression than log10

TooWaBoo:
@Thanatomanic , @everyone
Found a css solution for auto size the histogram depending on dpi/font size.
The histogram is higher on the left pane as on the right.
Can you test this theme with the  dev version of RT, please?

[HistAutoHight-GTK3-20_.css.txt](https://github.com/Beep6581/RawTherapee/files/2108727/HistAutoHight-GTK3-20_.css.txt)
Remove .txt after downloading.

Thanatomanic:
@TooWaBoo We've been working on a solution in parallel... I've implemented a Gtk::Paned solution, which probably doesn't require your elegant CSS solution. I haven't tested your CSS, so I cannot judge which approach works best.

For the rest: I've made a big update.

1. The histogram can now be scaled to arbitrary sized (with some lower limits), both on the left and the right side.
2. Based on @iliasg 's suggestion, I've ditched the cube-scaling mode and implemented a double-log mode. What do you think?
3. Gridlines now magically multiply to aid the legibility of the histogram. Good idea?
4. The histogram is made up of lines again, and not filled.

_N.B. in the process of digging through the GTK3+ documentation, I found out that RT uses deprecated widgets in many places (i.e. VBox, HBox, VPaned, HPaned). I modified them in this commit, because I was changing stuff with the layout anyway. Hope this is okay..._

TooWaBoo:
I've testet your changes and it looks good except the default size of the histogram.
I prefer the smallest size as default or the last used size should be stored in the option file.
I adapted my theme to your changes.
[TooWaBlueHiDPI-GTK3-20_.css.txt](https://github.com/Beep6581/RawTherapee/files/2108978/TooWaBlueHiDPI-GTK3-20_.css.txt)

Thanatomanic:
@TooWaBoo Your CSS works great!
I also tried to go back to the default 2:1 aspect ratio, but `Gtk::Paned` keeps resetting stuff to a default aspect ratio and ignoring `HistogramArea::get_preferred_height_for_width_vfunc()`. I will try to look into it the coming days, but help from a Gtk guru would be welcome.

Also, what do others think about storing the last used aspect ratio of the histogram in the options?

Beep6581:
+1 for storing the last-used settings across RT sessions.

TooWaBoo:
@Thanatomanic 
I'm not a Gtk Guru but I found the culprit.
At line 747 change

```
    if (gHeight < 100) {
        gHeight = 100;
    }
```
to
```
    if (gHeight > 150) {
        gHeight = 150;
    } else if (gHeight < 100) {
        gHeight = 100;
    }
```

Beep6581:
@Thanatomanic what is the status of the work in this PR - ready for merge?

Thanatomanic:
I need a little more time to work on this PR to test and include the fix suggested by @TooWaBoo and to make sure the histogram size is stored across sessions. Estimated time is roughly another week from now.

Thanatomanic:
@TooWaBoo That does not seem to work. I can set `histogramArea->set_vexpand(false);` in the beginning of `HistogramPanel` constructor, but then it looks like this:
image

I need to find a way that the parent `Gtk::Paned` follows the size of its children. But so far the Gtk docs haven't helped me a lot.

TooWaBoo:
@Thanatomanic 
Try this: [histogrampanel.cc.txt](https://github.com/Beep6581/RawTherapee/files/2142514/histogrampanel.cc.txt)
The changes start at line 465 and 747.
grafik

grafik

Thanatomanic:
I think I have the problem fixed now. The default aspect ratio for the histogram is now the golden ratio and the aspect ratio should be stored between sessions. 
In the process I had to change the scaling behavior of the histogram, because I didn't get it to work otherwise. Now there is a decoupling of the height and width of the histogram, i.e. the aspect ratio is not maintained when you resize the left or right toolbars. I hope this is an acceptable change.

One oddity that could be fixed: when the histogram switches sides, it keeps the aspect ratio and not the height. This might look weird at first.

Please test and comment :-)

TooWaBoo:
@Thanatomanic 
Looks good so far but there is an issue when you switch the window from "normal" to "maximize" or "fullscreen". The histogramm height increases by 18px. If you start RT in maximize the histogramm will grow 18px each time.

Thanatomanic:
Damn that's a bad side-effect, thanks for testing. It actually also happens when you start multiple times without having the window maximized.
I have an idea how this happens, but I need to investigate.

Thanatomanic:
@TooWaBoo Could you test again please? I now simply store the height of the `histogramPanel` instead of the aspect ratio.

TooWaBoo:
@Thanatomanic 
Same behaviour except now it's 11px off.

btw. the chroma curve is black and hard to see. I suggest to change the color to orange.
grafik

heckflosse:
> Same behaviour except now it's 11px off.

I know why I don't like gui coding.....

Thanatomanic:
@heckflosse All is new for me, so I'm not spoiled yet ;-)

@TooWaBoo, can you check if this happens with all themes? I really thought I didn't see the effect anymore on my machine. G'night!

TooWaBoo:
@Thanatomanic
I've tested with all themes and all themes are affected.

Thanatomanic:
Could others please test this as well? I cannot reproduce the behavior @TooWaBoo is seeing.
@Beep6581 @heckflosse @iliasg @agriggio ?

Please test if you get a consistent height of the histogram when you restart RT. Try this for different themes, having the RGB-bar on/off, having the histogram left/right.

Edit: could this be OS dependent? I'm on Win 7, and using a fully updated MSYS2 environment.

TooWaBoo:
I'm on Win10 and MSYS2 is fully updated too.

EDIT: When running RT not maximized it works fine.
I guess it has to do with the height of the title bar on Win10.
normal
grafik
maximized
grafik

heckflosse:
@TooWaBoo 
>If you start RT in maximize the histogramm will grow 18px each time.

Confirmed on Win7. I didn't count the px, but it grows each time.

TooWaBoo:
Here you can download a compiled version for Windows.
https://filebin.net/ee5kc16zl0162s9r

Hombre57:
@Thanatomanic Just in case you want to simplify life, I created a utility function named `setExpandAlignProperties` added in `guiutils.cc`, that let you specify those properties in a single line.

I know that they are separately set in `histogrampanel.cc`, I did it :sweat_smile: , but  one day I'll simplify the code using that function.

Also, why [setting a null size here](https://github.com/Thanatomanic/RawTherapee/blob/da5fd6fac850fa964a881139b98aa5c7f805274c/rtgui/histogrampanel.cc#L741) ? `get_preferred_height_vfunc(minimum_height, natural_height);` should be used instead, to set minimal values.

At some point, we'll add a function that will tell if the RTWindow is displayed on HiDPI resolution, w/o upscaling, if that can be of any interest for choosing the size of the Histogram.

Thanatomanic:
@TooWaBoo Thanks for the compiled version. I am puzzled though. On my main display (3440x1440) I really have _zero_ change in the histogram height when RT is maximized. However, on my secondary display (1920x1080) I now see the same changes everybody here is seeing. I am beginning to dislike this :smile:  Back to the drawing board.

@Hombre57 Good suggestions, I'll take the helper function it into account.
Setting null values makes no difference, since I changed the return value of `get_request_mode_vfunc()` to `Gtk::SIZE_REQUEST_CONSTANT_SIZE`. This means `get_preferred_height_for_width_vfunc()` doesn't even get called anymore.

Thanatomanic:
Another attempt, please test yet again. The trick seemed to be in the properties of the `Gtk::Paned` children. They no longer resize when the pane-size changes (e.g. by maximizing).

Interestingly, when RT is started maximized, I found out that RT (or GTK) calculates the sizes of the histogram _twice_ (and probably more GUI elements). First for the un-maximized window, and then for the maximized window. Odd?

heckflosse:
@Thanatomanic Fix confirmed

TooWaBoo:
@Thanatomanic
Didn't find any issue. Works fine.

ps: Don't forget the color of the chroma curve, please

Beep6581:
I tested `thanatonic/histogram-scaling` 293d0d7e using `gtk+-3.22.29`, RT theme.
* The histogram scales well.
* It seems to have a horizontal scaling limit:
screenshot_20180630_192725
screenshot_20180630_192738
* The scaling mode tooltip mentions only two modes - full or scaled, while there are now 3 modes. Please update the tooltip only in `rtdata/languages/default`, I'll take care of the rest. We can work on a new icon later once my icon cleanup branch is ready.
* Agreed about making the chromaticity curve some other color. I'm also happy that the chromaticity curve is now better visibile using one of the new modes.
* I'm also happy to see the grid changes, can you explain what the intervals are? Is it stops?

TooWaBoo:
@Beep6581 
Confirmed
grafik

Changing "natural_width" from 400 to 2000 fixed it.

```
void HistogramArea::get_preferred_width_vfunc (int &minimum_width, int &natural_width) const
{
    minimum_width = 200;
--> natural_width = 2000;
}
```

Thanatomanic:
@Beep6581 The horizontal scaling is fixed now. It was a stupid typo. 

image
What do you think about yellow for chroma?

Thanatomanic:
> The scaling mode tooltip mentions only two modes - full or scaled, while there are now 3 modes. Please update the tooltip only in rtdata/languages/default, I'll take care of the rest. We can work on a new icon later once my icon cleanup branch is ready.

@Beep6581 Would it be okay to change the tooltip name too? `HISTOGRAM_TOOLTIP_FULL` is not a 100% correct descriptor anymore. Or would changing the name have annoying consequences?

As for the grid: in normal mode, they represent simple linear divisions (quarters, eighths, sixteenths). In log mode I simply take the log of these values. So you still can pinpoint halves, quarters, etc. I haven't really put much more thought in it. Suggestions are more than welcome :smiley:

TooWaBoo:
@Thanatomanic
Chroma color in RGB bar is still black, should be changed to yellow too

Thanatomanic:
Oh my sloppiness :roll_eyes:  Sorry for all the extra work. You are very gentle in my first FOSS contribution, which is much appreciated.

heckflosse:
@Thanatomanic 
>In log mode I simply take the log of these values.

In non float raw files the first stop is represented by 2 values (0 to 1), the second stop is represented by 2 values (2 to 3), the third stop is represented by 4 values (4 to 7), the fourth stop is represented by 8 values (8 to 15) and so on. imho the horizontal scaling in log mode should reflect this (at least in raw histogram mode it does not make sense to have a fine granularity for first stops and a rough granularity for larger stops)

heckflosse:
@Thanatomanic 
> Sorry for all the extra work. You are very gentle in my first FOSS contribution, which is much appreciated.

No need to apologize. We appreciate your work contributing to RT :+1:

Thanatomanic:
@heckflosse It's my understanding from `RawImageSource::getRAWHistogram()` that the RAW histogram shows values corrected for gamma, blackpoint and gain, collected over 256 bins. I have difficulty to imagine  how stops should look like in that case, and which gridlines you would then expect to see. Can you maybe draw up an sketch, or do you have an example of another piece of software that I can compare with?

TooWaBoo:
Right click on the image freezes/unfreezes the RGB bar. This feature is usless 'cause right click on the image is also used to disable the tools colorpicker, rotate, crop etc. I suggest to remove the RGB freeze feature.

Beep6581:
> Would it be okay to change the tooltip name too? HISTOGRAM_TOOLTIP_FULL is not a 100% correct descriptor anymore. 

Agreed. Make those changes only in `default` please.

> I suggest to remove the RGB freeze feature.

Agreed, I don't see how it could be useful.

Beep6581:
@Thanatomanic could you explain what scaling modes there are now? I'm busy designing icons and knowing how they work would help.

Thanatomanic:
@Beep6581 There is a linear/linear mode, which is simply the 'old' histogram, and then a log/linear and log/log mode where either the y-axis or both the x- and y-axes are transformed by a simple logarithmic function. The log-scaling offsets all values by 1 to prevent log(0) = -inf.
I am unsure how to represent those in a tiny icon...

heckflosse:
@Thanatomanic 
> which gridlines you would then expect to see?

In horizontal log mode I would expect this grid lines:

grafik

Thanatomanic:
Frozen mode is now gone. @Beep6581 The default language file is updated in two places. I replaced `HISTOGRAM_TOOLTIP_FULL` with `HISTOGRAM_TOOLTIP_MODE` and I updated the text for `HISTOGRAM_TOOLTIP_BAR`.

I'll have to think a bit more about the logarithm scaling.

Thanatomanic:
@heckflosse I've been thinking about your expected histogram and the definition of stops, but I doubt what you show is actually what you would want to see. When I transform the histogram data by taking the logarithm (log2) of the x-values, I should get a graph that consists of only a few data points at the left and more data points on the right. Correct? However, stops should then be shown at _regular_ x-intervals if I'm not mistaken (i.e. taking the log2 of 1,2,4,8,16,32,64,128).

Showing stops in the RAW histogram is a bit misleading in any case, because the histogram is limited to 8-bit values. The actual number of stops in the raw data is usually higher.

heckflosse:
@Thanatomanic 
> I should get a graph that consists of only a few data points at the left and more data points on the right. Correct? However, stops should then be shown at regular x-intervals if I'm not mistaken (i.e. taking the log2 of 1,2,4,8,16,32,64,128).

Yes, for a not limited to 8-bit values raw histogram that would be fine (at least for non float raw images)

Thanatomanic:
I think I tackled the issue, so please tell me what you think.
What I do now is to always limit the log-log histogram to 8 vertical gridlines (8 stops), and I place them at the values (1,2,4,8,16,32,64,128) within the histogram. This happens both in the RAW and normal histogram. It now looks like this:
image

The fact that you don't see the gridlines equally spaced in log-log mode, but a bit squashed towards the left is due to two things. First, I have to shift the log function by 1 to prevent problems with log(0). Additionally, I 'soften' the log scale by a factor, so that the middle stops get more space (I assume this is usually the most interesting part of the histogram).

image
**Figure 1.** Transfer function used in log mode, with the position of the gridlines indicated (dashed lines). Red = simple log (not used), blue = 'softened' log (used).

heckflosse:
@Thanatomanic 
> 8 vertical gridlines (8 stops)

Wouldn't 8 stops mean 7 vertical gridlines?

I think I found a bug:
If you look at the raw histogram in log mode you won't find an image which gives you y > 0 for the second zone
grafik

Thanatomanic:
7 gridlines, indeed.
And about the bug, unless something funky happens in my math, my first explanation would be that there are no values of 1 in the RAW data. But this would be odd... I'll investigate.

heckflosse:
@Thanatomanic 
> I'll investigate

I as well. It's somehwere in `RawImageSource::getRAWHistogram`

heckflosse:
@Thanatomanic 
I think I found the reason for zone 2 being empty.
It's because I tested with raw files < 16bit.
With 16-bit raw files from Phase One IQ3 it works fine.

Thanatomanic:
My conclusion here as well. It's simply a cutoff happening because of the combination of the gamma and binning to 256 items. So, not a bug :smile:
However, I do see that many (if not all) of my RAW histograms stop at a value of 252. After that they're empty. Do you see that too?

heckflosse:
@Thanatomanic Can you try whether this patch fixes the issue with raw histogram stopping at 252?

```diff
diff --git a/rtengine/rawimagesource.cc b/rtengine/rawimagesource.cc
index 77cb6a52..71f42dc3 100644
--- a/rtengine/rawimagesource.cc
+++ b/rtengine/rawimagesource.cc
@@ -4740,10 +4740,10 @@ void RawImageSource::getRAWHistogram (LUTu & histRedRaw, LUTu & histGreenRaw, LU
     histRedRaw.clear();
     histGreenRaw.clear();
     histBlueRaw.clear();
-    const float mult[4] = { 65535.0f / ri->get_white(0),
-                            65535.0f / ri->get_white(1),
-                            65535.0f / ri->get_white(2),
-                            65535.0f / ri->get_white(3)
+    const float mult[4] = { 65535.0f / (ri->get_white(0) - cblacksom[0]),
+                            65535.0f / (ri->get_white(1) - cblacksom[1]),
+                            65535.0f / (ri->get_white(2) - cblacksom[2]),
+                            65535.0f / (ri->get_white(3) - cblacksom[3])
                           };
 
     const bool fourColours = ri->getSensorType() == ST_BAYER && ((mult[1] != mult[3] || cblacksom[1] != cblacksom[3]) || FC(0, 0) == 3 || FC(0, 1) == 3 || FC(1, 0) == 3 || FC(1, 1) == 3);

```

Thanatomanic:
@heckflosse  That does help. But on close inspection I think the multiplier is being applied wrongly. It is included in the gamma-scaling, but should be separate (it's a bit hard to explain... but I tried to do the math).

Currently, the `idx` for binning is calculated as follows:

    CLIP((int)Color::gamma(std::min(mult[0] * (i - (cblacksom[0])), gammaLimit))) >> 8

By printing out the original values and comparing, I convinced myself that this is fully equivalent to the following:

    (int) std::min(255.0f, 256.0f / Color::gamma(ri->get_white(0)) * Color::gamma(std::max(0.0f, i - cblacksom[0])))

Similar to your suggestion, now the value for the raw black point should be subtracted from the white point. However, this is mathematically different from subtracting it in the definition of `mult`. If you do it in my alternative formula, you _will_ end up at a histogram index value of 256 when you have values that match the whitepoint. What do you think? Do I make sense?

Extra benefit: `gammaLimit` is no longer needed, because all values passed to `Color::gamma()` will be smaller than 65535.

Edit: the values for `256.0f / Color::gamma(ri->get_white(0) - cblacksom[0])` could be tabulated, similar to `mult` to save some computations.

heckflosse:
@Thanatomanic Can you post a patch of your fix for ramimagesource.cc::getRAWHistogram please?

I tried to guess what you mean, but ended with different histograms:

My version:

grafik

your version:

grafik

Thanatomanic:
If you're asking to post a nice diff like you did, I don't know how to do that - and I don't even know if that would even be useful. I can just push the changes and then maybe revert the commit if necessary.

heckflosse:
@Thanatomanic 

```
git diff rtengine/rawimagesource.cc > test.patch
```

Thanatomanic:
Thanks! Learned something today :smiley:
Don't mind the testing code for output checking...

````diff
index 77cb6a52..6362ff43 100644
@@ -18,6 +18,7 @@
  */
 #include <cmath>
 #include <iostream>
+#include <fstream> // DEBUG
 
 #include "rtengine.h"
 #include "rawimagesource.h"
@@ -4740,10 +4741,10 @@ void RawImageSource::getRAWHistogram (LUTu & histRedRaw, LUTu & histGreenRaw, LU
     histRedRaw.clear();
     histGreenRaw.clear();
     histBlueRaw.clear();
-    const float mult[4] = { 65535.0f / ri->get_white(0),
-                            65535.0f / ri->get_white(1),
-                            65535.0f / ri->get_white(2),
-                            65535.0f / ri->get_white(3)
+    const float mult[4] = { 65535.0f / (ri->get_white(0)/* - cblacksom[0]*/),
+                            65535.0f / (ri->get_white(1)/* - cblacksom[1]*/),
+                            65535.0f / (ri->get_white(2)/* - cblacksom[2]*/),
+                            65535.0f / (ri->get_white(3)/* - cblacksom[3]*/)
                           };
 
     const bool fourColours = ri->getSensorType() == ST_BAYER && ((mult[1] != mult[3] || cblacksom[1] != cblacksom[3]) || FC(0, 0) == 3 || FC(0, 1) == 3 || FC(1, 0) == 3 || FC(1, 1) == 3);
@@ -4850,22 +4851,57 @@ void RawImageSource::getRAWHistogram (LUTu & histRedRaw, LUTu & histGreenRaw, LU
     } // end of parallel region
 
     constexpr float gammaLimit = 32767.f * 65536.f; // Color::gamma overflows when the LUT is accessed with too large values
+
+    //Testcode - REMOVE BEFORE FINAL COMMIT
+    //ofstream tempfile;
+    //tempfile.open("D:/out.txt",ios::out);
+    //tempfile << "Mult: " << mult[0] << "," << mult[1] << "," << mult[2] << "," << mult[3] << "\n";
+    //tempfile << "Blacksom: " << cblacksom[0] << "," << cblacksom[1] << "," << cblacksom[2] << "," << cblacksom[3] << "\n";
+    //tempfile.precision(1);
+    //for (int i = 0; i < 65536; i++) {
+    //    //tempfile << i << " (" << histRedRaw[i] << "," << histGreenRaw[i] << "," << histBlueRaw[i] << "\n";
+    //   
+    //    tempfile << i << std::fixed << " " 
+    //    //<< (CLIP((int)Color::gamma(std::min(mult[0] * (i - (cblacksom[0]/*+black_lev[0]*/)), gammaLimit))) >> 8) << "," 
+    //    //<< (CLIP((int)Color::gamma(std::min(mult[1] * (i - (cblacksom[1]/*+black_lev[1]*/)), gammaLimit))) >> 8) << "," 
+    //    //<< (CLIP((int)Color::gamma(std::min(mult[3] * (i - (cblacksom[3]/*+black_lev[3]*/)), gammaLimit))) >> 8) << "," 
+    //    //<< (CLIP((int)Color::gamma(std::min(mult[2] * (i - (cblacksom[2]/*+black_lev[2]*/)), gammaLimit))) >> 8) << " : "
+    //    //<< (int)std::min(255.0f, 256.0f / Color::gamma(ri->get_white(0)) * Color::gamma(std::max(0.0f, i - cblacksom[0]))) << "," 
+    //    //<< (int)std::min(255.0f, 256.0f / Color::gamma(ri->get_white(1)) * Color::gamma(std::max(0.0f, i - cblacksom[1]))) << "," 
+    //    //<< (int)std::min(255.0f, 256.0f / Color::gamma(ri->get_white(2)) * Color::gamma(std::max(0.0f, i - cblacksom[2]))) << "," 
+    //    //<< (int)std::min(255.0f, 256.0f / Color::gamma(ri->get_white(3)) * Color::gamma(std::max(0.0f, i - cblacksom[3]))) << " : "
+    //    << (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(0) - cblacksom[0]) * Color::gamma(std::max(0.0f, i - cblacksom[0]))) << "," 
+    //    << (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(1) - cblacksom[1]) * Color::gamma(std::max(0.0f, i - cblacksom[1]))) << "," 
+    //    << (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(2) - cblacksom[2]) * Color::gamma(std::max(0.0f, i - cblacksom[2]))) << "," 
+    //    << (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(3) - cblacksom[3]) * Color::gamma(std::max(0.0f, i - cblacksom[3]))) << " ("
+    //    << hist[0][i] << "," << hist[1][i] << "," << hist[2][i] << ")\n";
+    //}
+    //tempfile.close();
+    
     for(int i = 0; i < 65536; i++) {
         int idx;
-        idx = CLIP((int)Color::gamma(std::min(mult[0] * (i - (cblacksom[0]/*+black_lev[0]*/)), gammaLimit)));
-        histRedRaw[idx >> 8] += hist[0][i];
+        //idx = CLIP((int)Color::gamma(std::min(mult[0] * (i - (cblacksom[0]/*+black_lev[0]*/)), gammaLimit)));
+        //histRedRaw[idx >> 8] += hist[0][i];
+        idx = (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(0) - cblacksom[0]) * Color::gamma(std::max(0.0f, i - cblacksom[0])));
+        histRedRaw[idx] += hist[0][i];
 
         if (ri->get_colors() > 1) {
-            idx = CLIP((int)Color::gamma(std::min(mult[1] * (i - (cblacksom[1]/*+black_lev[1]*/)), gammaLimit)));
-            histGreenRaw[idx >> 8] += hist[1][i];
+            //idx = CLIP((int)Color::gamma(std::min(mult[1] * (i - (cblacksom[1]/*+black_lev[1]*/)), gammaLimit)));
+            //histGreenRaw[idx >> 8] += hist[1][i];
+            idx = (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(1) - cblacksom[1]) * Color::gamma(std::max(0.0f, i - cblacksom[1])));
+            histGreenRaw[idx] += hist[1][i];
 
             if (fourColours) {
-                idx = CLIP((int)Color::gamma(std::min(mult[3] * (i - (cblacksom[3]/*+black_lev[3]*/)), gammaLimit)));
-                histGreenRaw[idx >> 8] += hist[3][i];
+                //idx = CLIP((int)Color::gamma(std::min(mult[3] * (i - (cblacksom[3]/*+black_lev[3]*/)), gammaLimit)));
+                //histGreenRaw[idx >> 8] += hist[3][i];
+                idx = (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(3) - cblacksom[3]) * Color::gamma(std::max(0.0f, i - cblacksom[3])));
+                histGreenRaw[idx] += hist[3][i];
             }
 
-            idx = CLIP((int)Color::gamma(std::min(mult[2] * (i - (cblacksom[2]/*+black_lev[2]*/)), gammaLimit)));
-            histBlueRaw[idx >> 8] += hist[2][i];
+            //idx = CLIP((int)Color::gamma(std::min(mult[2] * (i - (cblacksom[2]/*+black_lev[2]*/)), gammaLimit)));
+            //histBlueRaw[idx >> 8] += hist[2][i];
+            idx = (int)std::min(255.0f, 255.0f / Color::gamma(ri->get_white(2) - cblacksom[2]) * Color::gamma(std::max(0.0f, i - cblacksom[2])));
+            histBlueRaw[idx] += hist[2][i];
         }
     }
````

**Comment by heckflosse** [comment]
@Thanatomanic Thanks for the diff. That clearly showed that I wrongly guessed what you meant ;-)
Works fine (except the gap at zone 2, though that's an old bug)

**Comment by Beep6581** [comment]
> If you're asking to post a nice diff like you did, I don't know how to do that - and I don't even know if that would even be useful. I can just push the changes and then maybe revert the commit if necessary.

Or my even lazier way:
```bash
git diff | xclip -i
```

Then just middle-click to paste the diff anywhere.

Thanatomanic:
These two pushes could be the last two necessary for this PR to be merged and closed. 
Who do I need to call upon to do a final check?

Floessie:
@Beep6581 

> Or my even lazier way:
> `git diff | xclip -i`

Nice one. Didn't know it. Now I do. :+1: You can even leave out the '-i'.

heckflosse:
@Thanatomanic 
> 7 gridlines, indeed.

I miss the fix for this.

Thanatomanic:
Done!

heckflosse:
@Thanatomanic I don't see the visual impact of your latest commit. histogram before and after looks equal to me

Thanatomanic:
@heckflosse I thought too quickly, my last commit makes no sense. Things were actually already how they should be (afaik). This is how it was before, with gridlines at exactly the right places. So, revert this change?

image

heckflosse:
@Thanatomanic As it makes no sense, it's confusing when looking at the code. So, yes, please revert. Maybe one suggestion for the grid lines:

0(left border), 1, 3, 7, 15, 31, 63, 127, 255(right border

That would lead to 7 grid lines and 8 zones.

Thanatomanic:
> 0(left border), 1, 3, 7, 15, 31, 63, 127, 255(right border
That would lead to 7 grid lines and 8 zones.

So, combining 2 and 4 into 3, discarding one line ~~and getting rid of the 2=empty problem? That last part sounds good, but~~ it does go against the idea of showing all stops, wouldn't it?

TooWaBoo:
Is this right? Shouldn't be the right side streched?

Thanatomanic:
@TooWaBoo This is correct behavior as far as logarithms work. Low values get stretched, high values compressed.

TooWaBoo:
But then the vertical lines are wrong.

Thanatomanic:
Some posts earlier @heckflosse suggested to show the stops instead of converting the linear divisions to their logarithmic counterparts. That's what's implemented here, so the lines correspond to (binned) raw values of 1, 2, 4, etc.

TooWaBoo:
This is confusing me. The horzontal lines represents the log and the vertical lines the stops. This has to be explained in a tooltip.

heckflosse:
@TooWaBoo Concerning the vertical grid lines I expected for the raw histogram:
https://github.com/Beep6581/RawTherapee/pull/4618#issuecomment-401559062
https://github.com/Beep6581/RawTherapee/pull/4618#issuecomment-401753752

Background for my expectation. Most raw files are integer.
The first (darkest) stop can have values 0 and 1, the second darkest stop can have values 2 and 3, the third darkest stop can have values 4 to 7, then 8 to 15, then 16  to 31 and so on until 128 to 255.

For that reason it does make sense to increase the horizontal space for the 128 to 255 range at the right (as it can hold 128 values) compared to the 4 to 7 range which can hold only 4 values.

Thanatomanic:
I agree that stuff is a little confusing now and should be explained in RawPedia and/or a tooltip. Some preliminary write-up:

> The histogram is 8-bit, so it always shows you a maximum of 256 values.
> 1. Linear-linear mode. Everything is simple. You find gridlines at halves, quarters, eighths and sixteenths, depending on the size of the histogram.
> 2. Linear-log mode. The x-axis is the same as in (1). The y-axis and the horizontal gridlines are now scaled logarithmically. The position of the gridlines still corresponds to the halves, quarters, etc. 
> 3. Log-log mode. The y-axis is the same as in (2). The x-axis is now scaled logarithmically. Importantly however, the gridlines are not scaled logarithmically, but correspond to 'stops': with every gridline the value doubles, so there are lines for the values 1, 3, 7, 15, 31, 63, and 127 (pow(2.0,i) - 1))

heckflosse:
@Thanatomanic What's your opinion about my suggestion for the horizontal log vertical grid line positions from left to right?

0(left border), 1, 3, 7, 15, 31, 63, 127, 255(right border)

For me it looks more logical to use this

Thanatomanic:
@heckflosse it took me a good night's sleep to realize yours might actually be a little more correct because it ends in 255 (and not 256 as does mine). It now looks like this 

image

Beep6581:
Documentation updated:
http://rawpedia.rawtherapee.com/The_Image_Editor_Tab#Main_Histogram

Screenshots and icons to be updated once this is merged.

TooWaBoo:
To make a visual difference between the log lines and stops, I suggest to change the stop lines from dash to dots. What do you think?

Thanatomanic:
@TooWaBoo It would look like this
image

My personal opinion is that it's not really necessary, and even might add to the confusion of different modes. What do the others think?

TooWaBoo:
@Thanatomanic 
Yes, you're right.

heckflosse:
@Thanatomanic For the raw histogram (and only for the raw histogram) I would prefer if the brightes stop at right would have a width of 128 pixels, the one left to the brightes 64 pixels and so on. That should give a more detailed curve in the bright regions with same granularity as before in dark regions. What do you think? That's for a curve (histogram) width of 256 of course...

Ingo

heckflosse:
@Thanatomanic 
> Who do I need to call upon to do a final check?

For code review I suggest @Floessie ;-)

Thanatomanic:
@heckflosse If I understand what you're asking, this would mean that the histogram curves return to their un-logarithmic form. That kind of defies the purpose of the whole logarithmic scaling?

heckflosse:
@Thanatomanic No, I mean a real logarithmic scaling for the log-log-mode of raw histogram, not one which is widened in the middle as it is now

Beep6581:
See these raw histograms for reference:
https://github.com/Beep6581/RawTherapee/issues/4667#issuecomment-403274554

Thanatomanic:
@heckflosse Also with the reference histogram I am confused about your request. I am doing an honest logarithmic scaling of the RAW RGB-curves. That will - by definition - compact the curves towards the right.
There are 128 binned RAW values in the highest stop. If you want to make them correspond to 128 pixels again, you arrive back at the linear scale. In the linear-linear mode the first pixel corresponds to the first stop, the next 2 pixels correspond to the second stop, the next 4 pixels ... etc.

heckflosse:
@Thanatomanic 
> If you want to make them correspond to 128 pixels again, you arrive back at the linear scale.

True, but the user has to know that and count vertical grid lines.
Wouldn't it be better to offer a (maybe additional) mode where the curve is in linear mode but the vertical grid lines show the stops?

Currently it's like this in linear mode
grafik

Simply not drawing the lines in between the ones where the arrows point to, would do that (or drawing a zero, a -1, a -2 and so on if we want to keep all the vertical lines)

Thanatomanic:
I would be totally fine if the vertical gridlines would _always_ correspond to stops. My impression always was that the linear-linear mode just has divisions to "guide the eye" as they say. To make the divisions meaningful (i.e. stops) seems like a good plan.

Thanatomanic:
I could even add proper axis labels - but that would need to wait for a next time I think. FYI: on Saturday I'm heading off for a month's vacation and won't be able to comment or work on this PR. So I would be glad to see this finished to a state that I can be properly merged into dev.

heckflosse:
@Thanatomanic 
> FYI: on Saturday I'm heading off for a month's vacation and won't be able to comment or work on this PR.

I'm also (though with internet and laptop) on vacations from Friday on for three weeks :)

I have no objections to merge the current state into dev and let the pr and the branch open for further improvements :+1:

Thanatomanic:
So, if the gridlines always represent stops, it would look like this:

Linear-linear mode
image

Linear-log mode
image

Log-log mode
image

TooWaBoo:
Why not merge it now to get feedback from users?

Thanatomanic:
Since I got two thumbs up from @Beep6581 and @heckflosse, I pushed the changed. 
Then, as @TooWaBoo suggests, I would vote for a merge now, and ask questions later :wink:

Thanatomanic:
@heckflosse I did something interesting: I made the log-mode factor modifiable by click-and-drag. That way the user can choose whichever region of the histogram is worthwhile to magnify.
This works only by moving your mouse horizontally, even though it also modifies the y-axis. The mechanism is also a bit fuzzy, so I appreciate both general and specific feedback on this.

heckflosse:
@Thanatomanic I've no time atm to test because of vacations. I will test probably on sunday or monday (when I have internet access again)

TooWaBoo:
Found a bug.
The left Navigator pane can't be opened when it's hidden at launch time of RT.
Latest dev works fine.

How to reproduce.
    Run RT and open an image
    Hide left navigator pane
    Quit RT
    Run RT and open an image
    Show left Navigator pane.
    --> Navigator pane is still hidden.

Thanatomanic:
@TooWaBoo I see that `ShowHistory=true` is getting set in the options if you click the button, and I see that the panel actually doesn't open even though `HistoryPanelWidth` has some width. It does get opened when you restart RT again.
What I don't see is how I caused this. Don't get me wrong: I might have, but I just don't know how...

~~Edit: it actually matters where you place the histogram. If it's on the right, there is no bug.~~
It has nothing to do with the position of the histogram. The fact remains that the left pane stays hidden. The right pane does not have this bug.

TooWaBoo:
Mine is on the right. :-)

Thanatomanic:
@TooWaBoo My bad, I've edited my post. I still don't know what is going on - and I'm a little low on time to check for a fix.

RawConvert:
I just added to #1598  (maybe I should have put it here.)

Beep6581:
@Thanatomanic now that `dev` has had its icon cleanup, don't hesitate to let me know if you need help with resolving conflicts between `dev` and your branch!

Thanatomanic:
@Beep6581 I tried merging, but my modifications require three icons for the three different modes for the scaling. For this I use the `modeImage`, `modeImage_g` and `modeImage_g2` currently. While these names might be changed easily, I would kindly request icons to accompany the modes. Or am I overlooking something you already added?

Thanatomanic:
@Beep6581 Disregard my last message. I found the three icons. Attempting the merger now!

Beep6581:
Yeah I already included 3 histogram-mode icons for you :)

TooWaBoo:
When horizontal log is active the RGB bar still shows linear scaling. Should be changed to log too.

btw. Question: How does the stops in the histogram helps me by developing an image?

Thanatomanic:
I would like some programming advice on how to properly make the `HistogramArea::scalingFunctionLog` also available to `HistogramRGBArea`. How can give me a push in the right direction?

Hombre57:
@Thanatomanic We used to create call back class, however one dev (Adam) told us that it would be better to use signals. I'm preparing a patch using them.

Hombre57:
@Thanatomanic Here is a patch that create a parent class for the `factor` parameters that is used in `scalingFunctionLog`. It also create a signal in `HistogramArea` that will be connected to `HistogramRGBArea` in the `HistogramPanel`'s ctor.

Additionally, it adds `movingPosition` in the `HistogramArea` ctor, it was missing and gcc complained about it.

You can download it here [scalingFunctionLog.patch.txt](https://github.com/Beep6581/RawTherapee/files/2216159/scalingFunctionLog.patch.txt)

I'll let you complete it (and remove the debug output), since it's not sufficient to have the indicators at the right position depending on the selected mode.

TooWaBoo:
@Thanatomanic 
Anything new here?

Thanatomanic:
Hi all, @TooWaBoo in particular, I've been away for a month on a good long holiday with few opportunities to read e-mail and so on. Coming Sunday I'm back home and will pick up where I left off and work with @Hombre57 's patch! Stay tuned :-)

Thanatomanic:
@TooWaBoo The `HistogramRGBArea` now correctly shows the (scaled) RGB value of the cursor. Thanks to the patch by @Hombre57.

And with that the only thing left is the bug where the left side-panel does not open anymore after hiding. But I really have no clue why this happens, or how to figure out why it happens. It must be a consequence of my changes to `editorpanel.cc` to make the histogram resizable. Who can help with this?
Otherwise, maybe one final check, and then merge?

Hombre57:
@Thanatomanic I have some free time left today, I'll look at the left lane bug.

TooWaBoo:
@Thanatomanic 
Everything looks fine (except the left lane bug). 👍

Thanatomanic:
Just for reference: if this PR is merged, #1596 can be closed

Hombre57:
@Thanatomanic Test this patch, it should solve the left pane bug :

```patch
diff --git a/rtgui/editorpanel.cc b/rtgui/editorpanel.cc
index dbcc6fd..2db1925 100644
--- a/rtgui/editorpanel.cc
+++ b/rtgui/editorpanel.cc
@@ -522,6 +522,7 @@
     leftsubbox->show_all ();
     
     leftbox->pack2 (*leftsubbox, true, true);
+    leftbox->show_all();
 
     // build the middle of the screen
     Gtk::Box* editbox = Gtk::manage (new Gtk::Box (Gtk::ORIENTATION_VERTICAL));
```

TooWaBoo:
@Hombre57 
Tested, works fine.

Thanatomanic:
Thanks @Hombre57, works like a charm!

ff2000:
ninja install fails:

```
CMake Error at rtdata/cmake_install.cmake:295 (file):                                                                  
  file INSTALL cannot find
  "/home/franz/src/RawTherapee/rtdata/images/non-themed/png/empty.png".
Call Stack (most recent call first):
  cmake_install.cmake:69 (include)


FAILED: CMakeFiles/install.util 
cd /home/franz/src/RawTherapee/build_gcc && /usr/bin/cmake -P cmake_install.cmake
ninja: build stopped: subcommand failed.
```

The scaling seems to work as advertised. I have issues with changing the log scaling factor by dragging. It didn't do anything at my first try, Then worked only when dragging vertically, then started working horizontally, too. But unfortunately didn't pick up changing the drag direction correctly.

In general I think this should use a modifier (SHIFT or STRG). Lightroom and Darktable (and likely others, too) change brightness when dragging the histogram. You can control highlights and shadows and midtones separately. I think it would be great to have this feature also in RawTherapee!

Beep6581:
I tested `5.4-701-g`af4217d2 and enjoy the new histogram very much, good work!
Looking forward to the merge.

Some ideas:
1. Currently the dashed grid is drawn over the histograms. Maybe it would be better if it was drawn under all of the histograms except for the luminance histogram?
1. Shift+clicking a histogram button, or right-clicking it to be consistent with the toolbox, could show that histogram while hiding all other histograms.

Thanatomanic:
@ff2000 I don't understand why your ninja install is failing, but it does not seem to have to do with this PR.

>The scaling seems to work as advertised. I have issues with changing the log scaling factor by dragging. It didn't do anything at my first try, Then worked only when dragging vertically, then started working horizontally, too. But unfortunately didn't pick up changing the drag direction correctly.

Currently only horizontal dragging should change the factor. When you don't see a change, you might be moving the wrong way (i.e. the factor is maxed out).

> In general I think this should use a modifier (SHIFT or STRG). Lightroom and Darktable (and likely others, too) change brightness when dragging the histogram. You can control highlights and shadows and midtones separately. I think it would be great to have this feature also in RawTherapee!

I agree that having this feature would be great, but currently it is not feasible to make that work. In the meanwhile, I don't think there is any harm done to leave the behaviour without a modifier key.

> Currently the dashed grid is drawn over the histograms. Maybe it would be better if it was drawn under all of the histograms except for the luminance histogram?

That would be a relatively easy change for a nice polishing up.

> Shift+clicking a histogram button, or right-clicking it to be consistent with the toolbox, could show that histogram while hiding all other histograms.

@Beep6581 I had to read that three times in order to understand it, but now I do :smile: Do you think this is essential to do before a merger?

Beep6581:
@Thanatomanic neither of my suggestions are necessary before a merge, they are just "would be nice if".

Thanatomanic:
@Beep6581 I pushed the graphic change you suggested. As far as I'm concerned this can actually be merged with dev. But please review the code is necessary (it is my first pr after all...) :smile:

TooWaBoo:
@Thanatomanic 
Tested and everything works fine. I think it's ready to merge. 👍

Beep6581:
I tested `5.4-745-g`f919b0b8

* The histogram lines in reference to the grid lines look great.
* Log-log mode seems very useful for cases where the area of interest is squashed up against one side, such as often is the case when dealing with the chromaticity histogram.
* The RGB indicator bar does not work at all. I tried deleting the `options` file, no difference.

heckflosse:
@Beep6581 
> The RGB indicator bar does not work at all. I tried deleting the options file, no difference.

That may be my fault. I will check that now.

Thanatomanic:
@heckflosse Must be to do with `pointermotionlistener.h` see [here](https://github.com/Beep6581/RawTherapee/pull/4618/commits/f919b0b873fd58e623fd5da834f09f7cc2048e67). If you can come up with a patch, that would be excellent.

heckflosse:
@Thanatomanic I'm already compiling

heckflosse:
I pushed a fix to dev.

Thanatomanic:
@heckflosse Thanks, is fixed!

heckflosse:
I let @Beep6581 press the merge button :)

Beep6581:
@Thanatomanic while I'm compiling the fix, please don't forget to add yourself to `AUTHORS.txt`.

Thanatomanic:
:heart:

Beep6581:
Thank you very much :)

casta:
Hi,

Tested with master commit 5471b34a312315d45119eff2c23c87981a490bda
There is an issue with the histogram state memorization: enable back luminance histogram (I prefer this way :p), restart RT, luminance is again disabled.
Other params seems correctly saved.

Thank’s !

Thanatomanic:
Could you please open a new issue for that?

Op za 25 aug. 2018 15:30 schreef Guillaume Castagnino <
notifications@github.com>:

casta:
Yes I can. I will. But it seemed to me more logic to reopen the pull request that causes the issue ;)

Here you are: #4749

PR #4623, opened 2018-06-17
Initial creation of Control Spot panel

Pandagrapher:
Creation of a Control Spot panel that allows to manage several spots at same time

Remember: For the moment, this new panel is independent from the rest of the locallab GUI (original one still present) or locallab tools which means that:
- Control spots created with this panel won't have any effect of picture
- Control spots created with this panel won't be saved in .mip file (just stored in memory)

### DIFF
--- a/rtgui/CMakeLists.txt
+++ b/rtgui/CMakeLists.txt
@@ -34,7 +34,7 @@ set(NONCLISOURCEFILES
     colorappearance.cc
     coloredbar.cc
     colortoning.cc
-	controlspotpanel.cc
+    controlspotpanel.cc
     coordinateadjuster.cc
     crop.cc
     crophandler.cc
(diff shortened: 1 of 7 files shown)


rtgui/CMakeLists.txt
```diff
@@ -34,6 +34,7 @@ set(NONCLISOURCEFILES
     colorappearance.cc
     coloredbar.cc
     colortoning.cc
+	controlspotpanel.cc
```
Beep6581: @Pandagrapher this line and other lines are incorrectly indented - is your editor configured to follow RT's formatting style? i.e. no tabs, 4 spaces, etc.

Pandagrapher: @Beep6581 I saw it after the commit: I have now configured my IDE to run AStyle to avoid that
I have also pushed a correction for the pull request

PR #4638, opened 2018-06-25
Update Deutsch locale

TooWaBoo:
ping @heckflosse :-)

### DIFF
--- a/rtdata/languages/Deutsch
+++ b/rtdata/languages/Deutsch
@@ -63,6 +63,7 @@
 #62 22.06.2018 Korrektur (TooWaBoo) RT 5.4
 #63 24.06.2018 DCB/RCD+VNG4 (TooWaBoo) RT 5.4
 #64 24.06.2018 Erweiterung/Korrektur (TooWaBoo) RT 5.4
+#65 25.06.2018 Korrekturen (TooWaBoo) RT 5.4
 
 ABOUT_TAB_BUILD;Version
 ABOUT_TAB_CREDITS;Danksagungen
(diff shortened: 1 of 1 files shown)


rtdata/languages/Deutsch
```diff
@@ -1832,9 +1833,9 @@ TP_RAWEXPOS_LINEAR;Korrekturfaktor
 TP_RAWEXPOS_PRESER;Weißpunkt: Lichter schützen
 TP_RAWEXPOS_RGB;Rot, Grün, Blau
 TP_RAWEXPOS_TWOGREEN;Grün-Werte automatisch angleichen
-TP_RAW_1PASSMEDIUM;1-Pass (Medium)
-TP_RAW_3PASSBEST;3-Pass (Beste)
-TP_RAW_4PASS;4-Pass
+TP_RAW_1PASSMEDIUM;1-Pass (Markesteijn)
+TP_RAW_3PASSBEST;3-Pass (Markesteijn)
```
ff2000, line 1837: Seems the above were reverted to strings from default language file. Is that intentional?
[And I had to google Markesteijn - seems to be an algorithm, is that correct?]

TooWaBoo: Yes, it's all correct and tested with latest "dev". @heckflosse changed it yesterday. There is also a discussion at pixel.us about that.
https://discuss.pixls.us/t/my-sharpening-workflow-for-base-iso-mages-with-the-new-features-in-rt-aka-extreme-pixel-peeping/8034/68

TooWaBoo:
ping @heckflosse :-)

heckflosse:
@TooWaBoo Thanks for the ping :+1:

Issue #4576, opened 2018-05-28
Hang when switching directories while thumbnails are loading


samtygier:
When switching between folders in the file browser view I sometimes get freezes. The OS eventually realizes and asks if want to force quit or wait. I can reproduce by switching rapidly back and forth between to directories.

This is on Ubuntu 18.04 with rawtherapee 5.3, and also if I update to the 5.4 version by installing the packages from https://launchpad.net/ubuntu/+source/rawtherapee/5.4-1 .

```
Version: 5.4
Branch: releases
Commit: a5e8eb9c
Commit date: 2018-03-20
Compiler: cc 7.3.0
Processor: generic
System: Linux
Bit depth: 64 bits
Gtkmm: V3.22.2
Lensfun: V0.3.2.0
Build type: Release
Build flags: -g -O2 -fdebug-prefix-map=/build/rawtherapee-x3NK8B/rawtherapee-5.4=. -fstack-protector-strong -Wformat -Werror=format-security -fPIC -Wdate-time -D_FORTIFY_SOURCE=2 -std=c++11  -Werror=unused-label -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -O3 -DNDEBUG -ftree-vectorize
Link flags: -Wl,-Bsymbolic-functions -Wl,-z,relro -Wl,-z,now -Wl,--as-needed
OpenMP support: ON
MMAP support: ON
```

here is a gdb log, while it is hung
[log.txt](https://github.com/Beep6581/RawTherapee/files/2045869/log.txt)

Floessie:
I would say that [`ThumbImageUpdater::removeJobs()`](https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/thumbimageupdater.cc#L242) is one end of this Gordian knot... And I guess the other one is the global GUI mutex held here:

```
Thread 1 (Thread 0x7ffff7fa2b00 (LWP 21935)):
#0  __lll_lock_wait () at ../sysdeps/unix/sysv/linux/x86_64/lowlevellock.S:135
No locals.
#1  0x00007ffff2a6f098 in __GI___pthread_mutex_lock (mutex=0x5555562e7390) at ../nptl/pthread_mutex_lock.c:113
        ignore1 = <optimised out>
        ignore2 = <optimised out>
        ignore3 = <optimised out>
        id = 21935
        type = 1
        __PRETTY_FUNCTION__ = "__pthread_mutex_lock"
[...]

Issue #4416, opened 2018-02-28
Multithreaded noise reduction leaks memory


PkmX:
To reproduce:

* In `Perferences`, `Performance & Quality`, set `Max number of threads for Noise Reduction and Wavelet Levels` to `0` or `8`.
* Load a image (JPG is fine, and faster to process), use neutral profile and turn on noise reduction and leave all sliders at default.
* Repeatedly export the image.

On my system (Arch Linux, i7-8550U, 4 cores, 8 threads), each save makes Rawtherapee use 200~400 MB more memory until it gets killed by OOM. Setting number of threads to 1 produces no (or very little of) leaks.

```
Version: 5.3-656-gb8440087
Branch: makepkg
Commit: b8440087
Commit date: 2018-02-28
Compiler: cc 7.3.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.22.2
Lensfun: V0.3.2.0
Build type: Release
Build flags: -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -std=c++11  -Werror=unused-label -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -O3 -DNDEBUG -ftree-vectorize
Link flags: -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now
OpenMP support: ON
MMAP support: ON
```

PkmX:
Note: even with the number of threads set to 1, enabling the wavelet module will also leak about 200 MB of memory per save as well.

heckflosse:
@PkmX Can not reproduce

```
Version: 5.3-657-gbad28bb0
Branch: rcd-speedup
Commit: bad28bb0
Commit date: 2018-02-27
Compiler: gcc 7.3.0
Processor: undefined
System: Windows
Bit depth: 64 bits
Gtkmm: V3.22.0
Lensfun: V0.3.2.0
Build type: Release
Build flags:  -std=c++11 -march=native -Werror=unused-label -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -O3 -DNDEBUG -ftree-vectorize
Link flags:  -march=native
OpenMP support: ON
MMAP support: ON
```

Floessie:
Confirmed, though the leak varies. I'll try to find something with the sanitizer...

gaaned92:
[...]

PR #4645, opened 2018-06-27
Try to fix thumbbrowser deadlock (#4576)

Floessie:
Hi,

The title says it all. See #4576 for the details and reproduction.

Best,
Flössie

### DIFF
--- a/rtgui/filebrowserentry.cc
+++ b/rtgui/filebrowserentry.cc
@@ -211,19 +211,8 @@ void FileBrowserEntry::procParamsChanged (Thumbnail* thm, int whoChangedIt)
 
 void FileBrowserEntry::updateImage (rtengine::IImage8* img, double scale, rtengine::procparams::CropParams cropParams)
 {
-
-    {
-        GThreadLock lock;
-
-        if ( feih == nullptr ||
-                feih->destroyed ) {
-            img->free();
-            return;
-        }
-
-        redrawRequests++;
-        feih->pending++;
-    }
+    redrawRequests++;
+    feih->pending++;
 
     struct tiupdate {
         FileBrowserEntryIdleHelper* feih;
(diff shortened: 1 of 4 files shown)


heckflosse:
@Floessie As I never managed to reproduce this deadlock I can only try whether it works as before. Shall I?

Floessie:
@heckflosse 
> As I never managed to reproduce this deadlock

Really? It's rather simple: Have let's say five (new, uncached) folders with 50 images each and switch between them while the thumbnails are rendering. It takes only a few seconds (<20) of trying until it deadlocks (on Linux). The original report was for Windows.

> I can only try whether it works as before. Shall I?

Yes, please. That at least would be encouraging.

Thanks,
Flössie

ff2000:
I could easily reproduce it usually at the first time. Now I tried really hard and jumped through probably 30-50 folders. No deadlock. YEAH! The only suspicious thing was that once the label still said "Loading thumbnails" although it clearly was "Ready". Changing once again and everything was "Ready" ;) Probably that is just a small issue with a signal connection between the thumbnailer and the label.

(OT^2: But this test showed me that at least I have issues with memory consumption, usage grew over time from 4% to 25% (of 4GB), loading and closing a raw file will add the size of the raw file to the overall consumed memory of RT. It's not just cache as it will start to swap when I launch gimp from within RT and start creating layers and masks... Is it just me or is there a real issue I should report? Running RT under valgrind with this "slightly" underpowered machine is  close to impossible... This is not an issue this PR created, I struggle with it since quite some time already.)

Floessie:
@ff2000 Thanks for testing!

Your memory problem reminds me of #4416. Please test @PkmX's workaround and see, if you experience the same problem. If so, report it there, otherwise create a new issue. Thanks!

heckflosse:
@Floessie Could reproduce with dev after clearing thumbnail cache. Now building `fix-thumbbrowser-deadlock`...

heckflosse:
@Floessie Can not reproduce anymore with your fix :+1:

heckflosse:
@Floessie Please push

PR #4646, opened 2018-06-28
Allow 0 to 16 pix border for bayer files instead of fixed 4 pix

heckflosse:


Floessie (approved):
No obvious problem found. :+1:

PR #4657, opened 2018-07-02
If available, use intrinsics to convert half precision float values t…

heckflosse:
…o single precision float values when loading float dng files

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -9937,6 +9937,7 @@ static void decodeFPDeltaRow(Bytef * src, Bytef * dst, size_t tileWidth, size_t 
 
 }
 
+#ifndef __F16C__
 // From DNG SDK dng_utils.h
 static inline uint32_t DNG_HalfToFloat(uint16_t halfValue) {
   int32_t sign     = (halfValue >> 15) & 0x00000001;
@@ -9970,6 +9971,7 @@ static inline uint32_t DNG_HalfToFloat(uint16_t halfValue) {
   // Assemble sign, exponent and mantissa.
   return (uint32_t) ((sign << 31) | (exponent << 23) | mantissa);
 }
+#endif
 
 static inline uint32_t DNG_FP24ToFloat(const uint8_t * input) {
   int32_t sign     = (input [0] >> 7) & 0x01;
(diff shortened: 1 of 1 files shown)


rtengine/dcraw.cc
```diff
@@ -10007,10 +10009,23 @@ static inline uint32_t DNG_FP24ToFloat(const uint8_t * input) {
...
   if (bytesps == 2) {
     uint16_t * dst16 = (uint16_t *) dst;
+#ifndef __F16C__
     uint32_t * dst32 = (uint32_t *) dst;
     for (int index = tileWidth - 1; index >= 0; --index) {
-      dst32[index] = DNG_HalfToFloat(dst16[index]);
+        dst32[index] = DNG_HalfToFloat(dst16[index]);
+    }
+#else
+    float * dst32 = (float *) dst;
```
Floessie: I'd like the following better:

```c++
float* const dst32 = static_cast<float*>(dst);
```

rtengine/dcraw.cc
```diff
@@ -10007,10 +10009,23 @@ static inline uint32_t DNG_FP24ToFloat(const uint8_t * input) {
...
+    int index = tileWidth - 4;
+    for (; index >= 0; index -= 4) {
```
Floessie: If I remember correctly, you once told me, that pre-caching works better, if you walk the cache line forward instead of backward. Shouldn't that be applied here, too?

rtengine/dcraw.cc
```diff
@@ -10007,10 +10009,23 @@ static inline uint32_t DNG_FP24ToFloat(const uint8_t * input) {
...
+        __m128i halfFloatv = _mm_loadu_si128((__m128i_u*)&dst16[index]);
+        STVFU(dst32[index], _mm_cvtph_ps(halfFloatv));
```
Floessie: Let's assume the load is expensive and the conversion is cheap. Then there's room for improvement, as you load 8 half-floats, but you convert and store only 4, always reloading 4 unused ones. Maybe walk in 8 float steps and use something like `_mm_srli_si128` on `halfFloatv`.

Floessie (changes requested):
Still compiles fine w/o `__F16C__` on Debian 9 with GCC 6.3 and Clang 3.9.1. Will test with `__F16C__` hardware later.

rtengine/dcraw.cc
heckflosse: I thought that too, but it won't work here because src and dst are the same buffer.

rtengine/dcraw.cc
heckflosse: good idea :+1:

rtengine/dcraw.cc
heckflosse: I will change that

heckflosse: @Floessie 
```
 error: invalid static_cast from type 'Bytef* {aka unsigned char*}' to type 'float*'
     float* const dst32 = static_cast<float*>(dst);
```

Floessie:
```c++
float* const dst32 = reinterpret_cast<float*>(dst);
```

Better?

heckflosse: Better. Thank you :+1:

Floessie:
@heckflosse 

```
/home/user/src/rawtherapee/rtengine/dcraw.cc: In function ‘void expandFloats(Bytef*, int, int)’:
/home/user/src/rawtherapee/rtengine/dcraw.cc:10021:47: error: ‘__m128i_u’ was not declared in this scope
         __m128i halfFloatv = _mm_loadu_si128((__m128i_u*)&dst16[index]);
                                               ^~~~~~~~~
/home/user/src/rawtherapee/rtengine/dcraw.cc:10021:57: error: expected primary-expression before ‘)’ token
         __m128i halfFloatv = _mm_loadu_si128((__m128i_u*)&dst16[index]);
                                                         ^
/home/user/src/rawtherapee/rtengine/dcraw.cc:10027:47: error: ‘__m128i_u’ was not declared in this scope
         __m128i halfFloatv = _mm_loadu_si128((__m128i_u*)&dst16[index]);
                                               ^~~~~~~~~
/home/user/src/rawtherapee/rtengine/dcraw.cc:10027:57: error: expected primary-expression before ‘)’ token
         __m128i halfFloatv = _mm_loadu_si128((__m128i_u*)&dst16[index]);
```

On Debian 9.4, GCC 6.3.

PR #4659, opened 2018-07-03
histmatching: use fewer points in the auto-matched curve

agriggio:
This makes editing the curve by hand much easier.

In my tests I see no degradation in the accuracy of the match, but I'd like to get more feedback before merging.

### DIFF
--- a/rtengine/histmatching.cc
+++ b/rtengine/histmatching.cc
@@ -51,7 +51,7 @@ CdfInfo getCdf(const IImage8 &img)
 
     for (int y = 0; y < img.getHeight(); ++y) {
         for (int x = 0; x < img.getWidth(); ++x) {
-            int lum = LIM(0, int(Color::rgbLuminance(float(img.r(y, x)), float(img.g(y, x)), float(img.b(y, x)))), 255);
+            int lum = LIM(int(Color::rgbLuminance(float(img.r(y, x)), float(img.g(y, x)), float(img.b(y, x)))), 0, 255);
             ++ret.cdf[lum];
         }
     }
@@ -206,6 +206,9 @@ void RawImageSource::getAutoMatchedToneCurve(const ColorManagementParams &cp, st
 
     int fw, fh;
     getFullSize(fw, fh, TR_NONE);
+    if (getRotateDegree() == 90 || getRotateDegree() == 270) {
+        std::swap(fw, fh);
+    }
     int skip = 3;
 
     if (settings->verbose) {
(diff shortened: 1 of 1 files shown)


sguyader:
@agriggio I think it can be a thing to make fewer points in those curves, particularly in the midtones. More points are often useful in the shadows, sometimes in the highlight.

agriggio:
@sguyader the current logic is to be denser on shadows and sparser on highlights. it seems to work fine for the curves produced by my camera (even with DRO turned on). but of course I need more feedback from users of other brands (mine is a Sony)...

heckflosse:
@agriggio :+1: for fewer points. With images from my D700 the old behaviour sometimes resulted in curves like this one

grafik

With the new code it's better

grafik

heckflosse:
Another example:

old code
grafik

new code
grafik
 :+1:

TooWaBoo:
Like it. 👍 Ready to merge?

Beep6581:
I compared the results of this branch to dev, and like it.

In a few cases the curves were identical between this branch vs dev and both were not-simple, e.g.:
* http://rawtherapee.com/shared/test_images/karin_library_2_4.pef
screenshot_20180708_130122

In a few cases the simple curve was further away from the embedded JPEG:
* http://rawtherapee.com/shared/test_images/karin_library_2.hdr.dng
screenshot_20180708_130347
screenshot_20180708_130352
screenshot_20180708_130357
* https://filebin.net/5vezeohornnn845g/2017-10_provence_160.arw?t=e7nti2wk

I'm showing these outliers only in case you want to tweak the algo, they are not an objection to merge.

heckflosse:
@agriggio Alberto, as already mentioned in another issue, histogram matching of images from same camera but with different orientation uses different amount of pixels for the matching. I think this is a bug.
To reproduce, use this patch and apply histogram matching on [these two files](https://filebin.net/toz2nxiboclbcptg) from my D700

```diff
diff --git a/rtengine/histmatching.cc b/rtengine/histmatching.cc
index 5db504816..1f2b4b118 100644
--- a/rtengine/histmatching.cc
+++ b/rtengine/histmatching.cc
@@ -49,6 +49,7 @@ CdfInfo getCdf(const IImage8 &img)
 {
     CdfInfo ret;
 
+    std::cout << img.getHeight() << " : " << img.getWidth() << std::endl;
     for (int y = 0; y < img.getHeight(); ++y) {
         for (int x = 0; x < img.getWidth(); ++x) {
             int lum = LIM(0, int(Color::rgbLuminance(float(img.r(y, x)), float(img.g(y, x)), float(img.b(y, x)))), 255);

```

agriggio:
yes, that's a bug for sure. I'll fix it and then merge, thanks!

agriggio:
btw now I see another bug there with `LIM`...

PR #4660, opened 2018-07-04
Update Deutsch locale

TooWaBoo:
@heckflosse 
There is no entry for key "PARTIALPASTE_RAW_BORDER" in "default".


TooWaBoo:
ping @heckflosse

PR #4670, opened 2018-07-07
Floating point dng files in [0;1] range have black thumbnail, fixes #…

heckflosse:
…4669


heckflosse:
Oups, forgot the xtrans case. Will follow soon...

PR #4694, opened 2018-07-22
Move r pix coords to arguments of amaze

CarVac:
This is an effort to modularize some of the RawTherapee functions so
they can be reused in a library to share with other open-source photo
editors.

The first step is to make explicit the dependencies needed by
amaze_demosaic_RT.

It compiles and seems to work normally, but I haven't tested it on pixel
shift stuff.

One concern I have is that by adding a bunch of arguments, it might clutter the code that calls Amaze, so I might end up adding a wrapper function that matches the original function signature.


heckflosse:
@CarVac Can you elaborate about the intention of your pr a bit more? Currently it just moves one occurence of `FC(..)` out of the amaze code. How about the other ones?

CarVac:
It's not done yet. Which is why I asked whether you wanted to pull the changes one at a time.

heckflosse:
@CarVac Ok, understood. I don't think we should pull (merge) change per change. But I think it's perfectly fine to have one pr which gets more commits until it's completed. Then we can merge.

CarVac:
Either way, I'd like your opinion.

Would you prefer that I write wrapper functions with the original arguments, or is it fine to break out all the arguments to, for example, calculate `ex` and `ey` like I did here?

heckflosse:
As ex and ey are values which can be calculated using the FC function, I would prefer a solution where the cfa info (rggb or grbg or similar) is passed to the amaze function. Then we could use a FC function in anonymous namspace

CarVac:
In my changes, I just use the code that used to reside inside amaze that calls FC and calculates ex and ey, but immediately preceding amaze, and then pass in the freshly calculated ex and ey.

Would you prefer it do something different?

heckflosse:
Currently we have a dependency on FC at a lot of places in amaze code (look at all the loops which use FC to calculate the starting point). To remove this dependency we should try to find a solution which

1) does not have negative impact on amaze processing time
2) leads to the simplest (and also clearly understandable) possible amaze interface.

A solution which fits 1) and 2) will be fine :)

CarVac:
I see, I completely missed that FC was used in a bunch of places.

More work to do...

Issue #4343, opened 2018-01-30
better spline interpolation


agriggio:
premise: I didn't look into this in detail (yet), these are just user-level observations so far.

The spline interpolation routines that RT uses are sometimes a bit too "unforgiving". What I mean is that often when editing curves in "custom" mode, moving a single point has a big impact on the whole curve. This makes them hard to use, to the point that I almost always prefer a "control cage" curve to the custom one.

But it doesn't have to be this way. A prominent example of a piece of software that does better is GIMP. Here are a couple of quick screenshots to illustrate what I mean:

![rt](https://user-images.githubusercontent.com/11982086/35566417-9c9d06dc-05c1-11e8-8a89-805fbc1eae9b.jpeg)

![gimp](https://user-images.githubusercontent.com/11982086/35566420-a0b86f72-05c1-11e8-82e6-dd2126b37d65.jpeg)

As you can see, the control points are essentially the same, but the GIMP curve is much nicer. In fact, moving the points in GIMP has a much more localized effect than doing the same in RT. I think it's worth looking into this and borrow the GIMP code if possible/feasible. 

(Obviously this is not urgent)

raceviper13:
I'd like this fixed too.  I actually don't like GIMP's either as the radius changes are fairly sudden at the points, but it is more manageable.

These splines could definitely look a whole lot more natural with a curve that has smooth and natural radius changes from end-to-end.

I wish i was a mathematician and could help you out, but i did not do well in my calculus classes.

Benitoite:
It appears that gimp is using a higher-order interpolation..  for example sine^3 vs sine.  I’m not with my graphing calculator at the moment. 🤓

agriggio:
@Benitoite yes I agree, I think they are simply using higher degree splines -- I didn't check though

Benitoite:
It’s a spline called Catmull Rom:
https://github.com/GNOME/gimp/search?utf8=✓&q=+Catmull&type=

agriggio:
Great, thanks for the link! Let's keep this in mind for after 5.4

raceviper13:
I see there is some work in the equation to make sure the curve doesn't double back (the slope should always be from lower left to upper right, and never reverse).  The only part of the GIMP curve that is really weird is the center point.  There seems to be a slope constraint that flattens the curve but there is an equal slope on each side.  it's almost like it's two spline curves joined together.

Hombre57:
[...]

PR #4701, opened 2018-07-26
diagonalcurves: replace cubic splines with (centripetal) Catmull-Rom splines

agriggio:
See https://en.wikipedia.org/wiki/Centripetal_Catmull%E2%80%93Rom_spline

This is my proposed fix for #4343.
It makes the "custom" curves much more controllable, IMHO at least.

### DIFF
--- a/rtengine/curves.h
+++ b/rtengine/curves.h
@@ -444,6 +444,7 @@ protected:
     DiagonalCurveType kind;
 
     void spline_cubic_set ();
+    void catmull_rom_set();
     void NURBS_set ();
 
 public:
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -317,7 +317,10 @@ inline void catmull_rom_spline(int n_points,
     double c, d, A1_x, A1_y, A2_x, A2_y, A3_x, A3_y;
     double B1_x, B1_y, B2_x, B2_y, C_x, C_y;
 
-    for (i = 0; i < n_points; ++i) {
+    res_x.push_back(p1_x);
+    res_y.push_back(p1_y);
+
+    for (i = 1; i < n_points-1; ++i) {
         t = t1 + space * i;
         
         c = (t1 - t)/(t1 - t0);
@@ -353,6 +356,9 @@ inline void catmull_rom_spline(int n_points,
         res_x.push_back(C_x);
         res_y.push_back(C_y);
     }
+
+    res_x.push_back(p2_x);
+    res_y.push_back(p2_y);
 }
 
 
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -289,7 +289,7 @@ inline double catmull_rom_tj(double ti,
                              double xi, double yi,
                              double xj, double yj)
 {
-    static constexpr double alpha = 0.25;
+    static constexpr double alpha = 0.5;
     return pow(sqrt(pow2(xj-xi) + pow2(yj-yi)), alpha) + ti;
 }
 
@@ -371,18 +371,23 @@ inline void catmull_rom_spline(int n_points,
 }
 
 
+inline void catmull_rom_reflect(double px, double py, double cx, double cy,
+                                double &rx, double &ry)
+{
+    double dx = px - cx;
+    double dy = py - cy;
+    rx = cx - dx;
+    ry = cy - dy;
+}
+
+
 void catmull_rom_chain(int n_points, int n_cp, double *x, double *y,
                        std::vector<double> &res_x, std::vector<double> &res_y)
 {
-    static const double epsilon = 1e-5;
-    double xr = x[1] - x[0];
-    double yr = y[1] - y[0];
-    double x_first = x[0] - xr * 0.01;
-    double y_first = xr > epsilon ? (yr / xr) * (x_first - x[0]) + y[0] : y[0];
-    xr = x[n_cp-1] - x[n_cp-2];
-    yr = y[n_cp-1] - x[n_cp-2];
-    double x_last = x[n_cp-1] + xr * 0.01;
-    double y_last = xr > epsilon ? (yr / xr) * (x_last - x[n_cp-1]) + y[n_cp-1] : y[n_cp-1];
+    double x_first, y_first;
+    double x_last, y_last;
+    catmull_rom_reflect(x[1], y[1], x[0], y[0], x_first, y_first);
+    catmull_rom_reflect(x[n_cp-2], y[n_cp-2], x[n_cp-1], y[n_cp-1], x_last, y_last);
 
     int segments = n_cp - 1;
     int points_segments = n_points / segments;
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -390,13 +390,12 @@ void catmull_rom_chain(int n_points, int n_cp, double *x, double *y,
     catmull_rom_reflect(x[n_cp-2], y[n_cp-2], x[n_cp-1], y[n_cp-1], x_last, y_last);
 
     int segments = n_cp - 1;
-    int points_segments = n_points / segments;
 
     res_x.reserve(n_points);
     res_y.reserve(n_points);
 
     for (int i = 0; i < segments; ++i) {
-        int n = points_segments + (i == 0 ? n_points % segments : 0);
+        int n = max(int(n_points * (x[i+1] - x[i]) + 0.5), 2);
         catmull_rom_spline(
             n, i == 0 ? x_first : x[i-1], i == 0 ? y_first : y[i-1],
             x[i], y[i], x[i+1], y[i+1],
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -320,41 +320,50 @@ inline void catmull_rom_spline(int n_points,
     res_x.push_back(p1_x);
     res_y.push_back(p1_y);
 
-    for (i = 1; i < n_points-1; ++i) {
-        t = t1 + space * i;
+    // special case, a segment at 0 or 1 is computed exactly
+    if (p1_y == p2_y && (p1_y == 0 || p1_y == 1)) {
+        for (i = 1; i < n_points-1; ++i) {
+            t = p1_x + space * i;
+            res_x.push_back(t);
+            res_y.push_back(p1_y);
+        }
+    } else {
+        for (i = 1; i < n_points-1; ++i) {
+            t = t1 + space * i;
         
-        c = (t1 - t)/(t1 - t0);
-        d = (t - t0)/(t1 - t0);
-        A1_x = c * p0_x + d * p1_x;
-        A1_y = c * p0_y + d * p1_y;
+            c = (t1 - t)/(t1 - t0);
+            d = (t - t0)/(t1 - t0);
+            A1_x = c * p0_x + d * p1_x;
+            A1_y = c * p0_y + d * p1_y;
 
-        c = (t2 - t)/(t2 - t1);
-        d = (t - t1)/(t2 - t1);
-        A2_x = c * p1_x + d * p2_x;
-        A2_y = c * p1_y + d * p2_y;
+            c = (t2 - t)/(t2 - t1);
+            d = (t - t1)/(t2 - t1);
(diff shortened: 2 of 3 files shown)


Beep6581:
Hi @agriggio !
I remembered that you wanted me to test a patch (the patch was to avoid having the curve leave the top edge once it hits it) and actually wanted to do that later today as it's weekend and I have a few photos to process. Is your latest commit the same as the patch? Anything specific to look out for when testing the branch?

agriggio:
forget about the patch, just test the latest commit

agriggio:
sorry for my brevity before. the latest commit should improve the situation with auto matching, and in general the accuracy of the new curve. everything should work more or less as in dev (minor differences are unavoidable) but the custom curves should be much easier to edit

Beep6581:
@agriggio no problem. I tested it and found two things worth noting:
1. Regarding the auto-matched tone curve, the shadows in this branch are generally lighter than in `dev` and in the embedded JPEG. I think these screenshots were from `blackcat.dng`
screenshot_20180810_224430
screenshot_20180810_224437
2. Regarding the behavior of the curve itself, the dip in the curve once it touches the top and then pulls down away again is still there and causes artifacts (solarize effect):
screenshot_20180810_224753
screenshot_20180810_224759
[soderasen_1.hdr.dng.pp3.txt](https://github.com/Beep6581/RawTherapee/files/2279428/soderasen_1.hdr.dng.pp3.txt)

agriggio:
thanks a lot for testing! still some work to do then :+1:

agriggio:
@Beep6581 I pushed further tweaks, let me know if they improve the situation.

Beep6581:
@agriggio the tweaks are good. It is now easy to get the white end to stick to the upper edge.

There is one issue. When the white point is in the corner and the second right-most point makes the curve touch the top, things are fine:
screenshot_20180812_144546
However, when you move the white point away from the corner, to the left, a dip appears
screenshot_20180812_144557
This can be easily prevented by adding a new corner node:
screenshot_20180812_144605

It's not a significant problem since the fix is so easy, but thought it was worth pointing out.

Beep6581:
Here is a better example, since in the previous example you would wonder why have three nodes when the middle one does nothing. Here the middle one does something:
screenshot_20180812_144854
screenshot_20180812_144903
screenshot_20180812_144908

Beep6581:
What are your thoughts on adding this new mode while still keeping the previous one?

agriggio:
thanks for the feedback! I'll see what I can do. I'm :-1: on adding this as an extra mode, I think RT is already over complicated and I'm only interested in making it simpler, sorry :-)
I'm fine with rejecting this pr though if it's not convincing

gaaned92:
- with last 5.4-634-g0d7d1cfc8, with amsterdam.pef and [amsterdam.pef.pp3.txt](https://github.com/Beep6581/RawTherapee/files/2281121/amsterdam.pef.pp3.txt), using Adobe DCP, I get 
screen-2018-08-12_18-44-24

The two last points are not connected. It does'nt happen with RT DCP. 


- I much prefer this new method as it avoids pathological curves and is easier to use.

- effectively, due to different interpolation, result of auto matched feature can be slightly different. But I really don't mind as, when I use it, I consider this only as a starting point. Thus  :-1: for extra mode and  :+1: to commit

agriggio:
@gaaned92 this is unexpected and most definitely a bug. I'll investigate

agriggio:
@gaaned92 I don't have the adobe profile at the moment: can you maybe upload it to filebin?

agriggio:
@Beep6581 the weird behaviour you reported in your last comment was due to a typo -- it should be fixed now

gaaned92:
- the adobe Pentax K10D DCP is here; https://filebin.net/rtr42latbwus5edj

- I don't know if it is useful, but I notice that the auto match in this branch generates a lot more points than in the dev branch.

agriggio:
@gaaned92 thanks! It should be ok now. Yes, the generated curve has more points than current dev, but that's expected with these new splines. It should still be easier to modify by hand though

gaaned92:
@agriggio it's ok now. Thanks.

Beep6581:
@agriggio the curve behaves very well now!

I want to be certain that when I move the white-point left, all areas to the right of it really are clipped, and that there is no reversal in brightness (solarisation) due to the catmull math.
screenshot_20180816_230259
I think that is the case (good), but...

A similar curve shape using different curve types should lead to a similar output, however something odd happens here (`overexposed_sky.pef`), NURBS:
screenshot_20180816_224955
Catmull (notice Tone Curve 1 has 4 points)
screenshot_20180816_225001
Why does a NURBS curve show clipped white while a similar Catmull curve does not, and why does adding an extra point which seems to make no difference in the curve lead to a funky difference in the preview? (notice Tone Curve 1 now has 5 points)
screenshot_20180816_225403
[overexposed_sky 4701.pp3.txt](https://github.com/Beep6581/RawTherapee/files/2295209/overexposed_sky.4701.pp3.txt)

The thing is, I can only reproduce that when the second curve is set to weighted mode. In all other modes, the two curve types lead to an almost identical result without solarisation, so I'm thinking this odd behavior is unrelated to your code and could be a bug in the weighting code. Besides, in this example the NURBS result seems incorrect - yes the first curve does clip the highlights, but the second curve drags those highlights down, so the highlights should be gray (as they are with the Catmull curve), but instead with NURBS they're clipped, as if the second curve's white point had no effect.

What do you make of this?

agriggio:
@Beep6581 thanks for testing! I have to think about what this means in practice... I assume that the behaviour in `dev` is different, right?

Beep6581:
@agriggio in `dev` using the custom curve with 4 points leads to clipped highlights, with 5 points still clipped highlights, but if I move the 5th point from 100.000 to 99.999 then I get gray highlights. What happens inside RT is a black box to me, but the fact that 100.000 leads to white while 99.999 leads to gray makes me worry that the allegedly-100.000 node in catmull is not really 100.000, since it too leads to gray at allegedly-100.000.

screenshot_20180817_110456
screenshot_20180817_110506

I worry about this because I often work with HDR images, and sometimes I output them to 32-bit TIFF files for stitching and subsequent post-processing in LHDR and RT. To HDR software the difference between 1 and 0.999999 can be significant. It would be bad if RT introduced solarisation in the highlights in the first step if I want to rely on the highlights being correct when tone-mapping in the second step. I want to be certain that when the curve touches the ceiling, the values really are maxed-out.

But I also want to re-state that there's a good chance that everything's fine here and it's the weighting code which leads to those odd results, though that doesn't account for why adding a 5th node to Catmull changes the result.

agriggio:
ok, I'll see what is going on

agriggio:
@Beep6581, I just took a look. There were two different issues here:

1. There was indeed a problem in the Catmull-Rom evaluation, due to the inherent imprecision of floating-point ops. Specifically, in your case even if the control point was set at (say) `[0.7, 1]`, the curve at that point was evaluated to something like `[0.7, 0.99999]`

2. it could still happen that the curve was a bit "wobbly" between two control points `[x1, 1]` and `[x2, 1]`, leading to some intermediate values to be strictly smaller than 1.

I have now fixed 1, and added a special case for 2 to make sure that segments between two control points at 0 or 1 evaluate to straight lines. Now everything seems to work as you expected, but can you please double check? Thanks!

Beep6581:
Thank you for looking into this @agriggio , I will test promptly.

Beep6581:
@agriggio it works fine now as far as I can tell. Just one thing... :]
screenshot_20180817_165500
The line appears to be reversing and going down as it flows into the node. Is that just a graphical glitch?

Beep6581:
@agriggio the behavior above is a symptom of this:
imgur-2018_08_17-17 38 33
It just looks odd when it happens at the ceiling.

I played more with this and it seems to have no visible effect on the image, I see no sudden step nor any solarisation. For certainty's sake it could be worth testing whether feeding that curve with values 0-1000 returns only increasing values, or whether any value is larger than any previous value.

[4701.rtc.txt](https://github.com/Beep6581/RawTherapee/files/2297850/4701.rtc.txt)

agriggio:
@Beep6581 as briefly explained on IRC, this is due to the placement of the (hidden) last control point needed by catmull-rom. I can try tweaking the behaviour, but first you'd have to tell me why it is wrong... :-)

Beep6581:
@agriggio I don't know whether it's wrong, but it was worth pointing out. The curve screenshot from two posts ago was of a very typical shape which users are sure to reproduce, and the curve line makes a small dip right before the node so I wanted to make sure it has no unexpected side-effects. I noticed none in the image.

Being the masters of the universe, some things we can improve, while others we just learn to live with :]

I'm happy with the way the curve now behaves, +1 for merge.

agriggio:
@Beep6581 I just tweaked the placement of the hidden points, here is a quick screenshot of the difference:

* old one:
old

* new one:
new

See if you like this better (notice it might slightly change the shape of the overall curves, since the two hidden points are put in a different place, and I also tweaked the weight of the "centripetal" factor alpha).

One thing (at least) before merging is deciding how to handle old pp3's that were using the cubic spline curves. If we just import them "as is", the curves will most definitely be different than what the user set. My proposal is to insert extra control points upon importing old pp3's, to make sure that the new curve is close to the old one. The code for this is already there (currently in `histmatching.cc`) so it should not be too hard to implement. The downside is that users will see more control points than what they entered, but at least they will get the curves they wanted... what do you think?

Beep6581:
> deciding how to handle old pp3's that were using the cubic spline curves. If we just import them "as is", the curves will most definitely be different than what the user set

I tested for that a few commits ago and found that the results were not identical but similar-enough, even when I used quite extreme curve adjustments. But I wouldn't object to more CPs to get an exact match. I'm neutral on the subject.

Beep6581:
I compared `5.4-638-g`ef57c5da0 to `5.4-640-g`e3ea0926c 
They're both fine. The former makes rounder curve segments near the ends, the latter makes flatter segments. I can't say which is better, feels more like just something one needs to adapt to one way or the other.

ef57c5da0
screenshot_20180818_140744

e3ea0926c
screenshot_20180818_140749

Odd things happen in both cases when two nodes are dragged onto each other, but odd things happen in `dev` too. https://filebin.net/tkwbcibsw40a3lv5/vid1-2018-08-18_14.12.40_x264_crf23.mp4?t=c5nhlk0u

Beep6581:
I don't like the cursor when editing a node, it obscures too much of what's happening underneath. Will change it in `dev` if no objections.

agriggio:
> I tested for that a few commits ago and found that the results were not identical but similar-enough, even when I used quite extreme curve adjustments

Even better then :-) I didn't test, so I trust you on this. Unless there are further objections, for me we can merge then.

agriggio:
@Beep6581 if you want to play with this, you can try tweaking the `alpha` value  [here](https://github.com/Beep6581/RawTherapee/blob/e3ea0926c22792edda2e4dffee3d59c9a7fb7c30/rtengine/diagonalcurves.cc#L292), keeping in mind it must be between `0` and `1`. (FYI, the previous version had `alpha = 0.25`, the current one `0.5`).

Beep6581:
No I'm fine :)

agriggio:
Here's a patch that makes the curves rounder (closer to the old version):

```diff
diff --git a/rtengine/diagonalcurves.cc b/rtengine/diagonalcurves.cc
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -289,7 +289,7 @@
                              double xi, double yi,
                              double xj, double yj)
 {
-    static constexpr double alpha = 0.5;
+    static constexpr double alpha = 0.375;
     return pow(sqrt(pow2(xj-xi) + pow2(yj-yi)), alpha) + ti;
 }
 
@@ -374,10 +374,18 @@
 inline void catmull_rom_reflect(double px, double py, double cx, double cy,
                                 double &rx, double &ry)
 {
+#if 0
     double dx = px - cx;
     double dy = py - cy;
     rx = cx - dx;
     ry = cy - dy;
+#else
+    static constexpr double epsilon = 1e-5;
+    double dx = px - cx;
+    double dy = py - cy;
+    rx = cx - dx * 0.01;
+    ry = dx > epsilon ? (dy / dx) * (rx - cx) + cy : cy;
+#endif    
 }
 
 
```

Beep6581:
@agriggio the patch is a great compromise between the two.
+1 for merge :)

PR #4712, opened 2018-08-01
Ahd demosaic speedup

heckflosse:


rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+
+    constexpr double xyz_rgb[3][3] = {        /* XYZ from RGB */
+        { 0.412453, 0.357580, 0.180423 },
+        { 0.212671, 0.715160, 0.072169 },
+        { 0.019334, 0.119193, 0.950227 }
+    };
+
+    constexpr float d65_white[3] = { 0.950456, 1, 1.088754 };
+
+    volatile double progress = 0.0;
```
Floessie: Don't know what you are trying to achieve with `volatile`. I can only only guess, it's some kind of thread safety. You're using `progress` in a critial section. Did you have problems with instruction reordering by the compiler? I think OMP mandates for reordering barriers in the right spot, but you're the expert.

Bottom line: There are only a few use cases for `volatile`, and I don't see one here. I hope you can enlighten me. :wink:

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+    if (plistener) {
+        plistener->setProgressStr (Glib::ustring::compose(M("TP_RAW_DMETHOD_PROGRESSBAR"), RAWParams::BayerSensor::getMethodString(RAWParams::BayerSensor::Method::AHD)));
+        plistener->setProgress (0.0);
+    }
+
+    for (int i = 0; i < 0x10000; i++) {
```
Floessie: `0x10000` -> `65536` for readability, as there's `65535.0` in the line below.

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+        double r = (double)i / 65535.0;
```
Floessie: `double` has a higher priority in coercion than `int`, so you can drop the ugly `(double)` C-style cast. As a bonus, `r` could be `const`.

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+
+#ifdef _OPENMP
+#pragma omp parallel
+#endif
+{
+    int progresscounter = 0;
+    float (*pix), (*rix)[3];
+    float ldiff[2][4], abdiff[2][4];
+    float (*lix)[3];
+    char *buffer = (char *) malloc (12 * TS * TS * sizeof(float) + 2 * TS * TS * sizeof(uint16_t)); /* 1053 kB per core */
```
Floessie: A raw `malloc()` in 2018? :grin:

With `<cstdint>` `uint16_t` is now part of namespace `std`.

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+                            rix[0][2 - cng] = CLIP(pix[0] + (0.5f * (pix[-1] + pix[1]
+                                                       - rix[-1][1] - rix[1][1] ) ));
+                            rix[0][cng] = CLIP(pix[0] + (0.5f * (pix[-width] + pix[width]
+                                                       - rix[-TS][1] - rix[TS][1])));
+                            rix[0][1] = pix[0];
+                        } else {
+                            rix[0][cng] = CLIP(rix[0][1] + (0.25f * (pix[-width - 1] + pix[-width + 1]
+                                                        + pix[+width - 1] + pix[+width + 1]
+                                                        - rix[-TS - 1][1] - rix[-TS + 1][1]
+                                                        - rix[+TS - 1][1] - rix[+TS + 1][1])));
```
Floessie, line 123: Didn't check this (and the above) relative indexing dance. I assume, you've double-checked it...

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+                            rix[0][2 - cng] = pix[0];
+                        }
+                        float xyz0 = 0.f;
```
Floessie: Why no array here?

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+                        lix = &lab[d][tr][tc];
+
+                        for (int i = 0; i < 4; i++) {
+                            ldiff[d][i] = std::fabs(lix[0][0] - lix[dir[i]][0]);
+                            abdiff[d][i] = SQR(lix[0][1] - lix[dir[i]][1])
+                                           + SQR(lix[0][2] - lix[dir[i]][2]);
+                        }
+                    }
+
+                    float leps = min(max(ldiff[0][0], ldiff[0][1]),
```
Floessie: Here you use `rtengine::(min|max)()`, while above it's `std::(min|max)()`. There's a unification due, IMHO.

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+                                     max(ldiff[1][2], ldiff[1][3]));
+                    float abeps = min(max(abdiff[0][0], abdiff[0][1]),
+                                      max(abdiff[1][2], abdiff[1][3]));
+
+                    for (int d = 0; d < 2; d++) {
+                        homo[d][tr][tc] = 0;
+                        for (int i = 0; i < 4; i++) {
+                            homo[d][tr][tc] += ((ldiff[d][i] <= leps) * (abdiff[d][i] <= abeps));
```
Floessie: Redundant parentheses.

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+
+            if(plistener) {
+                progresscounter++;
+
+                if(progresscounter % 32 == 0) {
+#ifdef _OPENMP
+                    #pragma omp critical (ahdprogress)
+#endif
+                    {
+                        progress += (double)32 * ((TS - 32) * (TS - 32)) / (height * width);
```
Floessie: Don't we have a `^2` function?

`(double)32` -> `32.0`.

rtengine/ahd_demosaic_RT.cc
```diff
@@ -0,0 +1,229 @@
...
+                        progress = progress > 1.0 ? 1.0 : progress;
```
Floessie: Use `std::min()`.

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
...
 }
 
+void RawImageSource::border_interpolate(unsigned int border, float (*image)[3], unsigned int start, unsigned int end)
+{
+    unsigned row, col, y, x, f;
+    float sum[8];
+    unsigned int width = W, height = H;
+    unsigned int colors = 3;
+
+    if (end == 0 ) {
```
Floessie: `if (!end) {` would express this nicer.

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
     free (image);
 }
 
+void RawImageSource::border_interpolate(unsigned int border, float (*image)[3], unsigned int start, unsigned int end)
+{
+    unsigned row, col, y, x, f;
```
Floessie: C-style!

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
...
+    unsigned int width = W, height = H;
+    unsigned int colors = 3;
+
+    if (end == 0 ) {
+        end = H;
+    }
+
+    for (row = start; row < end; row++)
+        for (col = 0; col < width; col++) {
+            if (col == border && row >= border && row < height - border) {
```
Floessie: Maybe there should be some checks for `border` relative to `start` and `height` when entering this function.

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
...
+                col = width - border;
+            }
+
+            memset (sum, 0, sizeof sum);
```
Floessie: Super-ugly.

-> `sum = {}`

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
...
+
+            for (y = row - 1; y != row + 2; y++)
```
Floessie: This assumes `start > 0`.

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
...
+                for (x = col - 1; x != col + 2; x++)
```
Floessie: This assumes `col > 0`, and I can't immediately derive if that's true from the above.

rtengine/demosaic_algos.cc
```diff
@@ -682,6 +682,41 @@ void RawImageSource::ppg_demosaic()
...
+                    if (y < height && x < width) {
+                        f = fc(y, x);
+                        sum[f] += image[y * width + x][f];
+                        sum[f + 4]++;
+                    }
+
+            f = fc(row, col);
+
+            FORCC if (c != f && sum[c + 4]) {
```
Floessie: Isn't `FORCC` dcraw specific?

Floessie (commented):
Just a few comments when reading the whole patch at once.

HTH,
Flössie

rtengine/ahd_demosaic_RT.cc
heckflosse: @Floessie You're right. I simply copied that from amaze code. Will change that now.

rtengine/ahd_demosaic_RT.cc
heckflosse: Done

rtengine/ahd_demosaic_RT.cc
heckflosse: Done

rtengine/ahd_demosaic_RT.cc
heckflosse: @Floessie Do you mean the indexing by cng (ColourNonGreen)?

rtengine/ahd_demosaic_RT.cc
heckflosse: Done

rtengine/ahd_demosaic_RT.cc
heckflosse: Fixed

rtengine/ahd_demosaic_RT.cc
heckflosse: Fixed

rtengine/demosaic_algos.cc
heckflosse: I eliminated the whole function as it now was obsolete

rtengine/ahd_demosaic_RT.cc
heckflosse: Changed

rtengine/ahd_demosaic_RT.cc
heckflosse: Changed

rtengine/ahd_demosaic_RT.cc
heckflosse: Done

Issue #3994, opened 2017-07-30
Make Preferences panel vertically scrollable


Stephan83c:
Hello,

I have a laptop on linux and the preference windows is to big for my screen.
I can't clic on "OK" for validate, how can I validate ?

Maybe you should add a  scrollbar on this window. 

best regards.

Stéphan.

Beep6581:
@Stephan83c 
1. Are you using RawTherapee 5.2?
2. Which desktop manager are you using? KDE Plasma for example allows you to alt+click a window to drag it around the screen.

sguyader:
I have the problem on my low resolution laptop screen. Try this workaround: alt + left click anywhere on the preferences window and drag it towards the top: this allows to move freely any window where want, and to reveal the hidden bottom part and buttons.

sguyader:
I have the problem on my low resolution laptop screen. Try this workaround: alt + left click anywhere on the preferences window and drag it towards the top: this allows to move freely any window where want, and to reveal the hidden bottom part and buttons.

Stephan83c:
Thank you, it's working.

afr-e:
`dev 5.4-641-g`c45c638ed on win10 from https://keybase.pub/gaaned92/RTW64NightlyBuilds/
Unfortunately, Windows doesn't have the `alt + left` trick. My workaround is to tab my way to the About-OK-Cancel row. I recall this wasn't a problem for me in `5.4-488...`.

Beep6581:
I will try to do this when time permits, but if anyone wants to jump in, feel free.

afr-e:
I took another look. It appears that

1. the padding is quite extreme
2. RT (or GTK3) ignores the space occupied by the Windows task bar and is obscured by it.

_Without the task bar. The tops of the 3 buttons are barely visible at the bottom._

![image](https://user-images.githubusercontent.com/42004791/43547874-8ba970f8-95aa-11e8-94d5-13bb0961b10e.png)

Beep6581:
@afr-e please show a screenshot of your entire desktop, uncropped.

afr-e:
Sure. With TooWaBlue, the padding is less but still too much.

_With task bar._

![image](https://user-images.githubusercontent.com/42004791/43549223-13597a36-95ae-11e8-9c10-f57684386396.png)

_Without task bar at bottom._

[...]

PR #4716, opened 2018-08-03
Makes Preferences window scrollable

Beep6581:
- The Preferences window is now scrollable.
- Cleanup of arcane variable names and obsolete comments.
- Issue #3994

### DIFF
--- a/rtgui/preferences.cc
+++ b/rtgui/preferences.cc
@@ -49,13 +49,6 @@ Preferences::Preferences (RTWindow *rtwindow)
 
     moptions.copyFrom (&options);
 
-    /*
-     * Do not increase height, since it's not visible on e.g. smaller netbook
-     * screens. The default height is about 620 pixels currently, that's why
-     * we do not set the height anymore. Netbook users will most certainly set
-     * a smaller font, so they'll be able to shrink the Preferences window and
-     * close it.
-     */
     set_size_request (650, -1);
     set_default_size (options.preferencesWidth, options.preferencesHeight);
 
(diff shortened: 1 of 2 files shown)


TooWaBoo:
Tested, works fine. 👍

afr-e:
Thanks, it works!

One nitpick is that the window starts with the smallest size possible. However, that is _much_ better than having an edge off-screen or hidden behind another object, or not being able to resize.

image

heckflosse:
Can we close?

Floessie:
@afr-e I stumbled upon this problem, too. It's only visible when deleting the options file.

This should kind of fix it:

```patch
diff --git a/rtgui/options.cc b/rtgui/options.cc
index dcba775b6..205dae821 100644
--- a/rtgui/options.cc
+++ b/rtgui/options.cc
@@ -331,7 +331,7 @@ void Options::setDefaults()
     dirBrowserHeight = 350;
     dirBrowserSortType = Gtk::SORT_ASCENDING;
     preferencesWidth = 800;
-    preferencesHeight = 0;
+    preferencesHeight = 600;
     toolPanelWidth = 400;
     browserToolPanelWidth = 465;
     browserToolPanelHeight = 600;
```

HTH,
Flössie

Beep6581:
Reproduced (had to turn my "remember previous window state" override off in KDE for RawTherapee), fix confirmed, committed.

Floessie:
Ah, the beauty of F/LOSS: Speed! :+1:

Issue #4718, opened 2018-08-06
Increase difference between enabled and disabled tools


Beep6581:
I wanted to increase the difference between enabled and disabled tool statuses, what do you think?
![screenshot_20180806_113651](https://user-images.githubusercontent.com/5844619/43709566-78ce5e32-996d-11e8-8e8f-3d4cf75086fc.png)

![screenshot_20180806_113643](https://user-images.githubusercontent.com/5844619/43709573-7c7eb91e-996d-11e8-9054-042e05685133.png)

Beep6581:
Branch `powericons`.

PR #4723, opened 2018-08-08
Power icons tweaked #4718

Beep6581:
Increases the visual difference between enabled and disabled tool status icons.


TooWaBoo:
Looks good. 👍

PR #4747, opened 2018-08-22
Update TooWaBlue theme / Adjustments for the scaled histogram.

TooWaBoo:


TooWaBoo:
@Beep6581 
Merge together with "Histogram update: different scaling modes and a new look", please.

Issue #4738, opened 2018-08-18
Navigator uses the same hand icon regardless whether panning red square or not


Beep6581:
The Navigator uses the same hand icon regardless whether panning the red square or not.
`5.4-672-g`7aeef817e

Beep6581:
Work-in-progress in branch [`navcursor`](https://github.com/Beep6581/RawTherapee/compare/navcursor).

Beep6581:
Work finished.

Beep6581:
Merged, closing.

PR #4754, opened 2018-08-28
Improvements to mouse cursor and icons

Beep6581:
Issue #4738


rtgui/cursormanager.cc
```diff
@@ -36,98 +36,133 @@ void CursorManager::init (Glib::RefPtr<Gdk::Window> mainWindow)
...
+        case CursorShape::CSResizeWidth:
+            window->set_cursor(cMoveX);
+            break;
+        case CursorShape::CSSpotWB:
+            window->set_cursor(cWB);
+            break;
+        case CursorShape::CSStraighten:
+            window->set_cursor(cRotate);
+            break;
+        case CursorShape::CSUndefined:
```
Beep6581, line 159: @Floessie is it good practice to include this `CSUndefined` item in the switch?

Floessie: This `switch` has a `default`. So, if you want `CSUndefined` to do something different than the default, you must include it.

The thing with `switch` is: If you want to catch unhandled `enum` values you must omit the `default`. Only then will the compiler warn you if you've added an `enum` value and forgot to handle it in the `switch`. Here's an example:

```c++
enum class MyEnum {
    ITEM_A,
    ITEM_B
};

void myFunction(int my_enum_value) // Notice the `int`
{
    switch (MyEnum(my_enum_value)) { // This is mandatory for `enum class` but not for `enum`. In any case, the compiler needs to know `my_enum_value` isn't a plain `int`.
        case MyEnum::ITEM_A: {
            // Do what must be done
            return;
        }

        case MyEnum::ITEM_B: {
            // Do what must be done
            return;
        }
    }
    
    // Unhandled: Default code here
}

void myOtherFunction(MyEnum my_enum_value)
{
    bool handled = false;
    
    switch (my_enum_value) {
        case MyEnum::ITEM_A: {
            // Do what must be done
            handled = true
            break;
        }

        case MyEnum::ITEM_B: {
            // Do what must be done
            handled = true
            break;
        }
    }

    if (!handled) {
        // Unhandled: Default code here
    }
    
    // Other code here
}
```

HTH,
Flössie

Beep6581: Thank you @Floessie 
I tried that, but a grep returned a whole lot of places where this enum is used, and as expected it caused compilation failures which I tried but am not skilled enough to fix. Would it be ok to commit the code as-is?

Floessie: @Beep6581 Of course! You asked about the need for including `CSUndefined` and I just noted that it must be included in this switch, because it has a different `default`. No need to change anything. The example given above just outlines good practice if you want to make sure that a `switch` contains all enum values, because the compiler will only warn you if the `switch` has no `default` and the enum type is known.

> and as expected it caused compilation failures which I tried but am not skilled enough to fix.

Did you change the `enum` to `enum class`, or what was the change you were after?

Beep6581: Thank you for explaining, it means a lot :)

> Did you change the enum to enum class, or what was the change you were after?

Yes, I used `enum class` and read about it in the docs, more out of curiosity.

TooWaBoo:
The style of the crosshair-small.png crosshair-small should be a cursor style crosshair-small

TooWaBoo:
@gaaned92
You can remove the window cursors from your build now. They aren't used anymore and there are no warnings of missing cursors in the debug window.

TooWaBoo:
Found an issue with the cursors in the navigator and curves.

If you click inside the red square in the navigator the cursor changes to the closed hand.
When you release the mouse button the cursor changes to a pointer (Should change to the open hand). 
When you move the mouse the curser changes to open hand.

If you click on a node in curves the cursor changes to a 4 arrow cursor.
When you release the mouse button the cursor still shows the 4 arrow cursor (Should change to the crosshair).
When you move the mouse the curser changes to crosshair.

PR #4756, opened 2018-08-30
modifications for portability to librtprocess

CarVac:
Here are the minimum changes needed specifically for amaze_demosaic.

I'll be doing one pull request per function made portable to break things up.

### DIFF
--- a/rtengine/demosaic/ahd.cc
+++ b/rtengine/demosaic/ahd.cc
@@ -71,7 +71,8 @@ void RawImageSource::ahd_demosaic()
             }
         }
 
-    bayerborder_demosaic(W, H, 5, rawData, red, green, blue);
+    unsigned cfarray[2][2] = {{FC(0,0), FC(0,1)},{FC(1,0),FC(1,1)}};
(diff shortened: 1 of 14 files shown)


Floessie:
@heckflosse Ingo, what's the intention of the `demosaic_source_folder` branch? Is it just a playground or intended to go mainline? If the latter, I'd like `cfarray` to be reconsidered and `setProgCancel` passed by `const&`.

heckflosse:
@Floessie The latter. My intention was to get more structure in that part of source code. What do you suggest for `cfarray`?

Floessie:
@heckflosse 
> What do you suggest for `cfarray`?

Either a `std::array<>` or a `CfArray` base class (interface) with virtual getters, if that's feasible.

Sorry for asking such a noob question, but what does `CF` and `FC` exactly stand for?

heckflosse:
@Floessie 
`CF` is Colour Filter.
`FC` is from dcraw. Don't know what it means

CarVac:
I think FC stands for Filter Color. I used CFarray because that corresponds to the acronym CFA for color filter array.

PR #4760, opened 2018-08-31
Change position of contrast threshold adjuster in demosaic and microc…

heckflosse:
…ontrast tool for consistency with sharpening tool


Beep6581 (approved):
Tested, looks good.

PR #4771, opened 2018-09-02
Highlight slider: changed default + disabled when unnecessary

### LINKED ISSUE #4562: Highlight compression threshold gets reset differently
A rather minor thing, but I noticed that when you press the general Reset button in the Exposure tab, the "Highlight compression threshold" gets set to 0. However, if you press the "Reset to Default" button paired with the slider it sets it to 33.

Thanatomanic:
This should resolve #4562 and put the highlight compression threshold to 0 by default. I haven't seen any downsides of that so far.
Additionally, it disables the Highlight compression threshold slider until a value != 0 is set for the Highlight compression. In this way it functions more similar to the Black slider and Shadow compression slider.


Beep6581:
I found two issues. The first:
1. Open amsterdam.pef,
1. Apply neutral,
1. Click Auto Levels. HC is not 0 but HCT is insensitive.

The second, and this probably applies to Shadows Compression as well:
1. Apply neutral to 3 consecutive images.
1. Open the first, set HC=0.
1. Open the second, set HC=1 HCT=0.
1. Open the third, set HC=1 HCT=1.
1. Go to the File Browser > the Batch Edit > Exposure
1. Click on the first image, shift+click on the second image. HC is disabled (good, because the value differs), HCT is enabled but insensitive (not good, I should be able to change it).
1. Shift+select all three images. HC is sensitive and disabled (good), but HCT is insensitive and disabled (bad, I should be able to change it).

To put it differently, I think in the Batch Edit tab sliders should not be insensitive, you should always be able to edit them, even if sometimes their values would be ignored in processing because some other condition causes them to be skipped.

Hombre57:
@Beep6581 Yes, there's `if (!batchMode)` all over the GUI code to prevent hiding or making the GUI insensitive in the Batch Editor. Alternatively, there's a second mechanism using the `multiImage` parameter ([see e.g. here](https://github.com/Beep6581/RawTherapee/blob/a7b7f20b837a49d1496a636e3da6f24e32e75a70/rtgui/blackwhite.cc#L656))

Thanatomanic:
@Beep6581 Can you check if the sliders behave as expected now?

Beep6581:
Fix confirmed, :+1: for merge.

P.S. We have a habit of waiting 24 hours before merging a PR to give others time to test.

Issue #4783, opened 2018-09-07
Windows installer: Add Registry entry with full executable path


muellesi:
Over at GIMP, there are multiple issues with the RawTherapee import plugin. One of them - RawTherapee's console not closing without user interaction is supposed to be fixed with #4514 I think?

But there is still one problem remaining: As I wrote in https://gitlab.gnome.org/GNOME/gimp/issues/1237, the plugin uses either PATH or a designated environment variable to find the RawTherapee executable which means, it will never work "out of the box". And if the user has no idea how to modify environment variables, it will just never work.

GIMP's darktable  plugin uses a registry entry, created by darktable to locate the corresponding binary - maybe this would be an option for RawTherapee, too? 

During installation, darktable creates two new keys in `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\App Paths`, see the image below
![darktable](https://user-images.githubusercontent.com/6770131/45205125-8d8aa880-b281-11e8-928c-63370a16e984.PNG)

The GIMP plugin is then able to read the `darktable.exe` key to start darktable.

Beep6581:
To be tested by someone with Windows:
```diff
diff --git a/tools/win/InnoSetup/WindowsInnoSetup.iss.in b/tools/win/InnoSetup/WindowsInnoSetup.iss.in
index 11601a53d..814e337a9 100644
--- a/tools/win/InnoSetup/WindowsInnoSetup.iss.in
+++ b/tools/win/InnoSetup/WindowsInnoSetup.iss.in
@@ -133,3 +133,4 @@ Root: HKCU; Subkey: "SOFTWARE\Classes\Directory\shell\Browse with RawTherapee";
 Root: HKCU; Subkey: "SOFTWARE\Classes\Directory\shell\Browse with RawTherapee\command"; ValueType: string; ValueData: "{app}\{#MyAppExeName} -w ""%1"""; Flags: uninsdeletekey; Tasks: regBrowseFolder
 Root: HKCU; Subkey: "SOFTWARE\Classes\*\shell\Open with RawTherapee"; ValueType: string; ValueData: "Open with RawTherapee"; Flags: uninsdeletekey; Tasks: regOpenFile
[...]

PR #4790, opened 2018-09-10
Add Single User/Multi User switch to Windows setup.

muellesi:
For context, see https://github.com/Beep6581/RawTherapee/issues/4783

This patch adds a new option for installing RawTherapee in Multi-User/Single-User mode to the installer:

**All changes in detail:**

* Location of the desktop icon changes depending on the selected option ( {commondesktop} vs. {userdesktop})
* Locations of context menu entries for starting RawTherapee are changed from HKLM to HKCU
* It also creates two new Registry entries that contain the full executable paths of rawtherapee.exe and rawtherapee-cli.exe for GIMP's RAW import plugin. Those new entries are also switched from HKLM to HKCU when installing in Single User mode. 

When run as a normal user, the option for the Multi-User Install is hidden:
unbenannt


Beep6581:
Fantastic!

@gaaned92 as you have been the official Windows package creator for our last two releases, could you confirm that it works well?

Beep6581:
Since @gaaned92 is away, I will go ahead and merge this in the evening.

PR #4796, opened 2018-09-11
Gray-out Retinex tool when editing non-raw image

### LINKED ISSUE #4795: Deactivate Retinex when editing non-raw image
#3704 deactivates tools which are unusable with the opened image type (non-raw, raw, or a specific type thereof), so in that fashion the Retinex tool should be deactivated for non-raw images.

Specifically, the tool should be disabled (so that the icon is not in a misleading "on" state), collapsed (so that it doesn't take up vertical space), and set insensitive.

Beep6581:
Closes #4795

### DIFF
--- a/rtgui/guiutils.cc
+++ b/rtgui/guiutils.cc
@@ -917,6 +917,11 @@ bool MyExpander::on_enabled_change(GdkEventButton* event)
     return false;
 }
 
+void MyExpander::setSensitive(bool makeSensitive)
+{
+    set_sensitive(makeSensitive);
+}
+
 /*
  *
  * Derived class of some widgets to properly handle the scroll wheel ;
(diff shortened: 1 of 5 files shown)


rtgui/guiutils.cc
```diff
@@ -917,6 +917,11 @@ bool MyExpander::on_enabled_change(GdkEventButton* event)
     return false;
 }
 
+void MyExpander::setSensitive(bool makeSensitive)
```
Floessie: @Beep6581 I don't quite understand this one. Isn't a `MyExpander` a `Gtk::VBox`, so that you can call `set_sensitive()` directly?

Beep6581: @Floessie you're right, it was unnecessary as-is.

Beep6581: Cleaned up in latest commit.

Beep6581:
@Floessie good to go?

Floessie:
@Beep6581 Sure, go ahead! :+1:

Issue #4777, opened 2018-09-05
Avoid desaturation and colour tint caused by RAW-CA correction


TooWaBoo:
The RAW-CA correction tool desaturates images. See https://github.com/Beep6581/RawTherapee/issues/4774#issuecomment-418732568
It would be great if the desaturation can be reduced. Ping @heckflosse 
CA off
![ca-off](https://user-images.githubusercontent.com/13800920/45104188-cae02080-b131-11e8-837a-3004d1555a45.png)
CA on
![ca-on](https://user-images.githubusercontent.com/13800920/45104205-d3d0f200-b131-11e8-83b6-d5c2f61c64c0.png)

CA off
![ca-off2](https://user-images.githubusercontent.com/13800920/45104334-28746d00-b132-11e8-9bbf-389739c15824.png)
CA on
![ca-on2](https://user-images.githubusercontent.com/13800920/45104350-34602f00-b132-11e8-8be8-0d602cc80b4c.png)

Upload the files later
`Filebin is currently unavailable (503) due to maintenance. Please retry later.`

TooWaBoo:
@heckflosse 
https://filebin.net/grdfzgriwg8tbsqy

heckflosse:
@TooWaBoo first result:

left is old code, middle is without ca correction, right is new code with `avoid desaturation`

![grafik](https://user-images.githubusercontent.com/13733487/45150038-3c1fe200-b1cb-11e8-8a0c-b8f2d4ebba60.png)

TooWaBoo:
This looks awesome, Great job. 👍

agriggio:
@heckflosse Ingo, do we really need an option for this? Who could possibly want the colour shift introduced by the "old" behaviour?

heckflosse:
@agriggio Alberto, I'm also fine without an option for this. @Beep6581 @Floessie @TooWaBoo @Hombre57 @Desmis @Thanatomanic @iliasg ?

Thanatomanic:
Agreed with @agriggio to avoid colour shift by default.

heckflosse:
@Thanatomanic It is default, but currently only for new files.

agriggio:
To me the colour shift is a bug, so it seems strage that we have a checkbox that says "do you want to apply the bug fix?" :-)

TooWaBoo:
Found a bug in Avoid color shift. I get a green circle in this picture .
Iteration = 2 and Avoid color shift = on.
[...]

PR #4797, opened 2018-09-11
Raw ca correction. Iterative processing and Avoid color shift.

heckflosse:


heckflosse:
I just found an old bug in raw ca correction code. Fix follows soon.

heckflosse:
@agriggio Alberto, the last commit fixes the artifacts with your file from [here](https://github.com/Beep6581/RawTherapee/issues/4777#issuecomment-420309818)

heckflosse:
I just found a case where `avoid colour shift` leads to a colour shift :(

left without raw ca correction, right with 2 iterations
grafik

heckflosse:
Merged to get more tests.

PR #4800, opened 2018-09-12
Make formats in `SaveFormatPanel` less error-prone

Floessie:
Hi,

This is the continuation of Ingo's work to fix the format list in `SaveFormatPanel`.

Best,
Flössie

### DIFF
--- a/rtgui/saveformatpanel.cc
+++ b/rtgui/saveformatpanel.cc
@@ -16,9 +16,89 @@
  *  You should have received a copy of the GNU General Public License
  *  along with RawTherapee.  If not, see <http://www.gnu.org/licenses/>.
  */
+#include <array>
+
 #include "saveformatpanel.h"
 #include "multilangmgr.h"
 #include "guiutils.h"
+
+namespace
+{
+
+const std::array<SaveFormat, 7> sf_templates = {
+     SaveFormat(
+         "jpg",
+         8,
+         90,
+         2,
+         8,
+         false,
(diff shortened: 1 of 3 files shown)


rtgui/saveformatpanel.cc
```diff
@@ -124,31 +124,29 @@ SaveFormatPanel::~SaveFormatPanel ()
...
-    } else if (sf.format == "tif" && sf.tiffBits == 16 && sf.tiffFloat) {
-        format->set_active (3);
-    } else if (sf.format == "tif" && sf.tiffBits == 16) {
-        format->set_active (2);
-    } else if (sf.format == "tif" && sf.tiffBits == 8) {
-        format->set_active (1);
+    std::pair<int, std::size_t> index;
+
+    for (std::size_t i = 0; i < sf_templates.size(); ++i) {
+        const int weight =
```
heckflosse: @Floessie The calculation of `weight` needs to be explained imho

heckflosse (commented):
:+1: for merge

rtgui/saveformatpanel.cc
Floessie: @heckflosse I'll add some comments to the code...

Beep6581 (approved):

rtgui/saveformatpanel.cc
heckflosse: :+1:

PR #4823, opened 2018-09-20
Added option to use a non-linear response for Adjuster sliders

agriggio:
Some of the RT parameters that are currently associated to Adjusters are very
hard to edit precisely by dragging the sliders, because small changes to the
default produce quite visible results. Prominent examples include black level,
WB tint, raw black and white points, and lens correction parameters
(distortion, CA, vignetting, perspective). The problem is made worse for those
settings in which not only small changes are significant, but also the
associated Adjusters have a very large range (again, think of black point and
WB tint). This is due to the fact that the current Adjusters have a linear
response. This commit adds an option to use a non-linear (specifically
logarithmic) response, which causes the sliders to move "slowly" around a
designated pivot point, and progressively faster the further you move away
from the pivot.

Besides adding the functionality to the Adjuster class, this changeset also
enables this behaviour for the following adjusters:

- exposure compensation
- black point
- lightness/contrast/saturation/chromaticity (both in exposure and in L*a*b*)
- WB tint
- channel mixer
- lens corrections (perspective, distortion, CA)
- rotation
- raw black and white points
- raw CA correction

### DIFF
--- a/rtgui/adjuster.cc
+++ b/rtgui/adjuster.cc
@@ -317,8 +317,8 @@ void Adjuster::resetPressed (GdkEventButton* event)
 
 double Adjuster::shapeValue (double a)
 {
-
-    return round(a * pow(double(10), digits)) / pow(double(10), digits);
+    double val = round(a * pow(double(10), digits)) / pow(double(10), digits);
+    return val == -0.0 ? 0.0 : val;
 }
 
 void Adjuster::setLimits (double vmin, double vmax, double vstep, double vdefault)
@@ -413,7 +413,7 @@ void Adjuster::sliderChanged ()
     }
 
     spinChange.block (true);
-    double v = getSliderValue();
+    double v = shapeValue(getSliderValue());
     spin->set_value (addMode ? v : slider2value(v));
     spinChange.block (false);
 
--- a/rtgui/adjuster.cc
+++ b/rtgui/adjuster.cc
@@ -46,6 +46,10 @@ Adjuster::Adjuster (Glib::ustring vlabel, double vmin, double vmax, double vstep
     grid = NULL;
     imageIcon1 = imgIcon1;
 
+    logBase = 0;
+    logPivot = 0;
+    logAnchorMiddle = false;
+
     if (imageIcon1) {
         setExpandAlignProperties(imageIcon1, false, false, Gtk::ALIGN_CENTER, Gtk::ALIGN_CENTER);
     }
@@ -289,13 +293,13 @@ void Adjuster::resetValue (bool toInitial)
 
     if (toInitial) {
         // resetting to the initial editing value, when the image has been loaded
-        slider->set_value (addMode ? defaultVal : value2slider(defaultVal));
+        setSliderValue(addMode ? defaultVal : value2slider(defaultVal));
     } else {
         // resetting to the slider default value
         if (addMode) {
-            slider->set_value (0.);
+            setSliderValue(0.);
         } else {
-            slider->set_value (value2slider(ctorDefaultVal));
+            setSliderValue(value2slider(ctorDefaultVal));
         }
     }
 }
@@ -333,7 +337,7 @@ void Adjuster::setLimits (double vmin, double vmax, double vstep, double vdefaul
     slider->set_digits (digits);
     slider->set_increments (vstep, 2.0 * vstep);
     slider->set_range (addMode ? vmin : value2slider(vmin), addMode ? vmax : value2slider(vmax));
-    slider->set_value (addMode ? shapeValue(vdefault) : value2slider(shapeValue(vdefault)));
+    setSliderValue(addMode ? shapeValue(vdefault) : value2slider(shapeValue(vdefault)));
     //defaultVal = shapeValue (vdefault);
     sliderChange.block (false);
     spinChange.block (false);
@@ -369,7 +373,7 @@ void Adjuster::spinChanged ()
     }
 
     sliderChange.block (true);
-    slider->set_value (addMode ? spin->get_value () : value2slider(spin->get_value ()));
+    setSliderValue(addMode ? spin->get_value () : value2slider(spin->get_value ()));
     sliderChange.block (false);
 
     if (delay == 0) {
@@ -409,7 +413,8 @@ void Adjuster::sliderChanged ()
     }
 
     spinChange.block (true);
-    spin->set_value (addMode ? slider->get_value () : slider2value(slider->get_value ()));
+    double v = getSliderValue();
+    spin->set_value (addMode ? v : slider2value(v));
     spinChange.block (false);
 
     if (delay == 0 || afterReset) {
@@ -447,7 +452,7 @@ void Adjuster::setValue (double a)
     spinChange.block (true);
     sliderChange.block (true);
     spin->set_value (shapeValue (a));
-    slider->set_value (addMode ? shapeValue(a) : value2slider(shapeValue (a)));
+    setSliderValue(addMode ? shapeValue(a) : value2slider(shapeValue (a)));
     sliderChange.block (false);
     spinChange.block (false);
     afterReset = false;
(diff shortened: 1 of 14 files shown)


Thanatomanic:
Wonderful idea!

heckflosse:
@agriggio I also like the idea! Though I could easily get empty fields using the red/blue adjusters in raw ca correction

grafik

Not that I care about. In fact I think, manual raw ca correction is not really needed anymore. Just wanted to point that out. Maybe it happens for other adjusters as well.

I can also get the empty fields in dev branch, but it's much harder to get them there. So it's not an issue caused by your changes, but an issue which is increased by your changes.

heckflosse:
@agriggio Alberto, I tried a bit more using your pr. Currently I see no real improvement about the currently available behaviour which is to wait a bit between clicking the adjuster and moving it (which automagically gives smaller adjustment steps). But I may miss somesting of course...

agriggio:
@heckflosse I never experienced empty fields -- but I'm still on gtk 3.18. which version are you using?

regarding the behaviour: I don't like the "small increments after delay" provided by gtk, I think it's cumbersome and typically it's too slow (meaning the steps are too small). for me the new behaviour is a massive improvement in workflow. but if that's just for me, no problem in dropping this :+1:

Beep6581:
@agriggio I love it!

Beep6581:
GTK3's GtkScale has a fine-tuning mode which turns on when you hold a slider for a while or when you hit shift. I actually find it a bit annoying, since when you're focused on the preview and slowly move the slider it has a habit of unexpectedly going into fine-tuning mode.

This log mode makes these sliders more easily tunable without relying on fine-tuning mode.

heckflosse:
@agriggio I didn't intend dropping this. Just wanted to point out some things. About the empty fields: V3.22.2 here.

agriggio:
@Beep6581 that's exactly my experience as well. In fact, I'd like to find out whether it is possible to turn off the fine-tuning mode completely, as I find it more annoying than useful. If so, would you accept a patch that adds an option in preferences to enable/disable this?

agriggio:
unfortunately, it seems the fine tuning mode is built deeply into gtk, with no obvious way to disable it...

Beep6581:
@agriggio haha, I asked the GTK+ devs before you suggested that, but they said it's not possible.

agriggio:
@heckflosse sorry, my bad. I thought you meant the added code complexity was not worth it -- and I just wanted to stress that this would be ok for me :-) (even though I disagree of course)

heckflosse:
@agriggio This fixes the empty fields. To reproduce, move the raw ca correction blue or red slider slowly to the left.

```diff
diff --git a/rtgui/adjuster.cc b/rtgui/adjuster.cc
index e0eb62a1f..7bcaeb8e5 100644
--- a/rtgui/adjuster.cc
+++ b/rtgui/adjuster.cc
@@ -413,7 +413,7 @@ void Adjuster::sliderChanged ()
     }
 
     spinChange.block (true);
-    double v = getSliderValue();
+    double v = shapeValue(getSliderValue());
     spin->set_value (addMode ? v : slider2value(v));
     spinChange.block (false);
 
```

agriggio:
@heckflosse Ingo, I tried but can't reproduce on GTK+ 3.18. However, the fix looks good to me, can you please push?

heckflosse:
@agriggio Alberto, my fix was not complete. There was one case missing. This one fixes the missing case as well.

```diff
diff --git a/rtgui/adjuster.cc b/rtgui/adjuster.cc
index e0eb62a1f..3f12b0cfb 100644
--- a/rtgui/adjuster.cc
+++ b/rtgui/adjuster.cc
@@ -317,8 +317,8 @@ void Adjuster::resetPressed (GdkEventButton* event)
 
 double Adjuster::shapeValue (double a)
 {
-
-    return round(a * pow(double(10), digits)) / pow(double(10), digits);
+    double val = round(a * pow(double(10), digits)) / pow(double(10), digits);
+    return val == -0.0 ? 0.0 : val;
 }
 
 void Adjuster::setLimits (double vmin, double vmax, double vstep, double vdefault)
@@ -413,7 +413,7 @@ void Adjuster::sliderChanged ()
     }
 
     spinChange.block (true);
-    double v = getSliderValue();
+    double v = shapeValue(getSliderValue());
     spin->set_value (addMode ? v : slider2value(v));
     spinChange.block (false);
 
```

Ok to push this one?

agriggio:
sure, go ahead :+1:

heckflosse:
@agriggio done

agriggio:
any objections to merging?

heckflosse:
@agriggio :+1: for merge

micheal65536:
This breaks backwards-compatibility for old profiles and also breaks my workflow as I am familiar with what sort of values give what sort of results and the way that the values are interpreted has changed as a result of this.

There needs to be a configuration option to disable this "feature".

heckflosse:
@micheal65536 
> This breaks backwards-compatibility for old profiles

How can this break backwards compatibility for old profiles?
At least when using rawtherapee-cli with old profiles nothing should be broken.
Or do I miss something?

agriggio:
I also don't understand, this is purely a GUI thing

micheal65536:
Suppose I have a profile that specifies a contrast of "20". When the profile is loaded, the slider value is set to "20".

With the old code, this value was used as-is. With the new code, the logarithmic scale is being applied to this value.

At least, I think that's the case. All I know is that loading old profiles in the new RawTherapee version gives a vastly different result for the basic parameters such as lightness, contrast, and saturation. I will need to conduct a more thorough investigation to double-check if this commit is to blame but it seemed to be the most likely cause of the issue when I looked at the commit history.

agriggio:
Hi,

> At least, I think that's the case. All I know is that loading old profiles in the new RawTherapee version gives a vastly different result for the basic parameters such as lightness, contrast, and saturation. 

Which version of RT? Do you have a pp3 and a raw file to reproduce?

> I will need to conduct a more thorough investigation to double-check if this commit is to blame but it seemed to be the most likely cause of the issue when I looked at the commit history.

that will be very helpful, thanks for your time!

micheal65536:
OK, I've done git bisect and the commit that actually introduced the difference is f490c6468221913e72626def9d9e09178c42a06b. I should probably open a separate issue for that but I'm not sure if it's worth the effort for me.

The old version that I was using was 4.2. While I am aware that there can be minor changes between results from different versions, this is a more significant difference. I'm not 100% sure exactly what's happening internally but the difference seems to be mainly that everything has a "red" color tint as opposed to the "blue/green" color tint from the old version.

I don't have a PP3 or a raw file that I am able to share however I can say that the camera in question is a Canon EOS 1300D and the issue was present with multiple raw files and processing profiles that I tested, including a profile that used only the most basic adjustments (no curves, no white balance adjustment, etc.). This was in a clean Ubuntu 16.04 virtual machine with RawTherapee compiled from source.

I haven't looked at the commit in question but I will look at it now and see if I can find what's changed that might be causing this.

EDIT: I see that camera data for the EOS 1300D has been added on line 781 in the commit (currently line 982 in the dev branch). I would appreciate it if someone with an understanding of how this file is used could indicate what data would have been used before this data was added and how I might go about using the older data in the current version of RawTherapee.

Thanatomanic:
It's unsurprising that your edits look different before and after the commit, since it introduces a custom colour matrix for the EOS 1300D.

micheal65536:
@Thanatomanic See ninja edit.

Thanatomanic:
Also, this should definitely be discussed in a separate issue.

micheal65536:
Opened issue #5154.

Bezierr:
May I ask what has happened to this "log scale" functionality? It doesn't seem to be part of the dev branch?!?

Thanatomanic:
@Bezierr It's definitely present in the dev branch, but not on every slider. Take a look: https://github.com/Beep6581/RawTherapee/search?q=setLogScale%28
Play for example with the Raw Black Point slider and see what it does. The scale runs up to 2048, but definitely not linearly.

Issue #4446, opened 2018-03-19
Save As - Wrong file suffix when clicking on existing file then changing file format


blitzgneisserin:
Hi, basically this bug works like this: I export a photo directy to JPEG or TIFF, and right after that I do another direct export and change the file type. Result is the same file type, but if it is a JPEG it is as big as the TIFF. E.g. first I export to JPEG. Then I switch file type to TIFF and export again. Result is a JPEG with usually more than 50 MB. I do not know what that file is. It is called JPEG but how can it be this big? I can open it and view it but I have no idea what it actually is.
I tried this with 5.3 (portable) from portableapps and 5.4. rc-3 (https://gaaned92.keybase.pub/RTW64NightlyBuilds/RawTherapee_5.4-rc3_WinVista_64.zip?dl=1).
OS is Win 10. When using 5.4-rc3 I also got a strange screen:
![screenshot 103](https://user-images.githubusercontent.com/16564210/37627306-19daa432-2bd4-11e8-861b-80200a250297.png)
And here the strange file in the Windows Explorer:
![screenshot 104](https://user-images.githubusercontent.com/16564210/37627418-95093574-2bd4-11e8-9899-53d5405595a4.png)
I opened the file in Affinity and it appears to be a 16-bit-JPEG
![screenshot 105](https://user-images.githubusercontent.com/16564210/37627511-0c2e4810-2bd5-11e8-8b3e-61464cbd6f0e.png)
Here the data of my PC:
![screenshot 102](https://user-images.githubusercontent.com/16564210/37627706-f5667dfe-2bd5-11e8-9be5-b5ea082d36d8.png)
It must be a wrong file ending, right?

Beep6581:
Could not reproduce.

@blitzgneisserin please post the *exact* steps to reproduce, or even better a video screen recording.

Floessie:
Could not reproduce as well.

When I hit Ctrl+S on let's say "DSC_1234.NEF" the save dialog proposes "DSC_1234" in the top line. Selecting JPEG produces a JPEG. Even if I hit Ctrl+S again while saving, it still proposes "DSC_1234", and selecting TIFF yields a TIFF. The only way to create a TIFF file with a ".jpg" extension is to select the "DSC_1234.jpg" in the file selection and TIFF as output format. But that is expected.

blitzgneisserin:
open an ORF, strg + s, choose file type (e.g. jpeg), save
[...]

PR #4857, opened 2018-10-09
Turn Listeners into abstract interfaces

Floessie:
Hi,

When working on #4446 I deliberately changed `SaveAsDialog::formatChanged(Glib::ustring)` to `SaveAsDialog::formatChanged(const Glib::ustring&)` unaware of the fact it was derived. It compiled without a warning and I had to find out for myself, why the dialog behaved strangely.

The problem was the fact, that the `FormatChangeListener` had a default implementation for `formatChanged()`. I then found out, that many of the Listeners are implemented in the same way, making it impossible to detect a signature miss.

One solution would be the marking of all derived methods with `override`, which can be automatically and globally done by `clang-tidy`. But this would only manifest the status quo and not reveal signature misses we have today or tomorrow.

So I went for a IMHO better solution: I made almost all Listeners abstract. This uncovered Listeners that weren't used anymore and will force future changes to comply with the base class. Fortunately, there were no signature misses in the code so far.

"Almost all" means, there were two or three Listeners that are more than just callback classes. I'll have to think a bit more about them.

During this code revision I also made the usual `const` and whitespace corrections. One outcome is that `LUT<T>::operator =()` is finally `const` correct. :tada:

There are still things to improve, and I'll probably add some commits here. Please feel free to do the same.

I know this is a big one, but it's certainly worth it. :smile: 

Best,
Flössie


heckflosse (commented):
@Floessie is rtetest.cc still needed or can we throw it away?

Floessie:
@heckflosse Ingo, I think you're right, I found no reference to it at first glance.

Floessie:
I won't be at home next week, so this will lose momentum. If someone likes to chime in, these are the items still left on my agenda:

- [`CurveListener`](https://github.com/Beep6581/RawTherapee/blob/abstract-listeners/rtgui/curvelistener.h#L26): Moving `multi` down to the implementations is easy. `blendPipetteValues()` is more of a problem.
- [`PointerMotionListener`](https://github.com/Beep6581/RawTherapee/blob/abstract-listeners/rtgui/pointermotionlistener.h#L22): Ideas welcome how to cope with `sig_cycle_(rgb|hsv)` (@Thanatomanic?).
- [`ThesholdAdjusterListener`](https://github.com/Beep6581/RawTherapee/blob/abstract-listeners/rtgui/thresholdadjuster.h#L32): This is already turned into an abstract interface, but this interface is too broad. IMHO an `enum class Type` and four `double`s would do, but I've not yet digged deeper.

Best,
Flössie

Thanatomanic:
@Floessie Interesting you ping me, because I have difficulty grasping the point of your PR. It looks good if it got rid of unused stuff and is 'better' code now, but to say I can contribute here is a stretch 😉 

FWIW, if your changes make no functional difference but improve the code nevertheless, it could be merged now and improve the remaining points later?

Floessie:
@Thanatomanic 
> It looks good if it got rid of unused stuff and is 'better' code now

Well spotted. :smile: I pinged you, because I had a vage recollection that you're well involved into sigc++ signals and might have a good idea on how to organize this differently (and didn't want to bother Hombre). Never mind. After all, it's more an organizational than a signalling problem.

> FWIW, if your changes make no functional difference but improve the code nevertheless, it could be merged now and improve the remaining points later?

I wouldn't mind. It shouldn't make a functional difference and is purely janitorial.

Beep6581:
From a 5.5 release point of view, and considering that it affects 159 files, it would be better to merge this sooner rather than later.

Floessie:
@Beep6581 If there's a consensus it's up to you to hit the merge button.

Beep6581:
@agriggio @heckflosse @Hombre57 @Thanatomanic please :+1: the comment if you agree to merge.

Beep6581:
Ping @agriggio

agriggio:
sorry, I didn't notice this, my bad :-(
no objections to merge :+1:

Beep6581:
Merged.

@Floessie I left the branch in case you needed it, feel free to hit the "Delete branch" button otherwise.

Floessie:
@Beep6581 I can open a new branch if necessary. Thanks!

PR #4870, opened 2018-10-16
Changed min. width of entries in filter panel (filterpanel.cc)

TooWaBoo:
old
old

new
neu


TooWaBoo:
Made a mistake on merging from Beep.
Will do it again.

PR #4872, opened 2018-10-17
Changed min. width of entries in filter panel (filterpanel.cc)

TooWaBoo:
old

new


TooWaBoo:
Made a mistake when I merged from Rep. Beep. I've made it new.

Beep6581:
Merged as there were no objections in https://github.com/Beep6581/RawTherapee/pull/4870#issue-223115530

PR #4879, opened 2018-10-19
Double-click resets slider

### LINKED ISSUE #4612: Functional middle/double click on sliders: reset to default
A feature that I use quite often in darktable, is double clicking on a slider to reset it to its default value. Would it be beneficial to have the same behavior in RT?
Edit: Could also be implemented by using a middle mouse button click.

It ~~would~~ could also clean up the GUI a little, as it would make the reset buttons superfluous.

Thanatomanic:
Learning on the job: I'm trying to fix #4612, but I would like some advice on how to proceed. 

I've managed (rather easily) to catch the double-click event by modifying the `MyHScale` class. But now I am unsure how I can trigger the actual reset functionality that is required. Of course this reset function only exists in the class that actually creates a `MyHScale` object, in my case the `Adjuster` class which has a `resetPressed` function. So.. uhm... how to proceed from here? Maybe there is even a C++ or GTK+ guide that can help me understand this from a purely programmatic pov?

https://github.com/Thanatomanic/RawTherapee/blob/doubleclick/rtgui/guiutils.cc#L1254

### DIFF
--- a/rtgui/guiutils.cc
+++ b/rtgui/guiutils.cc
@@ -1247,6 +1247,17 @@ bool MyHScale::on_scroll_event (GdkEventScroll* event)
     return false;
 }
 
+bool MyHScale::on_button_press_event (GdkEventButton* event)
+{
+
+    if (event->type == GDK_2BUTTON_PRESS) {
+        // How do I trigger the reset event ... from here?
+        return false;
+    } else {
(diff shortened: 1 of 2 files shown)


Beep6581:
@Thanatomanic it seems this PR was never ready for merging, and now it has merge conflicts. I will close it - please handle it in an issue and open a PR once ready for merging.

Issue #4889, opened 2018-10-23
Automatically determined parameter should always be able to be overriden by user


Thanatomanic:
There is some discussion on how the GUI should look and behave when a checkbox exists that enables some automatic behavior. See e.g. @heckflosse and @agriggio in https://github.com/Beep6581/RawTherapee/issues/4866#issuecomment-432278164_

My opinion is that whenever an 'Auto' checkbox exists it should have the following simple behavior:
1) If it is checked, the corresponding slider should move to the automatically calculated value. _The slider should remain active._
2) If the user disagrees with the automatic value he can immediately move the slider. This uncheckes the 'Auto' checkbox.

This is different from the current behavior, because the slider actually becomes inactive when 'Auto' is checked, and the automatically calculated value is never shown to the user.

Reason enough for a change in usability?

ff2000:
I don't really like that unlabeled checkbox. You have to hover it in order to get to know what it does. A togglebutton labeled "Auto" would make more sense to me.
Probably could be replaced with a normal button for things that really get calculated just once.

Thanatomanic:
@ff2000 Do you know DxO PhotoLab and how they handle the automatic stuff? I'm quite a fan of their approach:
![ddtmo52](https://user-images.githubusercontent.com/6567747/47430851-cd0e4500-d79a-11e8-876f-ba083f465f6e.jpg)

ff2000:
I don't know DxO (Linux is all I have). But that is different!
It uses a wand to signal that some automagic stuff is happening, I am quite sure this wand is only used for this specific use of an "auto" state. A checkbox is just a general widget that, especially if it isn't placed near a label. Of course we could use a specially styled checkbox. Or a togglebutton with that "wand" icon instead of an "auto" label ;)

Beep6581:
I agree that a checkbox by itself is vague, even though one quickly learns what it does from the tooltip.

I don't know about DxO, but our "auto" state is persistent, so a checkbox is semantically perfect for the role.

I see three paths we could take:
1. Leave it as is.
2. Replace it with a toggle-button icon.
3. Prepend an "Auto" label.

My opinions:
* I don't mind 1.  
* The problem with 2 is that there is no universal symbol for "automatic", so no matter which symbol we come up with, it will still not clarify what it does for everyone, some people will still fall back to checking the tooltip.
* 3 seems ideal to me. "Auto" is nice and short, and is understood in all the languages I know, so the risk of ending up with a long label is small.

[...]

PR #4890, opened 2018-10-23
Auto-calculate dual-demosaic contrast threshold

heckflosse:


rtengine/ipsharpen.cc
```diff
@@ -176,7 +176,8 @@ BENCHFUN
 
     // calculate contrast based blend factors to reduce sharpening in regions with low contrast
     JaggedArray<float> blend(W, H);
-    buildBlendMask(luminance, blend, W, H, sharpenParam.contrast / 100.f, sharpenParam.deconvamount / 100.0);
+    float contrast = sharpenParam.contrast / 100.f;
+    buildBlendMask(luminance, blend, W, H, contrast, sharpenParam.deconvamount / 100.0);
```
Thanatomanic: Tiny thing here maybe: 100.f to be consistent with the rest?

rtengine/ipsharpen.cc
```diff
@@ -222,7 +223,8 @@ void ImProcFunctions::sharpening (LabImage* lab, const SharpeningParams &sharpen
     if(showMask) {
         // calculate contrast based blend factors to reduce sharpening in regions with low contrast
         JaggedArray<float> blend(W, H);
-        buildBlendMask(lab->L, blend, W, H, sharpenParam.contrast / 100.f, sharpenParam.method == "rld" ? sharpenParam.deconvamount / 100.0 : 1.f);
+        float contrast = sharpenParam.contrast / 100.f;
+        buildBlendMask(lab->L, blend, W, H, contrast, sharpenParam.method == "rld" ? sharpenParam.deconvamount / 100.0 : 1.f);
```
Thanatomanic: Here as well 100.f ?

rtengine/rt_algo.cc
```diff
@@ -289,7 +382,7 @@ int calcContrastThreshold(float** luminance, float **blend, int W, int H) {
         }
     }
 
-    const float limit = (W - 2) * (H - 2) / 100.f;
+    const float limit = (W - 4) * (H - 4) / 100.f;
```
Thanatomanic, line 385: Is this a raw-crop thing hardcoded? Or are W and H already influenced by the `border` parameter in the demosaicer?

rtengine/ipsharpen.cc
Floessie: :+1: but not because of consistency, but `buildBlendMask()` taking a `float` in that position.

heckflosse: Done

rtengine/ipsharpen.cc
heckflosse: Done

rtengine/rt_algo.cc
heckflosse: No. It's not related to a raw crop. It's because the algorithm needs a border. Have a look at the loop before (luminance[j][i+2] ...)

ff2000:
Something that came to my mind when commenting on #4889:
Do you take into account if the image is cropped? Because the cropped region may result in a different threshold. (which might be an argument for keeping the checkbox/togglebutton to automatically re-calculate the value when crop is changed)

heckflosse:
@ff2000 No, crop and transformations (straighten, rotate ...) are not taken into account.

heckflosse:
@ff2000 That's what the rawcrop branch is for. Unfortunately is stalled atm.

Thanatomanic:
That might be something to improve later anyway, but I'm not sure if it makes _that_ much of a difference if you take the entire image or only a crop. Not in the general use at least.

PR #4891, opened 2018-10-23
Flat field: Show auto calculated clip control value in ui

heckflosse:
This pr is to show the auto-calculated flatfield clipcontrol value in ui

### DIFF
--- a/rtgui/flatfield.cc
+++ b/rtgui/flatfield.cc
@@ -104,6 +104,11 @@ FlatField::FlatField () : FoldableToolPanel(this, "flatfield", M("TP_FLATFIELD_L
     }
 }
 
+FlatField::~FlatField ()
+{
+    idle_register.destroy();
+}
+
 void FlatField::read(const rtengine::procparams::ProcParams* pp, const ParamsEdited* pedited)
 {
     disableListener ();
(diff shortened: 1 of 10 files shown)


rtgui/flatfield.cc
```diff
@@ -403,3 +403,21 @@ void FlatField::setShortcutPath(const Glib::ustring& path)
...
+        FlatField *me;
+        int n;
+    };
+    const auto func = [](gpointer data) -> gboolean {
+        Data *d = static_cast<Data *>(data);
+        FlatField *me = d->me;
+        me->disableListener();
+        me->flatFieldClipControl->setValue (d->n);
+        me->enableListener();
+        return FALSE;
```
Floessie, line 425: @heckflosse Ingo, you should delete `d` as a last step, as the `IdleRegister` won't touch it. It will still leak in case `func` wasn't called due to early destruction, and this is something I have to think about. Maybe an optional deleter (as a `std::function<>`) would be the right approach.

Furthermore, you should call `idle_register.destroy()` early in `~FlatField()`. Its automatic destruction is more like a safety net, other members of the containing class could already be gone.

Best,
Flössie

heckflosse: Done

Issue #4804, opened 2018-09-14
Add color wheels


eprovst:
Color wheels are a feature commonly found in color grading and correction software for film, however can be of great use for photographers too and is often more intuitive as it uses hue instead of RGB values.

Color wheels (see Speedgrade, Davinci Resolve, CameraBag etc.) are probably the easiest way to correct and grade color in an image. they allow you to push a certain color into the image using a different function (lift, gamma, gain and sometimes offset) per wheel. Most programs also have a luminance slider for every wheel which gives you the ability to quickly correct exposure.

They are a cruder tool than curves, but are a lot faster and usually get the job done, very quickly.

One of the best implementations of this is found in Davinci Resolve --- of which there is a (only slightly limited) free version (also runs on Linux).

The color wheels look something like this:
![davinci](https://user-images.githubusercontent.com/5394549/44935963-17fb7580-ad73-11e8-9e46-f21c79c5119d.jpg)

The slider under the wheels is for luminance.

(Note: originally part of issue #4763)

Thanatomanic:
I assigned myself because I am very fond to include this in RT. It's going to be a longer term project... Are you okay with that @ElecProg ?

eprovst:
@Thanatomanic Please take as long as you need, the thing I'm enjoying the most about RT is the near perfection of all the features it has, and getting software (especially a side project) to such a point takes time.

It would speed my edits up by quite a bit when it's finished of course, but it's not stopping me from using RT :smile:

eprovst:
Potentially useful: https://docs.blender.org/manual/en/dev/compositing/types/color/color_balance.html

eprovst:
About #4893:

@Thanatomanic As far as I can tell it's actually fairly difficult to achieve the same effect as the color wheels using this method (and as far as I can tell even impossible to achieve the exact effect of gamma).

Does this lower the necessity of color wheels? Yes it does, however I think that color wheels have a use as a primary color correction, the fine-grained control that this option provides is a bit overwhelming and in general I want to get the image roughly where I want before I get to the details :slightly_smiling_face:

agriggio:
I didn't follow the conversation fully, but did you have a look at the new colour correction module of darktable? maybe that's something we can borrow...

PR #4893, opened 2018-10-25
Added new color toning method "L*a*b* regions"

agriggio:
Allows to specify various "regions" of the image with masks, and to correct for hue, saturation and lightness.

Inspired by the existing L*a*b* grid (in turn taken from darktable)

A little demo:
https://filebin.net/plebyt2fqlrgto16/lab-regions.mp4?t=tmmbwxfr

(note: the PR is mainly for getting feedback at the moment, there might still be some corner cases)

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -398,7 +398,7 @@ HISTORY_MSG_145;Microcontrast - Uniformity
 HISTORY_MSG_146;Edge sharpening
 HISTORY_MSG_147;ES - Luminance only
 HISTORY_MSG_148;Microcontrast
-HISTORY_MSG_149;Microcontrast - 33 matrix
+HISTORY_MSG_149;Microcontrast - 3×3 matrix
 HISTORY_MSG_150;Post-demosaic artifact/noise red.
 HISTORY_MSG_151;Vibrance
 HISTORY_MSG_152;Vib - Pastel tones
@@ -728,14 +728,14 @@ HISTORY_MSG_492;RGB Curves
 HISTORY_MSG_493;L*a*b* Adjustments
 HISTORY_MSG_CLAMPOOG;Clip out-of-gamut colors
 HISTORY_MSG_COLORTONING_LABGRID_VALUE;CT - Color correction
-HISTORY_MSG_COLORTONING_LABREGION_LIST;CT - L*a*b* region list
+HISTORY_MSG_COLORTONING_LABREGION_LIST;CT - List
 HISTORY_MSG_COLORTONING_LABREGION_AB;CT - Color correction
-HISTORY_MSG_COLORTONING_LABREGION_SATURATION;CT - L*a*b* region saturation
-HISTORY_MSG_COLORTONING_LABREGION_LIGHTNESS;CT - L*a*b* region lightness
-HISTORY_MSG_COLORTONING_LABREGION_HUEMASK;CT - L*a*b* region H mask
-HISTORY_MSG_COLORTONING_LABREGION_CHROMATICITYMASK;CT - L*a*b* region C mask
-HISTORY_MSG_COLORTONING_LABREGION_LIGHTNESSMASK;CT - L*a*b* region L mask
-HISTORY_MSG_COLORTONING_LABREGION_SHOWMASK;CT - L*a*b* region show mask
+HISTORY_MSG_COLORTONING_LABREGION_SATURATION;CT - Saturation
+HISTORY_MSG_COLORTONING_LABREGION_LIGHTNESS;CT - Lightness
+HISTORY_MSG_COLORTONING_LABREGION_HUEMASK;CT - H mask
+HISTORY_MSG_COLORTONING_LABREGION_CHROMATICITYMASK;CT - C mask
+HISTORY_MSG_COLORTONING_LABREGION_LIGHTNESSMASK;CT - L mask
+HISTORY_MSG_COLORTONING_LABREGION_SHOWMASK;CT - Show mask
 HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
@@ -816,7 +816,7 @@ IPTCPANEL_CITY;City
 IPTCPANEL_CITYHINT;Enter the name of the city pictured in this image.
 IPTCPANEL_COPYHINT;Copy IPTC settings to clipboard.
 IPTCPANEL_COPYRIGHT;Copyright notice
-IPTCPANEL_COPYRIGHTHINT;Enter a Notice on the current owner of the Copyright for this image, such as 2008 Jane Doe.
+IPTCPANEL_COPYRIGHTHINT;Enter a Notice on the current owner of the Copyright for this image, such as ©2008 Jane Doe.
 IPTCPANEL_COUNTRY;Country
 IPTCPANEL_COUNTRYHINT;Enter the name of the country pictured in this image.
 IPTCPANEL_CREATOR;Creator
@@ -1375,9 +1375,9 @@ TP_COARSETRAF_TOOLTIP_ROTLEFT;Rotate left.\n\nShortcuts:\n<b>[</b> - Multiple Ed
 TP_COARSETRAF_TOOLTIP_ROTRIGHT;Rotate right.\n\nShortcuts:\n<b>]</b> - Multiple Editor Tabs Mode,\n<b>Alt-]</b> - Single Editor Tab Mode.
 TP_COARSETRAF_TOOLTIP_VFLIP;Flip vertically.
 TP_COLORAPP_ADAPTSCENE;Scene absolute luminance
-TP_COLORAPP_ADAPTSCENE_TOOLTIP;Absolute luminance of the scene environment (cd/m).\n1) Calculated from the Exif data:\nShutter speed - ISO speed - F number - camera exposure correction.\n2) Calculated from the raw white point and RT's Exposure Compensation slider.
-TP_COLORAPP_ADAPTVIEWING;Viewing absolute luminance (cd/m)
-TP_COLORAPP_ADAPTVIEWING_TOOLTIP;Absolute luminance of the viewing environment\n(usually 16cd/m).
+TP_COLORAPP_ADAPTSCENE_TOOLTIP;Absolute luminance of the scene environment (cd/m²).\n1) Calculated from the Exif data:\nShutter speed - ISO speed - F number - camera exposure correction.\n2) Calculated from the raw white point and RT's Exposure Compensation slider.
+TP_COLORAPP_ADAPTVIEWING;Viewing absolute luminance (cd/m²)
+TP_COLORAPP_ADAPTVIEWING_TOOLTIP;Absolute luminance of the viewing environment\n(usually 16cd/m²).
 TP_COLORAPP_ADAP_AUTO_TOOLTIP;If the checkbox is checked (recommended) RawTherapee calculates an optimum value from the Exif data.\nTo set the value manually, uncheck the checkbox first.
 TP_COLORAPP_ALGO;Algorithm
 TP_COLORAPP_ALGO_ALL;All
@@ -1414,7 +1414,7 @@ TP_COLORAPP_FREE;Free temp+green + CAT02 + [output]
 TP_COLORAPP_GAMUT;Gamut control (L*a*b*)
 TP_COLORAPP_GAMUT_TOOLTIP;Allow gamut control in L*a*b* mode.
 TP_COLORAPP_HUE;Hue (h)
-TP_COLORAPP_HUE_TOOLTIP;Hue (h) - angle between 0 and 360.
+TP_COLORAPP_HUE_TOOLTIP;Hue (h) - angle between 0° and 360°.
 TP_COLORAPP_LABEL;CIE Color Appearance Model 2002
 TP_COLORAPP_LABEL_CAM02;Image Adjustments
 TP_COLORAPP_LABEL_SCENE;Scene Conditions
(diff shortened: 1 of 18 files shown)


rtdata/languages/default
```diff
@@ -398,7 +398,7 @@ HISTORY_MSG_145;Microcontrast - Uniformity
 HISTORY_MSG_146;Edge sharpening
 HISTORY_MSG_147;ES - Luminance only
 HISTORY_MSG_148;Microcontrast
-HISTORY_MSG_149;Microcontrast - 3×3 matrix
+HISTORY_MSG_149;Microcontrast - 33 matrix
```
Floessie: Somehow your editor killed the characters outside the basic latin code block.

rtengine/iplabregions.cc
```diff
@@ -0,0 +1,159 @@
...
+namespace rtengine {
+
+void ImProcFunctions::labColorCorrectionRegions(LabImage *lab)
+{
+    if (!params->colorToning.enabled || params->colorToning.method != "LabRegions") {
+        return;
+    }
+
+    const float factor = ColorToningParams::LABGRID_CORR_MAX * 3.f;
+    const float scaling = 1.f;
```
Floessie: Could be both `constexpr`.

rtengine/improcfun.cc
```diff
@@ -4151,6 +4151,42 @@ void ImProcFunctions::chromiLuminanceCurve (PipetteBuffer *pipetteBuffer, int pW
         }
     }
 
+    //-------------------------------------------------------------------------
+    // support for pipettes for the new LabRegions color toning mode this is a
+    // hack to fill the pipette buffers also when
+    // !params->labCurve.enabled. It is ugly, but it's the smallest code
+    // change that I could find
+    //-------------------------------------------------------------------------
```
Floessie, line 4159: Indeed. :wink:

rtengine/iplabregions.cc
```diff
@@ -0,0 +1,159 @@
...
+
+    int n = params->colorToning.labregions.size();
```
Floessie, line 36: Wouldn't a `const std::size_t` fit better here? The `i`s below are used as indices, the `n` as vector sizes.

rtengine/iplabregions.cc
```diff
@@ -0,0 +1,159 @@
...
+        if (!r.lightnessMask.empty() && r.lightnessMask[0] != FCT_Linear) {
+            lmask[i].reset(new FlatCurve(r.lightnessMask, false));
+        }
+    }
+
+    array2D<float> guide(lab->W, lab->H, lab->L, ARRAY2D_BYREFERENCE);
+    std::vector<array2D<float>> abmask(n);
+    std::vector<array2D<float>> Lmask(n);
+    for (int i = 0; i < n; ++i) {
+        abmask[i](lab->W, lab->H);
```
Floessie, line 64: I know, you didn't think that up yourself, but this is a freaky syntax. I'd expected something like `abmask[i].resize(lab->W, lab->H)`.

rtengine/iplabregions.cc
```diff
@@ -0,0 +1,159 @@
...
+    for (int y = 0; y < lab->H; ++y) {
+        for (int x = 0; x < lab->W; ++x) {
+            float l = lab->L[y][x];
+            float a = lab->a[y][x];
+            float b = lab->b[y][x];
+            float c, h;
+            Color::Lab2Lch(a, b, c, h);
+            float c1 = lin2log(c * (327.68f / 48000.f), 10.f);
+            float h1 = Color::huelab_to_huehsv2(h);
+            h1 = h1 + 1.f/6.f;
```
Floessie: Magic constants.

rtengine/iplabregions.cc
```diff
@@ -0,0 +1,159 @@
...
+            }
+        }
+
+        return;
+    }
+
+    const auto abcoord =
+        [](float x) -> float
+        {
+            const float m = ColorToningParams::LABGRID_CORR_MAX;
```
Floessie: `constexpr`.

rtengine/procparams.cc
```diff
@@ -614,6 +614,66 @@ bool LocalContrastParams::operator!=(const LocalContrastParams &other) const
...
+            0.,
+            1.,
+            0.35,
+            0.35,
+            1.,
+            1.,
+            0.35,
+            0.35
+            }
+{
```
Floessie, line 655: Bogus indentation.

rtengine/procparams.cc
```diff
@@ -4740,6 +4818,49 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
...
+                if (assignFromKeyfile(keyFile, "ColorToning", Glib::ustring("LabRegionChromaticityMask_") + n, pedited, cur.chromaticityMask, pedited->colorToning.labregions)) {
+                    found = true;
+                    done = false;
+                }
+                if (assignFromKeyfile(keyFile, "ColorToning", Glib::ustring("LabRegionLightnessMask_") + n, pedited, cur.lightnessMask, pedited->colorToning.labregions)) {
+                    found = true;
+                    done = false;
+                }
+                if (!done) {
+                    lg.emplace_back(cur);
```
Floessie, line 4857: `emplace_back()` has no benefit here. Better use `push_back()` to make that clear.

Floessie: Maybe a bit odd: Couldn't those `if`s be compacted by

```c++
if (
    assignFromKeyfile(...)
    + assignFromKeyfile(...)
    + assignFromKeyfile(...)
    ...
) {
    found = true;
    done = false;
}
```

? That would at least make it a bit less verbose. Just an idea.

rtgui/colortoning.cc
```diff
@@ -11,6 +11,8 @@ using namespace rtengine;
 using namespace rtengine::procparams;
 
 
+static constexpr int ID_LABREGION_HUE = 5;
```
Floessie: I did some research when I first encountered a  `static constexpr` in your code, but didn't come to a final conclusion. Is the `static` really needed? If I remember correctly, Bjarne Stroustrup voted for "no" while some others said "yes". What's your source for this?

rtgui/colortoning.cc
```diff
@@ -336,6 +339,116 @@ ColorToning::ColorToning () : FoldableToolPanel(this, "colortoning", M("TP_COLOR
     pack_start(*labgridBox, Gtk::PACK_EXPAND_WIDGET, 4);
     //------------------------------------------------------------------------
 
+    //------------------------------------------------------------------------
+    // LAB regions
+
+    const auto add_button =
+        [&](Gtk::Button *btn, Gtk::Box *box) -> void
```
Floessie, line 350: Broad captures are usually discouraged.

rtgui/colortoning.cc
```diff
@@ -336,6 +339,116 @@ ColorToning::ColorToning () : FoldableToolPanel(this, "colortoning", M("TP_COLOR
...
+    labRegionBox = Gtk::manage(new Gtk::VBox());
+
+    labRegionList = Gtk::manage(new Gtk::ListViewText(3));
+    labRegionList->set_size_request(-1, 100);
+    labRegionList->set_column_title(0, "#");
+    labRegionList->set_column_title(1, M("TP_COLORTONING_LABREGION_LIST_TITLE"));
+    labRegionList->set_column_title(2, M("TP_COLORTONING_LABREGION_MASK"));
+    labRegionList->set_activate_on_single_click(true);
+    labRegionSelectionConn = labRegionList->get_selection()->signal_changed().connect(sigc::mem_fun(this, &ColorToning::onLabRegionSelectionChanged));
+    Gtk::HBox *hb = Gtk::manage(new Gtk::HBox());
```
Floessie, line 378: Please be so kind to constify those pointers, as we had [problems](https://github.com/Beep6581/RawTherapee/commit/653f252b9c0bac8366e18e7cb2ebf345d0817d2c) with large sequences of assignments in the past.

rtgui/colortoning.cc
```diff
@@ -1038,6 +1194,17 @@ void ColorToning::colorForValue (double valX, double valY, enum ColorCaller::Ele
...
+    } else if (callerId == ID_LABREGION_HUE) {
+        // TODO
+        float x = valX - 1.f/6.f;
+        if (x < 0.f) {
+            x += 1.f;
+        }
+        x = log2lin(x, 3.f);
+        // float x = valX;
+        Color::hsv2rgb01(x, 0.5f, 0.5f, R, G, B);        
+    } else if (callerId == ID_LABREGION_HUE+1) {
```
Floessie, line 1208: This calls for a new `enum` value.

rtgui/colortoning.cc
```diff
@@ -1195,3 +1372,195 @@ bool ColorToning::resetPressed(GdkEventButton* event)
...
+        if (labRegionShowMask->get_active()) {
+            labRegionShowMaskChanged();
+        }
+    }
+}
+
+
+void ColorToning::labRegionGet(int idx)
+{
+    auto &r = labRegionData[idx];
```
Floessie, line 1394: Unchecked index.

rtgui/colortoning.cc
```diff
@@ -1195,3 +1372,195 @@ bool ColorToning::resetPressed(GdkEventButton* event)
...
+
+void ColorToning::labRegionShowMaskChanged()
+{
+    if (listener) {
+        listener->panelChanged(EvLabRegionShowMask, labRegionShowMask->get_active() ? M("GENERAL_ENABLED") : M("GENERAL_DISABLED"));
+    }
+}
+
+
+namespace {
```
Floessie: Please move the anonymous namespace with helpers to the top of the file, so that there's ideally only one such namespace. This is where everyone will look for it when adding other helpers.

rtgui/colortoning.cc
```diff
@@ -1195,3 +1372,195 @@ bool ColorToning::resetPressed(GdkEventButton* event)
...
+
+bool hasMask(const std::vector<double> &dflt, const std::vector<double> &mask)
+{
+    if (mask.empty() || mask[0] == FCT_Linear || mask == dflt) {
+        return false;
+    } else {
+        return true;
+    }
+}
```
Floessie: This could be a single `return`.

rtgui/colortoning.cc
```diff
@@ -1195,3 +1372,195 @@ bool ColorToning::resetPressed(GdkEventButton* event)
...
+                hasMask(dflt.hueMask, r.hueMask) ? "H" : "",
+                hasMask(dflt.chromaticityMask, r.chromaticityMask) ? "C" : "",
+                hasMask(dflt.lightnessMask, r.lightnessMask) ? "L" : ""));
+    }
+}
+
+
+void ColorToning::labRegionShow(int idx, bool list_only)
+{
+    bool disable = listener;
```
Floessie: Could be `const`. (Same applies to many other instances I've read above, but this here is most obvious.)

rtgui/colortoning.h
```diff
@@ -117,6 +130,33 @@ class ColorToning final :
...
+    rtengine::ProcEvent EvLabRegionList;
+    rtengine::ProcEvent EvLabRegionAB;
+    rtengine::ProcEvent EvLabRegionSaturation;
+    rtengine::ProcEvent EvLabRegionLightness;
+    rtengine::ProcEvent EvLabRegionHueMask;
+    rtengine::ProcEvent EvLabRegionChromaticityMask;
+    rtengine::ProcEvent EvLabRegionLightnessMask;
+    rtengine::ProcEvent EvLabRegionShowMask;
+
+    Gtk::VBox *labRegionBox;
```
Floessie, line 138: Please be so kind to constify the members.

heckflosse:
@agriggio There are artifacts in less than 100% view:
grafik

[yellow_poppy.nef.pp3.txt](https://github.com/Beep6581/RawTherapee/files/2519500/yellow_poppy.nef.pp3.txt)

Thanatomanic:
@agriggio Wonderful addition! My only comment so far would be that it's a shame we can only apply this once (right?). Multiple masks with different modifications would be a dream to have.

agriggio:
> My only comment so far would be that it's a shame we can only apply this once (right?). Multiple masks with different modifications would be a dream to have.

Well, look closer :-)
The tool definitely does support multiple masks...

agriggio:
@heckflosse I know, that's why I wrote that this is mostly for gathering feedback at the moment. I'm investigating but didn't find the culprit yet. BTW, in some cases the artifacts are also in the output, so it's not just a matter of preview...

agriggio:
@Floessie thanks for the review. I hope you don't mind if I postpone all actions though, I'd like to have the tool working first, then I can (try to) make you happy about the code :-)

agriggio:
found the culprit for the artifacts, now they should be gone

Floessie:
@agriggio No, I don't mind. Please continue this great addition at the pace that suits you best. :+1:

heckflosse:
@agriggio Alberto, fix confirmed. Artifacts are gone. This is a really great addition :+1:

Thanatomanic:
> > My only comment so far would be that it's a shame we can only apply this once (right?). Multiple masks with different modifications would be a dream to have.
> 
> Well, look closer :-)
> The tool definitely does support multiple masks...

@agriggio Sorry, I just looked at the screenshot from Ingo 🙄 

@elecprog Do you think that if this feature is introduced, the color wheels are still necessary (see #4804)? I believe this is actually more powerful than the 'simple' color wheels?

eprovst:
It is definitely more powerful, but one could argue that the simplicity of the color wheels are their strength. I'd have to play a bit with this to give you a definite answer. :slightly_smiling_face:

Beep6581:
Easy to use, intuitive, and supports multiple adjustments \o/

Compiled and ran fine.

When "Show mask" is enabled, the PP3 contains:
```
LabRegionsShowMask=0
```
and when disabled, the value is `-1`. Are `0=true` and `-1=false` consistent with other true/false settings key values?

Currently the default mask curves are "apply to everything". Wouldn't "apply to nothing" be a better default? That way, when the mask curve is disabled (linear) it applies to everything, but when enabled then it applies to nothing, saving a few clicks when one wants to limit the mask only to a certain H/C/L value.
i.e. by default, instead of this:
screenshot_20181027_175034
it could be this:
screenshot_20181027_175037

Beep6581:
@agriggio here is a corrected `default` with the unintentionally affected values restored and the history message "L\*a\*b\* region" part removed:
[default.tar.zip](https://github.com/Beep6581/RawTherapee/files/2521965/default.tar.zip)

agriggio:
@Beep6581 'show mask' is not a Boolean; it's -1 when off, otherwise it's the index of the individual region for which you want to see the mask. example, if you have 3 regions, it can range from -1 to 2... 
I'm +0 about your proposal to start with empty masks, if you like it better I can easily change the default. what do the others think?

Beep6581:
@agriggio I'm having doubts about my proposal for default empty masks - the proposal was based on my having used the tool as a tester, but I'm not sure that's what I want during real use. Time will tell.

Beep6581 (approved):

Beep6581:
@agriggio would you like me to commit the fixed `default` file?

agriggio:
sure, thanks!

Desmis:
@agriggio 
Alberto, it is a very very good improvment to colortoning, easy to use.
Congratulations :)
jacques

agriggio:
Are we all ok for this to go in 5.5? Feature-wise it's ready, I only need to clean up the code to comply with (some of :-) @Floessie's requests.

heckflosse:
Don't know about all, but :+1: from me

Desmis:
No problem for me :)

Beep6581:
Yes please.

Thanatomanic:
Oh yes!

Issue #4882, opened 2018-10-20
Crash when stopping output queue


thirtythreeforty:
I get a crash (SIGABRT) when I have a series of images in the output queue, and flip the big switch in the top left to OFF.  It seems RT finishes its current image, hangs for a few seconds, then crashes.

Steps:
1. Put several images to the output queue.  I am using raws from a Fuji X-T10.  The profile does not seem to make much difference.
2. Go to the Queue tab and flip the switch to ON.
3. Wait for at least one image to finish.
4. Flip the switch to OFF.
5. Wait a few seconds for RT to crash.

Backtrace info is attached as [log.txt](https://github.com/Beep6581/RawTherapee/files/2498519/log.txt).

I can provide a GDB core file if needed.

Version info:
```
Version: 5.4-964-g18f812d96
Branch: dev
Commit: 18f812d96
Commit date: 2018-10-19
Compiler: cc 8.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.22.2
Lensfun: V0.3.2.0
Build type: Debug
Build flags:  -std=c++11 -march=native -Werror=unused-label -flto -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -g -ftree-vectorize
Link flags:  -march=native -flto
OpenMP support: ON
MMAP support: ON
```
...hot off my compiler on Arch.  Same behavior with 5.4 release.

thirtythreeforty:
It looks like GTK code is being called from a separate thread.  I am not all that familiar with GTK, but with most GUI toolkits I am familiar with, that is usually disallowed.

thirtythreeforty:
Simple patch which seems to fix the issue is as follows.  This is my first time touching RT code so it may not be the best place to stick the lock.  But it works for me:

```patch
diff --git a/rtgui/batchqueuepanel.cc b/rtgui/batchqueuepanel.cc
index b698ee9d6..bc03c2386 100644
--- a/rtgui/batchqueuepanel.cc
+++ b/rtgui/batchqueuepanel.cc
@@ -322,7 +322,7 @@ void BatchQueuePanel::addBatchQueueJobs(const std::vector<BatchQueueEntry*>& ent
 
 bool BatchQueuePanel::canStartNext ()
 {
-
+    GThreadLock lock;
     if (qStartStop->get_active()) {
         return true;
     } else {
```

heckflosse:
[...]

PR #4895, opened 2018-10-26
Properly handle queue start/stop switch

thirtythreeforty:
This is a small series of patches to clean up the GUI callback for the queue worker.  It no longer deadlocks or crashes when cancelling the batch.  See #4882 for more details.

### DIFF
--- a/rtgui/batchqueue.cc
+++ b/rtgui/batchqueue.cc
@@ -45,7 +45,7 @@ namespace
 struct NLParams {
     BatchQueueListener* listener;
     int qsize;
-    bool queueEmptied;
+    bool queueRunning;
     bool queueError;
     Glib::ustring queueErrorMessage;
 };
@@ -53,7 +53,7 @@ struct NLParams {
 int bqnotifylistenerUI (void* data)
 {
     NLParams* params = static_cast<NLParams*>(data);
-    params->listener->queueSizeChanged (params->qsize, params->queueEmptied, params->queueError, params->queueErrorMessage);
+    params->listener->queueSizeChanged (params->qsize, params->queueRunning, params->queueError, params->queueErrorMessage);
     delete params;
     return 0;
 }
@@ -229,7 +229,7 @@ void BatchQueue::addEntries (const std::vector<BatchQueueEntry*>& entries, bool 
         saveBatchQueue ();
 
     redraw ();
-    notifyListener (false);
+    notifyListener (true);
 }
 
 bool BatchQueue::saveBatchQueue ()
@@ -387,7 +387,7 @@ bool BatchQueue::loadBatchQueue ()
     }
 
     redraw ();
-    notifyListener (false);
+    notifyListener (true);
 
     return !fd.empty ();
 }
@@ -460,7 +460,7 @@ void BatchQueue::cancelItems (const std::vector<ThumbBrowserEntryBase*>& items)
     saveBatchQueue ();
 
     redraw ();
-    notifyListener (false);
+    notifyListener (true);
 }
 
 void BatchQueue::headItems (const std::vector<ThumbBrowserEntryBase*>& items)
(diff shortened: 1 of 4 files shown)


rtgui/batchqueue.cc
```diff
@@ -778,7 +775,8 @@ rtengine::ProcessingJob* BatchQueue::imageReady(rtengine::IImagefloat* img)
     }
 
     redraw ();
-    notifyListener (queueEmptied);
+    const bool queueRunning = (processing != nullptr);
```
Floessie: Feel free to write `const bool queueRunning = processing`.

rtgui/batchqueuepanel.cc
```diff
@@ -322,14 +326,9 @@ void BatchQueuePanel::addBatchQueueJobs(const std::vector<BatchQueueEntry*>& ent
...
-    if (qStartStop->get_active()) {
-        return true;
-    } else {
-        fdir->set_sensitive (true);
-        fformat->set_sensitive (true);
-        return false;
-    }
+    // This function is called from the background BatchQueue thread.
+    // It cannot call UI functions, so grab the stored state of qStartStop.
+    return qStartStopState;
```
Floessie: :+1:

rtgui/batchqueuepanel.h
```diff
@@ -53,29 +55,32 @@ class BatchQueuePanel : public Gtk::VBox,
 
     IdleRegister idle_register;
 
+    std::atomic<bool> qStartStopState;
```
Floessie, line 56: Excellent. Please move this before `idle_register`, which should always be the last member in case someone forgets to call `IdleRegister::destroy()` in the d'tor.

rtgui/batchqueuepanel.h
```diff
@@ -60,22 +64,23 @@ class BatchQueuePanel : public Gtk::VBox,
     void init (RTWindow* parent);
 
     void addBatchQueueJobs(const std::vector<BatchQueueEntry*>& entries , bool head = false);
+    void saveOptions ();
+
+    bool handleShortcutKey (GdkEventKey* event);
```
Floessie, line 69: There must be a reason to make those accessible, but I can't see it from the diff.

Anyway, as one last nitpicking: Could you adapt the function order in the implementation to the one given in the header?

Floessie (approved):
The patch looks sane code-wise. I haven't tested it, but I really like the solution details and quick round trip.

heckflosse:
I tested

1) no freezes :+1: 
2) when the queue is stopped while the second last image is processed, I can't start it again to process the last image.

rtgui/batchqueuepanel.h
thirtythreeforty: There's only these two usages, I think:
https://github.com/Beep6581/RawTherapee/blob/b37d88fb2245fc99c529a377c57f3475c9da3c54/rtgui/rtwindow.cc#L674-L676
https://github.com/Beep6581/RawTherapee/blob/b37d88fb2245fc99c529a377c57f3475c9da3c54/rtgui/rtwindow.cc#L614-L616

thirtythreeforty:
> when the queue is stopped while the second last image is processed, I can't start it again to process the last image.

Great catch, thank you.  The old code was (mis)using another GUI item's visibility as an indicator of idleness, which was now getting updated in the wrong order.

heckflosse:
@thirtythreeforty Works fine now. Thank you. :+1: for merge from me

heckflosse:
@thirtythreeforty Thank you very much for this fix :+1:

thirtythreeforty:
Glad to contribute!

ff2000:
I just pulled dev and rebuilt.
Unfortunately I can only start the queue if more than one image is there to being processed.
With just one image the start button is deactivated.
Auto start seems to work, it starts as soon as I add an image.

heckflosse:
@ff2000 Confirmed. I missed that case in my tests :(

thirtythreeforty:
There are a lot of edge cases with this :( I will look later!

thirtythreeforty:
@ff2000 #4902 should fix this (bug source: my own carelessness). Would you please test?

PR #4898, opened 2018-10-27
Dehaze

agriggio:
I think this is ready for consideration (and I'd propose to merge this before #4893)


ff2000:
Great addition!
Just one thing, could you extend the slider to negative values? So that I can add a tiny bit more haze to an image already full of atmosphere?

agriggio:
I think that won't work with the current implementation, sorry. it might be possible but would require more tweaking... I'd prefer to have a basic version merged first

agriggio:
can this go in for 5.5? what's the tentative release date?

heckflosse:
@agriggio no objection from me for 5.5

Beep6581:
@agriggio no objection from me either.

Beep6581:
I just noticed that the `default` language file is [also](https://github.com/Beep6581/RawTherapee/pull/4893) corrupt here.

agriggio:
sorry about the corruptions, I need to be more careful...

PR #4902, opened 2018-10-29
Fix GUI/batch state synchronization

thirtythreeforty:
Previously, this function required a `queueEmptied` bool, which I changed to `queueRunning`.  But I goofed the logic on this bool; it should always be "is `processing` nullptr."

See original report #4882, and initial fix PR #4895.

### DIFF
--- a/rtgui/batchqueue.cc
+++ b/rtgui/batchqueue.cc
@@ -597,6 +597,8 @@ void BatchQueue::startProcessing ()
             // start batch processing
             rtengine::startBatchProcessing (next->job, this);
             queue_draw ();
+
+            notifyListener();
         }
     }
 }
--- a/rtgui/batchqueue.cc
+++ b/rtgui/batchqueue.cc
@@ -229,7 +229,7 @@ void BatchQueue::addEntries (const std::vector<BatchQueueEntry*>& entries, bool 
         saveBatchQueue ();
 
     redraw ();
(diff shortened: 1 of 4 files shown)


heckflosse:
I think I found a bug:

1) disable auto-start of queue
2) put 5 images into the queue
3) enable auto-start of the queue => nothing happens, good!
4) put another image into the queue => queue starts, good!
but it stops after having processed one image => bad

thirtythreeforty:
I can confirm @heckflosse.  Thanks for testing, hang on.

thirtythreeforty:
So the problem now is that the UI gets a notification that the number of items has changed and that the queue is not running, so it stops the queue.  Separately, the autostart code is trying to start the queue, but control of `qStartStop`, which controls whether the queue should *keep* running, is split between the two pieces of UI code.

thirtythreeforty:
I am trying to work out what would be the cleanest way to reorganize this.

thirtythreeforty:
Hold off on merging this.  This is almost correct, but there is a problem where the UI is in the wrong state after auto-start but before the first image is finished processing.

This is because the batch queue notification should really be two different notifications: queue state changed, and queue size changed.  Currently the UI code must assume both changed even when only one changes.

thirtythreeforty:
OK please test.  This fixes all the behaviors described so far I think.

Issue #4894, opened 2018-10-25
Allow resizing to physical dimensions


Beep6581:
Every once in a while someone asks how to resize an image in RawTherapee to a physical dimension in centimeters or inches based on pixels per inch (PPI). I tried to solve that problem.

Beep6581:
Duplicates:
* #834
* #630
* #2426
* #2419
* https://discuss.pixls.us/t/save-jpg-file-for-printing/7675
* https://discuss.pixls.us/t/output-file-dpi-changeable/1046
etc.

Beep6581:
Branch [`hihowdoisaveaphotoat300dpi`](https://github.com/Beep6581/RawTherapee/tree/hihowdoisaveaphotoat300dpi) implements a working solution.

![screenshot_20181026_022533](https://user-images.githubusercontent.com/5844619/47537552-a3ead300-d8c6-11e8-925f-0f82b707d840.png)

The scaling is based on pixels, so dimensions in centimeters and inches are not stored, just calculated when needed. Only the PPI value is stored.

I know very little about C++ and would appreciate help with several things:
* What to do in the `if (pedited) {` sections is a mystery to me.
* The `XResolution` and `YResolution` Exif tags should be written to the saved image (all output formats) containing the PPI value:
http://www.exiv2.org/tags.html
I will create a PR once this is done.

Heads up @TooWaBoo and other theme designers:
RawTherapee still hard-codes most widget spacing values instead of leaving that up to the CSS the GTK3 way. The Resize tool now uses two grids without hard-coded spacing, and instead leaves the spacing up to CSS. All grids which require spacing to be set by CSS are assigned a class `grid-spacing`, see here:
https://github.com/Beep6581/RawTherapee/blob/hihowdoisaveaphotoat300dpi/rtdata/themes/RawTherapee-GTK3-20_.css#L1040

Beep6581:
Notes to self, review language keys when merging: `TP_RESIZE_ALLOW_UPSCALING` (capitalization), delete all `TP_RESIZE_W` and `TP_RESIZE_H`

Thanatomanic:
Just to say, love your branch name.
What I would prefer from a GUI point of view is to try and reduce the crowding. What I think I have seen elsewhere is a toggle (or tabs) that shows you either the pixel values, or the DPI + physical sizes, not both at the same time.

Hombre57:
[...]

PR #4913, opened 2018-11-01
Updated Preferences > File Browser > Cache Options

Beep6581:
- UI updated to use grid so that the options nicely align.
- "Clear only cached processing profiles" and "Clear all" buttons are
  hidden when Preferences > Image Processing > Processing profile saving
  location is set to "next to the input file", #4911

This started as PR #4911 but I changed things which complicated the merge, so I just made a new branch.


Beep6581:
Grid cell spacing will automagically happen when #4894 is merged.

PR #4919, opened 2018-11-02
Labregions speedup

heckflosse:


agriggio (approved):
looks good! I'd only remove the duplicate delcaration of `c_factor`, otherwise it's good to go for me :+1:

Beep6581 (commented):
No difference in output, all good.

PR #4920, opened 2018-11-02
Fix sundry memory problems found with Valgrind/LSan

thirtythreeforty:
This fixes several leaks and errors.  Feel free to go ahead and merge these, they're pretty tame I think.  Review welcome of course.

There are more leaks in the queue processing code that I am trying to track down.  None of the pointer ownership is clear there, unlike these "low-hanging fruit" cases.


thirtythreeforty:
Actually, the only time LSan complains about the queue code is when I quit RT with items still in the queue.  So that code looks OK upon further inspection.  This PR is unrelated and is still valid.

heckflosse:
Builds and runs fine here on Win7/64

rtengine/dcp.cc
```diff
@@ -419,7 +426,7 @@ std::map<std::string, std::string> getAliases(const Glib::ustring& profile_dir)
     buffer[read] = 0;
 
     cJSON_Minify(buffer.get());
-    const std::unique_ptr<cJSON> root(cJSON_Parse(buffer.get()));
+    const std::unique_ptr<cJSON, cJSON_deleter> root(cJSON_Parse(buffer.get()));
```
Floessie: If the `cJSON_deleter` was used in another place it had a right to exist. Here a lambda would fit better.

rtengine/dcp.cc
```diff
@@ -393,6 +393,13 @@ double xyCoordToTemperature(const std::array<double, 2>& white_xy)
     return res;
 }
 
+struct cJSON_deleter {
+    template<typename T>
+    void operator()(T *t) {
+        cJSON_Delete(const_cast<typename std::remove_const<T>::type*>(t));
```
Floessie: Sure the template is needed? `cJSON_Delete()` only takes a `cJSON*`.

See comment below.

rtengine/imagedata.cc
```diff
@@ -1118,9 +1113,7 @@ FramesData::FramesData (const Glib::ustring& fname, std::unique_ptr<RawMetaDataL
 
                 // creating FrameData
                 for (auto currFrame : exifManager.frames) {
-                    FrameData* fd = new FrameData(currFrame, currFrame->getRoot(), roots.at(0));
-
-                    frames.push_back(fd);
+                    frames.push_back(std::unique_ptr<FrameData>(new FrameData(currFrame, currFrame->getRoot(), roots.at(0))));
```
Floessie, line 1116: Should be `emplace_back(new FrameData...)`.

rtengine/imagesource.h
```diff
@@ -110,7 +110,6 @@ class ImageSource : public InitialImage
         return 0;
     }
 
-    virtual FrameData*     getImageData (unsigned int frameNum) = 0;
```
Floessie, line 113: Ha, that's surprising.

rtengine/rtlensfun.h
```diff
@@ -57,7 +57,7 @@ class LFModifier final :
     lfModifier *data_;
     bool swap_xy_;
     int flags_;
-    MyMutex *lfModifierMutex;
+    mutable MyMutex lfModifierMutex;
```
Floessie, line 60: :+1:

Floessie (changes requested):

rtengine/dcp.cc
thirtythreeforty: If I use a lambda, it will inflate the std::unique_ptr, whereas the `cJSON_deleter` class is stateless so the pointer is free.  In this particular case it won't make a performance difference, but that's why I did it this way.  I don't have a problem changing it.

rtengine/imagedata.cc
thirtythreeforty: `vector.emplace_back(new ...)` can leak the allocated memory if something throws during resizing the vector (admittedly unlikely here).  Would you still prefer the terser syntax?

As an aside, what is stopping RT from using C++14?  That would let me use `make_unique`.

rtengine/dcp.cc
thirtythreeforty: True, this version lets you pass a const pointer (unless I goofed it up).  Future proofing this if it gets reused.  Although, if you want the lambda, this is unneeded anyway.

thirtythreeforty: True, this lets you pass a `const` pointer I think.  Tried to future-proof this a little.  If you want the lambda, it'll all go away.

rtengine/dcp.cc
Floessie: > If I use a lambda, it will inflate the std::unique_ptr, whereas the cJSON_deleter class is stateless so the pointer is free.

Are you sure?

Quoting from Scott Meyers' "Effective Modern C++", item 18:
> I remarked earlier that, when using the default deleter (i.e., delete), you can reasonably assume that std::unique_ptr objects are the same size as raw pointers. When custom deleters enter the picture, this may no longer be the case. Deleters that are function pointers generally cause the size of a std::unique_ptr to grow from one word to two. For deleters that are function objects, the change in size depends on how much state is stored in the function object. Stateless function objects (e.g., from lambda expressions with no captures) incur no size penalty, and this means that when a custom deleter can be implemented as either a function or a captureless lambda expression, the lambda is preferable.

So let's try:

```c++
#include <iostream>
#include <memory>

namespace
{

    struct cJSON_deleter {
        template<typename T>
        void operator()(T *t) {
            delete const_cast<typename std::remove_const<T>::type*>(t);
        }
    };

    void deleteFunc(int* p)
    {
        delete p;
    }

}

int main()
{
    std::cout << sizeof(void*) << std::endl;

    std::unique_ptr<int> default_deleter(new int(5));
    std::cout << sizeof(default_deleter) << std::endl;

    std::unique_ptr<int, cJSON_deleter> cjson_deleter(new int(5));
    std::cout << sizeof(cjson_deleter) << std::endl;

    const auto lambda =
        [](int* p)
        {
            delete p;
        };

    std::unique_ptr<int, decltype(lambda)> lambda_deleter(new int(5), lambda);
    std::cout << sizeof(lambda_deleter) << std::endl;

    std::unique_ptr<int, decltype(&deleteFunc)> function_deleter(new int(5), deleteFunc);
    std::cout << sizeof(function_deleter) << std::endl;

    return 0;
}
```

Here's the output from my 64b system:

```
8
8
8
8
16
```

> In this particular case it won't make a performance difference, but that's why I did it this way. I don't have a problem changing it.

Okay, I see your motivation. Performance (which doesn't matter here) aside, I honestly prefer the most concise solution which would read something like:

```c++
const std::unique_ptr<cJSON, decltype(&cJSON_Delete)> root(cJSON_Parse(buffer.get()), cJSON_Delete);
```

IMHO everything else is nice but needless.

rtengine/imagedata.cc
Floessie: Minimum GCC version is 4.9. Many of the [C++14 features](https://gcc.gnu.org/projects/cxx-status.html#cxx14) were implemented in 5.

On the Mac side I don't really know the minimum Clang version. [Clang 3.4](https://clang.llvm.org/cxx_status.html) fully supports C++14. I guess, @Benitoite can shed some light here.

Floessie: `emplace` vs. `push`: Good point about exception safety. :+1: But if we run out of memory there is nothing that can safe us today (and most likely in future), so that leak won't matter much.

How about reserving `size() + 1` and then emplacing? Best of both worlds. :wink:

thirtythreeforty: I'll just push.  I'll make a note near one of the calls.  It'll only be slightly wrong :) and since unique_ptrs are pretty much noexcept, about the only thing that could happen is OOM.

Makes sense about C++14.

rtengine/dcp.cc
thirtythreeforty: Yep, the stateful deleter is small in this case.  I've changed it!

rtengine/imagedata.cc
```diff
@@ -1118,9 +1113,8 @@ FramesData::FramesData (const Glib::ustring& fname, std::unique_ptr<RawMetaDataL
 
                 // creating FrameData
                 for (auto currFrame : exifManager.frames) {
-                    FrameData* fd = new FrameData(currFrame, currFrame->getRoot(), roots.at(0));
-
-                    frames.push_back(fd);
+                    // Note: could leak if emplace_back throws (below as well).  Unlikely in practice.
```
Floessie: Very nice! :+1:

Floessie (approved):
Thanks for taking the time to trace those leaks! I'm fine with the result.

@heckflosse Ingo, could you test this again?

thirtythreeforty:
Glad to help!  I had noticed 5.4 gradually eating a large amount of memory at a couple points; I hope I fixed the cause (probably ImageData if I had to guess).

Floessie:
@thirtythreeforty I've discussed the "push vs. emplace" topic with a colleague at work and we've come to the conclusion that a `std::unique_ptr<>` demands for the scheme you first implemented (the "push"). While emplacing is optimal for most temporaries, `std::unique_ptr<>` has a duty beyond that and you were right to reflect that in your code. Feel free to revert the `emplace_back()` to `push_back()`. Thanks for bringing this (accidentally) up.

thirtythreeforty:
I'm glad we agree. IMO almost all code should treat `new` pointers as hot potatoes that should be wrapped in RAII immediately, unless you're implementing said wrapper.

Thanks for being approachable for contributions!

heckflosse:
@Floessie Please push the merge button if you think it's good to go. I'm not an expert in this kind of stuff...

thirtythreeforty:
@heckflosse I think Floessie just wanted you to build it on Windows to see if it was still good there

heckflosse:
@thirtythreeforty Ok, building from your branch now.

heckflosse:
@thirtythreeforty Can you merge dev into your branch? That will make it easier to catch the reason for this segfault I get using your branch:

```
Thread 76 "pool" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 7344.0x12b4]
0x00000000007f1e2b in rtengine::(anonymous namespace)::copyAndClamp(rtengine::LabImage const*, unsigned char*, double const (*) [3], bool) [clone ._omp_fn.1]
    ()
(gdb) bt full
#0  0x00000000007f1e2b in rtengine::(anonymous namespace)::copyAndClamp(rtengine::LabImage const*, unsigned char*, double const (*) [3], bool) [clone ._omp_fn.1] ()
#1  0x00000000636054cf in ?? () from C:\msys64\mingw64\bin\libgomp-1.dll
#2  0x00000000007f3633 in rtengine::ImProcFunctions::lab2monitorRgb(rtengine::LabImage*, rtengine::Image8*) ()
#3  0x00000000008bcc50 in rtengine::Thumbnail::processImage(rtengine::procparams::ProcParams const&, rtengine::SensorType, int, rtengine::TypeInterpolation, rtengine::FramesMetaData const*, double&, bool, bool) ()
#4  0x00000000006cbc02 in Thumbnail::processThumbImage(rtengine::procparams::ProcParams const&, int, double&) ()
#5  0x0000000000a65590 in ThumbImageUpdater::Impl::processNextJob() ()
#6  0x00000000664db564 in ?? () from C:\msys64\mingw64\bin\libglibmm-2.4-1.dll
#7  0x0000000064a1d54b in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#8  0x0000000064a1cbb9 in ?? () from C:\msys64\mingw64\bin\libglib-2.0-0.dll
#9  0x0000000064944b90 in ?? () from C:\msys64\mingw64\bin\libwinpthread-1.dll
#10 0x000007fefeea415f in srand () from C:\Windows\system32\msvcrt.dll
#11 0x000007fefeea6ebd in msvcrt!_ftime64_s ()
   from C:\Windows\system32\msvcrt.dll
#12 0x0000000076ae59cd in KERNEL32!BaseThreadInitThunk ()
   from C:\Windows\system32\kernel32.dll
#13 0x0000000076d4383d in ntdll!RtlUserThreadStart ()
   from C:\Windows\SYSTEM32\ntdll.dll
#14 0x0000000000000000 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb) q
A debugging session is active.

        Inferior 1 [process 7344] will be killed.

Quit anyway? (y or n) [answered Y; input not from terminal]

```

thirtythreeforty:
Done

heckflosse:
@thirtythreeforty building again...

heckflosse:
@thirtythreeforty Works fine now :+1:

PR #4926, opened 2018-11-02
Fix startup ding by initializing variable

thirtythreeforty:
Again, apologies for all the edge cases.  At least the bugs are getting smaller!


rtgui/batchqueuepanel.cc
```diff
@@ -57,6 +57,8 @@ BatchQueuePanel::BatchQueuePanel (FileCatalog* aFileCatalog) : parent(nullptr)
     qAutoStart->set_tooltip_text (M("BATCHQUEUE_AUTOSTARTHINT"));
     qAutoStart->set_active (options.procQueueEnabled);
 
+    queueShouldRun = false;
```
Floessie, line 60: Hm, this should go to the member initializers. Next time...

thirtythreeforty:
I agree. In this case I was attempting to follow the existing code "style"
which was not using member initializers.

PR #4940, opened 2018-11-05
Save space in Preferences -> File Browser Tab

TooWaBoo:
Changed `Folder-open.png` to `Folder-open-small.png`. The TWB-theme is using the same height for buttons and entry fields. The big icon is increasing the size. I hope this is OK

Changed "Context Menu Options" to two columns

Changed the `cache button separator` to a real separater with ID `#PrefCacheSeparator` for styling in css.
```
separator#PrefCacheSeparator.horizontal {
    <any style>
}
```
Adapted the TWB

ping @Beep6581 

old
oldrt
new
newrt
old
oldtwb
new
newtwb

### DIFF
--- a/rtdata/themes/TooWaBlue-GTK3-20_.css
+++ b/rtdata/themes/TooWaBlue-GTK3-20_.css
@@ -473,7 +473,7 @@ separator,
     background-color: transparent;
 }
 grid separator.horizontal, box separator.horizontal {
-    margin: 0.166666666666666666em;
+    margin: 0.333333333333333333em 0.166666666666666666em;
     padding: 0;
 }
 grid separator.vertical, box separator.vertical {
@@ -550,7 +550,7 @@ menu separator {
     margin: 0.166666666666666666em;
 }
 
-#MyExpander separator {
+#MyExpander separator.horizontal {
     background-color: @view-grid-border;
     margin: 0.333333333333333333em 0.166666666666666666em;
 }
--- a/rtdata/themes/TooWaBlue-GTK3-20_.css
+++ b/rtdata/themes/TooWaBlue-GTK3-20_.css
@@ -473,7 +473,7 @@ separator,
     background-color: transparent;
 }
 grid separator.horizontal, box separator.horizontal {
-    margin: 0.166666666666666666em 0;
+    margin: 0.166666666666666666em;
     padding: 0;
 }
 grid separator.vertical, box separator.vertical {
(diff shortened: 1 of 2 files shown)


Beep6581:
@TooWaBoo looks good, but I'd like to request one change before we merge.

As it's likely we'll need to use grid row/column separators in other places, they should be assigned a class and not an ID:
```
cacheSeparator->get_style_context()->add_class("grid-row-separator");
```
Paving way for `grid-column-separator` if needed.

RawTherapee currently uses IDs incorrectly in other places, but I'll leave those for 5.6.

TooWaBoo:
In general your are right.
eg. in html names are used as an unique ID of an object.
Classes are used:
- to group objects for styling
- if you you need more than on class so the same object can be in more than one group
- if a class changes during runtime.

In the case of a `separator` it's not necessary to have a name ( ID ) or a class. You can simply identify the separator by `grid > separator.horizontal` instead of the class `grid-row-separator` or `grid > separator.vertical`  instead of `grid-column-separator`.

I've created the `#PrefCacheSeparator` ID to make this separater unique, as you wrote: "...so that a button is not accidentally clicked...". This ID allows you to change e.g. the height of this separator without changing the default height.

In the TWB theme I've set this as default style:
```
grid separator.horizontal {
    margin: 0.166666666666666666em 0;
    padding: 0;
}
```

... but for the cache button separator and only for this I want this style:
```
separator#PrefCacheSeparator.horizontal {
    margin: 0.25em 0.166666666666666666em;
    padding: 0;
}
```
This is the reason why I've created an unique ID and not a class.

TooWaBoo:
@Beep6581 
... but if you want I can add a `grid-row-separator` class too.

Thanatomanic:
@TooWaBoo Can you use [`calc`](https://developer.mozilla.org/en-US/docs/Web/CSS/calc) here to make things a bit cleaner? So, `calc(1/6)` instead of the repeating digits?

Beep6581:
I just want this to be simple and maintainable, and don't currently understand why there is a need to have a unique ID for this separator. Why is it special? It seems that all horizontal separators used in a grid should be identical.

TooWaBoo:
@Beep6581 
Because you wrote this comment in the code: "Separation is needed so that a button is not accidentally clicked when one wanted to click a spinbox", I thought you want this huge separater between the `spinbutton` and `'clear'-button.` To give you the possibility to define any hight in this special case I added the ID.  I'm OK with the a default height of the separator.

Removed the `#PrefCacheSeparator` name in the TWB theme.
Replaced the `#PrefCacheSeparator` name with class `grid-row-separator` in `preferences.cc`

@Thanatomanic 
I'm using so many decimals to avoid rounding issues between GTK 3.22 and 3.24. If I use calc GTK is rounding different in combination of padding, margin and min-height.

PR #4963, opened 2018-11-09
Support for Blackmagic and Canon MagicLantern lj92 encoded files

heckflosse:
I would like to merge that asap to get more tests before we releast 5.5


heckflosse:
Merged to get more tests before releasing 5.5

PR #4967, opened 2018-11-10
Make WB-Picker button look like other toolbuttons

TooWaBoo:
Replaced `hbox` with `grid` and added the classes `grid-spacing` and `grid-row-separator`.
The PR contains light and dark `color-picker-small.png` images which have to be replaced with generated ones.

unbenannt

PR #4986, opened 2018-11-14
Add classes to curves for css styling

TooWaBoo:
Added classes for css styling and removed all hardcoded sizes.
@Beep6581  I didn't add the class `grid-spacing` 'cause these grids need to be handeld individual. If you want to add this class, tell me which grid.

**List of classes:**
```
grid-curve
grid-curve-flowboxes
grid-curve-sliders
grid-curve-buttons
Top
Bottom
Left
Right
```
all


Beep6581:
Hi @TooWaBoo 

> I didn't add the class grid-spacing

But you did.

> 'cause these grids need to be handeld individual

Why?

> If you want to add this class, tell me which grid.

That depends on what you want to do, which I don't know.

```text
grid-curve-buttons-top
grid-curve-buttons-bottom
grid-curve-buttons-left
grid-curve-buttons-right
```

The intention behind _cascading_ style sheets is to rely on hierarchy and inheritance, to have the styles apply top-down, to not repeat the same thing over and over. Working against that makes for unmaintainable CSS.

Looking at the changes I don't see why those four classes are needed.
The buttons grid when `options.curvebboxpos == 0` is identical to `options.curvebboxpos == 2`,
and `options.curvebboxpos == 1` is identical to `options.curvebboxpos == 3`. Instead of creating 4 classes, one `grid-curve-buttons` class will do. The buttons grid already gets a `horizontal` or `vertical` class assigned to it based on its orientation, so in the CSS file instead of `.grid-curve-buttons-top {` and `.grid-curve-buttons-bottom {` you could use `.grid-curve-buttons.horizontal {`.

It makes for non-maintainable CSS to create hundreds of non-reusable classes.

It's likely we could simplify further. In the above example I assumed there was a need for `grid-curve-buttons` and for `grid-curve`, but maybe there isn't, or if there is then maybe they could be broken up.

TooWaBoo:
Class `grid-spacing` is a copy/paste mistake. I will remove ist.

For me it's not enough to know if the buttons grid is `horizontal` or `vertical`, I want to know if it's on the `top`, `left`, `right` or at the `bottom`.
I will change `grid-curve-buttons-top` to `top` ...

Existing RT-classes which uses top, left ... are written Top, Left... I will use the same syntax.

I don't need the class `grid-spacing` at all.

TooWaBoo:
Changes done.

TooWaBoo:
Ahhhhh, now I got you. You want the classes `top`,` left`, `right`, `bottom` at the top level of the curves.
I'll also change the name e.g. `grid-curve-buttons` to `grid-buttons`
 so you can point to the objects like this:
```
Specific position of the button grid
.grid-curve.Top .grid-buttons
.grid-curve.Top .grid-flowboxes
.grid-curve.Top .grid-sliders

Position of the button grid doesn't matter
.grid-curve .grid-buttons
.grid-curve .grid-flowboxes
.grid-curve .grid-sliders
```
This makes sense, I'll change it.

Beep6581:
@TooWaBoo I won't have time to look into this in detail, I just wanted to point out that our use of CSS should be in line with how CSS ought to be used and keep things inheritable and reusable if possible. I trust you'll do the right thing.

Regarding CSS class names, things are currently a mess, but in my opinion we should use lowercase semantically-hyphenated names, not capitalized and definitely not camel-case, i.e. `grid-curve-top` and not `gridCurveTop`. This will require a cleanup for 5.6.

Let me know when you're ready to merge, and whether you'd like to squash the commits into one or not.

TooWaBoo:
@Beep6581 
Finished
List of classes
```
grid-curve
     top
     bottom
     left
     right
---------------------------
grid-curve-flowboxes
grid-curve-sliders
grid-curve-buttons
```
I didn't change the name of `grid-curve-flowboxes, ...` I have that feeling that this could break somthing in the future.

The updated TWB theme ist here: "Make WB-Picker button look like other toolbuttons" https://github.com/Beep6581/RawTherapee/pull/4967
**Before you merge this you must generate the icon `color-picker-small.png`. If it doesn't exist RT will not start.**
This is how the curves looks like in the TWB theme now.
unbenannt

Beep6581:
Great, I'll add the icon and merge tomorrow as I'm one foot out the door now.

TooWaBoo:
@Beep6581 
Please wait, I want to make some improvments. You can not use margin or padding on the curve itself. I will pack the curve inside the curve box in its own container. This makes styling much easier and gives you you more options.

TooWaBoo:
@Beep6581 
I changed the names to something more useful. 🙂
Shall I add the class `.diagonal` and `.flat` to the .curve-mainbox too?
Maybe it's useful 'cause the flat curve doesn't have a vertical colorbar and maybe you want to have different padding for that. With a colorbar it looks like the curve has an offset 'cause of the dot. But I don't know now. 🤔

I'll tell you when I'm finished

**List of classes**
```
.curve-mainbox ( .top .left .bottom .right .diagonal? .flat?)
.curve-curvebox
.curve-sliderbox
.curve-buttonbox
.curve-spinbuttonbox
```
grafik

TooWaBoo:
@Beep6581 
Finished and tested. Everything works fine.

**List of classes**
```
.curve-mainbox ( .top .left .bottom .right )
.curve-curvebox
.curve-sliderbox
.curve-buttonbox
.curve-spinbuttonbox
```
grafik

Beep6581:
Let's merge tomorrow if no objections.

Beep6581:
@TooWaBoo would you like to squash these commits before merging?

TooWaBoo:
Yes, you can do that.

PR #4992, opened 2018-11-18
Review of image classes interfaces

heckflosse:


Hombre57 (approved):
Thanks for the `float` to `bool` fix ! Everything else is beyond my understanding, but I guess welcomed.

rtengine/image16.h
```diff
@@ -40,62 +40,70 @@ class Image16 : public IImage16, public ImageIO
...
-    virtual void         getScanline(int row, unsigned char* buffer, int bps, bool isFloat = false);
-    virtual void         setScanline(int row, unsigned char* buffer, int bps, unsigned int numSamples);
+    virtual void getScanline(int row, unsigned char* buffer, int bps, bool isFloat = false) const;
+    virtual void setScanline(int row, unsigned char* buffer, int bps, unsigned int numSamples);
 
     // functions inherited from IImage16:
-    virtual MyMutex&     getMutex()
+    virtual MyMutex& getMutex()
     {
         return mutex();
```
Floessie, line 65: Hmm, I don't get this one (at least from the diff): Are we returning a temporary here, or is that some non-intuitive `operator ()`? And why non-const?

rtengine/image16.h
```diff
@@ -40,62 +40,70 @@ class Image16 : public IImage16, public ImageIO
...
+    virtual int getBPS() const
     {
         return 8 * sizeof(unsigned short);
     }
-    virtual void         getScanline(int row, unsigned char* buffer, int bps, bool isFloat = false);
-    virtual void         setScanline(int row, unsigned char* buffer, int bps, unsigned int numSamples);
+    virtual void getScanline(int row, unsigned char* buffer, int bps, bool isFloat = false) const;
+    virtual void setScanline(int row, unsigned char* buffer, int bps, unsigned int numSamples);
 
     // functions inherited from IImage16:
```
Floessie, line 62: If they are inherited, better drop `virtual` and add `override`.

rtengine/image8.cc
```diff
@@ -89,15 +89,15 @@ void Image8::setScanline (int row, unsigned char* buffer, int bps, unsigned int
     }
 }
 
-Image8* Image8::copy ()
+Image8* Image8::copy () const
```
Floessie, line 92: Sometime in the future this should become a `std::unique_ptr<>`.

rtengine/imageio.cc
```diff
@@ -1621,3 +1621,57 @@ int ImageIO::save (Glib::ustring fname)
...
+
+void ImageIO::getEmbeddedProfileData (int& length, unsigned char*& pdata) const
+{
+    length = loadedProfileLength;
+    pdata = (unsigned char*)loadedProfileData;
+}
+
+MyMutex& ImageIO::mutex ()
+{
+    return imutex;
```
Floessie, line 1663: Aha, now I get it. Crappy name. How about `MyMutex& ImageIO::getMutex() const`?

Floessie (approved):
:+1:, Ingo. Constification has many benefits in the long run. Could be committed as is, some optional fixes:

- De-inlining for decoupling header and implementation
- Whitespace unification (`void func()` vs. `void      func ( )`)
- `override`

Thanks for the fixes!

Best,
Flössie

rtengine/image16.h
heckflosse: Done

rtengine/imageio.cc
Floessie: Sorry, can't be `const` when returning a non-const reference. Missed that...

heckflosse: Yes, nevertheless I changed the code for part three. Will be pushed in a few minutes.

heckflosse: Scratch my last comment

PR #4999, opened 2018-11-19
Improvements to Profiled Lens Correction

### LINKED ISSUE #4995: Profiled Lens Correction > Camera and Lens combos should be grayed-out when not using manual correction parameters
imgur-2018_11_19-13 39 45

In the Profiled Lens Correction tool, the camera and lens comboboxes only apply to the "manual correction parameters" option, so they should be grayed-out when not using that option so as to not confuse the user.

Reference:
https://discuss.pixls.us/t/hasselblad-16-bits-files/9905/7?u=morgan_hardwood

### LINKED ISSUE #4997: Rename Profiled Lens Correction options
Since the Profiled Lens Correction panel does not currently allow us to manually specify the actual correction parameters (lens distortion parameters a, b and c, image shift parameters d and e, image shear parameters g and t, etc.), the current naming is misleading. I want to rename:
* "Auto-matched correction parameters" > "Automatically select lens profile"
* "Manual correction parameters" > "Manually select a lens profile"
* "LCP File" > either "LCP file" or "Manually select LCP file".

Any objections or suggestions?

Ping @agriggio

Beep6581:
Closes #4995
Closes #4997


agriggio:
I haven't tested nor looked at the code, but assuming the log message is accurate, I'd be -1 on hiding the widgets in automatic mode; seeing what has been auto selected is quite useful IMHO

heckflosse:
:+1: for hiding the widgets only if `None` or `Lcp` is selected.

Beep6581:
Done.

TooWaBoo:
@agriggio 

> seeing what has been auto selected is quite useful IMHO

... but when you select a profile with auto mode, the widget shows the data which are stored in the profile and not what is auto selected.

agriggio:
> ... but when you select a profile with auto mode, the widget shows the data which are stored in the profile and not what is auto selected.

and I would call that a bug :-)

Floessie:
@Beep6581 May I constify and shuffle `LensProfilePanel` a bit?

Beep6581:
@Floessie go right ahead!

Floessie:
@Beep6581 [Here](https://filebin.net/spvcfxb9actpkmbt) it is.

- Replaced unchecked `dynamic_cast<>`s with `static_cast<>`
- Constified members
- Shuffled header
- Aligned implementation with header
- Compile-tested :wink: 

HTH,
Flössie

Beep6581:
@Floessie thank you, compiled and tested in batch and single modes. I found two bugs related to batch mode #5001 #5002 but they are not related to my or your changes so they are handled separately.

I found one issue which was caused by my changes - the "(Unchanged)" option is missing in Batch mode. The fix is simple - just needs to be added to `modesGrid` (though even then it still does nothing due to pre-existing bug #5002).

Beep6581:
Your patch + missing Unchanged fix committed. In `dev` the "(Unchanged)" option is never selected, here it is now always selected.

Floessie:
@Beep6581 There was an accidental change to `rtengine/rt_math.h` in my patch. Could you revert that part? Sorry and thanks.

Beep6581:
@Floessie `rt_math.h` changes reverted. 

What do we do with (Unchanged) being always on in Batch mode?

PR #5006, opened 2018-11-21
Fix some cppcheck warnings

heckflosse:
More fixes will follow...

### DIFF
--- a/rtengine/coord.h
+++ b/rtengine/coord.h
@@ -114,22 +114,22 @@ inline void Coord::set (const int x, const int y)
 
 inline Coord& Coord::operator+= (const Coord& other)
 {
-    x += other.x;
-    y += other.y;
+    this->x += other.x;
+    this->y += other.y;
     return *this;
 }
 
 inline Coord& Coord::operator-= (const Coord& other)
 {
-    x -= other.x;
-    y -= other.y;
+    this->x -= other.x;
+    this->y -= other.y;
     return *this;
 }
 
 inline Coord& Coord::operator*= (const double scale)
 {
-    x *= scale;
-    y *= scale;
+    this->x *= scale;
+    this->y *= scale;
     return *this;
 }
 
@@ -218,7 +218,7 @@ inline PolarCoord &PolarCoord::operator-= (const PolarCoord &other)
 
 inline PolarCoord &PolarCoord::operator*= (const double scale)
 {
-    radius *= scale;
+    this->radius *= scale;
     return *this;
 }
 
--- a/rtengine/imagefloat.h
+++ b/rtengine/imagefloat.h
@@ -218,14 +218,14 @@ public:
         return (uint32_t) ((sign << 31) | (exponent << 23) | mantissa);
     }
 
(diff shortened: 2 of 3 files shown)


rtengine/coord.h
```diff
@@ -114,22 +114,22 @@ inline void Coord::set (const int x, const int y)
 
 inline Coord& Coord::operator+= (const Coord& other)
 {
-    x += other.x;
-    y += other.y;
+    this->x += other.x;
+    this->y += other.y;
```
Floessie, line 118: @heckflosse What was the warning? I see no obvious reason for this change...

rtengine/imagefloat.h
```diff
@@ -218,14 +218,14 @@ class Imagefloat : public IImagefloat, public ImageIO
...
-    void                 ExecCMSTransform(cmsHTRANSFORM hTransform);
-    void                 ExecCMSTransform(cmsHTRANSFORM hTransform, const LabImage &labImage, int cx, int cy);
+    void normalizeFloat(float srcMinVal, float srcMaxVal) override;
+    void normalizeFloatTo1();
+    void normalizeFloatTo65535();
+    void calcCroppedHistogram(const ProcParams &params, float scale, LUTu & hist);
+    void ExecCMSTransform2(cmsHTRANSFORM hTransform);
+
+    void ExecCMSTransform(cmsHTRANSFORM hTransform);
+    void ExecCMSTransform(cmsHTRANSFORM hTransform, const LabImage &labImage, int cx, int cy);
```
Floessie, line 228: :+1: If you like, I can fire up clang-tidy to add `override`s everywhere.

heckflosse: Please do. :+1:

Floessie: Processing...

rtengine/coord.h
heckflosse: It was a false positive
`Assignment of function parameter has no effect outside the function.`
But using `this->` silenced it.

Floessie: We shouldn't code against bugs in cppcheck...

rtengine/imagefloat.h
Floessie: [Here](https://filebin.net/j87h1dkc7lb65ej2) it is. Take it with a grain of salt, though: `clang-tidy` has some built-in default checks you can't disable. So there's always a bit more done than expected. See the grouping of `NAN`s in `dcraw.cc` in the patch, where one grouping was missed.

HTH,
Flössie

heckflosse: @Floessie Great. Shall I push?

rtengine/coord.h
heckflosse: agreed

rtengine/imagefloat.h
Floessie: @heckflosse Ingo, sure. Maybe also add the braces around the first `NAN` block in `dcraw.cc` to make it symmetric.

heckflosse: Ok, I'm just compiling it before I will push it...

PR #4877, opened 2018-10-18
Add reset button for Crop

Thanatomanic:
This resolves issue #4668

I've introduced this as a simple textual button below the 'Select crop' button, because I didn't see any immediate position where I could put a reset 'arrow' like for other tools.

PR #5015, opened 2018-11-22
Add Crop reset button - 2nd attempt

Thanatomanic:
This adds a 'Reset crop' button. Original discussion can be seen in #4877 where something broke the PR history.


Thanatomanic:
Note that there are duplicate commits for some odd reason, but the actual changes seem to be fine. A squash commit should do the trick here.

Beep6581:
If you would like this in 5.5 then please commit by tomorrow, else we will have to bump it to 5.6.

PR #5031, opened 2018-11-25
custom trc, speedup and reduced memory usage

heckflosse:


Desmis:
@heckflosse 

Ingo
I just test this branch
All works fine
Very goog job, as usual

jacques

Desmis:
After compilation, all works fine 

jacques

Desmis:
@heckflosse 
Ingo, do you know "Lucky Luke" ? he is shooting faster than his shadow...
you are the Lucky Luke of RAwtherapee 

Works fine
jacques

heckflosse:
@Desmis :rofl: Jacques, would it be possible to make this in one cmstransform instead of two? That would give ~ double speed.

Desmis:
@heckflosse 

it seems impossible, because you cannot calculated (or with big difficulties ???) the double transformation
1) trc sRGB ==> 1.
2) 1 => user
jacques

heckflosse:
ok

heckflosse:
@Desmis Jacques, I found another possibility to speedup custom trc, but only in gui this time.
The first `cmsCreateTransform()` is really expensive (~50 ms). Means it takes ~50 ms for the main preview + ~50 ms for each detail window. Fortunately it's possible the save the transform in improccoordinator and use it in dcrop. I will post a patch soon.

PR #5034, opened 2018-11-26
custom trc: speedup for preview and detail windows

heckflosse:


Desmis:
@heckflosse 

I just test your new (big) modification !

Works fine, no problem
Thank you for this good job

Since tomorrow, I will be away for about 14 days

Jacques

heckflosse:
@Desmis Jacques, thanks for testing so quickly :+1:

Desmis:
@heckflosse 
De rien, merci

PR #5035, opened 2018-11-26
Fixed monitor gamut check

### LINKED ISSUE #5019: Gamutcheck for monitor
Actually "Gamut check" to see Out of gamut colors only works well with RTv2 and gamma simple.

If you use ICC v2  and TRC as sRGB or BT709, the image is full of  green dots
if you use ICCv4 nothing works

At first I think a problem with LUT, I tried many solutions and always same problem

I tried RTv2 and Rtv4 with TRC sRGB or BT709 with Photoshop and all works well==>it is not a LUT problem

I think it is a LCMS bug, the "same" as I implemented in "iplab2rgb.cc" to sove bad result in histogram for low light
I use the same solution in "improcfun.cc", but always dots
I think another bug, I solved by a strange action, only for blue tag, I replace TRC sRGB or same by gamma simple, and result seems good

For ICCv4, I used the same strange action, and now RTv4 seems to work well !!!

I don't understand why this very strange change works,...but it seems to work in  all cases with files RTv2 or RTv4

I create a branch "gamutcheck_monitor"

jacques

Beep6581:
Checking out-of-gamut colors worked well before only with ICC v2 profiles using a simple gamma. Now it should work well for both v2 and v4 profiles with any TRC.
Closes #5019


Beep6581:
As @Desmis is away and @agriggio and @heckflosse were somehow involved, @agriggio @heckflosse  could you please decide whether this should go into 5.5 and if so then merge?

We do have 2-3 weeks for testing and bug fixes so I don't see why not.

agriggio:
@Beep6581 I'll test the branch asap.

rtengine/improcfun.cc
```diff
@@ -298,6 +298,18 @@ void ImProcFunctions::updateColorProfiles (const Glib::ustring& monitorProfile,
...
         monitor = ICCStore::getInstance()->getProfile (options.rtSettings.srgb);
 #endif
+//    }
+    Glib::ustring outtest = monitorProfile ;
+    
+    std::string fileis_RTv2 = outtest.substr(0, 4);
+
+    if((fileis_RTv2 == "RTv4" || fileis_RTv2 == "RTv2")  && gamutCheck) {
+            cmsToneCurve* GammaTRC[3];
+            GammaTRC[2] = cmsBuildGamma(NULL, 2.4);//if GammaTRC[2] used by v4 or v2 gamut check does not work...why ??
```
agriggio, line 308: I think we need to dig deeper into this, and understand exactly why it works... this is so odd that I suspect we are doing something not quite right here.

agriggio:
I think this branch doesn't really solve the issue. The reason why it works is that it somehow forces LCMS to not use the "unbounded" mode, *which is the right thing to do* in this case, because we do want it to clamp in the intermediate transform, since we are actually checking for out of gamut. If you don't clamp, going through the "soft proofing" transform doesn't lose any data, and that's why you don't see any out-of-gamut indicator. The code was buggy from the beginning, it was just a coincidence that it was working with the output profiles shipped by previous versions of RT, because those were using a tabulated TRC, which made LCMS clamp the output of the transforms.

Here's a patch that should fix this properly (at least, it seems to work fine here). Apply directly on `dev`, not on this branch. @Beep6581, can you please test and push to `dev` if you think it's ok? (I'm on a different branch at the moment)

```diff
diff --git a/rtengine/gamutwarning.cc b/rtengine/gamutwarning.cc
--- a/rtengine/gamutwarning.cc
+++ b/rtengine/gamutwarning.cc
@@ -34,7 +34,7 @@
     softproof2ref(nullptr)
 {
     if (cmsIsMatrixShaper(gamutprof) && !cmsIsCLUT(gamutprof, intent, LCMS_USED_AS_OUTPUT)) {
-        cmsHPROFILE aces = ICCStore::getInstance()->getProfile("RTv4_ACES-AP0");
+        cmsHPROFILE aces = ICCStore::getInstance()->workingSpace("ACESp0");
         if (aces) {
             lab2ref = cmsCreateTransform(iprof, TYPE_Lab_FLT, aces, TYPE_RGB_FLT, INTENT_ABSOLUTE_COLORIMETRIC, cmsFLAGS_NOOPTIMIZE | cmsFLAGS_NOCACHE);
             lab2softproof = cmsCreateTransform(iprof, TYPE_Lab_FLT, gamutprof, TYPE_RGB_FLT, INTENT_ABSOLUTE_COLORIMETRIC, cmsFLAGS_NOOPTIMIZE | cmsFLAGS_NOCACHE);
@@ -81,6 +81,12 @@
         
         float delta_max = lab2ref ? 0.0001f : 4.9999f;
         cmsDoTransform(lab2softproof, srcbuf, buf2, width);
+        // since we are checking for out-of-gamut, we do want to clamp here!
+        for (int iy = 0, j = 0; j < width; ++j, iy += 3) {
+            buf2[iy] = LIM01(buf2[iy]);
+            buf2[iy+1] = LIM01(buf2[iy+1]);
+            buf2[iy+2] = LIM01(buf2[iy+2]);
+        }
         cmsDoTransform(softproof2ref, buf2, buf1, width);
         
         float *proofdata = buf1;
```

agriggio:
Not valid anymore (fixed in 6ba989ef1cb058b11d56b658489e9de7ed014ee3)

PR #5056, opened 2018-12-02
Lens profile bug fix

TooWaBoo:
When changed distortion the profile changed too.
Fixed


Beep6581:
@TooWaBoo could you explain the problem a bit more clearly and provide steps to reproduce?

TooWaBoo:
1. Set an image to Lensfun Auto
2. Set another image to Lensfun Manually
3. Highlight both images in the file browser
4. Check or uncheck any distortion in batch mode
--> both images are set to Auto or Manually depending on the the last highlighted image.

Beep6581:
Fix confirmed.

PR #5063, opened 2018-12-04
fix #4968 proposal

### LINKED ISSUE #4968: Context menu spacing issue with TooWa* themes
Currently using current master ( b488e20 ), but this occurs since some time.
Menu and sub-menu have incorrect spacing between them when using TooWa* themes. This does not occur with default RT theme.

See the screenshot, this is self explanatory.
toowagrey-bug

casta:
This is my proposal to address the issue I encounter on bug #4968


TooWaBoo:
see https://github.com/Beep6581/RawTherapee/issues/4968#issuecomment-444147074

PR #5064, opened 2018-12-04
Improve calculation of dual demosaic contrast threshold

heckflosse:
Calculation of dual demosaic contrast threshold value has up to two passes.
The first pass analyzes the image in tiles of 80x80 pixels. If there is at least one tile with variance <= 1 it uses the tile with the min. variance to calculate the treshold and skips pass two.

Because there are not always flat regions of 80x80 pixels pass two analyzes the image in tiles of 40x40 pixels. If there is at least one tile with variance <= 4 it uses the tile with the min. variance to calculate the threshold, else it returns a threshold of 0.

I improved pass 2 by changing the tile distance from 40 to 10. This increases processing time of pass 2 a bit but gives a better detection rate, means less threshold = 0 values.

### DIFF
--- a/rtengine/rt_algo.cc
+++ b/rtengine/rt_algo.cc
@@ -203,20 +203,19 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
             }
         }
     } else {
-        constexpr float scale = 0.0625f / 327.68f;
         if (autoContrast) {
             for (int pass = 0; pass < 2; ++pass) {
                 const int tilesize = 80 / (pass + 1);
                 const int skip = pass < 1 ? tilesize : tilesize / 4;
                 const int numTilesW = W / skip - 3 * pass;
                 const int numTilesH = H / skip - 3 * pass;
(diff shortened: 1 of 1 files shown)


rtengine/rt_algo.cc
```diff
@@ -203,19 +203,19 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
...
         }
     } else {
-        constexpr float scale = 0.0625f / 327.68f;
         if (autoContrast) {
             for (int pass = 0; pass < 2; ++pass) {
                 const int tilesize = 80 / (pass + 1);
-                const int numTilesW = W / tilesize;
-                const int numTilesH = H / tilesize;
-                std::vector<std::vector<std::pair<float, float>>> variances(numTilesH, std::vector<std::pair<float, float>>(numTilesW));
+                const int skip = pass < 1 ? tilesize : tilesize / 4;
```
Floessie: Minor: Better `pass == 0` or `const bool first_pass == !pass`?

Floessie (approved):
Looks good from the diff. :+1:

PR #5067, opened 2018-12-05
Get rid of special windows code

heckflosse:


heckflosse:
Merged to get more tests.

PR #5070, opened 2018-12-06
Improve detection of flat regions for calculation of dual demosaic co…

heckflosse:
…ntrast threshold

### DIFF
--- a/rtengine/rt_algo.cc
+++ b/rtengine/rt_algo.cc
@@ -304,6 +304,7 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
     if (autoContrast) {
         constexpr float minLuminance = 2000.f;
         constexpr float maxLuminance = 20000.f;
+        constexpr float minTileVariance = 0.5f;
         for (int pass = 0; pass < 2; ++pass) {
             const int tilesize = 80 / (pass + 1);
             const int skip = pass == 0 ? tilesize : tilesize / 4;
@@ -325,6 +326,8 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
                         continue;
                     } else {
                         variances[i][j] = tileVariance(luminance, tileY, tileX, tilesize, avg);
+                        // exclude tiles with a variance less than minTileVariance
+                        variances[i][j] = variances[i][j] < minTileVariance ? RT_INFINITY_F : variances[i][j];
                     }
                 }
             }
@@ -341,10 +344,9 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
                 }
             }
 
-            const int minY = skip * minI;
-            const int minX = skip * minJ;
-
             if (minvar <= 1.f || pass == 1) {
+                const int minY = skip * minI;
+                const int minX = skip * minJ;
                 if (pass == 0) {
                     // a variance <= 1 means we already found a flat region and can skip second pass
                     contrastThreshold = calcContrastThreshold(luminance, minY, minX, tilesize);
@@ -374,6 +376,8 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
                                 continue;
                             } else {
                                 variances[i][j] = tileVariance(luminance, tileY, tileX, tilesize, avg);
+                            // exclude tiles with a variance less than minTileVariance
+                            variances[i][j] = variances[i][j] < minTileVariance ? RT_INFINITY_F : variances[i][j];
                             }
                         }
                     }
--- a/rtengine/rt_algo.cc
+++ b/rtengine/rt_algo.cc
@@ -20,6 +20,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cmath>
+#include <cstddef>
 #include <cstdint>
 #include <vector>
 #ifdef _OPENMP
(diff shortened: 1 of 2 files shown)


rtengine/rt_algo.cc
```diff
@@ -216,47 +261,14 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
...
-#endif
-                        avg /= SQR(tilesize);
-                        if (avg < 2000.f || avg > 20000.f) {
+                        const float avg = tileAverage(luminance, tileY, tileX, tilesize);
+                        if (avg < minLuminance || avg > maxLuminance) {
                             // too dark or too bright => skip the tile
                             variances[i][j] = RT_INFINITY_F;
                             continue;
+                        } else {
+                            variances[i][j] = tileVariance(luminance, tileY, tileX, tilesize, avg);
```
Floessie: Much cleaner. :+1:

rtengine/rt_algo.cc
```diff
@@ -51,6 +51,49 @@ vfloat calcBlendFactor(vfloat valv, vfloat thresholdv) {
...
+    return avg / rtengine::SQR(tilesize);
+}
+
+float tileVariance(float **data, size_t tileY, size_t tileX, size_t tilesize, float avg) {
+
+#ifdef __SSE2__
+    vfloat varv = ZEROV;
+    const vfloat avgv = F2V(avg);
+    for (size_t y = tileY; y < tileY + tilesize; ++y) {
+        for (size_t x = tileX; x < tileX + tilesize; x +=4) {
```
Floessie: `x += 4`.

rtengine/rt_algo.cc
```diff
@@ -51,6 +51,49 @@ vfloat calcBlendFactor(vfloat valv, vfloat thresholdv) {
     return onev / (onev + xexpf(c16v - c16v * valv / thresholdv));
 }
 #endif
+
+float tileAverage(float **data, size_t tileY, size_t tileX, size_t tilesize) {
```
Floessie, line 55: AFAIK, `size_t` -> [`std::size_t`](https://en.cppreference.com/w/cpp/types/size_t).

rtengine/rt_algo.cc
```diff
@@ -276,17 +288,72 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
...
-                    // in second pass we allow a variance of 2
-                    JaggedArray<float> Lum(tilesize, tilesize);
-                    JaggedArray<float> Blend(tilesize, tilesize);
-                    for (int i = 0; i < tilesize; ++i) {
-                        for (int j = 0; j < tilesize; ++j) {
-                            Lum[i][j] = luminance[i + minY][j + minX];
+                    if (pass == 0) {
+                        // a variance <= 1 means we already found a flat region and can skip second pass
+                        JaggedArray<float> Lum(tilesize, tilesize);
+                        JaggedArray<float> Blend(tilesize, tilesize);
```
Floessie: `Lum` and `Blend` are instances. Shouldn't they be named `lum` and `blend`?

rtengine/rt_algo.cc
```diff
@@ -276,17 +288,72 @@ void buildBlendMask(float** luminance, float **blend, int W, int H, float &contr
...
+                            JaggedArray<float> Lum(tilesize, tilesize);
+                            JaggedArray<float> Blend(tilesize, tilesize);
+                            const int minY = topLeftYStart + minI;
+                            const int minX = topLeftXStart + minJ;
+                            for (int i = 0; i < tilesize; ++i) {
+                                for (int j = 0; j < tilesize; ++j) {
+                                    Lum[i][j] = luminance[i + minY][j + minX];
+                                }
+                            }
+                            contrastThreshold = calcContrastThreshold(Lum, Blend, tilesize, tilesize) / 100.f;
```
Floessie: Isn't that a repetition of the above? I mean: Shouldn't that go into a function or lambda?

Floessie (approved):

ff2000:
I had several RAW files now where I got a threshold of "0" though I think there are several flat regions bigger than 80x80. One of them I uploaded:
https://www.dropbox.com/s/3dx87k4l93jji8q/_1110596.RW2?dl=0
The improvements didn't change anything here. I manually set 20 (might still be higher than neccessary) which improves the water but keeps details in the ferns and lichens.

heckflosse:
@ff2000 The current improvements are about finding a better flat region (means a region with less or equal variance than before). If the variance of this region is still > 4, the threshold still will be set to 0.
With your example I get the following variances:

pass 1 : variance of best 80x80 tile: 64.9854. Greater than 1.0, will not be used
pass 2 : variance of best 40x40 tile when scanning in steps of 10 pixels: 6.98647
pass 3 : variance of best 40x40 tile in the 20x20 neighbourhood to the tile from pass 2: 6.38742

becasue the variance from pass 3 is still > 4, the threshold will be set to 0


I just push some console output. Maybe you can help to find a better value than 4 by looking at the output and/or playing with [this value in code](https://github.com/Beep6581/RawTherapee/blob/improve_dualdemosaic_flatregion_detection/rtengine/rt_algo.cc#L397)

ff2000:
@heckflosse I will give it a try tomorrow. But I think you have more experience, and I am afraid this might break other cameras' threshold. This is micro four thirds and it could be just too noisy...
I still have the slider to adjust it in case.

heckflosse:
@ff2000 

> I am afraid this might break other cameras' threshold.

Don't worry about that. Four eyes see more than two. That's important ;-)

PR #5127, opened 2018-12-31
WIP: Ukrainian translation

ror6ax:


Beep6581:
@ror6ax a translation is meant to be done based on the `default` language file, while you are working from the `Polish` file which is very outdated and some strings are incorrect.

https://discuss.pixls.us/t/localization-how-to-translate-rawtherapee-and-rawpedia/2594

ror6ax:
Thanks, I'll fix accordingly.

ror6ax:
@Beep6581 - What's the minimum amount of lines for a translation to be merged into master? I'd love to split these 1500 lines in 2 takes if possible.

Beep6581:
@ror6ax there is no minimum. I will merge it into `dev`, run it through the script, then you can download the new version from https://github.com/Beep6581/RawTherapee/blob/dev/rtdata/languages/Ukrainian

Beep6581:
@ror6ax done, though the Ukrainian file still has many entries not prefixed by an exclamation mark which are in fact in Polish, so incorrect, and therefore I removed it from the repo again.
You can find it here:
https://filebin.net/4h32rv3ohg4v3d1z/Ukrainian?t=g8qktt73

To fix it you need to start over, following https://discuss.pixls.us/t/localization-how-to-translate-rawtherapee-and-rawpedia/2594 :
1. Rename that file to `invalid`.
2. Download the latest [`default`](https://raw.githubusercontent.com/Beep6581/RawTherapee/dev/rtdata/languages/default)
3. Create a copy of `default` called `Ukrainian: `cp default Ukrainian`
3. Open all three files in a text editor. Translate each line in `Ukrainian` again. You can reference `invalid` if you want to, but since `invalid` is based on `Polish` which is incomplete and outdated and some strings are wrong, that means some of your translated lines in `invalid` are wrong, so verify each line against `default`.

Please open a new issue with just the translated file attached. It does not have to be 100% translated, but I will not be able to accept it if any of the lines without an exclamation mark are invalid.

PR #5146, opened 2019-01-16
Fix standard tonecurve for non sse builds, also inluced speedup for s…

heckflosse:
…se builds


Floessie (approved):
Builds and runs fine on Debian 10 i386 non-SSE and Debian unstable PPC. :+1:

PR #5159, opened 2019-01-30
Speedup for hphd demosaic

heckflosse:
Measured on my rusty 8-core AMD FX 8350 using a 36 MP file

before: ~1150 ms
after: ~ 560 ms


heckflosse:
With last commit processing time reduced to ~470 ms

rtengine/demosaic_algos.cc
```diff
@@ -62,88 +62,100 @@ extern const Settings* settings;
...
+    float* temp = new float[W];
+    float* avg = new float[W];
+    float* dev = new float[W];
 
-    memset (temp, 0, max(W, H)*sizeof(float));
-    memset (avg, 0, max(W, H)*sizeof(float));
-    memset (dev, 0, max(W, H)*sizeof(float));
+    memset(temp, 0, W * sizeof(float));
+    memset(avg, 0, W * sizeof(float));
+    memset(dev, 0, W * sizeof(float));
```
Floessie: @heckflosse I came across an interesting [detail](https://en.cppreference.com/w/cpp/language/new) (under *Construction*):

> * If type is an array type, an array of objects is initialized.
>   * If initializer is an empty pair of parentheses, each element is value-initialized.

As `new()` under Linux apparently zero-initialized by default, I had a look at the assembly:

```
        int* const test = new int[items]();
  400896:       bf 00 00 04 00          mov    $0x40000,%edi
  40089b:       e8 70 fe ff ff          callq  400710 <operator new[](unsigned long)@plt>
  4008a0:       48 89 c1                mov    %rax,%rcx
  4008a3:       48 89 c8                mov    %rcx,%rax
  4008a6:       ba ff ff 00 00          mov    $0xffff,%edx
  4008ab:       eb 0e                   jmp    4008bb <main+0x35>
  4008ad:       c7 00 00 00 00 00       movl   $0x0,(%rax)
  4008b3:       48 83 c0 04             add    $0x4,%rax
  4008b7:       48 83 ea 01             sub    $0x1,%rdx
  4008bb:       48 83 fa ff             cmp    $0xffffffffffffffff,%rdx
  4008bf:       75 ec                   jne    4008ad <main+0x27>
  4008c1:       48 89 4d e8             mov    %rcx,-0x18(%rbp)

       int* const test = new int[items];
  400896:       bf 00 00 04 00          mov    $0x40000,%edi
  40089b:       e8 70 fe ff ff          callq  400710 <operator new[](unsigned long)@plt>
  4008a0:       48 89 45 e8             mov    %rax,-0x18(%rbp)
```

:tada: 

So it looks like you can spare the `memset()`s by simply adding  `()` to `new float[W]`. That will not make the code faster but less verbose.

Best,
Flössie

heckflosse: @Floessie Thanks for the review :+1: 
I included your suggested change in the [librtprocess hphd code](https://github.com/CarVac/librtprocess/commit/5889d38a0b977a331f65ea99bf6ae866ce3f95ad#diff-5990c9321c7b70ce2e381ae019e6b06dR119)

Floessie: @heckflosse And I learned about `new(std::nothrow)` from you now. :smile: :+1:

Issue #5174, opened 2019-02-11
Filmstrip moves active image to the far left


Thanatomanic:
I think #5125 actually had a side-effect, although @Beep6581 said it didn't. Or maybe there was another commit that changed the filmstrip behavior, but I'm not seeing it. In any case, the issue below is not present in 5.5 (release version) but it is in dev. For what it's worth, I'm using SETM.

What's the issue: whenever I select an image from the filmstrip, it moves the strip so that the selected image is the one on the far left. I would very much prefer the selected image to be (somewhat) in the middle. This is especially annoying when working on a sequence of images where you repeatedly have to scroll to get the earlier images in view.

TooWaBoo:
This will fix it. Check it out.
Change yellow to blue. 😀
### editorpanel.cc
![unbenannt](https://user-images.githubusercontent.com/13800920/52899103-6fd54e00-31e6-11e9-97d8-b5200cbb678a.png)

heckflosse:
@TooWaBoo Your modification breaks this:
```
    // When passing a photo as an argument to the RawTherapee executable, the user wants
    // this auto-loaded photo's thumbnail to be selected and visible in the Filmstrip.
```

heckflosse:
@TooWaBoo scratch my comment from above.

TooWaBoo:
> @TooWaBoo scratch my comment from above.

I will. 😁

heckflosse:
@TooWaBoo But F3/F4 does not behave as before...

heckflosse:
@TooWaBoo before modification after F4 the current thumb in filmstrip is always the leftmost (if there are enough thumbs on the right)
after modification sometimes after F4 the current thumb in filmstrip is the rightmost even if there are enough thumbs on the right.

TooWaBoo:
This should fix the F3/F4 behaviour.
![grafik](https://user-images.githubusercontent.com/13800920/52900712-6dc8ba80-31f9-11e9-9a61-07016f8f5d28.png)

EDIT: Oh sh..., it doesn't. I've to play a bit more.

TooWaBoo:
Ok, now all changes together.
### editorpanel.cc
![52899103-6fd54e00-31e6-11e9-97d8-b5200cbb678a](https://user-images.githubusercontent.com/13800920/52903221-a462fd00-321a-11e9-99f4-5995c921c9e1.png)

### filebrowser.cc
[...]

PR #5183, opened 2019-02-20
Center selected thumb in Filmstrip #5174

Beep6581:
When opening an image, selecting the next/previous image or syncing the
Filmstrip, the newly selected thumbnail is centered, unless it lies
within the visible area more than 1 thumbnails-width away from either
edge, in which case centering does not occur.

### DIFF
--- a/rtgui/filebrowser.cc
+++ b/rtgui/filebrowser.cc
@@ -1727,62 +1727,62 @@ void FileBrowser::buttonPressed (LWButton* button, int actionCode, void* actionD
     }
 }
 
-void FileBrowser::openNextImage ()
+void FileBrowser::openNextImage()
 {
     MYWRITERLOCK(l, entryRW);
 
     if (!fd.empty() && selected.size() > 0 && !options.tabbedUI) {
-
         for (size_t i = 0; i < fd.size() - 1; i++) {
             if (selected[0]->thumbnail->getFileName() == fd[i]->filename) { // located 1-st image in current selection
                 if (i < fd.size() && tbl) {
                     // find the first not-filtered-out (next) image
                     for (size_t k = i + 1; k < fd.size(); k++) {
                         if (!fd[k]->filtered/*checkFilter (fd[k])*/) {
+
                             // clear current selection
                             for (size_t j = 0; j < selected.size(); j++) {
                                 selected[j]->selected = false;
                             }
 
-                            selected.clear ();
+                            selected.clear();
 
                             // set new selection
                             fd[k]->selected = true;
-                            selected.push_back (fd[k]);
-                            //queue_draw ();
+                            selected.push_back(fd[k]);
+                            //queue_draw();
 
                             MYWRITERLOCK_RELEASE(l);
 
                             // this will require a read access
-                            notifySelectionListener ();
+                            notifySelectionListener();
 
                             MYWRITERLOCK_ACQUIRE(l);
 
-                            // scroll to the selected position
-                            double h1, v1;
-                            getScrollPosition(h1, v1);
+                            // scroll to the selected position, centered horizontally in the container
+                            double x1, y1;
+                            getScrollPosition(x1, y1);
 
-                            double h2 = selected[0]->getStartX();
-                            double v2 = selected[0]->getStartY();
+                            double x2 = selected[0]->getStartX();
+                            double y2 = selected[0]->getStartY();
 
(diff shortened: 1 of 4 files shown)


rtgui/filebrowser.cc
```diff
@@ -1727,62 +1727,62 @@ void FileBrowser::buttonPressed (LWButton* button, int actionCode, void* actionD
...
+                            if ((x2 > x1 + ww - 2 * tw) || (x2 - tw < x1)) {
+                                setScrollPosition(x2 - (ww - tw) / 2, y2);
                             }
 
                             // open the selected image
                             std::vector<Thumbnail*> entries;
-                            entries.push_back (thumb);
-                            tbl->openRequested (entries);
+                            entries.push_back(thumb);
+                            tbl->openRequested(entries);
```
Floessie: If I get that right from the diff, these three lines could be replaced by one:
```c++
tbl->openRequested({thumb});
```

rtgui/filebrowser.cc
```diff
@@ -1792,62 +1792,62 @@ void FileBrowser::openNextImage ()
...
```
Floessie: Same here.

rtgui/filebrowser.h
```diff
@@ -181,12 +181,13 @@ class FileBrowser  : public ThumbBrowserBase,
         return tbl ? tbl->isInTabMode() : false;
     }
 
-    void openNextImage ();
-    void openPrevImage ();
+    void openNextImage();
+    void openPrevImage();
+    void selectImage(Glib::ustring fname);
```
Floessie: This should really be `void selectImage(const Glib::ustring& fname)`.

rtgui/filebrowser.cc
Floessie: Hm, looks like a lot could be factored out here...

rtgui/filebrowser.h
```diff
@@ -181,12 +181,13 @@ class FileBrowser  : public ThumbBrowserBase,
...
+
     void copyProfile ();
     void pasteProfile ();
     void partPasteProfile ();
-    void selectImage (Glib::ustring fname);
     void openNextPreviousEditorImage (Glib::ustring fname, eRTNav eNextPrevious);
```
TooWaBoo: @Floessie 
...then this should be `void openNextPreviousEditorImage (const Glib::ustring fname, eRTNav eNextPrevious);` too

TooWaBoo:
The behavior between` F3/F4` and `<shift> F3/F4` or buttons `Left/Right` is different.

`F3/F4` centers the image on the last fully visible thumb.

 `<shift> F3/F4` or buttons `Left/Right` always scrolls the filmstrip and the selcted thumb is always one thumb left or right from the center.

TooWaBoo:
This will fix the navigatopn issue
### filebrowser.cc
old
```
void FileBrowser::openNextPreviousEditorImage (Glib::ustring fname, eRTNav nextPrevious)
{
    // let FileBrowser acquire Editor's perspective
    selectImage (fname);

    // now switch to the requested image
    if (nextPrevious == NAV_NEXT) {
        openNextImage();
    } else if (nextPrevious == NAV_PREVIOUS) {
        openPrevImage();
    }
}
```

new
```
void FileBrowser::openNextPreviousEditorImage (Glib::ustring fname, eRTNav nextPrevious)
{
    // switch to the requested image
    if (nextPrevious == NAV_NEXT) {
        openNextImage();
    } else if (nextPrevious == NAV_PREVIOUS) {
        openPrevImage();
    } else {
        // let FileBrowser acquire Editor's perspective
        selectImage (fname);
    }
}
```

rtgui/filebrowser.cc
Beep6581: Thank you @Floessie , I will do that.

Beep6581:
Thank you @TooWaBoo , I wasn't aware of the `<shift>` variant, will look into that. What do you mean by "buttons Left/Right", do you mean the   buttons below the main preview?

TooWaBoo:
> What do you mean by "buttons Left/Right", do you mean the   buttons below the main preview?

Yes, it's the same function as `<shift> F3/F4`

Beep6581:
I believe the branch is ready for merging. @TooWaBoo could you please test again?

Floessie (commented):
> Default to using defaults by default

:rofl:

TooWaBoo:
> 
> 
> I believe the branch is ready for merging. @TooWaBoo could you please test again?

Tested and it works fine.

PR #5197, opened 2019-02-28
Relax dependency from `procparams.h`

Floessie:
Hi there,

I've always wondered why pulling from GitHub often results in the complete tree being built. Especially `rtexif` stands out, as there isn't much progress. I always suspected it has something to do with `config.h.in`, but now I've found out that everything depends on `proparams.h`.

So here's my work to forward-declare the `proparams.h` members wherever possible. When now touching `procparams.h` recompiles 147 `.cc` files while it were 227 before. There are still some includes in `rtgui` which I missed. I'll try to finish them by tomorrow evening.

I had to make some trade-offs:
- Members had to be turned into `std::unique_ptr<>`s.
- `ToneCurveParams::TcMode` had to be split out as `ToneCurveMode`.
- `ExifPairs` and `IPTCPairs` are no longer `typedef`s but thin wrappers.
- For the sake of simplicity some `RenderingIntent` members were turned into `int`s.

I hope you also see some benefit in this and if not, let me know.

Best,
Flössie


Floessie:
Okay, ready for review.

heckflosse:
@Floessie I get this warnings:
```
Z:/H2/rt55_fix/rtgui/thumbnail.h: In constructor 'Thumbnail::Thumbnail(CacheManager*, const Glib::ustring&, CacheImageData*)':
Z:/H2/rt55_fix/rtgui/thumbnail.h:55:21: warning: 'Thumbnail::pparamsValid' will be initialized after [-Wreorder]
     bool            pparamsValid;
                     ^~~~~~~~~~~~
Z:/H2/rt55_fix/rtgui/thumbnail.h:54:66: warning:   'const std::unique_ptr<rtengine::procparams::ProcParams> Thumbnail::pparams' [-Wreorder]
     const std::unique_ptr<rtengine::procparams::ProcParams>      pparams;
                                                                  ^~~~~~~
Z:/H2/rt55_fix/rtgui/thumbnail.cc:40:1: warning:   when initialized here [-Wreorder]
 Thumbnail::Thumbnail (CacheManager* cm, const Glib::ustring& fname, CacheImageData* cf)
 ^~~~~~~~~
In file included from Z:/H2/rt55_fix/rtgui/thumbnail.cc:19:
Z:/H2/rt55_fix/rtgui/thumbnail.h: In constructor 'Thumbnail::Thumbnail(CacheManager*, const Glib::ustring&, const string&)':
Z:/H2/rt55_fix/rtgui/thumbnail.h:55:21: warning: 'Thumbnail::pparamsValid' will be initialized after [-Wreorder]
     bool            pparamsValid;
                     ^~~~~~~~~~~~
Z:/H2/rt55_fix/rtgui/thumbnail.h:54:66: warning:   'const std::unique_ptr<rtengine::procparams::ProcParams> Thumbnail::pparams' [-Wreorder]
     const std::unique_ptr<rtengine::procparams::ProcParams>      pparams;
                                                                  ^~~~~~~
Z:/H2/rt55_fix/rtgui/thumbnail.cc:68:1: warning:   when initialized here [-Wreorder]
 Thumbnail::Thumbnail (CacheManager* cm, const Glib::ustring& fname, const std::string& md5)
 ^~~~~~~~~

```

Floessie:
@heckflosse Should be fixed now.

Hombre57:
@Floessie Thanks for requesting my review, but honestly, I don't have a clue of what is better or not regarding coding practice. I can only tell that it touch a lot of file, but I'm also used to this kind of commit :slightly_smiling_face: . All I need to know is how to refactor the code from my other branches to comply to the new implementation (other devs will be interested too).

Apart from that, I'm happy that you managed to speedup the re-compile process. :+1:

Floessie:
> All I need to know is how to refactor the code from my other branches to comply to the new implementation (other devs will be interested too).

Hi Hombre,

No need to "comply" with anything. :smiley: To speed up compile times the only thing you should strive for is including as few headers as possible. Otherwise you'll get dependency chains where a single header change results in a full tree compilation. Thus you should forward declare instead of including.

Forward declarations are sufficient for:
1. Parameters or members passed or stored as references or pointers.
2. Return types.
3. Template parameters if the instances are passed or stored as references or pointer.

Maybe I've missed something, but those are the most important points.

You can forward declare:
- classes
- structs
- enums with an underlying type
- templates

You can't forward declare:
- typedefs
- inner classes / structs / enums

Again, some cases may be missing, but those are the basics.

So next time you write a header file, forward the types your interface needs instead of including the corresponding headers. STL types are hard to get right, so forward declaring them is usually not recommended. The same might be true for glibmm types.

One thing that really rocks concerning seperation of header and implementation is the [PIMPL](https://en.wikipedia.org/wiki/Opaque_pointer#C++). The Wikipedia entry is a full implementation. Most of the time, our objects are non-copyable, so a basic implementation is enough (not test-compiled, I'm on holiday atm):

```c++
// Header

#include <memory>

class OtherClass;

class MyClass final
{
public:
    MyClass();
    ~MyClass();
    
    void consume(const OtherClass& other);
    OtherClass returnOther() const;

private:
    class Implementation;
    
    const std::unique_ptr<Implementation> implementation;
};


// Implemenation

#include "OtherClass.h"
#include "ALotOfOtherStuff.h"

class MyClass::Implementation final
{
public:
    Implementation() :
        init_everything()
    {
    }
    
    ~Implementation()
    {
        cleanEverythingUp();
    }
    
    void consume(const OtherClass& other)
    {
        // Do everything that needs to be done
    }
    
    OtherClass returnOther() const
    {
        return {}; // Or do more ;)
    }

private:
    // Arbitrary members
};

MyClass::MyClass() :
    implementation(new Implementation)
{
}

MyClass::~MyClass() = default;

void MyClass::consume(const OtherClass& other)
{
    implementation->consume(other);
}

OtherClass MyClass::returnOther() const
{
    return implementation->returnOther();
}
```

HTH,
Flössie

Floessie:
Anyone against merging this for 5.6?

Hombre57:
@Floessie @Beep6581 Ok to commit for me, but your explanation just above should go in the coding guidelines.

Hombre57:
Hmmm, one thing I have to check is class visibility by the IDE. I'd like to be able to access the structure so it shouldn't be hidden. Is it possible to have real includes based on a preprocessor switch, that would be different in the IDE than the build script? Let me see if it's required first.

Floessie:
@Hombre57 Eclipse will take you to the definition by default. So no problem...

Hombre57:
@Floessie Great !

Hombre57 (approved):

Beep6581:
> but your explanation just above should go in the coding guidelines.

I added it to https://github.com/Beep6581/RawTherapee/blob/dev/CONTRIBUTING.md
For me this is akin to copy-pasting text in Aramaic, so please make fixes directly if needed.

Floessie:
@Beep6581 I think that's a bit too much, given that it's 101 of C++ programming. I had already added one line in cd566da.

Hombre57:
@Beep6581 @Floessie If this is something every C++ programmer should know, then we can remove the full text, I can live with(out) that.

Beep6581:
aaand it's gone.

Floessie:
@Beep6581 Waiting for your approval...

Beep6581 (approved):
Compiled successfully without warnings. Tested and no issues found.

Floessie:
Also successfully tested on Debian 10 i386 and powerpc.

PR #5208, opened 2019-03-10
Adobe (filmlike) tonecurve: SSE version

heckflosse:
No differences in output.
Measured in rgbproc() using a Nikon D850 file with a 'Film-like' tonecurve:

before: 530 ms
after: 455 ms


Thanatomanic:
I'm pretty sure you put in `curves.h.save-failed` by accident.
For the rest, I get much lower processing times on my machine (~ 50 ms at most) but everything seems to work properly.

heckflosse:
@Thanatomanic I guess you tested in preview, not in full processing using the queue

heckflosse:
Update:

After merging the tiles per thread branch and this branch I tested again using the same file.
Processing time of rgbproc before merging both branches: 530 ms
Processing time of rgbproc after merging both branches: 380 ms

Issue #3636, opened 2017-01-26
SIGABRT in FreeBSD when closing


mandree:
When quitting normally from RawTherapee 5.0 release GTK3 on FreeBSD 10.3 (amd64), I get this and a core dump:
```
GLib (gthread-posix.c): Unexpected error from C library during 'pthread_mutex_lock': Invalid argument.  Aborting.

(gdb) bt full
#0  0x00000008062b739a in thr_kill () from /lib/libc.so.7
No symbol table info available.
#1  0x00000008062b7386 in raise () from /lib/libc.so.7
No symbol table info available.
#2  0x00000008062b7309 in abort () from /lib/libc.so.7
No symbol table info available.
#3  0x0000000801d156d5 in ?? () from /usr/local/lib/libglib-2.0.so.0
No symbol table info available.
#4  0x0000000801d1569e in g_mutex_lock () from /usr/local/lib/libglib-2.0.so.0
No symbol table info available.
#5  0x0000000801fe4ba2 in Glib::ObjectBase::~ObjectBase() () from /usr/local/lib/libglibmm-2.4.so.1
No symbol table info available.
#6  0x0000000803c64dbd in Gdk::Pixbuf::~Pixbuf() () from /usr/local/lib/libgdkmm-3.0.so.1
No symbol table info available.
#7  0x0000000803c64e89 in virtual thunk to Gdk::Pixbuf::~Pixbuf() () from /usr/local/lib/libgdkmm-3.0.so.1
No symbol table info available.
#8  0x0000000801cb36d3 in g_datalist_clear () from /usr/local/lib/libglib-2.0.so.0
No symbol table info available.
#9  0x0000000801a4fa2a in g_object_unref () from /usr/local/lib/libgobject-2.0.so.0
No symbol table info available.
#10 0x0000000806316210 in __cxa_finalize () from /lib/libc.so.7
No symbol table info available.
#11 0x00000008062b729c in exit () from /lib/libc.so.7
No symbol table info available.
#12 0x00000000005638d6 in _start ()
No symbol table info available.
```

heckflosse:
@Floessie Do you have an idea?

Floessie:
[...]

PR #5215, opened 2019-03-13
Hidpi GUI

Hombre57:
This branch adds HiDPI support by scaling the whole interface, taking into account the real dpi and font size vs base font size of 9pt (see issue #3547). The icons are now bundled in svg format, and rendered on use (stored in a cache) whenever the resolution changes (restarting RT is required).

You can test this branch by using the native resolution of your HiDPI monitor (MacOS users will probably need the _RDM_ tool for that) or set this environment variable : `GDK_SCALE=2` to simulate HiDPI on FullHD monitors. 

Now that the full resolution of 4k monitors is exploited, you can expect longer rendering time for the preview because it has to crunch more data.

Also integrate a fix for #3636 from @Floessie.

Should be ready to merge, comments welcome.


heckflosse:
@Hombre57 Already compiling...

Beep6581:
No problems here in Linux/KDE using gtk+-3.24.1.

gaaned92:
W10, GTK3 3.22.3-1.
builds and runs ok.
No GUI glitch on low dpi display

Floessie:
Testing on low DPI display.

- Debian 10
  - amd64
    - GCC 8.3 :heavy_check_mark: 
    - GCC 9 (experimental) :heavy_check_mark: 
    - Clang 7 :heavy_check_mark: 
    - Clang 9 (experimental) :heavy_check_mark: 
  - i386 (w/o SSE2)
    - GCC 8.3 :heavy_check_mark: 
    - Clang 7 :heavy_multiplication_x: (`-ffloat-store` unsupported, SIGILL, also on `dev`)
  - powerpc
    - GCC 8.3 :heavy_check_mark: 
    - Clang 7 :heavy_check_mark: (slooooow compilation)
- FreeBSD 12
  - amd64
    - Clang 6 :heavy_check_mark: 

Observations:
- There are some warnings in `FTblockDN.cc` but I think they're also on `dev`.
- There is still some debug output on the console.
- Folder line spacing seems wider than before.
- Spacing around the icons on top wider.
- Needs `librsvg2-dev` on Debian and `librsvg2` on FreeBSD (:memo: RawPedia).

heckflosse:
One minor thing:

Start rt, it shows a progressbar. When all thumbs are loaded, the progressbar looks like this in dev:
grafik

In hdpi-icons it looks like this:
grafik

Because of the small blue part at the left it looks like thumb loading wasn't completed, which is a bit annoying.
Same for the progressbar below the editor preview.

Edit: vertical scrollbar when using SETM vertical tabs mode works fine

heckflosse:
As @Floessie already mentioned: Folder line spacing is different. I prefer the old folders look

dev
grafik
hidpi-icons
grafik

gaaned92:
on  W10, GTK3 3.22.3-1 Toowaboo-grey, average surround,
 I don't see the same things as Ingo

- progress bar: on dev and Hidpi-icons, the progress bar are thin. On both builds they become entirely grey when finished

- icon sizes are identical

icons

- folder panes are identical, except the window header is wider


foldertree

- you can notice that RT icones are different
RTicones

Beep6581:
@gaaned92 the issues relate to the RawTherapee theme, which now inherits much from the TWB themes.

As we discussed in IRC, the RT theme needs liposuction in a few places, but in my opinion we should go ahead with the merge and tweak these things in `dev`. The feature-freeze is this weekend and we could use the `dev` testing.

TooWaBoo:
The folder spacing is necessary 'cause it's not possible to size individual items in a treeview. If the spacing would be like the old dev-RT-theme, items would overlapping when increasing the font size.

EDIT: Forget this, with the new icon scaling standar spacing works fine. Changing the TWB theme

TooWaBoo:
WB Method icons don't grow with the font size.
grafik

TooWaBoo:
The icons of e.g. image buttons don't become grey when they are disabled.

Hombre57:
Thanks everybody for the testing. I'll try to adress most of them (the simplest ones) tomorrow. Don't expect RawTherapee theme to be exaaactly the same than for dev though.

Hombre57:
@Floessie 

> There is still some debug output on the console.

Could you paste theme here ?

Floessie:
@Hombre57 > Could you paste them here?

I wish I could, but I can't reproduce them anymore. It was something like "DPI: 96.0 / ...". Obviously they disappeared, so scratch that line of mine. :blush:

Hombre57:
@Floessie @heckflosse @Beep6581 @TooWaBoo I think I've solved the issues.

@Beep6581 When merge will be done, could you edit RawPedia to add the `librsvg` dependency ? (see @Floessie's comment, [last line](https://github.com/Beep6581/RawTherapee/pull/5215#issuecomment-472465942))

TooWaBoo:
Disabled icons still doesen't work.
grafik

Hombre57:
@TooWaBoo Disabled icons does work now, excepted special cases I didn't saw. I'll fix this in an hour.

Hombre57:
@TooWaBoo 

Disabled icons even works for reset button here.

disabled-icons

Where does you screenshot comes from ? I tried the `Save As` button, but it's not disabled here.

TooWaBoo:
@Hombre57 
I've found the culprit. In the TWB theme I'm using `-gtk-icon-effect: dim;` and this doesn't work anymore.
Modifying the TWB theme from `-gtk-icon-effect: dim;` to `opacity: .45;` for disabled images. This works fine.

Beep6581:
Looks good @Hombre57 

The `entry` in `FileBrowserQueryToolbar` needs to be fixed (the other `entry` are fine, but this one has the bottom part eaten away)
screenshot_20190316_163714

TooWaBoo:
@Beep6581 ; @Hombre57 

To fix the query toolbar change line `line 241` in `size.css` from:
```
#FileBrowserQueryToolbar{
    margin: 0 0 0.416666666666666666em 0;
    min-height: 0;
    min-width: 0;
    padding: 0;
}
```
to:
```
#FileBrowserQueryToolbar > viewport > box{
    margin: 0 0 0.416666666666666666em 0;
    min-height: 0;
    min-width: 0;
    padding: 0;
}
```

Hombre57:
@TooWaBoo Thanks for the fix, patch committed.

Any objection to the merge ?

TooWaBoo:
Can be merged.

Issue #1808, opened 2015-08-11
Overwriting existing file doesn't work when using Queue


Beep6581:
Originally reported on Google Code with ID 1824

```
Edit a photo, hit Ctrl+s and click on an existing file. Disable "Automatically add a
suffix if the file already exists". If you use the queue, the existing file will not
get overwritten. It should.

Works fine if you "save immediately".

Linux, noticed this for some time now. Not related to issue 1794.
```

Reported by `entertheyoni` on 2013-04-02 23:00:58

Beep6581:
Still broken in `5.4-1142-g`834f6bafe

okoncevoy:
I have created pull request with fix: https://github.com/Beep6581/RawTherapee/pull/5229

PR #5229, opened 2019-03-18
Fixing batchqueue file overwrite #1808

okoncevoy:


Beep6581:
I tested, and when the "Overwrite existing output files" option in Preferences > Batch Processing (it really should not be there but in the Queue tab) is ticked and you put the same file to queue several times, you still end up with `foo.jpg`, `foo-1.jpg`, `foo-2.jpg`, etc. Tried sending images to queue from the File Browser and from the Editor, it made no difference.

Beep6581:
When I restarted RT then saving via the queue would overwrite the files. But then I disabled "Overwrite existing output files" and the queue would still overwrite files. So I restarted RT, and saving via queue still overwrites files. So the saving overwrite mechanism is still broken and somewhat erratic.

My settings:
screenshot_20190319_132444
screenshot_20190319_132456
screenshot_20190319_132755

Beep6581:
I think I identified the problem. In the Save window you can see the option "Automatically add a suffix if the file exists". This setting should take priority when using the Save dialog, that makes sense. But it seems that RawTherapee uses this setting regardless how you save the image, so even if I start RT, DON'T open any image in the editor, just select a thumbnail in the File Browser and hit ctrl+b, then start the queue, then RT does whatever I last set that option to, regardless of what "Preferences > Batch Processing > Overwrite existing output files" is set to. And that is wrong, since I didn't even see the Save window so the checkbox from Preferences should take priority.

okoncevoy:
Yes. You are correct, I was focused on save dialog. I will update a pull request in next few days.

PR #5231, opened 2019-03-19
Suppress 'rtwindow.h' from main-cli.cc

Hombre57:
Should be safe, but I wouldn't mind a validation. Please thumbs up if ok to merge (for 5.6).

### DIFF
--- a/rtgui/main-cli.cc
+++ b/rtgui/main-cli.cc
@@ -28,15 +28,17 @@
 #include <giomm.h>
 #include <iostream>
 #include <tiffio.h>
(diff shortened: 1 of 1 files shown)


Floessie:
Still builds fine on Linux. :+1:

gaaned92:
W10, builds and runs

Beep6581:
Builds fine in Linux, tested GUI and CLI, no problems.

Floessie:
As it only shuffles includes a compile test is sufficient, IMHO.

PR #5237, opened 2019-03-21
Fixed file overwrite from queued items

### LINKED ISSUE #1808: Overwriting existing file doesn't work when using Queue

okoncevoy:
Fixes #1808


Beep6581:
Fix confirmed.

Issue #5194, opened 2019-02-25
Lensfun lens selection broken in latest windows builds.


heckflosse:
In latest windows builds (nightly) and in my own windows build, the lensfun lens selection is broken.
If I want to select a Nikkor Ais lens I can't do that anymore, because the scroll-arrow at the bottom of the lens list is not available. Maybe a gtk issue?

Edit: works fine in official windows build of RT 5.5

Can other windows builders confirm? I'm on Windows7/64

![grafik](https://user-images.githubusercontent.com/13733487/53340505-73b15080-3909-11e9-957b-05648fc9cd6e.png)

gaaned92:
Yes confirmed on W10 64.

TooWaBoo:
Confirmed W10 x64 Home

heckflosse:
@gaaned92 @TooWaBoo 
Do you know what to downgrade to make it working again?

TooWaBoo:
@heckflosse 
I think you have to downgrade to:
`mingw-w64-x86_64-gtk3-3.22.30-1` and `mingw-w64-x86_64-gtkmm3-3.22.3-1`
All higher versions have some issues.

heckflosse:
@TooWaBoo Thanks :+1: 
Works fine after downgrading

gaaned92:
@heckflosse @TooWaBoo 
Do you think:
-  it is a bug in GTK3 that will be corrected by GTK3 people
-  or a modification of behaviour (a feature), requiring a modif in RT.
In the mean time, I am going to downgrade!
Thanks.

heckflosse:
I think it is a bug in GTK3

LordPhoenixfr:
Hi folks
Just to say I have same problem here with rawtherapee 5.5 and Ubuntu 18.10.

Beep6581:
I need to know for the 5.6 release notes which version(s) of GTK+ are affected. @LordPhoenixfr @heckflosse @TooWaBoo @gaaned92 which versions are you using?

Beep6581:
No problem with GTK+ 3.24.1 in Linux.

TooWaBoo:
@Beep6581 
Windows 10, GTK+ 3.24.5

Floessie:
- Debian 10 (GTK+ 3.24.5): :x:
- Debian 9 (GTK+ 3.22.11): :heavy_check_mark: 
- FreeBSD 12 (GTK+ 3.22.30): :heavy_check_mark:

[...]

PR #5253, opened 2019-03-30
Warn if using problematic GTK+ version #5194

Beep6581:
GTK+ versions >=3.24.2 - <=3.24.6 are known to not draw combobox menu
scroll-arrows.


Beep6581:
I will merge this evening if noone objects.

Beep6581:
Merged manually in 458f47d.

PR #5262, opened 2019-04-03
Improvements on Locallab spot

Pandagrapher:
@Desmis 
Improvements:
- Replaces Arcellipse widget by Ellipse widget (to simplify spot management)
- Gets Locallab spots more discrete and add opacity according to selected Locallab spot

Bugfixes:
- Mouse cursor is now correctly managed according to widget when mouse-overing
- Ellipse radius are now saturated at minimum at 2 to avoid ellipse to disapear at small radius values
- Selection points have now priority on ellipse/rectangle and center circle which improves Locallab spot resizing
- When mouse-overing a Locallab spot, all the spot becames now orange (even selection spots)
- Avoids Locallab spot to disapear when mouse right-clicking

Pierre


rtgui/editwidgets.h
```diff
@@ -570,6 +593,12 @@ inline Line::Line (int beginX, int beginY, int endX, int endY) :
         begin (beginX, beginY), end (endX, endY) {
 }
 
+inline Ellipse::Ellipse (rtengine::Coord& center, int radYT, int radY, int radXL, int radX,
```
Floessie, line 565: @Pandagrapher Salut Pierre, I really like the way you code: It's a delight to read. :+1:

Here's one place (not your fault) that needs to be changed: `center` isn't an "out" parameter and should be `const rtengine::Coord&`.

Pandagrapher: @Floessie 
Sorry for my late answer... Thanks for the review :) I have corrected it with commit `5a58d32`
Pierre

Desmis:
@Pandagrapher 

Have you seen the patch from @Floessie , to avoid memory leaks here https://github.com/Beep6581/RawTherapee/issues/5261

jacques

PR #5264, opened 2019-04-04
Fix broken retinex in locallab. Also some cleanups

heckflosse:


Desmis:
@heckflosse 

Ingo
I will test :)

Thank you
jacques

Desmis:
@heckflosse 
Works well :)

I tested on file _DSC0653
open file - 100ms
zoom 100%  - 2100ms
export TIF - 3000ms
I think "normal" because call to "MSR"

The result is (for me) very good
* global image with "dehaze" : strength 64, depth 25
* then locallab on upper part
_DSC0653

PR #5273, opened 2019-04-08
Locallab blur cleanup and small speedup

heckflosse:


rtengine/CMakeLists.txt
```diff
@@ -79,6 +79,7 @@ set(RTENGINESOURCEFILES
     impulse_denoise.cc
     init.cc
     iplab2rgb.cc
+	iplocallab.cc
```
Beep6581, line 82: @heckflosse indentation seems off (when viewed in GitHub)

rtdata/languages/default
```diff
@@ -692,21 +692,37 @@ HISTORY_MSG_441;Retinex - Gain transmission
...
-HISTORY_MSG_465;PS - Blur radius
-HISTORY_MSG_468;PS - Fill holes
-HISTORY_MSG_469;PS - Median
-HISTORY_MSG_471;PS - Motion correction
-HISTORY_MSG_472;PS - Smooth transitions
-HISTORY_MSG_473;PS - Use LMMSE
-HISTORY_MSG_474;PS - Equalize
-HISTORY_MSG_475;PS - Equalize channel
+HISTORY_MSG_445;Raw Sub-Image
+HISTORY_MSG_446;EvPixelShiftMotion
```
Beep6581, line 696: Are these keys HISTORY_MSG_446 - 475 used in the "Raw" tab, or by the Local tool? If the latter, why aren't they prefixed with "Local - " same as HISTORY_MSG_515 - 661?

Beep6581:
A bit confused about all this. I haven't seen branch `locallab_burr_cleanup` before, but this PR makes it seem like this branch has been around since 2018, that it has a whopping 581 commits, and that you want to merge it into `dev`, none of which makes sense. Also unclear how this fits in with the other locallab speedup branches `newlocallab-speedup` `newlocallab-speedup2`.

heckflosse:
@Beep6581 This was nonsense. I forgot to change the destination branch to newlocallab. Then I set it to newlocallab and forgot to close this one.

PR #5274, opened 2019-04-08
locallab: cleanup and small speedup for blur

heckflosse:


Desmis:
@heckflosse 
Ingo
I tested this morning

No differences on TIF
Speed up about 30%
good work :)

jacques

PR #5280, opened 2019-04-12
locallab: cleanup for retinex code

heckflosse:


Desmis:
@heckflosse 
Ingo

No problem, no differences with TIF

jacques

heckflosse:
@Desmis Jacques, thanks for testing :+1:

Issue #5285, opened 2019-04-16
Gradients in themes are not shown on the Mac


TooWaBoo:
It looks like gradients are not shown on the Mac.
@Benitoite 
Can you test this theme, please?
[TooWaBlue-Gradient-Test-GTK3-20_.css.txt](https://github.com/Beep6581/RawTherapee/files/3085872/TooWaBlue-Gradient-Test-GTK3-20_.css.txt)

TooWaBoo:
ping @Benitoite

Benitoite:
@TooWaBoo sorry about the delay.
Here's a screenshot of the Gradient-Test theme (Helvetica 11):
![Screen Shot 2019-04-18 at 8 40 35 AM](https://user-images.githubusercontent.com/17919239/56373394-aebb5c00-61b5-11e9-8804-551ce4fcfcd8.png)

TooWaBoo:
@Benitoite 
Thanks, is it font size 12? If yes, can you provide a screenshot of font size 11 or 10, please?
If it's not 12, provide a screenshot in font size 12, please.

Benitoite:
Sure,
here is Helvetica 12:
![Screen Shot 2019-04-18 at 9 01 26 AM](https://user-images.githubusercontent.com/17919239/56374660-926cee80-61b8-11e9-9a7d-d3c5155363f8.png)

TooWaBoo:
Thanks, some fancy styles behave weird on the Mac.
Buttons, scale slider, scale slider through, progressbar have a gradient and a box shadow.
At font size e.g. **11** the gradient is shown on all these objects except for the ton curve combined button but the box shadow is not shown.

At font size **12** the gradient is shown only on the scale slider but the box shadow is shown on all other objects.

As a workaround I've add a background color (blue) to the scale through. If the gradient is not shown it will display the background color.

TooWaBoo:
This issue is a bug in GTK+ on MacOS. My workaround makes the TWB-theme usable and it behaves like the RT-theme.

Shall we close it? @Benitoite , @Beep6581

PR #5288, opened 2019-04-18
TWB theme - Mac-gradient-workaround

TooWaBoo:
Some fancy styles behave weird on the Mac.
Buttons, scale slider, scale slider through, progressbar have a gradient and a box shadow.
At font size e.g. 11 the gradient is shown on all these objects except for the ton curve combined button but the box shadow is not shown.

At font size 12 the gradient is shown only on the scale slider but the box shadow is shown on all other objects.

As a workaround I've add a background color (blue) to the scale through. If the gradient is not shown it will display the background color.

see: **Gradients in themes are not shown on the Mac** https://github.com/Beep6581/RawTherapee/issues/5285

### DIFF
--- a/rtdata/themes/TooWaBlue-GTK3-20_.css
+++ b/rtdata/themes/TooWaBlue-GTK3-20_.css
@@ -2,7 +2,7 @@
   This file is part of RawTherapee.
 
   Copyright (c) 2016-2019 TooWaBoo
-  Version 3.06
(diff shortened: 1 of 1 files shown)


Beep6581:
@TooWaBoo is this for 5.6?

TooWaBoo:
Yes, it's for 5.6

PR #5293, opened 2019-04-21
GUI fixes for newlocallab

### LINKED ISSUE #5265: locallab keyboard shortcuts issue
Using shortcuts (e.g. z to zoom in) does not work because of this (whatever that is):
grafik

### LINKED ISSUE #5266: locallab: dialog to rename a spot has no default button
When renaming a locallab spot I would like the ok button to be the default button of the rename dialog, so I can just press enter to commit the new name.
grafik

### LINKED ISSUE #5286: newlocallab: Spot WB, crop and straighten from toolbar are broken
To reproduce:

1) Open an image
2) click Spot WB button in toolbar. Result: WB tool is opened but editor is still in hand-mode
grafik

Works fine in dev.

Pandagrapher:
@Desmis 

Bugfixes:
- Took into account Floessie review remark
- Fixed #5266 (default button for rename dialog)
- Fixed #5265 (deactivation of treeview search popup)
- Fixed #5286 (fixed switching to another tool with toolbar)
- Some code cleanup

Pierre


Desmis:
@Pandagrapher 

All seems to work well :)

Thank you
jacques

PR #5321, opened 2019-05-13
Clut scan speedup

heckflosse:
Processing time of the function which scans the clut folder structure when starting rt (as usual all measures are median of 7):

dev: 52.5 ms
1st commit: 24.0 ms
2nd commit: 23.0 ms
3rd commit: 22.6 ms


heckflosse:
Edit: The 30 ms speedup when starting rt doesnt sound much, but when added to the speedup in the adjuster_review branch it's already a 150 ms speedup when starting rt in setm mode.

Small speedups sum up...

Floessie (approved):
LGTM. :+1:

PR #5349, opened 2019-06-07
Film negative processing feature

rom9:
Color film negative inversion, implemented using reciprocals. Works on raw files only. Thumbnails not working yet. See related discussion here:
https://discuss.pixls.us/t/any-interest-in-a-film-negative-feature-in-rt/12569


rtengine/filmnegativeproc.cc
```diff
@@ -255,9 +255,10 @@ void RawImageSource::filmNegativeProcess(const procparams::FilmNegativeParams &p
 #endif
         for (int row = 0; row < H; row ++) {
             for (int col = 0; col < W; col++) {
-                int c  = FC(row, col);                        // three colors,  0=R, 1=G,  2=B
-                // Apply the multipliers
-                rawData[row][col] *= mults[c];
+                int c = FC(row, col);                        // three colors,  0=R, 1=G,  2=B
+                // Apply the multipliers, clamp max output value to 65535
+                float out = rawData[row][col] * mults[c];
+                rawData[row][col] = out > 65535.f ? 65535.f : out;
```
heckflosse: rawData[row][col] = rtengine::min(out, 65535.f);

heckflosse (commented):
I suggest this patch if you like:

```diff
diff --git a/AUTHORS.txt b/AUTHORS.txt
index 15e3e13d4..160869776 100644
--- a/AUTHORS.txt
+++ b/AUTHORS.txt
@@ -28,6 +28,7 @@ Development contributors, in last name alphabetical order:
   Adam Reichold
   Philip Rinn
   Jan Rinze
+  Alberto Romei
   Ben S.
   Andrey Skvortsov
   Fabio Suprani

```

heckflosse:
Another one:

```diff
diff --git a/rtengine/filmnegativeproc.cc b/rtengine/filmnegativeproc.cc
index bd68314e3..f5a231c05 100644
--- a/rtengine/filmnegativeproc.cc
+++ b/rtengine/filmnegativeproc.cc
@@ -172,10 +172,12 @@ void rtengine::RawImageSource::filmNegativeProcess(const procparams::FilmNegativ
         return;
     }
 
+    // Exponents are expressed as positive in the parameters, so negate them in order
+    // to get the reciprocals.
     const std::array<float, 3> exps = {
-        static_cast<float>(params.redExp),
-        static_cast<float>(params.greenExp),
-        static_cast<float>(params.blueExp)
+        static_cast<float>(-params.redExp),
+        static_cast<float>(-params.greenExp),
+        static_cast<float>(-params.blueExp)
     };
     
     MyTime t1, t2, t3,t4, t5;
@@ -244,8 +246,8 @@ void rtengine::RawImageSource::filmNegativeProcess(const procparams::FilmNegativ
         // Find median values for each channel
         if (!cvs[c].empty()) {
             findMinMaxPercentile(cvs[c].data(), cvs[c].size(), 0.5f, medians[c], 0.5f, medians[c], true);
-            medians[c] = pow_F(rtengine::max(medians[c], 1.f), -exps[c]);
-            // Determine the channel multipler so that N times the median becomes 65k. This clips away
+            medians[c] = pow_F(rtengine::max(medians[c], 1.f), exps[c]);
+            // Determine the channel multiplier so that N times the median becomes 65k. This clips away
             // the values in the dark border surrounding the negative (due to the film holder, for example),
             // the reciprocal of which have blown up to stellar values.
             mults[c] = MAX_OUT_VALUE / (medians[c] * 24.f);
@@ -274,10 +276,9 @@ void rtengine::RawImageSource::filmNegativeProcess(const procparams::FilmNegativ
 #endif
         for (int row = 0; row < H; ++row) {
             int col = 0;
-            // Exponents are expressed as positive in the parameters, so negate them in order
-            // to get the reciprocals. Avoid trouble with zeroes, minimum pixel value is 1.
-            const float exps0 = -exps[FC(row, col)];
-            const float exps1 = -exps[FC(row, col + 1)];
+            // Avoid trouble with zeroes, minimum pixel value is 1.
+            const float exps0 = exps[FC(row, col)];
+            const float exps1 = exps[FC(row, col + 1)];
             const float mult0 = mults[FC(row, col)];
             const float mult1 = mults[FC(row, col + 1)];
 #ifdef __SSE2__
@@ -306,15 +307,14 @@ void rtengine::RawImageSource::filmNegativeProcess(const procparams::FilmNegativ
 #endif
         for (int row = 0; row < H; row ++) {
             int col = 0;
-            // Exponents are expressed as positive in the parameters, so negate them in order
-            // to get the reciprocals. Avoid trouble with zeroes, minimum pixel value is 1.
+            // Avoid trouble with zeroes, minimum pixel value is 1.
             const std::array<float, 6> expsc = {
-                -exps[ri->XTRANSFC(row, 0)],
-                -exps[ri->XTRANSFC(row, 1)],
-                -exps[ri->XTRANSFC(row, 2)],
-                -exps[ri->XTRANSFC(row, 3)],
-                -exps[ri->XTRANSFC(row, 4)],
-                -exps[ri->XTRANSFC(row, 5)]
+                exps[ri->XTRANSFC(row, 0)],
+                exps[ri->XTRANSFC(row, 1)],
+                exps[ri->XTRANSFC(row, 2)],
+                exps[ri->XTRANSFC(row, 3)],
+                exps[ri->XTRANSFC(row, 4)],
+                exps[ri->XTRANSFC(row, 5)]
             };
             const std::array<float, 6> multsc = {
                 mults[ri->XTRANSFC(row, 0)],

```

rom9:
Thanks ;-)

heckflosse:
@rom9 @Beep6581 @Floessie 

How do we handle the thumbnail stuff?
Personally, I don't mind to see the negatives in thumbnail view, though an indicator that filmnegative is enabled would be nice. What's your opinion?

rom9:
I've implemented the negative processing in rtthumbnail.cc, but the channel multipliers that are applied to the thumbnail are not the same of the main image. So the thumbnail has a strong tint. I could ignore the WB params completely and just show the thumbnail with an auto-WB based on channel medians. But this would not reflect the balance chosen by the user.
Right now i'm trying to track down where in the load->scale->preprocess->demosaic pipeline, the raw values of the main image start to diverge from the values of the thumbnail.
If we can't find it, we can give up and go with the auto-wb solution if it's ok for you...

heckflosse:
It would even be ok for me to still see the unprocessed negative (maybe with a small icon that shows filmnegative is enabled) in thumbnail view. But that's my opinion. Other contributors may differ.

Edit: I really like to see the negative in thumb view as it reminds me on old times :)

rom9:
I finally found the problem. It is caused by a difference in the call to RawImage::get_colorsCoeff, between the thumbnail and the main image processing.
This is the main image proc, rawimagesource.cc line 1644:
```
// ...then we re-get the constants but now with auto which gives us better demosaicing and CA auto-correct
// performance for strange white balance settings (such as UniWB)
ri->get_colorsCoeff( ref_pre_mul, scale_mul, c_black, true);
```
and this is rtthumbnail.cc line 570:
```
    ri->get_colorsCoeff (pre_mul, scale_mul, cblack, false);
```
last bool param is `forceAutoWB`: rawimagesource always applies a pre-AWB before preprocessing/demosaic, and then applies the user-specified WB.
Rtthumbnail, instead, uses the native multipliers and goes straight to final WB.
If i just change that last bool param from `false` to `true` in rtthumbnail, everything looks perfect. Problem is, that call is in `Thumbnail::loadFromRaw`, so i don't have the ProcParams instance and i can't check if filmNeg is enabled, and flip that bool accordingly.
So i guess i'll call it again inside Thumbnail::processImage, and rescale values before filmNeg processing.
...but now i'm definitely going to sleep :-D

rom9:
Here's a working solution. I could not call get_colorsCoeff again because after loadFromRaw the RawImage instance is destroyed, so i had to calculate channel averages in processing, and adapt the multipliers as if they were computed on a pre-AWB'ed image (as happens in RawImageSource).
This is a bit inefficient but seems to replicate the exact colors. Maybe the global gain is a bit low in some cases.
@heckflosse : i did not implement any SSE2 or other optimizations, does it make sense on a ~180px wide image? The whole film neg processing block takes about 1 to 5 ms per thumbnail on my machine.
@Floessie : Should i move that block in a separate `filmnegative_tumbnail.cc` compilation unit, or simply add a separate private method inside rtthumbnail.cc ?

Let me know what you think :-)

heckflosse:
@rom9 With my latest merge to dev, there is a merge conflict for your pr. My apologies for that :frowning_face: 

As I can not push to your fork I can not solve the conflict, but I can post a patch in case you want and didn't solve it already.

rom9:
Merged, everything looks fine ;-)

heckflosse:
@Floessie @Beep6581  @Hombre57 @Desmis @Thanatomanic  @agriggio 
I'm fine with merge. Your opinions?

Hombre57:
@heckflosse I'll have a look tonight.

Hombre57:
@rom9 @heckflosse From a GUI p.o.v., this is nicely done :ok_hand: . Can't say for the algorithm itself.

I'm fine for merging too.

rom9:
> From a GUI p.o.v., this is nicely done. Can't say for the algorithm itself.

Are you referring to the maths used to do the conversion, or to how it is implemented in code (like code style, etc.) ?

Regarding the GUI, actually i had a new idea that i was testing right now: i'd like to change the way the sliders work, so that the red and blue sliders do not express actual exponents, but instead the *ratio* between the red (or blue) exponent and the reference (green) one. So we would have these 3 sliders in sequence:
`Master exponent (contrast)`  < -- this is the actual green exponent
`Red ratio`    < -- ratio redExp / greenExp
`Blue ratio`    < -- ratio blueExp / greenExp
This way, when the user adjusts the master slider, the other sliders *do not move*, while red and blue exponents are still scaled accordingly. I think this is more intuitive to use, and is especially useful when the user clicks the "reset to default" button on the red or blue slider: the visible value would go to the usual default, but the actual exponent would still be scaled properly, based on master exp.
After using it a bit, i think this is nicer to use. What do you think?
Thanks! :-)

Hombre57:
@rom9 I'm referring to the math in rtengine.

Having red and blue ratio seems to be a good idea, but be careful with the slider's range/precision. I mean, the set of values that will be available for the red and blue exponent will depend on the green exponent.

[PS] I'm not saying that the red and blue slider must have a variable range, but must be sufficiently precise to fine tune the exponents.

rom9:
@Hombre57 actually, i don't know much about maths or physics (i just took the formula verbatim from that sentence in the Wikipedia article), so it's very likely that i'm missing something :-D
What is wrong? Do you mean the fact that i don't take into consideration the different behaviour of the film at the extremes of the film exposure range?
Regarding the ratio sliders, i lowered the exponent max from 6 to 4 (it was too large anyway), and the precision for the all sliders is .001 so, in the worst case we can set the red/blue exponents with a precision of .004 which doesn't seem noticeable.
The range should not be a problem: the worst film that i've tried until now (which has a very strong blue/purple tint), looks good with blueExp = 1.8 * greenExp. Ratio sliders range from 0.3 to 3 so that should be enough.
Just pushed the change, any feedback is appreciated :-)

rtengine/rtthumbnail.cc
```diff
@@ -1229,12 +1229,33 @@ IImage8* Thumbnail::processImage (const procparams::ProcParams& params, eSensorT
...
+            float *bline = baseImg->b(i);
+            int j = 0;
+#ifdef __SSE2__
+            for (; j < rwidth - 3; j +=4) {
+                STVFU(rline[j], vminf(rmultv * pow_F(LVFU(rline[j]), rexpv), clipv));
+                STVFU(gline[j], vminf(gmultv * pow_F(LVFU(gline[j]), gexpv), clipv));
+                STVFU(bline[j], vminf(bmultv * pow_F(LVFU(bline[j]), bexpv), clipv));
+            }
+#endif
+            for (; j < rwidth; ++j) {
```
heckflosse: use pow_F instead of powf here as well

Hombre57:
@rom9 

> What is wrong?

Nothing, and never said that.

rtengine/rtthumbnail.cc
rom9: Done.

heckflosse:
Works fine. I especially like that changing the exponents in batch edit can be used to get close to good exponents in short time, as the thumb processing is very fast :+1:

rom9:
@Hombre57
>  and never said that.

Yes, sorry, that's true actually. But what do you like less, or what do you think is less "nicely done", then? i'll try to make it better if i can :-)
Don't know if this can help, but in my last commit, i made things a bit cleaner in `channelsAvg`, and more importantly, i was able to re-introduce the clipping to 65535 without losing the ability to pick spots from clipped areas (since i now read samples from the original RawImage data).
Calculated exponents are the same as before, so i think it's better this way.

heckflosse:
What's the reason, the exponent defaults in procparams and gui are different?

heckflosse:
Ok, I think the reason is the log scale of the adjuster. Could you add a comment in code?

rom9:
> What's the reason, the exponent defaults in procparams and gui are different?

The processing profile contains the actual R,G,B exponents, while the sliders in the GUI express the master exponent (Green) and the ratios of R and B exponents to the master (just for better usability).

rom9:
> Ok, I think the reason is the log scale of the adjuster. Could you add a comment in code?

No, the adjuster shows the true value anyway, the log-ness only affects the handle position.

heckflosse:
Ok, then I was wrong with the log scale. But please add a comment in code about the defaults being different between gui adjuster value and procparams

rom9:
i used expressions as the slider defaults, to hint at the same default exponents found in procparams. Is it clearer?

heckflosse:
@rom9 :+1: 

I vote for merge this into dev to get more tests and feedback. It's really very well usable atm and if changes have to be made, we are still in dev, not in release. Any objections?

rom9:
@heckflosse still working on extracting thumbnail processing to own compilation unit. I'm almost finished, expect to push that change too in a few minutes...

rom9:
Done. Please pardon my C++  :-D

Hombre57:
@heckflosse @rom9 I don't really like setting different parameters in the GUi vs the ProcParams. Why not having the same behavior ?

heckflosse:
@Hombre57 
> I don't really like setting different parameters in the GUi vs the ProcParams. Why not having the same behavior ?

I think you are right with this, though still thinking about it... (with a 90% tendency to agree with you) :+1: 
Let me sleep about that.

rom9:
@Hombre57 initally, the sliders showed the plain exponents. That way, though, slider values are less intuitive to set manually. Then i made the green slider the "lead" and the other two followed automatically as the user changed the green. This was better but it doesn't play well with the "reset to default" buttons: if i move the lead slider to increase contrast for example, then i change the red slider to alter colors, and finally i change my mind and want to go back to the default red/green ratio, hitting the reset button on the red slider sets the _absolute_ default, which is the value it should be when the green slider is in the default position. Instead, with ratios, slider are fixed while exponents follow naturally.
On the other hand, using ratios in ProcParams would make little sense as they are there just for a GUI usability issue, and are never used by the rtengine calculations.

heckflosse:
I agree with @Hombre57 . Values in pp3 file should be the values in ui and values in ui should be consistent, means the values shown in history message should be the values of the adjusters.

Here's a patch, you can also download it from [here](https://filebin.net/81hbei3q9ue5qyba)

```diff
diff --git a/rtdata/languages/Catala b/rtdata/languages/Catala
index f9bcf25b2..a6b104be2 100644
--- a/rtdata/languages/Catala
+++ b/rtdata/languages/Catala
@@ -1315,7 +1315,7 @@ ZOOMPANEL_ZOOMOUT;Allunya\nDrecera: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1853,12 +1853,12 @@ ZOOMPANEL_ZOOMOUT;Allunya\nDrecera: <b>-</b>
 !TP_EXPOSURE_TCMODE_PERCEPTUAL;Perceptual
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/Chinese (Simplified) b/rtdata/languages/Chinese (Simplified)
index 78807ddd4..9f0e4b365 100644
--- a/rtdata/languages/Chinese (Simplified)	
+++ b/rtdata/languages/Chinese (Simplified)	
@@ -1362,7 +1362,7 @@ ZOOMPANEL_ZOOMOUT;缩放拉远\n快捷键: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1800,12 +1800,12 @@ ZOOMPANEL_ZOOMOUT;缩放拉远\n快捷键: <b>-</b>
 !TP_EXPOSURE_TCMODE_PERCEPTUAL;Perceptual
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/Deutsch b/rtdata/languages/Deutsch
index fd2d0d07a..e3aa6ea44 100644
--- a/rtdata/languages/Deutsch
+++ b/rtdata/languages/Deutsch
@@ -2365,11 +2365,11 @@ ZOOMPANEL_ZOOMOUT;Herauszoomen\nTaste: <b>-</b>
 !!!!!!!!!!!!!!!!!!!!!!!!!
 
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
diff --git a/rtdata/languages/English (UK) b/rtdata/languages/English (UK)
index 24bffbb0a..677dc1818 100644
--- a/rtdata/languages/English (UK)	
+++ b/rtdata/languages/English (UK)	
@@ -839,7 +839,7 @@ TP_WBALANCE_EQBLUERED_TOOLTIP;Allows to deviate from the normal behaviour of "wh
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1667,12 +1667,12 @@ TP_WBALANCE_EQBLUERED_TOOLTIP;Allows to deviate from the normal behaviour of "wh
 !TP_EXPOSURE_TCMODE_WEIGHTEDSTD;Weighted Standard
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/English (US) b/rtdata/languages/English (US)
index 5dad9591a..7d2c5ff43 100644
--- a/rtdata/languages/English (US)	
+++ b/rtdata/languages/English (US)	
@@ -749,7 +749,7 @@
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1638,12 +1638,12 @@
 !TP_EXPOSURE_TCMODE_WEIGHTEDSTD;Weighted Standard
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/Italiano b/rtdata/languages/Italiano
index 98d86b69d..2390c0d81 100644
--- a/rtdata/languages/Italiano
+++ b/rtdata/languages/Italiano
@@ -1540,7 +1540,7 @@ ZOOMPANEL_ZOOMOUT;Rimpicciolisci.\nScorciatoia: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1905,12 +1905,12 @@ ZOOMPANEL_ZOOMOUT;Rimpicciolisci.\nScorciatoia: <b>-</b>
 !TP_EXPOSURE_TCMODE_PERCEPTUAL;Perceptual
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/Magyar b/rtdata/languages/Magyar
index 7aa5b30c3..c3272133a 100644
--- a/rtdata/languages/Magyar
+++ b/rtdata/languages/Magyar
@@ -1249,7 +1249,7 @@ ZOOMPANEL_ZOOMOUT;Kicsinyítés <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1818,12 +1818,12 @@ ZOOMPANEL_ZOOMOUT;Kicsinyítés <b>-</b>
 !TP_EXPOSURE_TCMODE_WEIGHTEDSTD;Weighted Standard
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/Nederlands b/rtdata/languages/Nederlands
index 508f06384..179ba9b3f 100644
--- a/rtdata/languages/Nederlands
+++ b/rtdata/languages/Nederlands
@@ -2037,7 +2037,7 @@ ZOOMPANEL_ZOOMOUT;Zoom uit\nSneltoets: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -2232,12 +2232,12 @@ ZOOMPANEL_ZOOMOUT;Zoom uit\nSneltoets: <b>-</b>
 !TP_EXPOSURE_CLAMPOOG;Clip out-of-gamut colors
 !TP_EXPOSURE_HISTMATCHING;Auto-Matched Tone Curve
 !TP_EXPOSURE_HISTMATCHING_TOOLTIP;Automatically adjust sliders and curves (except exposure compensation) to match the look of the embedded JPEG thumbnail.
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_ICM_WORKING_TRC;Tone response curve:
 !TP_ICM_WORKING_TRC_CUSTOM;Custom
diff --git a/rtdata/languages/Polish b/rtdata/languages/Polish
index e35f583ff..daca63dd6 100644
--- a/rtdata/languages/Polish
+++ b/rtdata/languages/Polish
@@ -1622,7 +1622,7 @@ ZOOMPANEL_ZOOMOUT;Oddal\nSkrót: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1931,12 +1931,12 @@ ZOOMPANEL_ZOOMOUT;Oddal\nSkrót: <b>-</b>
 !TP_EXPOSURE_HISTMATCHING_TOOLTIP;Automatically adjust sliders and curves (except exposure compensation) to match the look of the embedded JPEG thumbnail.
 !TP_EXPOSURE_TCMODE_LUMINANCE;Luminance
 !TP_EXPOSURE_TCMODE_PERCEPTUAL;Perceptual
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
 !TP_ICM_APPLYBASELINEEXPOSUREOFFSET;Baseline exposure
diff --git a/rtdata/languages/Portugues (Brasil) b/rtdata/languages/Portugues (Brasil)
index 0bafe632f..0a9f4fe89 100644
--- a/rtdata/languages/Portugues (Brasil)	
+++ b/rtdata/languages/Portugues (Brasil)	
@@ -2259,7 +2259,7 @@ ZOOMPANEL_ZOOMOUT;Menos Zoom\nAtalho: <b>-</b>
 !HISTORY_MSG_COLORTONING_LABREGION_OFFSET;CT - region offset
 !HISTORY_MSG_COLORTONING_LABREGION_POWER;CT - region power
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !MAIN_FRAME_PLACES_DEL;Remove
 !PREFERENCES_APPEARANCE_PSEUDOHIDPI;Pseudo-HiDPI mode
 !PREFERENCES_CACHECLEAR_SAFETY;Only files in the cache are cleared. Processing profiles stored alongside the source images are not touched.
@@ -2273,12 +2273,12 @@ ZOOMPANEL_ZOOMOUT;Menos Zoom\nAtalho: <b>-</b>
 !TP_COLORTONING_LABREGION_OFFSET;Offset
 !TP_COLORTONING_LABREGION_POWER;Power
 !TP_CROP_PPI;PPI
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_LENSPROFILE_CORRECTION_AUTOMATCH;Automatically selected
 !TP_LENSPROFILE_CORRECTION_MANUAL;Manually selected
diff --git a/rtdata/languages/Russian b/rtdata/languages/Russian
index 9e43f1098..dd4c501d7 100644
--- a/rtdata/languages/Russian
+++ b/rtdata/languages/Russian
@@ -1707,7 +1707,7 @@ ZOOMPANEL_ZOOMOUT;Отдалить\nГорячая клавиша: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
 !HISTORY_MSG_ICM_OUTPUT_TYPE;Output - Type
@@ -1976,12 +1976,12 @@ ZOOMPANEL_ZOOMOUT;Отдалить\nГорячая клавиша: <b>-</b>
 !TP_DISTORTION_AUTO_TIP;Automatically corrects lens distortion in raw files by matching it against the embedded JPEG image if one exists and has had its lens disortion auto-corrected by the camera.
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
 !TP_FLATFIELD_CLIPCONTROL;Clip control
diff --git a/rtdata/languages/Serbian (Cyrilic Characters) b/rtdata/languages/Serbian (Cyrilic Characters)
index ce1452821..966213d1d 100644
--- a/rtdata/languages/Serbian (Cyrilic Characters)	
+++ b/rtdata/languages/Serbian (Cyrilic Characters)	
@@ -1515,7 +1515,7 @@ ZOOMPANEL_ZOOMOUT;Умањује приказ слике <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -1904,12 +1904,12 @@ ZOOMPANEL_ZOOMOUT;Умањује приказ слике <b>-</b>
 !TP_EXPOSURE_TCMODE_PERCEPTUAL;Perceptual
 !TP_EXPOS_BLACKPOINT_LABEL;Raw Black Points
 !TP_EXPOS_WHITEPOINT_LABEL;Raw White Points
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FILMSIMULATION_LABEL;Film Simulation
 !TP_FILMSIMULATION_SLOWPARSEDIR;RawTherapee is configured to look for Hald CLUT images, which are used for the Film Simulation tool, in a folder which is taking too long to load.\nGo to Preferences > Image Processing > Film Simulation\nto see which folder is being used. You should either point RawTherapee to a folder which contains only Hald CLUT images and nothing more, or to an empty folder if you don't want to use the Film Simulation tool.\n\nRead the Film Simulation article in RawPedia for more information.\n\nDo you want to cancel the scan now?
diff --git a/rtdata/languages/Swedish b/rtdata/languages/Swedish
index cd3b9b3fc..4fb3e301b 100644
--- a/rtdata/languages/Swedish
+++ b/rtdata/languages/Swedish
@@ -1864,7 +1864,7 @@ ZOOMPANEL_ZOOMOUT;Förminska.\nKortkommando: <b>-</b>
 !HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 !HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 !HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-!HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+!HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 !HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 !HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 !HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
@@ -2119,12 +2119,12 @@ ZOOMPANEL_ZOOMOUT;Förminska.\nKortkommando: <b>-</b>
 !TP_EXPOSURE_CLAMPOOG;Clip out-of-gamut colors
 !TP_EXPOSURE_HISTMATCHING;Auto-Matched Tone Curve
 !TP_EXPOSURE_HISTMATCHING_TOOLTIP;Automatically adjust sliders and curves (except exposure compensation) to match the look of the embedded JPEG thumbnail.
-!TP_FILMNEGATIVE_BLUE;Blue exponent
+!TP_FILMNEGATIVE_BLUE;Blue ratio
 !TP_FILMNEGATIVE_GREEN;Green exponent
 !TP_FILMNEGATIVE_GUESS_TOOLTIP;Calculate exponents by picking 2 neutral reference spots in the image; one white (light gray) and one black (dark gray).\nThe order does not matter. The exponents will be updated after the second spot is picked.
 !TP_FILMNEGATIVE_LABEL;Film Negative
 !TP_FILMNEGATIVE_PICK;Pick white and black spots
-!TP_FILMNEGATIVE_RED;Red exponent
+!TP_FILMNEGATIVE_RED;Red ratio
 !TP_FILMNEGATIVE_REF_SPOTS;Film negative reference spots
 !TP_FLATFIELD_CLIPCONTROL;Clip control
 !TP_FLATFIELD_CLIPCONTROL_TOOLTIP;Clip control avoids clipped highlights caused by applying the flat field. If there are already clipped highlights before applying the flat field, clip control can lead to color cast.
diff --git a/rtdata/languages/default b/rtdata/languages/default
index 63c877bbf..d5e3b13fa 100644
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -748,7 +748,7 @@ HISTORY_MSG_DEHAZE_STRENGTH;Dehaze - Strength
 HISTORY_MSG_DUALDEMOSAIC_AUTO_CONTRAST;Dual demosaic - Auto threshold
 HISTORY_MSG_DUALDEMOSAIC_CONTRAST;Dual demosaic - Contrast threshold
 HISTORY_MSG_FILMNEGATIVE_ENABLED;Film Negative
-HISTORY_MSG_FILMNEGATIVE_EXPONENTS;Film negative exponents changed
+HISTORY_MSG_FILMNEGATIVE_VALUES;Film negative values
 HISTORY_MSG_HISTMATCHING;Auto-matched tone curve
 HISTORY_MSG_ICM_OUTPUT_PRIMARIES;Output - Primaries
 HISTORY_MSG_ICM_OUTPUT_TEMP;Output - ICC-v4 illuminant D
diff --git a/rtengine/filmnegativeproc.cc b/rtengine/filmnegativeproc.cc
index db5705a3c..58d87d103 100644
--- a/rtengine/filmnegativeproc.cc
+++ b/rtengine/filmnegativeproc.cc
@@ -101,9 +101,9 @@ bool channelsAvg(
 bool rtengine::RawImageSource::getFilmNegativeExponents(Coord2D spotA, Coord2D spotB, int tran, const FilmNegativeParams &currentParams, std::array<float, 3>& newExps)
 {
     newExps = {
-        static_cast<float>(currentParams.redExp),
+        static_cast<float>(currentParams.redRatio * currentParams.greenExp),
         static_cast<float>(currentParams.greenExp),
-        static_cast<float>(currentParams.blueExp)
+        static_cast<float>(currentParams.blueRatio * currentParams.greenExp)
     };
 
     constexpr int spotSize = 32; // TODO: Make this configurable?
@@ -171,9 +171,9 @@ void rtengine::RawImageSource::filmNegativeProcess(const procparams::FilmNegativ
     // Exponents are expressed as positive in the parameters, so negate them in order
     // to get the reciprocals.
     const std::array<float, 3> exps = {
-        static_cast<float>(-params.redExp),
+        static_cast<float>(-params.redRatio * params.greenExp),
         static_cast<float>(-params.greenExp),
-        static_cast<float>(-params.blueExp)
+        static_cast<float>(-params.blueRatio * params.greenExp)
     };
     
     MyTime t1, t2, t3,t4, t5;
diff --git a/rtengine/filmnegativethumb.cc b/rtengine/filmnegativethumb.cc
index 8889915e6..51ab08feb 100644
--- a/rtengine/filmnegativethumb.cc
+++ b/rtengine/filmnegativethumb.cc
@@ -17,22 +17,13 @@
  *  along with RawTherapee.  If not, see <http://www.gnu.org/licenses/>.
  */
 #include <cmath>
-#include <iostream>
-
-#ifdef _OPENMP
-#include <omp.h>
-#endif
 
 #include "rtengine.h"
 #include "rtthumbnail.h"
-
-#include "mytime.h"
 #include "opthelper.h"
-#include "procparams.h"
 #include "rt_algo.h"
 #include "rtengine.h"
 #include "settings.h"
-#include "procparams.h"
 #define BENCHMARK
 #include "StopWatch.h"
 
@@ -44,7 +35,9 @@ extern const Settings* settings;
 }
 
 void rtengine::Thumbnail::processFilmNegative(
-    const procparams::ProcParams &params,
+    const double redRatio,
+    const double greenExp,
+    const double blueRatio,
     const Imagefloat* baseImg,
     const int rwidth, const int rheight,
     float &rmi, float &gmi, float &bmi
@@ -52,9 +45,9 @@ void rtengine::Thumbnail::processFilmNegative(
     StopWatch stop1("Thumbnail film negative", true);
 
     // Channel exponents
-    const float rexp = -params.filmNegative.redExp;
-    const float gexp = -params.filmNegative.greenExp;
-    const float bexp = -params.filmNegative.blueExp;
+    const float rexp = -redRatio * greenExp;
+    const float gexp = -greenExp;
+    const float bexp = -blueRatio * greenExp;
 
     // Need to calculate channel averages, to fake the same conditions
     // found in rawimagesource, where get_ColorsCoeff is called with
diff --git a/rtengine/procparams.cc b/rtengine/procparams.cc
index b35faca94..45c71b03a 100644
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -2736,9 +2736,9 @@ bool MetaDataParams::operator!=(const MetaDataParams &other) const
 
 FilmNegativeParams::FilmNegativeParams() :
     enabled(false),
-    redExp(2.04),
+    redRatio(1.36),
     greenExp(1.5),
-    blueExp(1.29)
+    blueRatio(0.86)
 {
 }
 
@@ -2746,9 +2746,9 @@ bool FilmNegativeParams::operator ==(const FilmNegativeParams& other) const
 {
     return
         enabled == other.enabled
-        && redExp   == other.redExp
+        && redRatio   == other.redRatio
         && greenExp == other.greenExp
-        && blueExp  == other.blueExp;
+        && blueRatio  == other.blueRatio;
 }
 
 bool FilmNegativeParams::operator !=(const FilmNegativeParams& other) const
@@ -3589,9 +3589,9 @@ int ProcParams::save(const Glib::ustring& fname, const Glib::ustring& fname2, bo
 
 // Film negative
         saveToKeyfile(!pedited || pedited->filmNegative.enabled, "Film Negative", "Enabled", filmNegative.enabled, keyFile);
-        saveToKeyfile(!pedited || pedited->filmNegative.redExp, "Film Negative", "RedExponent", filmNegative.redExp, keyFile);
+        saveToKeyfile(!pedited || pedited->filmNegative.redRatio, "Film Negative", "RedRatio", filmNegative.redRatio, keyFile);
         saveToKeyfile(!pedited || pedited->filmNegative.greenExp, "Film Negative", "GreenExponent", filmNegative.greenExp, keyFile);
-        saveToKeyfile(!pedited || pedited->filmNegative.blueExp, "Film Negative", "BlueExponent", filmNegative.blueExp, keyFile);
+        saveToKeyfile(!pedited || pedited->filmNegative.blueRatio, "Film Negative", "BlueRatio", filmNegative.blueRatio, keyFile);
 
 // EXIF change list
         if (!pedited || pedited->exif) {
@@ -5138,9 +5138,9 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
 
         if (keyFile.has_group("Film Negative")) {
             assignFromKeyfile(keyFile, "Film Negative", "Enabled", pedited, filmNegative.enabled, pedited->filmNegative.enabled);
-            assignFromKeyfile(keyFile, "Film Negative", "RedExponent", pedited, filmNegative.redExp, pedited->filmNegative.redExp);
+            assignFromKeyfile(keyFile, "Film Negative", "RedRatio", pedited, filmNegative.redRatio, pedited->filmNegative.redRatio);
             assignFromKeyfile(keyFile, "Film Negative", "GreenExponent", pedited, filmNegative.greenExp, pedited->filmNegative.greenExp);
-            assignFromKeyfile(keyFile, "Film Negative", "BlueExponent", pedited, filmNegative.blueExp, pedited->filmNegative.blueExp);
+            assignFromKeyfile(keyFile, "Film Negative", "BlueRatio", pedited, filmNegative.blueRatio, pedited->filmNegative.blueRatio);
         }
 
         if (keyFile.has_group("MetaData")) {
diff --git a/rtengine/procparams.h b/rtengine/procparams.h
index 2af018035..46eb530bc 100644
--- a/rtengine/procparams.h
+++ b/rtengine/procparams.h
@@ -1501,9 +1501,9 @@ struct RAWParams {
   */
 struct FilmNegativeParams {
     bool enabled;
-    double redExp;
+    double redRatio;
     double greenExp;
-    double blueExp;
+    double blueRatio;
 
     FilmNegativeParams();
 
diff --git a/rtengine/rtthumbnail.cc b/rtengine/rtthumbnail.cc
index 4b0baa685..1f3f29850 100644
--- a/rtengine/rtthumbnail.cc
+++ b/rtengine/rtthumbnail.cc
@@ -1168,7 +1168,7 @@ IImage8* Thumbnail::processImage (const procparams::ProcParams& params, eSensorT
     Imagefloat* baseImg = resizeTo<Imagefloat> (rwidth, rheight, interp, thumbImg);
 
     if(isRaw && params.filmNegative.enabled) {
-        processFilmNegative(params, baseImg, rwidth, rheight, rmi, gmi, bmi);
+        processFilmNegative(params.filmNegative.redRatio, params.filmNegative.greenExp, params.filmNegative.blueRatio, baseImg, rwidth, rheight, rmi, gmi, bmi);
     }
 
     if (params.coarse.rotate) {
diff --git a/rtengine/rtthumbnail.h b/rtengine/rtthumbnail.h
index 09a0510eb..9ebda1479 100644
--- a/rtengine/rtthumbnail.h
+++ b/rtengine/rtthumbnail.h
@@ -68,7 +68,7 @@ class Thumbnail
     bool gammaCorrected;
     double colorMatrix[3][3];
 
-    void processFilmNegative(const procparams::ProcParams& params, const Imagefloat* baseImg, int rwidth, int rheight, float &rmi, float &gmi, float &bmi);
+    void processFilmNegative(const double redRatio, const double greenExp, const double blueRatio, const Imagefloat* baseImg, int rwidth, int rheight, float &rmi, float &gmi, float &bmi);
 
 public:
 
diff --git a/rtgui/filmnegative.cc b/rtgui/filmnegative.cc
index 4956968e3..598c9463a 100644
--- a/rtgui/filmnegative.cc
+++ b/rtgui/filmnegative.cc
@@ -49,7 +49,7 @@ Adjuster* createExponentAdjuster(AdjusterListener* listener, const Glib::ustring
 FilmNegative::FilmNegative() :
     FoldableToolPanel(this, "filmnegative", M("TP_FILMNEGATIVE_LABEL"), false, true),
     EditSubscriber(ET_OBJECTS),
-    evFilmNegativeExponents(ProcEventMapper::getInstance()->newEvent(FIRST, "HISTORY_MSG_FILMNEGATIVE_EXPONENTS")),
+    evFilmNegativeExponents(ProcEventMapper::getInstance()->newEvent(FIRST, "HISTORY_MSG_FILMNEGATIVE_VALUES")),
     evFilmNegativeEnabled(ProcEventMapper::getInstance()->newEvent(FIRST, "HISTORY_MSG_FILMNEGATIVE_ENABLED")),
     fnp(nullptr),
     greenExp(createExponentAdjuster(this, M("TP_FILMNEGATIVE_GREEN"), 0.3, 4, 1.5)),  // master exponent (green channel)
@@ -123,45 +123,45 @@ void FilmNegative::read(const rtengine::procparams::ProcParams* pp, const Params
     disableListener();
 
     if (pedited) {
-        redRatio->setEditedState(pedited->filmNegative.redExp ? Edited : UnEdited);
+        redRatio->setEditedState(pedited->filmNegative.redRatio ? Edited : UnEdited);
         greenExp->setEditedState(pedited->filmNegative.greenExp ? Edited : UnEdited);
-        blueRatio->setEditedState(pedited->filmNegative.blueExp ? Edited : UnEdited);
+        blueRatio->setEditedState(pedited->filmNegative.blueRatio ? Edited : UnEdited);
         set_inconsistent(multiImage && !pedited->filmNegative.enabled);
     }
 
     setEnabled(pp->filmNegative.enabled);
-    redRatio->setValue(pp->filmNegative.redExp / pp->filmNegative.greenExp);
+    redRatio->setValue(pp->filmNegative.redRatio);
     greenExp->setValue(pp->filmNegative.greenExp);
-    blueRatio->setValue(pp->filmNegative.blueExp / pp->filmNegative.greenExp);
+    blueRatio->setValue(pp->filmNegative.blueRatio);
 
     enableListener();
 }
 
 void FilmNegative::write(rtengine::procparams::ProcParams* pp, ParamsEdited* pedited)
 {
-    pp->filmNegative.redExp = greenExp->getValue() * redRatio->getValue();
+    pp->filmNegative.redRatio = redRatio->getValue();
     pp->filmNegative.greenExp = greenExp->getValue();
-    pp->filmNegative.blueExp = greenExp->getValue() * blueRatio->getValue();
+    pp->filmNegative.blueRatio = blueRatio->getValue();
     pp->filmNegative.enabled = getEnabled();
 
     if (pedited) {
-        pedited->filmNegative.redExp = greenExp->getEditedState() || redRatio->getEditedState();
+        pedited->filmNegative.redRatio = redRatio->getEditedState();
         pedited->filmNegative.greenExp = greenExp->getEditedState();
-        pedited->filmNegative.blueExp = greenExp->getEditedState() || blueRatio->getEditedState();
+        pedited->filmNegative.blueRatio = blueRatio->getEditedState();
         pedited->filmNegative.enabled = !get_inconsistent();
     }
 }
 
 void FilmNegative::setDefaults(const rtengine::procparams::ProcParams* defParams, const ParamsEdited* pedited)
 {
-    redRatio->setValue(defParams->filmNegative.redExp / defParams->filmNegative.greenExp);
+    redRatio->setValue(defParams->filmNegative.redRatio);
     greenExp->setValue(defParams->filmNegative.greenExp);
-    blueRatio->setValue(defParams->filmNegative.blueExp / defParams->filmNegative.greenExp);
+    blueRatio->setValue(defParams->filmNegative.blueRatio);
 
     if (pedited) {
-        redRatio->setDefaultEditedState(pedited->filmNegative.redExp ? Edited : UnEdited);
+        redRatio->setDefaultEditedState(pedited->filmNegative.redRatio ? Edited : UnEdited);
         greenExp->setDefaultEditedState(pedited->filmNegative.greenExp ? Edited : UnEdited);
-        blueRatio->setDefaultEditedState(pedited->filmNegative.blueExp ? Edited : UnEdited);
+        blueRatio->setDefaultEditedState(pedited->filmNegative.blueRatio ? Edited : UnEdited);
     } else {
         redRatio->setDefaultEditedState(Irrelevant);
         greenExp->setDefaultEditedState(Irrelevant);
@@ -189,10 +189,10 @@ void FilmNegative::adjusterChanged(Adjuster* a, double newval)
                 listener->panelChanged(
                     evFilmNegativeExponents,
                     Glib::ustring::compose(
-                        "R=%1 ; G=%2 ; B=%3",
-                        greenExp->getValue() * redRatio->getValue(),
+                        "Ref=%1 ; R=%2 ; B=%3",
                         greenExp->getValue(),
-                        greenExp->getValue() * blueRatio->getValue()
+                        redRatio->getValue(),
+                        blueRatio->getValue()
                     )
                 );
             }
@@ -269,10 +269,10 @@ bool FilmNegative::button1Pressed(int modifierKey)
                     listener->panelChanged(
                         evFilmNegativeExponents,
                         Glib::ustring::compose(
-                            "R=%1 ; G=%2 ; B=%3",
-                            greenExp->getValue() * redRatio->getValue(),
+                            "Ref=%1 ; R=%2 ; B=%3",
                             greenExp->getValue(),
-                            greenExp->getValue() * blueRatio->getValue()
+                            redRatio->getValue(),
+                            blueRatio->getValue()
                         )
                     );
                 }
diff --git a/rtgui/paramsedited.cc b/rtgui/paramsedited.cc
index 7d52d309a..952b08161 100644
--- a/rtgui/paramsedited.cc
+++ b/rtgui/paramsedited.cc
@@ -581,9 +581,9 @@ void ParamsEdited::set(bool v)
     dehaze.depth = v;
     metadata.mode = v;
     filmNegative.enabled = v;
-    filmNegative.redExp = v;
+    filmNegative.redRatio = v;
     filmNegative.greenExp = v;
-    filmNegative.blueExp = v;
+    filmNegative.blueRatio = v;
 
     exif = v;
     iptc = v;
@@ -1147,9 +1147,9 @@ void ParamsEdited::initFrom(const std::vector<rtengine::procparams::ProcParams>&
         dehaze.depth = dehaze.depth && p.dehaze.depth == other.dehaze.depth;
         metadata.mode = metadata.mode && p.metadata.mode == other.metadata.mode;
         filmNegative.enabled = filmNegative.enabled && p.filmNegative.enabled == other.filmNegative.enabled;
-        filmNegative.redExp = filmNegative.redExp && p.filmNegative.redExp == other.filmNegative.redExp;
+        filmNegative.redRatio = filmNegative.redRatio && p.filmNegative.redRatio == other.filmNegative.redRatio;
         filmNegative.greenExp = filmNegative.greenExp && p.filmNegative.greenExp == other.filmNegative.greenExp;
-        filmNegative.blueExp = filmNegative.blueExp && p.filmNegative.blueExp == other.filmNegative.blueExp;
+        filmNegative.blueRatio = filmNegative.blueRatio && p.filmNegative.blueRatio == other.filmNegative.blueRatio;
 
 //      How the hell can we handle that???
 //      exif = exif && p.exif==other.exif
@@ -3187,16 +3187,16 @@ void ParamsEdited::combine(rtengine::procparams::ProcParams& toEdit, const rteng
         toEdit.filmNegative.enabled = mods.filmNegative.enabled;
     }
 
-    if (filmNegative.redExp) {
-        toEdit.filmNegative.redExp = mods.filmNegative.redExp;
+    if (filmNegative.redRatio) {
+        toEdit.filmNegative.redRatio = mods.filmNegative.redRatio;
     }
 
     if (filmNegative.greenExp) {
         toEdit.filmNegative.greenExp = mods.filmNegative.greenExp;
     }
 
-    if (filmNegative.blueExp) {
-        toEdit.filmNegative.blueExp = mods.filmNegative.blueExp;
+    if (filmNegative.blueRatio) {
+        toEdit.filmNegative.blueRatio = mods.filmNegative.blueRatio;
     }
 
     // Exif changes are added to the existing ones
@@ -3245,5 +3245,5 @@ bool RetinexParamsEdited::isUnchanged() const
 
 bool FilmNegativeParamsEdited::isUnchanged() const
 {
-    return enabled && redExp && greenExp && blueExp;
+    return enabled && redRatio && greenExp && blueRatio;
 }
diff --git a/rtgui/paramsedited.h b/rtgui/paramsedited.h
index 57e6f2cdd..08a41fc7a 100644
--- a/rtgui/paramsedited.h
+++ b/rtgui/paramsedited.h
@@ -668,9 +668,9 @@ struct MetaDataParamsEdited {
 
 struct FilmNegativeParamsEdited {
     bool enabled;
-    bool redExp;
+    bool redRatio;
     bool greenExp;
-    bool blueExp;
+    bool blueRatio;
 
     bool isUnchanged() const;
 };
diff --git a/rtgui/partialpastedlg.cc b/rtgui/partialpastedlg.cc
index 34f9a414d..52adcfbf7 100644
--- a/rtgui/partialpastedlg.cc
+++ b/rtgui/partialpastedlg.cc
@@ -976,9 +976,9 @@ void PartialPasteDlg::applyPaste (rtengine::procparams::ProcParams* dstPP, Param
 
     if (!filmNegative->get_active ()) {
         filterPE.filmNegative.enabled   = falsePE.filmNegative.enabled;
-        filterPE.filmNegative.redExp   = falsePE.filmNegative.redExp;
+        filterPE.filmNegative.redRatio   = falsePE.filmNegative.redRatio;
         filterPE.filmNegative.greenExp  = falsePE.filmNegative.greenExp;
-        filterPE.filmNegative.blueExp   = falsePE.filmNegative.blueExp;
+        filterPE.filmNegative.blueRatio   = falsePE.filmNegative.blueRatio;
     }
 
     if (dstPE) {

```

rom9:
@heckflosse oh, i didn't think about the values in the history, actually that wasn't very cool. Thank you for the patch! Applying it right now...
One question: why passing the 3 separate values to `Thumbnail::processFilmNegative` and not the whole `params` object? Is that a general rule, like "only pass what is strictly necessary" ?
Sorry again for my C++ ignorance :-D

heckflosse:
@rom9 
> by passing the 3 separate values

That was nonsense by me, please revert

rom9:
Here's a little awk script to migrate existing pp3's , if needed:

```
/^RedExponent=/ { split($0, a, "="); re=a[2]; next }
/^GreenExponent=/ {
  if(re) {
    split($0, a, "=");
    ge=a[2];
    print "RedRatio="(re / ge);
  }
  print $0;
  next
}
/^BlueExponent=/ { split($0, a, "="); be=a[2]; print "BlueRatio="(be / ge); next }
{ print $0 }
```

@heckflosse 
> That was nonsense by me, please revert

ok, putting back to the params object

rom9:
Done. You guys convinced me, i have to say this version is way more clear, it made no sense to have the GUI speak a different language than the pp3. Thanks :-)

heckflosse:
:+1: for merge. @Beep6581 @Floessie @Desmis @Hombre57 ?

Floessie:
@heckflosse Absolutely. :+1: Maybe the benchmarking code should be removed after that.

heckflosse:
@rom9 Alberto, can you please push a white space change to your pr? This way we would get an LGTM test ;-)

Beep6581:
Merged, thank you @rom9 !

rom9:
> Merged, thank you @rom9 !

Cool! Thanks to all of you for the support :-)

heckflosse:
@rom9 Thank you for contributing :+1:

Beep6581:
@rom9 could you take a look at #5381 please?

PR #5370, opened 2019-07-07
Speedup for color propagation

heckflosse:
This is a first mockup to speed up hl reconstruction color propagation.
The speedup depends on the image. In worst case there is no speedup, but on average there are reasonable speedups.

There are small differences in output.
@Floessie please no code review atm. I first want to know if it works as expected.

### DIFF
--- a/rtengine/hilite_recon.cc
+++ b/rtengine/hilite_recon.cc
@@ -28,14 +28,11 @@
 #include "rawimagesource.h"
 #include "rt_math.h"
 #include "opthelper.h"
-namespace rtengine
-{
+#define BENCHMARK
+#include "StopWatch.h"
 
-extern const Settings* settings;
-
-//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-//%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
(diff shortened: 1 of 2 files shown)


heckflosse:
About memory usage:

In current version there is a console output writing out a factor.
Memory usage is reduced by (1 - (1 / factor)) * (width * height * 28 byte)

Desmis:
@heckflosse 
I have tested with Amsterdam.pef

Thank you

jacques

Desmis:
@heckflosse 
I just tested new commit with "yellow_poppy"

No differences in TIF
Huge speedup 
* without 370ms
* with 18ms

good job

jacques

Desmis:
@heckflosse 
Ingo

I just tested this new commit
For me no speedup with patch above, or very small

No differences in TIF

jacques

heckflosse:
@Desmis Thanks for testing. The (really minor) speedup this time was in last loop, where I could make `totwt` a bool. Formerly it was a float, and was used above and below the fraction line. This way I could turn some divisions into multiplications and also eliminate some divisions completely.

heckflosse:
Too bad that I can not fix typos in commit comments :frowning_face:

heckflosse:
@Floessie I think it's a good time for a code review now :)

Desmis:
@heckflosse 
I tested last commit, all works fine :)

jacques

Floessie (approved):
Good work, Ingo. Builds and runs fine.

Cleanup done.

Issue #5375, opened 2019-07-15
speedup for FileCatalog::reparseDirectory()


heckflosse:
FileCatalog::reparseDirectory() is very slow because of
[this loop](https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/filecatalog.cc#L1706) is O(n^2) and uses Glib::ustring for comparison, means it internally does n^2 expensive ustring::collate_key calls and (n^2)/2 std::string comparisons.

This can be changed to 2*n ustring::collate_key calls and n * log2(n) std::string comparisons.

I tested on a folder with 6934 copies of amsterdam.pef by renaming the first file.
Takes 454 seconds with old code
Takes 1.5 seconds with improved code

pr follows

heckflosse:
Closed as fixed

PR #5376, opened 2019-07-15
Filecatalog speedups

heckflosse:
#5375

### DIFF
--- a/rtgui/filecatalog.cc
+++ b/rtgui/filecatalog.cc
@@ -566,44 +566,44 @@ void FileCatalog::closeDir ()
     redrawAll ();
 }
 
-std::vector<Glib::ustring> FileCatalog::getFileList ()
+std::vector<Glib::ustring> FileCatalog::getFileList()
 {
+    BENCHFUN
     std::vector<Glib::ustring> names;
 
-    std::set<Glib::ustring> extensions;
+    std::set<std::string> extensions;
     for (const auto& parsedExt : options.parsedExtensions) {
-        extensions.emplace (parsedExt.lowercase ());
+        extensions.emplace(parsedExt.lowercase());
     }
 
     try {
 
-        auto dir = Gio::File::create_for_path (selectedDirectory);
+        const auto dir = Gio::File::create_for_path(selectedDirectory);
 
-        auto enumerator = dir->enumerate_children ("standard::name");
+        auto enumerator = dir->enumerate_children("standard::name");
 
         while (true) {
             try {
-                auto file = enumerator->next_file ();
+                const auto file = enumerator->next_file();
                 if (!file) {
                     break;
                 }
 
-                const Glib::ustring fname = file->get_name ();
+                const Glib::ustring fname = file->get_name();
 
-                auto lastdot = fname.find_last_of ('.');
+                const auto lastdot = fname.find_last_of ('.');
                 if (lastdot >= fname.length () - 1) {
                     continue;
                 }
 
-                const auto fext = fname.substr (lastdot + 1).lowercase ();
-                if (extensions.count (fext) == 0) {
+                if (extensions.count(fname.substr(lastdot + 1).lowercase()) == 0) {
                     continue;
                 }
 
-                names.emplace_back (Glib::build_filename (selectedDirectory, fname));
+                names.emplace_back(Glib::build_filename(selectedDirectory, fname));
             } catch (Glib::Exception& exception) {
                 if (options.rtSettings.verbose) {
-                    std::cerr << exception.what () << std::endl;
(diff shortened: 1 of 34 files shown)


rtgui/filecatalog.cc
```diff
@@ -563,44 +566,44 @@ void FileCatalog::closeDir ()
...
                 }
 
-                const auto fext = fname.substr (lastdot + 1).lowercase ();
-                if (extensions.count (fext) == 0) {
+                if (extensions.count(fname.substr(lastdot + 1).lowercase()) == 0) {
                     continue;
                 }
 
-                names.emplace_back (Glib::build_filename (selectedDirectory, fname));
+                names.emplace_back(Glib::build_filename(selectedDirectory, fname));
```
Floessie: Don't think, `emplace_back()` helps here...

rtgui/filecatalog.cc
```diff
@@ -1704,17 +1707,15 @@ void FileCatalog::reparseDirectory ()
...
-        bool found = false;
-
-        for (size_t j = 0; j < fileNameList.size(); j++)
-            if (nfileNameList[i] == fileNameList[j]) {
-                found = true;
-                break;
-            }
+    // build a set of collate-keys for faster search
+    std::set<std::string> oldNames;
+    for (size_t j = 0; j < fileNameList.size(); j++) {
```
Floessie: RB `for` loop?

rtgui/filecatalog.cc
```diff
@@ -1704,17 +1707,15 @@ void FileCatalog::reparseDirectory ()
...
-    for (size_t i = 0; i < nfileNameList.size(); i++) {
-        bool found = false;
-
-        for (size_t j = 0; j < fileNameList.size(); j++)
-            if (nfileNameList[i] == fileNameList[j]) {
-                found = true;
-                break;
-            }
+    // build a set of collate-keys for faster search
+    std::set<std::string> oldNames;
```
Floessie, line 1713: Have you considered `std::unordered_set<>`? Would be interesting, if it's faster for your workload...

rtgui/filecatalog.cc
```diff
@@ -1704,17 +1707,15 @@ void FileCatalog::reparseDirectory ()
...
+    for (size_t j = 0; j < fileNameList.size(); j++) {
+        oldNames.insert(fileNameList[j].collate_key());
+    }
 
-        if (!found) {
-            checkAndAddFile (Gio::File::create_for_parse_name (nfileNameList[i]));
+    for (size_t i = 0; i < nfileNameList.size(); i++) {
+        if (oldNames.count(nfileNameList[i].collate_key()) == 0) {
```
Floessie: I'd go for `find()` here... And a RBFL for `nfileNameList`.

rtgui/filecatalog.cc
heckflosse: I just tested using an unordered_set. No difference in processing time. I guess the most expensive part here still are the collate_key() calls.

heckflosse:
Last commit speeds up the following case (not only this case, but especially):

1) Open a folder with only a few files and apply a find-filter
2) then open a folder with a large amount of files without resetting the find-filter before

heckflosse:
Imho the pr has reached a state where it can be reviewed for a merge.

@Floessie @Beep6581 @Hombre57 Please test and review if time permits

Beep6581:
I created a folder with 5000 JPEGs on an SSD drive using the EXT4 filesystem in Linux.

I opened the folder using `dev` `5.6-184-g`2dee88428, took a few minutes to generate the cache. I navigated to a different folder, and then back to this one. Re-opening it took about 1 minute for the "File Browser [5000]" counter to reach 5000, at which point I saw filenames but had to wait another minute for thumbnails to appear. Deleting one of the files also froze RT for about 2 minutes. I went to a different folder, then went back to this one and waited for everything to settle, then using the "Find" filename filter is instant (I searched for "123"). Clearing the filter is instant too. Clicking the "Find" textbox the first time took a few seconds before the flashing cursor appeared.

I opened the folder using `filecatalog_speedups` `5.6-186-g`d9c93e77a. It took ~5 seconds for the file counter to show 5000 and for filenames to appear, and another 30 seconds for thumbnails to appear. Clicking the "Find" textbox also took a few seconds for the flashing cursor to appear. Finding "123" was instant. Clearing was instant. Deleting a file was instant. I searched for "123", went to a different folder, and went back to this one - this time loading all thumbs only took about 6 seconds.

heckflosse:
@Beep6581 Thanks for testing :+1: . I would like to merge the pr to get more tests. Any objections?

Hombre57:
No objection :+1:

PR #5383, opened 2019-07-18
Film Negative - Fix History view of values

TooWaBoo:
Setting values with the white - black button, the History shows a different format of the values.
grafik

This fix make look it equal to the standard view.
grafik


Beep6581:
Oops, thanks @TooWaBoo

PR #5386, opened 2019-07-22
use empty() instead of comparison with an empty string

heckflosse:


Beep6581 (approved):
All instances of `""` gone, except for one in a comment:
```
rtgui/navigator.cc:106:    if (0) { // (fontname!=""){
```

PR #5390, opened 2019-07-27
Further speedups for filebrowser/catalog

heckflosse:


heckflosse:
Some comments to the latest improvements:

1) filtering using the find string was improve by ~factor 15
2) to `feel` the improvments you need to use folders with a lot of files (>= 500 or so)

PR #5395, opened 2019-07-30
Make thumbnail range-selection more similar to OS behaviour, fixes #5393

### LINKED ISSUE #5393: Make thumbnail range-selection more similar to OS behaviour
Related to #5389: the usual behaviour in Windows or Linux when selecting multiple files with Shift, is that the very first image selected is always the anchor point for the range, regardless of selecting forward or backward.

After https://github.com/Beep6581/RawTherapee/commit/13207910528e2f553f3f875503b20f104941d958 the behaviour in RT still differs from the common behaviour in the OS.

1. Forward selection works fine. I.e. select one image, press shift and then a later one, and the range is selected. You can also reduce the range by shift-clicking one of the highlighted thumbnails in the selected range.
2. Backward selection is broken. If you select an image and shift-select an earlier one, the range is still fine. However...
a. ... if you then do a forward-selection, the backward-selected images stay selected and the forward selected ones are simply added to the selection.
b. ... if you want to reduce the backward-selection, it is reduced with respect to the thumbnail with the lowest index, and not with respect to the original first selected thumbnail.

In essence, RawTherapee needs to remember the first selected thumbnail when a modifier key is pressed, and should use that thumbnail as a reference point to handle ranges.

heckflosse:


Thanatomanic:
@heckflosse Works beautifully! +1 for merge

heckflosse:
@Thanatomanic I'm still working on some corner cases.

For example set anchor to first entry when selecting all files (ctrl-a) and no anchor is set...

heckflosse:
@Thanatomanic Now I think it's good to merge. I will wait for the LGTM analysis...

PR #5401, opened 2019-08-06
cbdl code cleanup

heckflosse:


Beep6581 (approved):
Compiled fine, and output with CbDL skin tone targetting from `cbdl-code-review` 5e9a409db was identical to that from `dev` 972d6907d.

PR #5421, opened 2019-08-21
improcfun.cc: cleanup

shlyakpavel:
The pointers allocated with new can never be null. Exception will be thrown in case of memory allocation error.


Floessie (approved):
This is right s long as we don't have an `operator new` which won't throw (would be non-compliant).

Note also, that `delete x; x = nullptr;` is a nice target for C++14's `std::exchange()`.

PR #5427, opened 2019-08-27
Speedup for transform

heckflosse:
This is a speedup for transformations like rotation, lensfun distortion correction...
I measured using a D800 file in full processing

1) only 1 degree rotation applied: before 400 ms, after 275 ms
2) 1 degree rotation + lensfun distortion correction: before 700 ms, after 570 ms

No differences in output


rtengine/iptransform.cc
```diff
@@ -87,6 +88,114 @@ float normn (float a, float b, int n)
     }
 }
 
+#ifdef __SSE2__
+inline void interpolateTransformCubic(rtengine::Imagefloat* src, int xs, int ys, float Dx, float Dy, float &r, float &g, float &b, float mul)
+{
+    constexpr float A = -0.85f;
+
+    // Vertical
+    const float t1Vert = A * (Dy - Dy * Dy);
```
Floessie, line 99: Don't think this makes a difference after optimization by the compiler, but there are still common subterms used more than once (like ' Dy * Dy' or 't1Vert * Dy').

rtengine/iptransform.cc
```diff
@@ -740,83 +849,70 @@ void ImProcFunctions::transformLuminanceOnly (Imagefloat* original, Imagefloat*
 
 void ImProcFunctions::transformGeneral(bool highQuality, Imagefloat *original, Imagefloat *transformed, int cx, int cy, int sx, int sy, int oW, int oH, int fW, int fH, const LensCorrection *pLCPMap)
 {
+    BENCHFUN
```
Floessie: Should be removed before merging.

rtengine/iptransform.cc
```diff
@@ -740,83 +849,70 @@ void ImProcFunctions::transformLuminanceOnly (Imagefloat* original, Imagefloat*
...
-    float** chOrig[3];
-    chOrig[0] = original->r.ptrs;
-    chOrig[1] = original->g.ptrs;
-    chOrig[2] = original->b.ptrs;
-
-    float** chTrans[3];
-    chTrans[0] = transformed->r.ptrs;
-    chTrans[1] = transformed->g.ptrs;
-    chTrans[2] = transformed->b.ptrs;
+    float** chOrig[3] = {original->r.ptrs, original->g.ptrs, original->b.ptrs};
```
Floessie: `const float* const * const` doesn't really look good, so I'll try to make use of std::array<>.

Floessie (approved):
Test-run on i386-non-SSE. All fine. I'll push some more cleanups.

Issue #5426, opened 2019-08-26
Trackpad gestures


Davide-sd:
I just installed Rawtherapee on a Windows 10 laptop, which comes with a decent trackpad and a handful of useful gestures. I'm most interested in the zoom and swipe to scroll between photos.

Looking at the settings, I did not find any references to gestures. At the moment, it is possible to zoom by moving two fingers in the trackpad (both vertical and horizontal), but the zoom velocity is very very high even using medium sensitivity. Would love a slower zoom. 

About scrolling photos, would it be possible to implement something like a two-finger swipe (left-right) two change to the previous or next photo?

Entropy512:
Two-finger scrolling is a fairly default behavior usually implemented in the OS trackpad driver.

That said - I can confirm that on an Ubuntu 19.04 laptop with an ELAN touchpad (Acer Aspire R15), zoom gain is waaaay too high - if anything one of the first things I noticed when giving RT a try after working with another application for a while.  I haven't compared against a proper hardware mouse yet...  I've been doing most of my work on the living room floor lately.  :)

Same goes for the file browser - a slight movement of the trackpad goes from the beginning of the directory to the end almost instantly.

Scroll response is OK within the Settings screen.

This is on a build from the git "dev" branch as of early Saturday.

On your machine, do you know if two-finger swipe right and left gets mapped at a low level to a mouse axis?  (Some mice do support left/right scroll in addition to up/down)  I'm not sure if any trackpad drivers actually return individual multitouch data to the application, at least for the touchpad?

Entropy512:
If you place your mouse cursor over the upper film roll and two-finger swipe left/right - do you see it scroll?

It scrolls for me, but at similarly incredibly high gain as zooming in/out.

Also - while the thumbnails in the file browser have extremely high scroll gain with my trackpad, the folder list on the left does not.

Thanatomanic:
I'm pretty sure we're all very inexperienced in handling any other input method than a simple mouse and keyboard. At least afaik RT hasn't been optimized for use with a trackpad.
If anyone can point to a good resource on how to handle those (including maybe ease-in and ease-out functions for movement) that would be a start in the right direction.

Entropy512:
Good question.  I'll see what I can find, and see if I can see something obvious in RT's code that explains why only certain uses of scrollwheel input misbehave with this particular variation of scrollwheel input.

[...]

Issue #4784, opened 2018-09-07
GTK+3.24.1 issues


TooWaBoo:
**This is a list of issues with GTK+3.24.1 and RT**
W = Windows, M = MacOS, L = Linux
1. ~~Filename tooltip on thumbnails and History tooltip are flickering. (W/M)~~
2. ~~Scrolling with the mouse wheel in the file browser, queue and filmstrip doesn't work anymore. (W)~~
3. ~~Scrolling with the mouse wheel when hovering a combobox (e.g. Toolpanel) doesn't work anymore. (W)~~
4. ~~Zooming with the mouse wheel in the preview doesn't work anymore. (W)~~
5. ~~Rounding of "em" values in themes has changed. (this affects only the TWB theme) (W)~~
6. ~~Scrolling with the mouse wheel only possible by page. (W)~~

**EDIT: All bugs are fixed in GTK+3.24-2**

Benitoite:
There is a Mac-only issue involving the quartz backend in 3.24.0, RT occasionally crashes when dragging the window around, etc.  The gtk dev who wrote the backend code there has reported it to gitlab.gnome.org: https://gitlab.gnome.org/GNOME/gtk/issues/1312

TooWaBoo:
@Benitoite 
Thanks, what about the other issues?

Benitoite:
2, 3, and 4 don’t seem to be any problem on Mac.

Hombre57:
I've opened an issue in Gtk project for n°3 : https://gitlab.gnome.org/GNOME/gtk/issues/1330.

N°2, 4 and 6 are solved here : the scroll event is a `GTK_SMOOTH_SCROLL` which wasn't handled (and it may be a Gtk bug IMHO : it shouldn't emulate a smooth scroll when using a mouse). I'll post a patch soon.

@Benitoite What's your pointing device ? Standard mouse ?

Benitoite:
@hombre standard Logitech wired usb mouse with scroll wheel.

Hombre57:
@TooWaBoo (ping @Beep6581) You've added W/M/L tag to each entry, Does it mean that OS not listed are not affected, or that we don't know ? I have to answer in GNOME's issue about n°3 in Linux.

Hombre57:
About n°2, 4 and 6 : this may evolve soon with [a change in Gtk](https://gitlab.gnome.org/GNOME/gtk/merge_requests/335), hopefully in a good way (but that's not sure).

[...]

Issue #4663, opened 2018-07-04
Thumbnail scrolling speed under linux too fine.


abertens:
Hi there!

Today I git clone the dev branch to try the latest functionality and I found an enhancement in the gui that not works well on my linux desktop.

Merge pull request #4652 from Thanatomanic/scroll-speed (Scrolling speed changed to height of thumbnail) 8a9f102cd252c26629dd2bc314e8b3ff9230e0ca, from Jul 1, 2018 makes thumbnail scrolling almost unusable on a directory with 200 images. The slightest touch on the scroll button moves the page upward/downward that makes me stop using it and start using keys instead to have some control over it.

May I ask if you tested this merge on linux? I don't know about Windows...

Information about my build:
Version: 5.4-470-gb419bb2d
Branch: dev
Commit: b419bb2d
Commit date: 2018-07-04
Compiler: cc 5.5.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.22.2
Lensfun: V0.3.2.0
Build type: Release
Build flags:  -std=c++11 -march=native -Werror=unused-label -flto -fopenmp -Werror=unknown-pragmas -Wall -Wno-unused-result -Wno-deprecated-declarations -O3 -DNDEBUG -ftree-vectorize
Link flags:  -march=native -flto
OpenMP support: ON
MMAP support: ON

Gtk version is gtk+-3.22.30.

Thank you for your effort to bring us this magnificent tool.

Yours,
Andres Bertens

abertens:
Additional info:
Testing on a directory with a 25 images behaves the same. Seems that image number is irrelevant.

Forgot to mention that reverting merge 8a9f102cd252c26629dd2bc314e8b3ff9230e0ca makes it usable again.

Thanks.

Thanatomanic:
Hi! Thanks for reporting. Can you confirm if scrolling in your 200 image folder still goes in discrete steps? I mean, it should scroll now in steps of the thumbnail height. Does it do that?

ff2000:
"The slightest touch on the scroll button moves the page upward/downward"
Are you talking of the scroll wheel of your mouse? Or the slider on the scrollbar?
I am also on gtk+-3.22.30 on Gentoo linux and have no issues. I only noticed a small inconsistency: Turning the scroll wheel while mouse is over the thumb view each turn scrolls by one row. But turning while over the scrollbar reveals the old "micro-scrolling".

[...]

PR #5430, opened 2019-08-29
Smooth scrolling improvements

Entropy512:
The thumbnail browser (thumbbrowser.cc), right-side panels (guiutils.cc), and editor zooming (cropwindow.cc) all had discretized scroll behaviors that quantized GDK_SMOOTH_SCROLL events up to +/- 1.0, greatly amplifying the response to some smooth scrolling events (most notably touchpad scroll gestures)

This patchset:
Allows the thumb browser and right-side panels to scroll more smoothly, proportionally to the scroll event delta
Accumulates smooth scroll deltas when zooming the editor to prevent high-frequency small-delta events from rapidly traversing between minimum and maximum zoom

Smooth scrolling without delta accumulation in the zoom window would be an ideal goal but would also require a major rewrite of most of the cropwindow.cc zooming mechanisms in order to not use discretized zoom steps.

This fixes the most egregious issues raised in issue #5426 - implementing the requested swipe right/left in editor to move to prev/next photo will require additional work, and there are GUI design considerations to be made since this might break behavior for some existing users who rely on left/right smooth scrolling being treated identically to up/down by the zoom scrolling code.

### DIFF
--- a/rtgui/cropwindow.cc
+++ b/rtgui/cropwindow.cc
@@ -51,7 +51,7 @@ CropWindow::CropWindow (ImageArea* parent, bool isLowUpdatePriority_, bool isDet
       backColor(options.bgcolor), decorated(true), isFlawnOver(false), titleHeight(30), sideBorderWidth(3), lowerBorderWidth(3),
       upperBorderWidth(1), sepWidth(2), xpos(30), ypos(30), width(0), height(0), imgAreaX(0), imgAreaY(0), imgAreaW(0), imgAreaH(0),
       imgX(-1), imgY(-1), imgW(1), imgH(1), iarea(parent), cropZoom(0), zoomVersion(0), exposeVersion(0), cropgl(nullptr),
-      pmlistener(nullptr), pmhlistener(nullptr), observedCropWin(nullptr),
+      pmlistener(nullptr), pmhlistener(nullptr), scrollAccum(0.0), observedCropWin(nullptr),
       crop_custom_ratio(0.f)
 {
     initZoomSteps();
@@ -295,11 +295,16 @@ void CropWindow::scroll (int state, GdkScrollDirection direction, int x, int y, 
     } else {
         delta = deltaY;
     }
-    if (delta == 0.0 && direction == GDK_SCROLL_SMOOTH) {
-        // sometimes this case happens. To avoid zooming into the wrong direction in this case, we just do nothing
-        return;
+
+    if (direction == GDK_SCROLL_SMOOTH) {
+        scrollAccum += delta;
+        //Only change zoom level if we've accumulated +/- 1.0 of deltas.  This conditional handles the previous delta=0.0 case
+        if (abs(scrollAccum) < 1.0) {
+            return;
+        }
     }
-    bool isUp = direction == GDK_SCROLL_UP || (direction == GDK_SCROLL_SMOOTH && delta < 0.0);
+    bool isUp = direction == GDK_SCROLL_UP || (direction == GDK_SCROLL_SMOOTH && scrollAccum < 0.0);
+    scrollAccum = 0.0;
     if ((state & GDK_CONTROL_MASK) && onArea(ColorPicker, x, y)) {
         // resizing a color picker
         if (isUp) {
--- a/rtgui/cropwindow.h
+++ b/rtgui/cropwindow.h
@@ -102,6 +102,7 @@ class CropWindow : public LWButtonListener, public CropDisplayHandler, public Ed
     PointerMotionListener* pmlistener;
     PointerMotionListener* pmhlistener;
     std::list<CropWindowListener*> listeners;
+    double scrollAccum;
 
     CropWindow* observedCropWin;  // Pointer to the currently active detail CropWindow
 
--- a/rtgui/guiutils.cc
+++ b/rtgui/guiutils.cc
@@ -971,9 +971,8 @@ bool MyScrolledWindow::on_scroll_event (GdkEventScroll* event)
                 scroll->set_value(value2);
             }
         } else if (event->direction == GDK_SCROLL_SMOOTH) {
-            if (abs(event->delta_y) > 0.1) {
-                value2 = rtengine::LIM<double>(value + (event->delta_y > 0 ? step : -step), lowerBound, upperBound);
-            }
+            value2 = rtengine::LIM<double>(value + event->delta_y * step, lowerBound, upperBound);
+
             if (value2 != value) {
                 scroll->set_value(value2);
             }
--- a/rtgui/thumbbrowserbase.cc
+++ b/rtgui/thumbbrowserbase.cc
@@ -79,10 +79,19 @@ void ThumbBrowserBase::scroll (int direction, double deltaX, double deltaY)
         delta = deltaY;
     }
     if (direction == GDK_SCROLL_SMOOTH && delta == 0.0) {
-        // sometimes this case happens. To avoid scrolling the wrong direction in this case, we just do nothing    
+        // sometimes this case happens. To avoid scrolling the wrong direction in this case, we just do nothing
+        // This is probably no longer necessary now that coef is no longer quantized to +/-1.0 but why waste CPU cycles?
         return;
     }
-    double coef = direction == GDK_SCROLL_DOWN || (direction == GDK_SCROLL_SMOOTH && delta > 0.0) ? +1.0 : -1.0;
+    //GDK_SCROLL_SMOOTH can come in as many events with small deltas, don't quantize these to +/-1.0 so trackpads work well
+    double coef;
+    if(direction == GDK_SCROLL_SMOOTH) {
+        coef = delta;
+    } else if (direction == GDK_SCROLL_DOWN) {
+        coef = +1.0;
+    } else {
+        coef = -1.0;
+    }
 
     // GUI already acquired when here
     if (direction == GDK_SCROLL_UP || direction == GDK_SCROLL_DOWN || direction == GDK_SCROLL_SMOOTH) {


heckflosse:
@Entropy512 
Hi, I will test your pr tomorrw on my PC using a mouse wheel. Unfortuntately I don't have a pad which knows geastures. But I will ne able to agree or disagree with your PR then. I guess, it won'e be the latter :+1:

Entropy512:
Sounds good to me.  I have no experience with building for Windows such that @davide-sd can test.
Ideally this should have no change for mouse wheel users, unless your wheel happens to be from the category of mice mentioned in my pull request (ones that have scroll detents but emit smooth scroll events with deltas less than +/-1.0 - I'm not sure if these actually exist).

I'm not sure what other types of input devices might emit smooth scroll events and whether RT handled these well or not.  (Such as a jog wheel???)

heckflosse:
Works fine here on Win7 using a Logitech mouse with scroll wheel

Floessie:
@Entropy512 Andrew, some minor remarks:
- Please indent with 4 spaces.
- Please have a space after `if`.
- Please add curly braces even to one-liners after `if` etc.

We also have an [`astyle.rc`](https://github.com/Beep6581/RawTherapee/blob/dev/rawtherapee.astylerc). :wink: 

Thanks for your contribution! :+1: 

Best,
Flössie

Entropy512:
OK, thanks for the input!

I'll try to get to these sometime this weekend, but it's now crunch time for cleaning my apartment and a family wedding.  :)

Also, would your preference be for additional commits, or rebase/squash the style changes to keep commit history cleaner?

Thanatomanic:
I'll also give this a test run soon, as I was the one who did the last modification of the scroll behaviour (#4562) and not everyone agreed with it (#4663).

Entropy512:
Hmm, if I'm reading this correctly, the issue raised by @abertens is exactly the behavior this submission is intended to fix.  Although the behavior he describes is the exact opposite of "too fine".

It looks like you increased the increment of non-smooth scrolls in such a way that it amplified the effect of quantizing smooth scroll deltas upwards to +/- 1.0

As far as different behaviors based on hardware, what I've seen (and what I assume based on how the code is written):
Some mice with scroll wheels that have detents emit GDK_SCROLL_UP and GDK_SCROLL_DOWN events.  I assume these exist but haven't yet encountered one.  I need to dig up some of my other mice though.  This pull request won't change behavior with these
Some mice with scroll wheels that have detents emit GDK_SCROLL_SMOOTH with deltas equal to +/- 1.0 - the mouse scroll behavior in this case won't change.  (abertens didn't like this either)
Trackpads/touchpads emit frequent GDK_SCROLL_SMOOTH events with very small deltas.  These were quantized upwards to +/-1.0 which led to highly undesirable behavior (abertens' other complaint) - this patch should resolve those complaints.

It remains to be seen whether anyone has a mouse that has scrollwheel detents that emit GDK_SCROLL_SMOOTH with values less than 1.0, these will also change behavior with this patch.

Some of the discussion in https://github.com/Beep6581/RawTherapee/issues/3923 is puzzling me.  There's code in https://github.com/Beep6581/RawTherapee/issues/3923#issuecomment-374196785 that appears to respond to GDK_SMOOTH_SCROLL events effectively assigning v=0 regardless of its delta.  This is clearly not what is present now, as reading that code, it would respond to GDK_SMOOTH_SCROLL with no action at all.

It appears that @Beep6581 has a trackpad driver that emits GDK_SCROLL_UP and GDK_SCROLL_DOWN events at a reduced rate...  It may internally be performing something similar to the accumulate/coalesce operation I implemented for cropwindow.  Edit:  At least on another machine I just quickly checked out running Ubuntu 19.04 + KDE Plasma 5.15.4 + KDE Frameworks 5.56.0, the "scrolling distance" and "coasting" settings are all greyed out on a Lenovo E470, which is different than one of the screenshots @Beep6581 provided.  I expect software versions are the same on my Acer Aspire R15 which is what I developed these patches on.

Same for Ubuntu 18.04 + KDE Plasma 5.12.8 + Frameworks 5.47.0 on another laptop (Dell E7440)

https://github.com/Beep6581/RawTherapee/commit/0d61f14bae5945fc3a33fd7bfab34d178cd3c6ab#diff-b4587767fa42796444a638e361d7b17d happened at some point after the discussion you linked, and it's primarily this behavior that has been altered.

It looks like potentially GDK_SCROLL_SMOOTH is new sometime after GTK+ 3.18?  When I get home I'll check to see where Ubuntu 19.04 is, but I'm guessing 3.22 or 3.24.  The fact that I'm running KDE may or may not be relevant as far as how the touchpad driver behaves.

Edit:  Looks like @Hombre57 should probably take a look at this.

Entropy512:
Since no one raised an objection to squash-and-rebase, I squashed the style fixes into the original commits to keep commit history clean.

Beep6581:
> Also, would your preference be for additional commits, or rebase/squash the style changes to keep commit history cleaner?

Since no one commented, my 2 cents: keep formatting changes in a separate commit from functional code changes if possible, and if squashing makes things more clear than go for it.

Beep6581:
Tried and tested using my laptop touchpad in KDE as well as the mouse scrollwheel. Scrolling in branch `smooth_scroll` works well with no regressions that I know of.

Beep6581:
+1 for merge.

Hombre57:
@Entropy512 Thanks for fixing my dirty workaround. Please see [this comment](https://github.com/Beep6581/RawTherapee/issues/4784#issuecomment-420432476) from issue #4784
and the [discussion here](https://gitlab.gnome.org/GNOME/gtk/merge_requests/335).

IIRC, the mouse wheel events thrown by a specific Gtk version was not what one would expect, and I've added the workaround to convert `GDK_SCROLL_SMOOTH` as plain `GDK_SCROLL` events. I'll let you take any action on this, since the it has probably been solved meanwhile in Gtk.

Entropy512:
@Hombre57 Ah, thanks for the history.

As to what I've seen, as described in your second link, GDK_SCROLL_SMOOTH with deltas of +/- 1.0 is exactly what I see from my mice.

It appears that as far as telling what hardware is sending the events, you can't tell that, however you can differentiate between hardware based on the nature of the data you receive.

If you set scrolling responses up such that a delta of +/- 1.0 maps to your original discrete scroll amounts for up/down/left/right, you get the same behavior for such hardware without having to identify it.  Then you automatically get smooth scrolling for hardware that does support it.

Based on what the GTK guys are saying, it looks like:
In thumbbrowserbase and guiutils, SCROLL_MASK should be removed in favor of SMOOTH_SCROLL_MASK - and the code here for smooth scroll events should be able to now handle all devices
In cropwindow, the opposite should be done, but this will require some testing on trackpads.  Eliminating SMOOTH_SCROLL_MASK may eliminate the need for the smooth scroll accumulator depending on how things behave if you have a trackpad and only SCROLL_MASK is set.

I'm not sure if my bugfix is under consideration for 5.7 - but the cleanup I describe almost surely needs to be wait for 5.8 or later because it has a lot more potential for side effects.  It also may mandate a minimum GTK+ version higher than what RT is currently demanding?

Beep6581:
Let's leave it for 5.8 and not rush things for 5.7.

Entropy512:
Fine for me.  (Really the only reason to try and at least get a band-aid in place for 5.7 is because if a new user with a smooth scrolling trackpad or other smooth scrolling device tries RT for the first time, these issues WILL be one of the first negative things they notice.  But we don't have any metrics on how many masochists actually use their laptop without an external mouse like I frequently do.  People like me who slowly chew through their backlog during Netflix pauses are probably pretty rare. 
 :) )

sguyader:
I'm one of those rare cats! 😅
Most of the time I use a mouse on my laptop, but once in a while I take the laptop in my bedroom to work on some shots, and then I don't always bother grabbing a mouse, and I use the 2-finger scrolling on the trackpad... Which has always been a pain in the arse indeed. In this case I may use the scrolling slider but it's also a pain to get the cursor on it using the trackpad.

Entropy512:
@sguyader That was exactly my first impression of RT when I started working with it.  It was oddly coincidental timing that someone posted #5426 a few days after I first gave RT a try due to the new film negative inversion tool.  (I can't remember why I selected darktable over RT ages ago, but at least as far as my current workflow needs - RT is doing a MUCH better job of meeting them as is.)

Hombre57:
@Beep6581 My comment was only to explain my changes at that time. I'm in favor of adding it to 5.7 since it's a pretty simple change and enhance the situation for touchpad users.

Beep6581:
Ok then, we add the band-aid for 5.7 (I prefer this too), since it has been tested by various people by now. Will merge in a few hours if no objections.

Entropy512:
Works for me.  After I get back from vacation in mid-September I'll look into the next phase of cleaning this up by looking at whether we can simplify things by removing some of the mask flags.

abertens:
Hi there!

I've just downloaded and compile RT with this merge and it works as expected.

Sorry to act as dissident, but as I said before, I really dislike the current scrolling thumbnail behaviour in file browser window or top panel. I see this behaviour on other applications, but not all. I see it is not fit to all of them.

I think its about my actual setup. My laptop screen is 1366x768 and I like to see thumbnails as large as possible, so two full rows only are shown. Scrolling (two fingers on touchpad or wheel on mouse) moves a row at a time with a discrete movement: next row or previous one replace actual and I get lost when following images.
If I reduce thumbnail size in order to show more rows, the effect is less dramatic because rows do not disappear at once.

As an alternative, I can move the mouse to the scrollbar and use with shift key simultaneously. The movement now is about 1/3 of thumbnail height.

But as patching is not so difficult, I have been modifying thumbbrowserbase.cc to assign a fixed increment of 32 pixels since the original behaviour was changed and I am happy with it :-)

I may see a problem with a fixed increment, many rows in file browser and a large monitor. I guess this was addressing the original change in behaviour.

Please, this is not a complaint; just another point of view. Finally, I can adapt using smaller thumbnails or use the shift key trick (if I decided not to patch thumbbrowserbase.cc).

Great software and thank you for your effort!

Regards,
Andres Bertens

Thanatomanic:
Your reply is well taken! Though, I would like to put the following comments:

> My laptop screen is 1366x768

This is an increasingly rare and small display size. That does not reduce the wish to have a good user experience for the GUI in that case, but a change may not be beneficial to many (or more) other users.

> As an alternative, I can move the mouse to the scrollbar and use with shift key simultaneously. The movement now is about 1/3 of thumbnail height.

There are always the arrow keys you can use. That should give you perfect control over the thumbnail movement.

Entropy512:
At least my intent with this patchset was to only address the most egregious flaws in trackpad behavior.

There's some potential further investigation that is merited, and I'm going to look at some of it when I get back from next week's vacation:
1)  Based on the link provided to that GTK+ change, there may be some significant code cleanup achievable by selectively removing either SCROLL_MASK or SMOOTH_SCROLL_MASK depending on the exact use case.  In general I prefer smooth scroll, however since cropwindow really can't be made "smooth" without a massive rework, removing SMOOTH_SCROLL_MASK may remove the need for the scrollAccum bandaid.  That's definitely something that needs to wait for the 5.8 cycle
2)  For your use case, I wonder if a compromise might be to expose a "scroll increments per image height" option in settings?  Default it to the current of 1, but also allow 2, 3, maybe even 4.  It would simply take the current scroll amount and divide it by the setting value.  I personally want to try an A/B comparison of the current 1 increment per image height and 2 increments when I get back, because I can sort of see your original complaint that the "flow" gets hard to visually track when you jump an entire image - and I have a 1080p screen.  Just a thought as to a possible approach.

For your screen, does the old hardcoded 32 happen to roughly map to 1/2 the image height?

abertens:
@Thanatomanic
According to Stats Counter (https://gs.statcounter.com/screen-resolution-stats/desktop/worldwide), on global desktop screen resolutions,  1366x768 is on top of the list (23.86%), followed by 1920x1080 (20.07%).

And using keys are the same as the new behaviour, jumps from one row to the next, moving one row at the time.

@Entropy512
On my laptop (Thinkpad X230), highest thumbnail size is 288x218 approx.
With the wheel in an external mouse, I've got 1/6 height approx for each roll movement (218/32 = 6).
With the integrated touchpad + two fingers (Debian 10, XFCE) I've got twice as much (I think there is a reduction factor somewhere), something like 1/12.

Yes, a configuration option will do it.
But I think perhaps an adaptive alternative could be easier: count how many rows are shown in file browser and apply according to it. Less equal than 3 rows, 1/4 of thumbnail size; else +1/-1 thumbnail size. Everyone happy :-)

Regards,
Andres

Entropy512:
Yeah, I was thinking of potentially something along those lines too - the risk of breaking "visual flow" is greater if there are fewer thumbnails.

PR #5435, opened 2019-09-03
Converting Glib's mutex (obsolete) to std::mutex

Hombre57:


Hombre57:
@Floessie I've made the conversion and didn't had performance loss or deadlock so far. I think that the conversion was pretty straightforward, but I prefer the review of an experienced dev :wink: . The review is about the `Glib` vs `std` equivalence, not the utility class as a whole, which anyone can deeply review on a later date...

heckflosse:
Without looking at the changes, I would see that for RT 5.8
Though I will @Floessie have the last word on this...

rtgui/threadutils.cc
```diff
@@ -136,12 +137,12 @@ void MyWriterLock::release ()
...
     }
 
-    Glib::Threads::Mutex::Lock lock (mutex.mutex);
+    std::lock_guard<std::mutex> lock (mutex.mutex);
 
     // Decrement the writer number first...
     if (--mutex.writerCount == 0) {
         // ...and if the writer count is zero again, we can wake up the next writer or reader.
-        mutex.cond.broadcast ();
+        mutex.cond.notify_one();  // notify_all ?
```
Floessie: Hm, we want multiple readers, so we should notify them all, don't we?

I asked my workmate to review this piece of code, too, and he noticed a possible imbalance in `MyReaderLock::acquire()`:

```c++
void MyReaderLock::acquire ()
{
    if (locked) {
        return;
    }

    std::lock_guard<std::mutex> lock (mutex.mutex);
// Precondition: A writer is writing, so writerCount != 0.
    if (mutex.writerCount == 0) {
        // There's no writer operating, we can increment the writer count which will lock writers.
        ++mutex.writerCount;
    } else if (mutex.readerCount == 0) {
// There is no active reader yet, so readerCount == 0.
        // The writer count is non null, but a reader can be the owner of the writer lock,
        // which will be the case if the reader count is not zero too.
// Now let two readers spin here.
        while (mutex.writerCount != 0) {
            mutex.cond.wait(mutex.mutex);
        }
// The writer leaves, both readers proceed.
        // Then, we can increment the writer count.
// The writerCount is incremented by 2. This leads to an imbalance in release().
        ++mutex.writerCount;
    }

    // Finally, we can increment the reader count as well.
    ++mutex.readerCount;

    locked = true;
}

void MyReaderLock::release ()
{
    if (!locked) {
        return;
    }

    std::lock_guard<std::mutex> lock (mutex.mutex);

    // decrement the writer number first...
    --mutex.readerCount;

    if (mutex.readerCount == 0) {
        // ...if no more reader, we decrement the writer count as well...
// Last reader leaves, but writerCount is only decremented by 1.
        --mutex.writerCount;

        // ...and signal the next waiting reader/writer that it's free
        mutex.cond.notify_one();  // notify_all ?
    }

    locked = false;
}
```

Is this plausible?

rtgui/threadutils.h
```diff
@@ -92,8 +90,8 @@ class MyRWMutex :
     friend class MyWriterLock;
 
 private:
-    Glib::Threads::Mutex mutex;
-    Glib::Threads::Cond cond;
+    std::mutex mutex;
+    std::condition_variable_any cond;
```
Floessie: [cppreference](https://en.cppreference.com/w/cpp/thread/condition_variable_any) states:
> If the lock is std::unique_lock, std::condition_variable may provide better performance.

Maybe something like this?

```patch
diff --git a/rtgui/threadutils.cc b/rtgui/threadutils.cc
index 4f8190c36..b7e18e955 100644
--- a/rtgui/threadutils.cc
+++ b/rtgui/threadutils.cc
@@ -68,7 +68,7 @@ void MyReaderLock::acquire ()
         return;
     }
 
-    std::lock_guard<std::mutex> lock (mutex.mutex);
+    std::unique_lock<std::mutex> lock (mutex.mutex);
 
     if (mutex.writerCount == 0) {
         // There's no writer operating, we can increment the writer count which will lock writers.
@@ -77,7 +77,7 @@ void MyReaderLock::acquire ()
         // The writer count is non null, but a reader can be the owner of the writer lock,
         // which will be the case if the reader count is not zero too.
         while (mutex.writerCount != 0) {
-            mutex.cond.wait(mutex.mutex);
+            mutex.cond.wait(lock);
         }
 
         // Then, we can increment the writer count.
@@ -96,7 +96,7 @@ void MyReaderLock::release ()
         return;
     }
 
-    std::lock_guard<std::mutex> lock (mutex.mutex);
+    const std::unique_lock<std::mutex> lock (mutex.mutex);
 
     // decrement the writer number first...
     --mutex.readerCount;
@@ -118,11 +118,11 @@ void MyWriterLock::acquire ()
         return;
     }
 
-    std::lock_guard<std::mutex> lock (mutex.mutex);
+    std::unique_lock<std::mutex> lock (mutex.mutex);
 
     // The writer count is not zero, so we have to wait for it to be zero again...
     while (mutex.writerCount != 0) {
-        mutex.cond.wait (mutex.mutex);
+        mutex.cond.wait (lock);
     }
 
     // ...then we can increment the writer count.
@@ -137,7 +137,7 @@ void MyWriterLock::release ()
         return;
     }
 
-    std::lock_guard<std::mutex> lock (mutex.mutex);
+    const std::unique_lock<std::mutex> lock (mutex.mutex);
 
     // Decrement the writer number first...
     if (--mutex.writerCount == 0) {
diff --git a/rtgui/threadutils.h b/rtgui/threadutils.h
index f54cbb845..9a7e810d8 100644
--- a/rtgui/threadutils.h
+++ b/rtgui/threadutils.h
@@ -91,7 +91,7 @@ public:
 
 private:
     std::mutex mutex;
-    std::condition_variable_any cond;
+    std::condition_variable cond;
 
     std::size_t writerCount = 0;
     std::size_t readerCount = 0;
```

rtgui/threadutils.h
```diff
@@ -195,18 +193,6 @@ inline MyMutex::MyLock::MyLock (MyMutex& mutex)
...
-    , locked (false)
-{
-}
-
-inline MyMutex::MyLock::MyLock (MyMutex& mutex, Glib::Threads::TryLock)
-    : mutex (mutex)
-    , locked (mutex.trylock ())
-{
-}
-
```
Floessie, line 208: :+1: for removing unused code.

rtgui/thumbimageupdater.cc
```diff
@@ -164,10 +164,10 @@ class ThumbImageUpdater::Impl :
         }
 
         if ( --active_ == 0 ) {
-            Glib::Threads::Mutex::Lock lock(mutex_);
+            std::lock_guard<std::mutex> lock(mutex_);
             if (inactive_waiting_) {
                 inactive_waiting_ = false;
-                inactive_.broadcast();
+                inactive_.notify_one();  // notify_all ?
```
Floessie: Don't know. Is a thundering herd allowed here?

rtgui/thumbimageupdater.cc
```diff
@@ -83,25 +83,25 @@ class ThumbImageUpdater::Impl :
...
+    std::mutex mutex_;
 
     JobList jobs_;
 
     std::atomic<unsigned int> active_;
 
     bool inactive_waiting_;
 
-    Glib::Threads::Cond inactive_;
+    std::condition_variable_any inactive_;
```
Floessie: Again, a `std::condition_variable` could be faster.

Floessie (changes requested):

Hombre57:
Thanks for the review @Floessie , I'll get back on this this w.e.

Beep6581:
Do we merge?

rtgui/threadutils.cc
```diff
@@ -17,6 +17,7 @@
  *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
  */
 #include "threadutils.h"
+#include <glibmm/threads.h>
```
hfiguiere: this shouldn't be added if you remove the `Glib::Threads` (here it builds without)

hfiguiere:
How can I help getting this merged? 

I was about to prepare something similar. This would greatly help build the rtengine with a more modern gilbmm.

rtgui/threadutils.cc
hfiguiere: `broadcast`  notifies all waiting threads, so `notify_all` is likely the proper answer.

rtgui/thumbimageupdater.cc
hfiguiere: `notify_all` is the proper equivalent here.

Hombre57:
Damn ! How can some dude start something useful and almost ready and leave the sh!t undone ? I apologize for this stalled branch, I'll resurrect it ASAP.

ghost:
Hello Hombre,

do you need any help? If so, just give me a short hint. After the switch to std has been merged, I will got on with a look at gtk4.

Thank you in advance.

Hombre57:
Hi Christian,

Hopefully I'll be able to work on this tomorrow. I'll ask for your (and others) help here if needed. I'll need reviewer anyway :wink:

rtgui/threadutils.cc
Hombre57: @Floessie 

About 
```
// Now let two readers spin here.
```

Isn't `std::lock_guard<std::mutex> lock (mutex.mutex);` protecting from this possibility ?

Hombre57:
I've made the changes locally, I'll have one next try for pushing the changes to `std-mutex` branch, unless I don't have commit rights anymore. For now it send's me back `not authorized`, but I don't know if it's related to some SSH public key trouble or repositories access rights. Hi @Beep6581 :slightly_smiling_face: , any idea ?

Hombre57:
Yay ! I managed to find the correct connection string in Eclipse, now waiting for your review.

ghost:
Hello everyone,

thank you so much for pushing this Hombre57. As far as I can judge - based on my knowledge which is limited to thread handling in gtk/gdk - it looks ok. Most of the changes have also been done and tested in my repo (the PL has been closed cause I have not seen, that one exists already here).

As far as I can see here, the review of Floessie is missing to get the commits into dev right?

Greetings

Hombre57:
Hi Christian,

I think I've done things the wrong way. I've added @Floessie's requested changes in my commits w/o validating it here. So I'll dismiss it but, in fact, the changes have been taken into account.

Hombre57:
Sorry but I'll commit this PR, it integrates the requested changes. I've used this branch for some times now w/o problem. In any case, a fix should be straightforward.

PR #5447, opened 2019-09-10
Soft Light: speedup and reduce memory usage

heckflosse:
This speeds up and reduces memory usage of soft light tool.

Memory usage is reduced by 12 * width * height byte
Speedup is ~2x

```
12 MP file before: 240 ms, after 120 ms
50 MP file before: 970 ms, after 485 ms
```

### DIFF
--- a/rtengine/ipsoftlight.cc
+++ b/rtengine/ipsoftlight.cc
@@ -22,8 +22,6 @@
 #include "improcfun.h"
 
 #include "procparams.h"
-#define BENCHMARK
-#include "StopWatch.h"
 
 namespace {
 
@@ -60,7 +58,6 @@ void rtengine::ImProcFunctions::softLight(LabImage *lab)
     if (!params->softlight.enabled || !params->softlight.strength) {
         return;
     }
-    BENCHFUN
 
     const TMatrix wprof = ICCStore::getInstance()->workingSpaceMatrix(params->icm.workingProfile);
     const float wp[3][3] = {
--- a/rtengine/ipsoftlight.cc
+++ b/rtengine/ipsoftlight.cc
@@ -20,17 +20,17 @@
  */
 
 #include "improcfun.h"
+
 #include "procparams.h"
 #define BENCHMARK
 #include "StopWatch.h"
-namespace rtengine {
 
 namespace {
 
 inline float sl(float blend, float x)
 {
-    if (!OOG(x)) {
-        float v = Color::gamma_srgb(x) / MAXVALF;
+    if (!rtengine::OOG(x)) {
+        float v = rtengine::Color::gamma_srgb(x) / rtengine::MAXVALF;
         // using Pegtop's formula from
         // https://en.wikipedia.org/wiki/Blend_modes#Soft_Light
         // const float orig = 1.f - blend;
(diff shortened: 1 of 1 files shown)


rtengine/ipsoftlight.cc
```diff
@@ -18,33 +19,40 @@
...
-        v = v2 + v22 - v22 * v;
-        x = blend * Color::igamma_srgb(v * MAXVALF) + orig * x;
+        // const float orig = 1.f - blend;
+        // float v2 = v * v;
+        // float v22 = v2 * 2.f;
+        // v = v2 + v22 - v22 * v;
+        // return blend * Color::igamma_srgb(v * MAXVALF) + orig * x;
+
+        // using optimized formula (heckflosse67@gmx.de)
+        return intp(blend, Color::igamma_srgb(v * v * (3.f - 2.f * v) * MAXVALF), x);
```
Floessie: I tried to wrap my head around the `3.f`, but I guess it's the `1.f` in `orig` somehow.

Floessie (approved):
All fine. :+1: 

I'll push some more `const`. :wink:

rtengine/ipsoftlight.cc
heckflosse:
```
        // using Pegtop's formula from
        // https://en.wikipedia.org/wiki/Blend_modes#Soft_Light
        // https://en.wikipedia.org/wiki/Blend_modes#Soft_Light
        // float v2 = v * v;
        // float v22 = v2 * 2.f;
        // v = v2 + v22 - v22 * v;
```
So
```
f(v) = v2 + v22 - v22 * v

f(v) = v * v + 2.f * v * v - 2.f * v * v * v

f(v) = 3.f * v * v - 2.f * v * v * v

f(v) = v * v * (3.f - 2.f * v)
```

Floessie: Ah, I forgot a v² when reducing. :facepalm: Sorry for the noise.

heckflosse: @Floessie We might go back to the original formula...

original formula on Bulldozer
```
        vmulss  xmm1, xmm0, xmm0
        vaddss  xmm2, xmm1, xmm1
        vaddss  xmm1, xmm1, xmm2
        vfnmadd132ss    xmm0, xmm1, xmm2
        ret
```

new formula on Bulldozer
```
        vmovss  xmm2, DWORD PTR .LC1[rip]
        vmulss  xmm1, xmm0, xmm0
        vfnmadd132ss    xmm0, xmm2, DWORD PTR .LC0[rip]
        vmulss  xmm0, xmm1, xmm0
        ret
```

Though on Bulldozer the sse code would be used anyway... and the new formula has less dependencies

Floessie: @heckflosse Only a micro-benchmark will tell. Although the new formula has two loads, those must be present earlier in the old formula, too. Less dependencies: You mean `vfnmadd132ss` only depending on `xmm0` and `xmm2`?

heckflosse: original has a dependency between the first mul and the add. New has no dependencies in the first two instructions.

> Although the new formula has two loads, those must be present earlier in the old formula, too.

Original code has only 2.f as a constant for the multiplication, which was replaced by an add.

Anyway, I won't spend more time on this. Any objections to merge this after removing the BENCHFUN?

Floessie: Of course not. :+1:

PR #5456, opened 2019-09-17
Speedup Dehaze

heckflosse:
PR created to get LGTM check.
More commits will follow.


heckflosse:
The first commit speeds up the guided filter, which should also speedup the shadows/highlights tool.

I measured dehaze on a 50 MP file
before: 2480 ms
after:  1790 ms

heckflosse:
Edit: I forgot to mention that the above commit also reduces memory usage of the guided filter by width * height * 4 byte

heckflosse:
Last commit reduces processing time of dehaze for the above test case to
1200 ms

heckflosse:
Hmm, there is a bug concerning right border. Looking...

heckflosse:
Oh, LGTM check failed. Seems I used some new omp features which are not supported in gcc 8.3...

heckflosse:
With last commit the processing time for the above test case is reduced to
1060 ms

rtengine/boxblur.h
```diff
@@ -312,6 +313,222 @@ template<class T, class A> void boxblur (T** src, A** dst, T* buffer, int radx,
...
+        }
+        return;
+    }
+
+    constexpr int numCols = 8; // process numCols columns at once for better usage of L1 cpu cache
+#ifdef _OPENMP
+    #pragma omp parallel if (multiThread)
+#endif
+    {
+        float* const buffer = new float[std::max(W, 8 * H)];
```
Floessie: Should we still allocate raw buffers on heap in new/refactored code? Would using a `std::vector<float>` hurt?

rtengine/boxblur.h
```diff
@@ -312,6 +313,222 @@ template<class T, class A> void boxblur (T** src, A** dst, T* buffer, int radx,
...
+                tempval += src[row][j];
+            }
+
+            tempval /= len;
+            dst[row][0] = tempval;
+
+            for (int col = 1; col <= radius; col++) {
+                lineBuffer[col] = src[row][col];
+                dst[row][col] = tempval = (tempval * len + src[row][col + radius]) / (len + 1);
+                len ++;
```
Floessie: `++len;`

I for one refrain from multiple assignments per line (`dst[row][col] = tempval =...`) but it's readable here...

rtengine/boxblur.h
```diff
@@ -312,6 +313,222 @@ template<class T, class A> void boxblur (T** src, A** dst, T* buffer, int radx,
...
+                dst[row][col] = tempval = (tempval * len - lineBuffer[col - radius - 1]) / (len - 1);
+                len --;
+            }
+        }
+
+        //vertical blur
+#ifdef __SSE2__
+        vfloat (* const rowBuffer)[2] = (vfloat(*)[2]) buffer;
+        vfloat leninitv = F2V(radius + 1);
+        vfloat onev = F2V(1.f);
```
Floessie: `const`

rtengine/boxblur.h
```diff
@@ -312,6 +313,222 @@ template<class T, class A> void boxblur (T** src, A** dst, T* buffer, int radx,
...
+#else
+        float (* const rowBuffer)[8] = (float(*)[8]) buffer;
+#ifdef _OPENMP
+        #pragma omp for nowait
+#endif
+
+        for (int col = 0; col < W - numCols + 1; col += 8) {
+            float len = radius + 1;
+
+            for(int k = 0; k < numCols; k++) {
```
Floessie: Space.

rtengine/boxblur.h
```diff
@@ -312,6 +313,222 @@ template<class T, class A> void boxblur (T** src, A** dst, T* buffer, int radx,
...
+                rowBuffer[0][k] = dst[0][col + k];
+            }
+
+            for (int i = 1; i <= radius; i++) {
+                for(int k = 0; k < numCols; k++) {
```
Floessie: Space.

rtengine/boxblur.h
```diff
@@ -312,6 +313,222 @@ template<class T, class A> void boxblur (T** src, A** dst, T* buffer, int radx,
...
+                    dst[0][col + k] += dst[i][col + k];
+                }
+            }
+
+            for(int k = 0; k < numCols; k++) {
```
Floessie, line 456: Again.

rtengine/guidedfilter.cc
```diff
@@ -78,15 +63,7 @@ int calculate_subsampling(int w, int h, int r)
...
-
-    if (subsampling <= 0) {
-        subsampling = calculate_subsampling(W, H, r);
-    }
-
-    enum Op { MUL, DIVEPSILON, ADD, SUB, ADDMUL, SUBMUL };
+    enum Op {MUL, DIVEPSILON, SUBMUL};
 
     const auto apply =
         [=](Op op, array2D<float> &res, const array2D<float> &a, const array2D<float> &b, const array2D<float> &c=array2D<float>()) -> void
```
Floessie: We should avoid "catch all" captures.

rtengine/ipdehaze.cc
```diff
@@ -229,100 +269,143 @@ void ImProcFunctions::dehaze(Imagefloat *img)
...
         extract_channels(img, R, G, B, patchsize, 1e-1, multiThread);
-    
-        patchsize = max(max(W, H) / 600, 2);
-        npatches = get_dark_channel(R, G, B, dark, patchsize, nullptr, false, multiThread);
-        DEBUG_DUMP(dark);
 
-        max_t = estimate_ambient_light(R, G, B, dark, patchsize, npatches, ambient);
+        {
+            constexpr int sizecap = 200;
+            float r = float(W)/float(H);
```
Floessie: Can't see any modification of `r` in the diff below. If so, it should read

```c++
const float r = static_cast<float>(W) / static_cast<float>(H);
```

rtengine/ipdehaze.cc
```diff
@@ -229,100 +269,143 @@ void ImProcFunctions::dehaze(Imagefloat *img)
 
     int patchsize = max(int(5 / scale), 2);
     float ambient[3];
-    array2D<float> &t_tilde = dark;
     float max_t = 0.f;
```
Floessie: `max_t` is a bad variable name, as `_t` usually denotes a type.

rtengine/ipdehaze.cc
```diff
@@ -229,100 +269,143 @@ void ImProcFunctions::dehaze(Imagefloat *img)
...
+        {
+            constexpr int sizecap = 200;
+            float r = float(W)/float(H);
+            const int hh = r >= 1.f ? sizecap : sizecap / r;
+            const int ww = r >= 1.f ? sizecap * r : sizecap;
+
+            if (W <= ww && H <= hh) {
+                // don't rescale small thumbs
+                array2D<float> D(W, H);
+                int npatches = get_dark_channel_downsized(R, G, B, D, 2, multiThread);
```
Floessie: `const` (same below).

rtengine/ipdehaze.cc
```diff
@@ -229,100 +269,143 @@ void ImProcFunctions::dehaze(Imagefloat *img)
...
-            // ... t >= tu to avoid values > 1
-            float tu = t0 - teps;
-            for (int c = 0; c < 3; ++c) {
-                if (ambient[c] < 1) {
-                    tu = max(tu, (rgb[c] - ambient[c])/(1.f - ambient[c]));
-                }
+            const vfloat tlv = onev - vminf(r / ambient0v, vminf(g / ambient1v, b / ambient2v));
+            const vfloat mtv = vmaxf(LVFU(dark[y][x]), vmaxf(tlv + tepsv, t0v));
+            if (params->dehaze.showDepthMap) {
+                const vfloat valv = vclampf(onev - mtv, ZEROV, onev) * cmaxChannelv;
```
Floessie, line 388: Adding a `ONEV` makes more and more sense.

rtengine/procparams.h
```diff
@@ -1343,7 +1343,7 @@ struct DehazeParams {
     int strength;
     bool showDepthMap;
     int depth;
-
+    bool luminance;
```
Floessie, line 1346: This killed a blank line which seperated data from functions.

rtengine/procparams.cc
```diff
@@ -3240,7 +3242,7 @@ int ProcParams::save(const Glib::ustring& fname, const Glib::ustring& fname2, bo
         saveToKeyfile(!pedited || pedited->dehaze.strength, "Dehaze", "Strength", dehaze.strength, keyFile);        
         saveToKeyfile(!pedited || pedited->dehaze.showDepthMap, "Dehaze", "ShowDepthMap", dehaze.showDepthMap, keyFile);        
         saveToKeyfile(!pedited || pedited->dehaze.depth, "Dehaze", "Depth", dehaze.depth, keyFile);        
-
```
Floessie, line 3243: Add this seperating blank line again, please.

rtengine/boxblur.h
heckflosse: I used raw buffers here because I later reinterprete them e.g. as vfloat buffers.
I dont know how's the best way to do this using std::vector<float>, std::vector<vfloat>.
Maybe you can help on this?

Floessie: Well, at least, it could be a `std::unique_ptr<>`...

Regarding `std::vector<>`: it has gained a [`data()`](https://en.cppreference.com/w/cpp/container/vector/data) function, which should be the one you're after.

rtengine/ipdehaze.cc
heckflosse: Changed

rtengine/ipdehaze.cc
heckflosse: Changed

rtengine/procparams.h
heckflosse: Changed

rtengine/procparams.cc
heckflosse: Done

rtengine/boxblur.h
heckflosse: Both fixed.

rtengine/boxblur.h
heckflosse: Changed to std::unique_ptr<>

rtengine/boxblur.h
heckflosse: Done

rtengine/boxblur.h
heckflosse: Done

rtengine/boxblur.h
heckflosse: Done

rtengine/boxblur.h
heckflosse: Done

rtengine/guidedfilter.cc
heckflosse: Fixed

rtengine/ipdehaze.cc
heckflosse: Fixed

rtengine/ipdehaze.cc
heckflosse: The ZEROV is an exception because it's a single instruction without memory access (a simple xor) while F2V(1.f) usually has a memory load.

heckflosse:
With last speedup processing time for the 50 MP file now is
~995 ms

Issue #5434, opened 2019-09-01
Hasselblad H6D-100cMS pixelshift files do not decode correctly


heckflosse:
I will try to fix that. I will start with a code review of the hasselblad_load_raw() function to get better readable code.

Thanatomanic:
The effect of the wrong decode is actually quite nice. Can't we keep it? 😉 

![image](https://user-images.githubusercontent.com/6567747/64322065-1ede4900-cfc2-11e9-9c7e-6c57046d4643.png)

ahoenselaar:
The right-shifting behavior in the multi-shot codepath looks strange:
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/dcraw.cc#L2469
Why encode the noisy LSB only to shift it out eventually?
It is also not clear how high local contrast can be encoded correctly. After the right-shift, the Huffman encoded diff can correct the predicted pixel value only by +/- 16,384, or half the required range.
In fact, the artifacts do appear after decoding highlights.

ahoenselaar:
Removing the right-shifting brings the levels in line with Phocus. Adjusting the predictor to match lossless JPEG predictor 5 fixes the artifact. It raises a new issue, though, as the predicted values wraps around in very bright parts of the image. Likely, the predictor used by Hasselblad has some additional tweaks.

https://github.com/ahoenselaar/RawTherapee/commit/7885e113b8835c367c8150c2a400ec755cbca1c2

![image](https://user-images.githubusercontent.com/3757431/66264972-16448280-e7bb-11e9-9349-493824bbd584.png)

heckflosse:
@ahoenselaar Thanks for working on this issue :+1:
Unfortunately I can not test atm because I'm on holidays and my laptop has only 3GB RAM, which is not enough for this huge files

ahoenselaar:
https://github.com/ahoenselaar/RawTherapee/commit/de1c04b6336aef0cf52a477b7924952995c7c32e  reverts the predictor to the same one used for non-pixelshift files and results in correct decoding without artifacts. I wonder why the predictor was changed in the original dcraw commit that added pixelshift support.
![Screenshot_2019-10-19_11-35-14](https://user-images.githubusercontent.com/3757431/67150505-b6db7d80-f264-11e9-8f99-01c41a2e130e.png)

PR #5464, opened 2019-09-20
Improve readability of hasselblad_load_raw() code, #5434

heckflosse:
I would like to merge this before I try to find the reason why the Hasselblad pixelshift files don't decode correctly. @Floessie Could you review it please?

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -2440,10 +2440,10 @@ void CLASS hasselblad_load_raw()
         }
         for (int col = 0; col < raw_width; col += 2) {
             for (int s = 0; s < tiff_samples * 2; s += 2) {
-                int len[2];
-                for (int c = 0; c < 2; ++c) {
-                    len[c] = ph1_huff(jh.huff[0]);
-                }
+                const int len[2]= {
+                    ph1_huff(jh.huff[0]),
+                    ph1_huff(jh.huff[0])
+                };
                 for (int c = 0; c < 2; ++c) {
                     diff[s + c] = hb_bits(len[c]);
                     if ((diff[s + c] & (1 << (len[c] - 1))) == 0) {
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -2417,59 +2417,78 @@ void CLASS hasselblad_correct()
 
 void CLASS hasselblad_load_raw()
 {
-  struct jhead jh;
-  int shot, row, col, *back[5], len[2], diff[12], pred, sh, f, s, c;
-  unsigned upix, urow, ucol;
-  ushort *ip;
+    struct jhead jh;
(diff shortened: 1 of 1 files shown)


rtengine/dcraw.cc
```diff
@@ -2417,59 +2417,78 @@ void CLASS hasselblad_correct()
...
+    cblack[6] >>= sh;
+    const int shot = LIM(shot_select, 1, tiff_samples) - 1;
+    for (int row = 0; row < raw_height; ++row) {
+        for (int c = 0; c < 4; ++c) {
+            back[(c + 3) & 3] = back[c];
+        }
+        for (int col = 0; col < raw_width; col += 2) {
+            for (int s = 0; s < tiff_samples * 2; s += 2) {
+                int len[2];
+	            for (int c = 0; c < 2; ++c) {
```
Floessie: Indentation.

Is `ph1_huff()` stateful, or why is it computed twice? Also, couldn't this just be

```c++
const int len[2]= {
    ph1_huff(jh.huff[0]),
    ph1_huff(jh.huff[0])
};

File: rtengine/dcraw.cc [hunk 2]
```diff
@@ -2417,59 +2417,78 @@ void CLASS hasselblad_correct()
...
+                    }
+                    if (diff[s + c] == 65535) {
+                        diff[s + c] = -32768;
+                    }
+                }
+            }
+            for (int s = col; s < col + 2; ++s) {
+	            int pred;
+	            if (col) {
+                    pred = back[2][s - 2];
```
Floessie, line 2460: Indentation.

Floessie (changes requested):
Indentation is odd but the code is much more readable. Didn't check every transposition...

rtengine/dcraw.cc
heckflosse: ph1_huff() reads from file changes filepos. I will change it to the const :+1:

heckflosse: @Floessie Is it guaranteed that the first ph1_huff call in the assignment is executed before the second?

Floessie: I guess so. From [here](https://en.cppreference.com/w/cpp/language/list_initialization):

> Every initializer clause is sequenced before any initializer clause that follows it in the braced-init-list. This is in contrast with the arguments of a function call expression, which are unsequenced.

And from [here](https://en.cppreference.com/w/c/language/array_initialization):

> The order of evaluation of subexpressions in an array initializer is indeterminately sequenced in C (but not in C++ since c++11)

HTH,
Flössie

heckflosse: Thank you :+1: Fixed

rtengine/dcraw.cc
heckflosse: Fixed.

PR #5477, opened 2019-09-26
reviewed boxblur code and usage

heckflosse:
I made a review and some speedups for boxblur code
PR is just to get the LGTM check

@Desmis Can you please test whether retinex still works as expected (it should also be a bit faster now)?


Desmis:
@heckflosse 
Ingo
I tested on 2 very different images
* No differences in TIF
* small speed-up about 10%

Good job  :)

Jacques

PR #5479, opened 2019-09-29
Add VERSION and SOVERSION to libraries.

### LINKED ISSUE #5402: Packaging: unversioned libraries installed in /usr/lib
RawTherapee version `5.6` OS `clearlinux` version `30700`.

Our packaging tool (autospec) detects that rawtherapee's cmake files are installing libraries (not `dlopen`) files as unversioned `.so`'s into `/usr/lib64`. 

libraries that are supposed to be linked should be versioned by default (libfoo.so.1). The following 2 libraries are included in rawtherapee and are not versioned:
`/usr/lib64/librtengine.so`
`/usr/lib64/librtexif.so`

Note that it's also not picking up the correct default `-DLIBDIR` and `CMAKE_INSTALL_LIBDIR` either.

Note: other distros seem to entirely ignore this problem, but that doesn't mean it's not an issue.

mattiaverga:
This will make `rtengine` and `rtexif` fully versioned.
I've set the starting VERSION to `5.7.0` and the starting SOVERSION to `5`.

Fixes #5402

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,14 @@ set(PROC_TARGET_NUMBER 0 CACHE STRING "Selected target processor from the list a
 # Some Linux distros build with -O2 instead of -O3. We explicitly enable auto vectorization by using -ftree-vectorize
 set(RTENGINE_CXX_FLAGS "-ftree-vectorize" CACHE STRING "Special compilation flags for RTEngine")
 
+# Set VERSION and SOVERSION for rtengine library
+set(RTENGINE_VERSION "5.7.0")
+set(RTENGINE_SOVERSION "5")
+
+# Set VERSION and SOVERSION for rtexif library
+set(RTEXIF_VERSION "5.7.0")
+set(RTEXIF_SOVERSION "5")
+
 # Loads the ProcessorTargets list:
 include(ProcessorTargets.cmake)
 set(PROC_LABEL "undefined" CACHE STRING "Target processor label, unused if PROC_TARGET_NUMBER = 0 or 2")
--- a/rtengine/CMakeLists.txt
+++ b/rtengine/CMakeLists.txt
@@ -172,7 +172,12 @@ if(BUILD_SHARED_LIBS)
     install(TARGETS rtengine DESTINATION ${LIBDIR})
 endif()
 
-set_target_properties(rtengine PROPERTIES COMPILE_FLAGS "${RTENGINE_CXX_FLAGS}")
+set_target_properties(rtengine
+    PROPERTIES
+        COMPILE_FLAGS "${RTENGINE_CXX_FLAGS}"
+        VERSION "${RTENGINE_VERSION}"
+        SOVERSION "${RTENGINE_SOVERSION}"
+    )
 
 target_link_libraries(rtengine rtexif
     ${EXPAT_LIBRARIES}
--- a/rtexif/CMakeLists.txt
+++ b/rtexif/CMakeLists.txt
@@ -15,3 +15,9 @@ include_directories(BEFORE "${CMAKE_CURRENT_BINARY_DIR}")
 if(BUILD_SHARED_LIBS)
     install(TARGETS rtexif DESTINATION ${LIBDIR})
 endif()
+
+set_target_properties(rtexif
+    PROPERTIES
+        VERSION "${RTEXIF_VERSION}"
+        SOVERSION "${RTEXIF_SOVERSION}"
+    )


Beep6581:
@heckflosse @CarVac @LebedevRI et al. what are your thoughts?

LebedevRI:
I don't have enough context to answer here, but seems harmless to me?

Floessie:
@mattiaverga I'm wondering why RT is compiled with SOs in your case. When building RT myself or even in the official Debian package there is not a single `.so`. After all, those libs are only meant for RT and there's no other consumer.

mattiaverga:
@Floessie from Fedora Packaging Guidelines:
`Whenever possible (and feasible), Fedora packages containing libraries SHOULD build them as shared libraries.`

I assume I could set `-DBUILD_SHARED_LIBS:BOOL=OFF`, but that's in contrast with the above paragraph. And I don't know much about C+ and libraries to know the pros and cons of one solution vs another.

CarVac:
I guess the only issue with this would be that it places a burden on the RT maintainers to care about ABI compatibility and version for these internal-use libraries where there was none before.

I'm not sure why the packaging guidelines want this, but it says an exception can be made so long as "the nature of the deviation and the reason behind it…are documented in the specfile".

Floessie:
@CarVac Exactly. IMHO the unfortunate aspect of having shared libs in the subdirs is due to the (mis)use of CMake which makes it easy to have libs per dir. I think we should change that or force static libs for the subdirs.

@Hombre57 @heckflosse @Beep6581 @cryptomilk Opinions?

Floessie:
@mattiaverga 
> I assume I could set -DBUILD_SHARED_LIBS:BOOL=OFF, but that's in contrast with the above paragraph. And I don't know much about C+ and libraries to know the pros and cons of one solution vs another.

What exactly are the parameters to CMake in your case?

cryptomilk:
Are those private or public libraries?

Floessie:
@cryptomilk The relevant code is [here](https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/CMakeLists.txt#L167), I guess.

cryptomilk:
a) Are the libraries meant to be used by other projects? If yes, they should have header files. If not, they should be private. No need for symbol versioning. Then put them in a dir and use rpath.

b) Start using include(GNUInstallDirs) -> https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

Floessie:
@cryptomilk 
> a) Are the libraries meant to be used by other projects? If yes, they should have header files. If not, they should be private. No need for symbol versioning. Then put them in a dir and use rpath.

No, they aren't. But why have shared private libs in the first place? Longer startup times and no global LTO? :wink: 

> b) Start using include(GNUInstallDirs) -> https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html

Okay.

cryptomilk:
> No, they aren't. But why have shared private libs in the first place? Longer startup times and no global LTO?

Cleaner APIs?

Floessie:
> Cleaner APIs?

My question was more targeted at the sense of shared private libs in RT. :grin: 

In general I agree with you. Good point.

mattiaverga:
@Floessie 
> What exactly are the parameters to CMake in your case?

```
%{cmake} %{?BUILD_TESTS} \
  -DCMAKE_INSTALL_PREFIX=%{_prefix} \
  -DLIBDIR=%{_libdir} \
  -DCMAKE_BUILD_TYPE=release \
  -DBUILD_SHARED_LIBS:BOOL=ON \
  -DAUTOMATED_BUILD_SYSTEM:BOOL=ON \
  -DCACHE_NAME_SUFFIX="" \
  -DCMAKE_CXX_FLAGS="$RPM_OPT_FLAGS" \
  -DCMAKE_C_FLAGS="$RPM_OPT_FLAGS" \
  -DWITH_SYSTEM_KLT=ON .

```
Note that `-DBUILD_SHARED_LIBS:BOOL=ON` seems to be the default, since if I omit that SOs are been built.

If you don't want to have the burden of care about ABI changes, it's not a problem, I can simply do that downstream with a script in the specfile. Or I can ask Fedora Committee for an exception to build RT libs not as shared libs.
I think the original reporter of #5402, which was on clearlinux, can also easily solve his problem with the second solution. @ahkok

cryptomilk:
@mattiaverga : Speaking with my Fedora hat on:

If those libraries are not public libraries (meant for consumption by 3rd parties), then they should be installed in `/usr/lib{,64}/rawtherapee/libx.so and linked using RPATH.

If they are public libraries and a header file is provided, then the ABI should be tracked using [abimap](https://github.com/ansasaki/abimap) and SO versions be set. Just setting SO verison number is not enough.

Floessie:
@mattiaverga Thanks for the info and offer.

@cryptomilk As those libraries are **not** (I can't stress this enough) meant for the public and @Beep6581 and me not even being aware that they can be SOs under certain circumstances, wouldn't it be sufficient to add `STATIC` to the `add_library()` calls?

cryptomilk:
Yes, then just add STATIC to the add_library()

mattiaverga:
@cryptomilk speaking about Fedora, I think the best way is like your first suggestion, to put those two .so files under a subdir, right? Guidelines strongly suggest to avoid static libraries.

@Floessie I'm closing this PR since it isn't the right way, I apologize for the noise

Floessie:
@mattiaverga Thanks for making us aware of the SOs. :+1: 

> Guidelines strongly suggest to avoid static libraries.

Of course, if they are meant as publicly available APIs with the corresponding headers. As mentioned earlier, those libs in `rtexif` and `rtengine` are merely a means to group the `.o` files for us. There are no headers for the public which you could use outside RT. Thus, please don't export any `.so`s or `.a` from RT to `/usr...`.

PR #5483, opened 2019-10-03
Whitespace cleanup

Beep6581:
Cleans up whitespace (tab -> space), alignment and DOS line endings.


Beep6581:
I also tested removal of duplicate klt entries from `rtengine/CMakeLists.txt` while test-compiling, and forgot to undo that before committing, so that snuck into this PR.
See:
https://github.com/Beep6581/RawTherapee/pull/3912#issuecomment-537849258

I can undo the klt removal when merging, or I can keep it...

Issue #5204, opened 2019-03-03
Panasonic Lumix DC-S1 raw support


ilya-varivchenko:
These are raw files from Panasonic DC-S1

https://filebin.net/ugxofhllq0zh51ho/Panasonic-S1-NR-OFF.zip?t=xepqu64i
https://filebin.net/y7tzpq5r31yrxlen/Panasonic-S1-NR-ON.zip?t=10njyogo

Maybe RawTherapee will be the first converter to support this cool camera. :)

Beep6581:
Ping @iliasg

LebedevRI:
> These are raw files from Panasonic DC-S1

If you have that camera, it would be great if you could contribute the full sample set to https://raw.pixls.us/

ilya-varivchenko:
I followed these instructions
http://rawpedia.rawtherapee.com/Adding_Support_for_New_Raw_Formats

what additional info is needed for camera support ?

Beep6581:
@ilya-varivchenko as far as RawTherapee is concerned the set you uploaded is fine. But RPU is a separate project which several open-source programs including RawTherapee benefit from, so it would be great if you uploaded a normal photo (not a clipped white-frame) to RPU. A description of what sort of photo(s) to upload can be found on the RPU site.

ilya-varivchenko:
This is ColorChecker shot in daylight
https://filebin.net/ubdjzh2f3pzrkx6u/P1032889.RW2?t=l5yibq26

Beep6581:
@ilya-varivchenko  do you/could you also shoot the target in incandescent light? http://rawpedia.rawtherapee.com/How_to_create_DCP_color_profiles#Shooting_the_Color_Target

Beep6581:
Ping @ilya-varivchenko

ilya-varivchenko:
I have several samples made in  incandescent light, but they are of poor quality. :(
I will upload these images later today.

ilya-varivchenko:
Try these ones pls:
https://drive.google.com/open?id=1wQEkXsAoD9aBkaBu2JhzCs-09-dt-g2N

Beep6581:
Hi @ilya-varivchenko 
Indeed the shots are not usable, the issues are:
* The target is not lit evenly as the light source is too close to the target.
* Some of the shots have motion blur so the program cannot isolate dust spots and scratches - use a tripod.
[...]

PR #5515, opened 2019-11-06
Decode Panasonic DC-S1 and DC-S1R files

heckflosse:
(pixelshift files are not supported yet), #5204

PR created to get lgtm check


heckflosse:
Merged to get more tests.

PR #5531, opened 2019-11-18
findHotDeadPixels: speedup and reduced memory usage

heckflosse:
I measured using a 189 MPixel file.

Before: ~840 ms
After: ~355 ms

Memory usage is reduced from width * height * 4 byte to width * 5 * t * 4 byte, where t is the number of parallel threads.

### DIFF
--- a/rtengine/badpixels.cc
+++ b/rtengine/badpixels.cc
@@ -22,7 +22,8 @@
 #include "pixelsmap.h"
 #include "rawimage.h"
 #include "rawimagesource.h"
-
+#define BENCHMARK
+#include "StopWatch.h"
 namespace
 {
 unsigned fc(const unsigned int cfa[2][2], int r, int c) {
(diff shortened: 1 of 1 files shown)


rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
 /*  Search for hot or dead pixels in the image and update the map
  *  For each pixel compare its value to the average of similar color surrounding
  *  (Taken from Emil Martinec idea)
- *  (Optimized by Ingo Weyrich 2013 and 2015)
+ *  (Optimized by Ingo Weyrich 2013, 2015 and 2019)
```
Floessie: ", and 2019"

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
-        #pragma omp for schedule(dynamic,16) nowait
+        // note, static scheduling is important in this implementation
+        #pragma omp for schedule(static) nowait
 #endif
 
         for (int i = 2; i < H - 2; i++) {
-            for (int j = 2; j < W - 2; j++) {
+            if (firstRow == -1) {
+                firstRow = i;
+                if (firstRow == 2) {
```
Floessie: Surely I'm missing something, but how can `firstRow` be something other than `-1` and `2`?

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
+                            const float temp = median(rawData[row - 2][j - 2], rawData[row - 2][j], rawData[row - 2][j + 2],
+                                                      rawData[row][j - 2], rawData[row][j], rawData[row][j + 2],
+                                                      rawData[row + 2][j - 2], rawData[row + 2][j], rawData[row + 2][j + 2]);
+                            cfablur[destRow][j] = rawData[row][j] - temp;
+                        }
+                    }
+                }
+            }
+            lastRow = i;
+            const int destRow = i % 5;
```
Floessie, line 514: Interestingly, `DIV` is still faster than `IDIV` on newer [Intels](https://www.agner.org/optimize/instruction_tables.pdf). Doesn't matter here, just FYI.

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
+
+                    if (pixdev == 0.f) {
+                        continue;
+                    }
 
-            for (int i = 2; i < H - 2; ++i) {
-                for (int j = 0; j < 2; ++j) {
-                    cfablur[i * W + j] = 0.f;
-                }
+                    if ((!findDeadPixels) && pixdev < 0) {
```
Floessie: `if (!findDeadPixels && pixdev < 0.f) {`

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
+                        continue;
+                    }
 
-                for (int j = W - 2; j < W; ++j) {
-                    cfablur[i * W + j] = 0.f;
-                }
-            }
+                    if ((!findHotPixels) && pixdev > 0) {
```
Floessie: Same here.

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
-                    cfablur[i * W + j] = 0.f;
-                }
+                    pixdev = fabsf(pixdev);
+                    float hfnbrave = -pixdev;
+
+#ifdef __SSE2__
+                    // sum up 5*4 = 20 values using SSE
+                    // 10 fabs function calls and 10 float additions with SSE
+                    vfloat sum1 = vabsf(LVFU(cfablur[rrm2][cc - 2])) + vabsf(LVFU(cfablur[rrm1][cc - 2]));
+                    vfloat sum2 = vabsf(LVFU(cfablur[rr0][cc - 2])) + vabsf(LVFU(cfablur[rrp1][cc - 2]));
```
Floessie: `const`

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
             }
         }
 
-#ifdef _OPENMP
-        #pragma omp barrier // barrier because of nowait clause above
-
-        #pragma omp for reduction(+:counter) schedule(dynamic,16)
+        if (lastRow > 0 && lastRow < H - 2) {
+            //cfa pixel heat/death evaluation
+            for (int rr = lastRow - 1; rr < lastRow + 1; ++rr) {
```
Floessie, line 551: So basically only two rows...

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
+                const int i = rr + 2;
+                const int destRow = i % 5;
+                if (i >= H - 2) {
+                    for (int j = 2; j < W - 2; j++) {
+                        cfablur[destRow][j] = 0.f;
+                    }
+                } else {
+                    int j = 2;
+#ifdef __SSE2__
+                    for (; j < W - 5; j += 4) {
```
Floessie: This looks familiar. Any chance to factor it out?

rtengine/badpixels.cc
```diff
@@ -445,126 +446,227 @@ int RawImageSource::interpolateBadPixelsXtrans(const PixelsMap &bitmapBads)
...
-                    continue;
-                }
+                    if (pixdev == 0.f) {
+                        continue;
+                    }
 
-                if ((!findDeadPixels) && pixdev < 0) {
-                    continue;
-                }
+                    if ((!findDeadPixels) && pixdev < 0) {
```
Floessie: Same as above.

rtengine/badpixels.cc
heckflosse: The scheduling splits the processing in chunks. Let's assume 800 rows and 8 threads. Because the scheduling is static we get 8 chunks of (in this case) equal size, each having its own i and so its own firstrow (0, 100, 200, ... , 700)

rtengine/badpixels.cc
heckflosse: Same here, each chunk has its own lastRow, so two rows per chunk

heckflosse:
@Floessie Thanks for the review. I already fixed some things localy and made another speedup.
I also detected that there is a bug somewhere, which leads to different output on the same raw. Have to check this...

rtengine/badpixels.cc
heckflosse: Done

rtengine/badpixels.cc
heckflosse: Done

rtengine/badpixels.cc
heckflosse: Done

rtengine/badpixels.cc
heckflosse: Done

rtengine/badpixels.cc
heckflosse: Done

rtengine/badpixels.cc
heckflosse: Done

heckflosse:
Output is consistent now.
Processing time for the 189 MPixel file now is ~295 ms

PR #5539, opened 2019-11-24
Fix linking against system klt

### LINKED ISSUE #5525: Linking against system version of KLT fails on Fedora 31
See here: https://discuss.pixls.us/t/problems-building-current-git-on-fedora-31

To reproduce:
1. Clean building environment, clean clone of `dev`
2. ```klt-devel-1.3.4``` installed (which is the latest version as of writing)
3. Build RT using ```-DWITH_SYSTEM_KLT="ON"```
4. Linking fails with ```undefined reference to `KLTCreateTrackingContext'``` or others.

Can anyone test on different Linux distro's?

mattiaverga:
This should fix #5525 and also assures that system KLT sources are used when the relevant option is specified.

I'm very unfamiliar with c++/cmake library linking, so please review this PR carefully... I did test locally building with and without system KLT and everything works, but...

### DIFF
--- a/rtengine/CMakeLists.txt
+++ b/rtengine/CMakeLists.txt
@@ -12,6 +12,11 @@ include_directories(${EXTRA_INCDIR}
     ${LENSFUN_INCLUDE_DIRS}
     ${RSVG_INCLUDE_DIRS}
 )
+if(NOT WITH_SYSTEM_KLT)
+    include_directories("${CMAKE_SOURCE_DIR}/rtengine/klt")
+else()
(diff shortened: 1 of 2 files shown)


Beep6581:
klt is not shipped by Sabayon so I can't check whether using it works, but I can confirm that the branch does not break compilation using the RT-provided klt, so +1 for merge from me.

Beep6581:
@heckflosse ok to merge?

Beep6581:
No objections - merged.

PR #5549, opened 2019-12-02
Multithread lensfun vignetting correction

heckflosse:
Processing time for a 24 MPixel file went down from ~230 to ~40 ms on my 8-core.

I tested for differences and races. Found none.
I also removed one Mutex because imho it's not needed.


Floessie (approved):
Compiles and runs fine on Debian 10 i386 and PPC. I tried to crash it on AMD64 but failed. :+1:

PR #5556, opened 2019-12-05
lmmse demosaic review

heckflosse:
PR created for LGTM check. More commits will follow


Floessie:
Tested on Debian PPC32: No difference in output. :+1:

Maybe you could ditch the `using namespace std;`...

heckflosse:
@Floessie Thanks for testing. Merging now for 5.8. Further work will be for 5.9

Issue #5550, opened 2019-12-02
macOS CI pipeline - automated builds


Krijger:
To support macOS development, this issue is to implement a Continuous Integration pipeline. This will allow us to create macOS builds for basically each commit, and to share these with the entire comunity, making testing easier.

The following needs to be done:

- [x] create a pipeline using GitHub Actions (service of choice for now)
- [x] output a bootable artifact
- [ ] codesign and notarize the app and DMG 
- [ ] artifact are ZIP files directly, as opposed to a ZIP within a ZIP 
- [ ] create a release flow for git tags - this will make the artifact appear in releases, and not only via github actions, which incidentally gets removed after 90 days

The following is nice to have

- [ ] make the scripts in the build fail fast (`set -euo pipefail`), and working
- [ ] include build succeeding/failing widget on the readme page

--------- ORIGINAL ISSUE POST BELOW ---------

Since there are no nightly builds for Mac OS, and I see on the website that the last build had issues, I guess a pipeline that builds for Mac would be quite useful. Have you considered this?

I can start work on this, given that both Travis CI and Azure DevOps provide Mac build agents and are completely free for open source projects. I would opt for Azure in this case, since I can use that experience. But first of all, let me ask you how the current pipeline is setup. Is there continuous integration on this project?

Beep6581:
@Krijger that would be nice, please talk with @aferrero2707 as he is working on that as well.

Krijger:
Sure. Shall we keep the communication in this issue?

Any answers to the other questions though? Where do the linux nightly builds come from? And are there unit tests being run on commit anywhere?

heckflosse:
@Krijger 

Thanks for you offer to help. Very much appreciated :+1: 

> Sure. Shall we keep the communication in this issue?

Yes, please.

> Where do the linux nightly builds come from?

From @aferrero2707 

> And are there unit tests being run on commit anywhere?

No, the only thing running automatically on any pr commit is LGTM

Beep6581:
For info on automated AppImage and Windows builds, see #5551

Krijger:
@heckflosse just to be sure, the builds are executed locally on machines using the scripts mentioned in #5551 ? Because I’ve also checked the repos of aferro, but the latest activity is quite some time ago on those.

If that is the case, and the binaries are uploaded manually to the /Downloads page, we have a green field here to set these things up automatically.

[...]

PR #5557, opened 2019-12-08
Ci macos

Krijger:
For issue #5550 .
This is a first step to create CI for Mac OS. I strongly believe having CI in place and allowing all supporting operating systems feel like first citizens is important for the user base of an open source project. Now all beta mac users have to build themselves which we can also channel into improving this build for all to use.

Included for now is the build on Github Actions (free for open source). The result is that by navigating to Github Actions you are now able to download the zipped artifact. <del>The resulting DMG for now crashes however, probably due to the patch on `libiconv.2.dylib` not having been implemented yet. So, maybe you may still want to accept this PR on a branch instead, however,</del> it does not interfere with anything, so `dev` is safe.

For a list of things to do related to this PR, check #5550 

We can also think about moving Travis CI builds into here as well later on, to have all CI and releases automated in the same way.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,17 @@
+name: CI
+
+on: [push]
+
+jobs:
+  build:
+
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Run a one-line script
+      run: echo Hello, world!
+    - name: Run a multi-line script
+      run: |
+        echo Add other actions to build,
+        echo test, and deploy your project.
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,7 +13,7 @@ jobs:
     - name: test getting the git branch of the triggering ref 
       run: echo ${GITHUB_REF##*/}
     - name: Install dependencies
-      run: brew install gtk+3 gtkmm3 gtk-mac-integration adwaita-icon-theme libsigc++ little-cms2 libiptcdata fftw lensfun wget llvm cmake expat pkgconfig libomp
+      run: brew install gtk+3 gtkmm3 gtk-mac-integration adwaita-icon-theme libsigc++ little-cms2 libiptcdata fftw lensfun llvm expat pkgconfig libomp
     - name: cmake
       env:
         CMAKE_CXX_STANDARD: 11
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -42,10 +42,11 @@ jobs:
         make --jobs 8
         make install
         sudo make macosx_bundle
-        ARTIFACT=$(ls | grep .dmg)
-        echo "===== Writing ${ARTIFACT} to the ARTIFACT environment variable for use in upload-artifact step ====="
-        echo "::set-env name=ARTIFACT::${ARTIFACT}"
+        ARTIFACT=$(ls | grep RawTherapee*.zip)
+        echo "=== artifact: ${ARTIFACT}"
+        echo "::set-env name=ARTIFACT_PATH::${GITHUB_WORKSPACE}/build/${ARTIFACT}"
+        echo "::set-env name=ARTIFACT_FILE::${ARTIFACT}"
     - uses: actions/upload-artifact@v1
       with:
-        name: ${{env.ARTIFACT}}
-        path: build/${{env.ARTIFACT}}
+        name: ${{env.ARTIFACT_FILE}}
+        path: ${{env.ARTIFACT_PATH}}
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,7 @@ jobs:
         sudo make macosx_bundle
         ARTIFACT=(RawTherapee*.zip)
         echo "=== artifact: ${ARTIFACT}"
+        # defining environment variables for next step as per https://github.com/actions/starter-workflows/issues/68
         echo "::set-env name=ARTIFACT_PATH::${GITHUB_WORKSPACE}/build/${ARTIFACT}"
         echo "::set-env name=ARTIFACT_FILE::${ARTIFACT}"
     - uses: actions/upload-artifact@v1
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,7 +17,7 @@ jobs:
         CMAKE_CXX_STANDARD: 11
       run: |
         cmake \
-          -DCMAKE_BUILD_TYPE="release"  \
+          -DCMAKE_BUILD_TYPE="release" \
           -DCACHE_NAME_SUFFIX="5-${GITHUB_REF##*/}" \
           -DPROC_TARGET_NUMBER="2" \
           -DWITH_LTO="OFF" \
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,8 +10,6 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
-    - name: test getting the git branch of the triggering ref 
-      run: echo ${GITHUB_REF##*/}
     - name: Install dependencies
       run: brew install gtk+3 gtkmm3 gtk-mac-integration adwaita-icon-theme libsigc++ little-cms2 libiptcdata fftw lensfun llvm expat pkgconfig libomp
     - name: cmake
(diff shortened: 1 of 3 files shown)


Beep6581:
It would be nice to stick to Travis CI as that's what we already use for AppImage and Windows builds.

@aferrero2707 have a look.

Krijger:
I understand that having the same System for all is nice. For now I chose Github actions because of the nice interfacing with github. Importantly, it allows me to ‘publish’ build artifacts without releasing them. I was able to easily do this without having to insert S3 credentials. Also, this now works on all commits on all branches, thus you would always clearly know if a commit broke the build, providing a quicker feedback cycle than nightly builds - and people can test the build from a specific commit (and earlier) instead of the dev-commit that happened to be HEAD at the time. Also, we can expect GitHub Actions to improve a lot still.

For now, the most important thing is getting a build working for Mac OS in the first place. Once we have that we can discuss going in either direction.

Beep6581:
Sounds good!

.github/workflows/main.yml
```diff
@@ -0,0 +1,70 @@
+name: Mac OS build
```
Beep6581: @Krijger we stick to the Apple's current (2019) way of writing that: macOS

.github/workflows/main.yml
```diff
@@ -0,0 +1,70 @@
...
+    - name: cmake
+      env:
+        CMAKE_CXX_STANDARD: 11
+        # clutching at the straw https://stackoverflow.com/questions/17912092/gtk-installed-with-brew-is-not-found-when-i-try-to-compile-other-applications with /opt/X11
+        PKG_CONFIG_PATH: /usr/local/opt/libffi/lib/pkgconfig:/usr/local/opt/expat/lib/pkgconfig:/opt/X11/lib/pkgconfig
+        RAW_THERAPEE_MAJOR: '5'
+        RAW_THERAPEE_MINOR: '7'
+        C_FLAGS: -Xpreprocessor -fopenmp /usr/local/lib/libomp.dylib -I/usr/local/include -I/usr/local/opt/gdk-pixbuf/include -I/usr/local/opt/libiconv/include -I/usr/local/opt/libxml2/include -I/usr/local/opt/expat/include -I/usr/local/opt/llvm/include
+      run: |
+        REF=${GITHUB_REF##*/}
```
Beep6581, line 37: @Krijger when debugging scripts in the future, it's nice when the code comments show examples of what the input and output are meant to be. What are GITHUB_REF and REF?

.github/workflows/main.yml
```diff
@@ -0,0 +1,70 @@
...
+          -DOpenMP_CXX_FLAGS=-fopenmp=libomp \
+          -DOpenMP_C_LIB_NAMES="libomp" \
+          -DOpenMP_CXX_LIB_NAMES="libomp" \
+          -DOpenMP_libomp_LIBRARY="/usr/local/lib/libomp.dylib" \
+          -DOpenMP_C_FLAGS="${C_FLAGS}" \
+          -DOpenMP_CXX_FLAGS="${C_FLAGS}" \
+          -DCMAKE_AR="/usr/local/opt/llvm/bin/llvm-ar" \
+          -DCMAKE_RANLIB="/usr/local/opt/llvm/bin/llvm-ranlib" \
+          ..
+        make --jobs 2
```
Beep6581: @Krijger can `jobs` not be left on automatic?

.github/workflows/main.yml
```diff
@@ -0,0 +1,70 @@
...
+        make install
+        sudo make macosx_bundle
+        ARTIFACT=$(ls | grep RawTherapee*.zip)
```
Beep6581: @Krijger 
https://mywiki.wooledge.org/ParsingLs

```bash
ARTIFACT=(RawTherapee*.zip)
```

tools/osx/macosx_bundle.sh
```diff
@@ -111,13 +112,13 @@ install -d  "${RESOURCES}" \
...
 ditto "${CMAKE_BUILD_TYPE}/MacOS" "${MACOS}"
-ditto "${CMAKE_BUILD_TYPE}/Resources" "${RESOURCES}"
+ditto "Resources" "${RESOURCES}"
 
 msg "Copying dependencies from ${GTK_PREFIX}:"
 CheckLink "${EXECUTABLE}"
 
 msg "Copying library modules from ${GTK_PREFIX}:"
-ditto --arch "${arch}" {"${GTK_PREFIX}/lib","${LIB}"}/gdk-pixbuf-2.0
+ditto --arch "${arch}" {"${GDK_PREFIX}/lib","${LIB}"}/gdk-pixbuf-2.0
```
Beep6581, line 121: @Krijger what do the outer curly braces do?
**{**"${GDK_PREFIX}/lib","${LIB}"**}**/gdk-pixbuf-2.0

.github/workflows/main.yml
Krijger: I don't know. I saw it specified in every example I could find, so decided to include it here as well and specify the number of cores I know the build agent has. I guess we can simply remove it and see if the build does not become slower :)

.github/workflows/main.yml
Krijger: Thanks for the nice read. Good point

.github/workflows/main.yml
heckflosse: afaik `make -j` uses all available cores

Krijger: Nope - testing shows that removing `--jobs 2` makes the make step approximately twice as slow

heckflosse: Yes, without `--jobs 2` only one core will be used.
I just tested with `-j` on my 8 core. In fact it starts much more than 8 jobs then...

.github/workflows/main.yml
Krijger: check

tools/osx/macosx_bundle.sh
Krijger: That was new for me as well, but it's Brace Expansion https://www.linuxjournal.com/content/bash-brace-expansion

```
echo {1,2}/test
1/test 2/test

echo 1,2/test
1,2/test
```

.github/workflows/main.yml
Krijger: added requested info

.github/workflows/main.yml
Krijger: Tested the `--jobs` option to be as fast as `--jobs 2`. Keeping the first, since this will automatically improve the build if the agents grow in CPUs in the future.

tools/osx/macosx_bundle.sh
Beep6581: Personally, I evolved to prefer explicit and simple script code to fancy expansion if the latter is unnecessary, as it's just more readable and easier for others (and myself 6 months down the line) to grasp and fix:
```
ditto --arch "${arch}" \
    "${GDK_PREFIX}/lib/gdk-pixbuf-2.0" \
    "${LIB}/gdk-pixbuf-2.0"
```

Krijger: I'm fine with that, but my aim was to be as least intrusive to the original code as possible. I think the goal of the original author may have been to keep things as DRY as possible.

Do you want me to change it in the scope of this PR?

Krijger:
For merging this PR (which I think is now possible, since it steadily builds a downloadable bootable app), please consider that the changes to the bundle script are backwards incompatible. I.e. it may be a good idea to check with the person that created the macOS releases before? Probably, that person used macports, while the agent uses homebrew, so I hope that this change is OK. Personally, I would say the reproduceability of the build outweighs a local dev laptop.
Note that once we make the scripts strict (`set -euo pipefail`), and have some simple integration test (is that possible? I don't know yet) we could with confidence refactor the code to support all of macports local machines, homebrew local machines (with possible customized HOMEBREW_PREFIX) and the build agent. But for now, my vote is the build agent as priority 1.

Krijger:
I've cleaned up the history a bit for the merge

tools/osx/macosx_bundle.sh
Beep6581: @krijger not necessary in the scope of this PR.

Beep6581:
I will merge this weekend if @Benitoite doesn't object.

Issue #5561, opened 2019-12-12
Review demosaicers


heckflosse:
Currently the demosaicers behave differently regarding clipping the output

Bayer demosaicers:

border_interpolate: does not clip but does not produce negative values
amaze: does not clip and produces negative values
rcd: clips into [0;65535]
dcb: does not clip and produces negative values
lmmse: does not clip and produces negative values
igv: clips into [0;65535]
ahd: does not clip but does not produce negative values
eahd: does not clip and produces negative values
hphd: does not clip and produces negative values
vng4: does not clip and produces negative values
fast: does not clip and produces negative values

xtrans demosaicers:

border_interpolate: does not clip but does not produce negative values
3-pass: does not clip and produces negative values
1-pass: does not clip and produces negative values
fast: does not clip but does not produce negative values

In my opionion we should make the behaviour consistent in the way that we clip negative values to zero but do not clip the highlights to 65535.

Opinions?

butcherg:
Definitely clip negative to zero, not sure about >65536.0. I'm the one who pointed out the behavior to Ingo; for me the egregious positive behavior is in AMAZE, where I see some highlight pixels pushed out an order of magnitude, which seems extreme for the semantics of the operation.  Specifically, I'm seeing values around 0.4 ending up 4.5 after AMAZE application.  My stats collector in rawproc has a few issues I'm fixing; will post some stats after they're fixed.

PR #5562, opened 2019-12-12
Clip negative values to zero in raw processing, #5561

heckflosse:


Floessie (approved):
Fine codewise. :+1: 

Maybe the different `max()` and `min()` calls in `rawimagesource.cc` should be unified while you're at it (theres are even `MAX()` calls, whereever that comes from).

I've tested all demosaicers with an HDRMerged DNG and couldn't spot any obvious outliers.

heckflosse:
@Floessie 
> MAX() calls, whereever that comes from

<glib/gmacros.h>

Beep6581:
I compared `dev` vs `review_demosaicers` using 61 test images and various PP3s. There were no visual differences. When using `compare` (imagemagick) the differences ranged from none to this:
screenshot_20191213_160717
(red means there is a difference, though it was only numerical - cannot see any difference in these images with my eyes)

Beep6581:
No problem with unclipped.

Beep6581 (approved):

Issue #5568, opened 2019-12-18
Generalize perspective correction


Lawrence37:
The flexibility of the perspective correction tool is limited by the current implementation's assumptions about the image. I feel there are two assumptions in particular which limit the tool's usefulness.

The first is the angular field of view. The current perspective tool does not consider the image's FOV. I have found that the tool works best for images with a 35mm equivalent focal length of around 20mm. For other focal lengths significantly different from 20mm, corrected images may appear stretched or compressed along one dimension. The circular depressions in the tower below are circular in reality, but appear elliptical in the corrected image:
![towers_85mm](https://user-images.githubusercontent.com/45837045/71046068-fc44eb80-20eb-11ea-9efd-7337eb3bbcb1.jpg)
(towers_85mm.tif)

The second assumption is that, for horizontal correction, the axis of rotation is parallel with the image's vertical axis. In practice, this could mean that horizontal correction works well only if the camera was pointed straight forward (not aiming up or down). This assumption often results in skew. This is the best I could do with the current tool:
![towers_28mm](https://user-images.githubusercontent.com/45837045/71046103-0f57bb80-20ec-11ea-9d4e-172c7cfa8f84.jpg)
(towers_28mm.tif)

I played around with the source code to see if removing the two assumptions would improve results. I am happy to report that it does! Here is a proof-of-concept at work:
![Screenshot](https://user-images.githubusercontent.com/45837045/71060162-a129ed80-2119-11ea-9f58-bde36be28faf.png)
The axis of rotation for horizontal correction is dependent on the value for Vertical. The Vertical bias value allows for vertical correction without affecting the axis of rotation. Field of view is the diagonal angle of view.

I've placed some of my test images (created in Blender) and pp3s here:
https://filebin.net/rmcnvpp9oj4ptla6
Also included is a patch containing the modifications I made. I grant permission to integrate the code into RawTherapee.

I would love it if the perspective tool becomes generalized by removing the two assumptions I described. I am willing to contribute to the coding.

Thanatomanic:
Nice work. Please make a PR with your modifications so that more people can test them!

gaaned92:
I was wondering why you don't borrow and perhaps improve ART automatic perspective correction code that I think was itself  borrowed from darktable.

Lawrence37:
[...]

Issue #4505, opened 2018-04-12
Enhance perspective correction


RawConvert:
Following my post here -
https://discuss.pixls.us/t/feature-suggestion-converging-verticals/5670
and more recent discussion here -
https://discuss.pixls.us/t/perspective-conversion/7206/9
this issue puts the suggestion in the correct place for possible consideration.

In summary, the current RT correction process is rather slow, involving iterations of horiziontal or vertical correction together with use of the straighten tool.  What would be good is a tool that allowed correction of converging verticals (or horizointal correction) in one action, for example by the user making two lines which would then be processed to parallel.

Beep6581:
The way perspective correction works in Autopano Pro/Giga is amazingly easy to use and efficient. Basically it's as @RawConvert described it in the last line - you draw two (or more, but two can be enough) lines, and these are made parallel. They can be horizontal, vertical, or both (4 lines).
https://youtu.be/voVBTiRxcLk?t=39

aferrero2707:
I implemented something similar in PhotoFlow, but the code is actually derived from Darktable:
https://www.youtube.com/watch?v=RCbpjyrtXLc

Here is where the perspective correction is computed: https://github.com/aferrero2707/PhotoFlow/blob/stable/src/vips/perspective.cc#L226

I can give you some hints if you are interested.

chaevvi:
It is not always convenient to carry out two lines.  I like the visual way of transforming perspective, as in the GIMP editor - pulling the image with the mouse. At the same time, you focus on a rectangular grid on the image.
I propose to implement this method in RawTherapee. It is also possible to implement other convenient ways - at the choice of the user. The current method of correcting the perspective requires a large number of actions, which is inconvenient.

Bezierr:
I support chaevvi in this. Perspective correction for anything more complex than a slight tilt of the camera is something that I, too, do in Gimp. I like the way it is done there, with one exception: when choosing "forward" correction, the grid should not move with the image but remain stable (i.e. rectangular), so the image can be adjusted to the grid.
So, it would be great if this could be done directly in RT.

chaevvi:
Regarding Bezierr comment: when adjusting the perspective in GIMP, you need to enable and select the grid size on the image. Then the undistorted rectangular grid and the mesh deformed during perspective correction will be displayed at the same time.

Bezierr:
Ah - didn't know that (I don't use Gimp all that often). Still, I think the following logic would be, ehm, more logical:

[...]

PR #5569, opened 2019-12-18
Generalize perspective correction

Lawrence37:
As suggested by @Thanatomanic, this is a pull request containing the modifications I referenced in #5568. The changes made allow the perspective tool to use the image's field of view and makes the horizontal transformation dependent on the vertical transformation.


heckflosse:
@Lawrence37 Thanks for your work :+1: 
I found an issue: Currently backwards compatibility is broken
I applied perspective correction in rt dev
grafik

and opened the same file in your branch
grafik

Thanatomanic:
I think these changes will probably require a toggle between two modes in the perspective correction module.

Lawrence37:
The units are different between the two versions. The new one uses degrees while the current one uses a unit equal to 0.45 degrees/unit. If I can find the exact field of view used by the current tool, I think it is possible to match the current output using the new implementation. I will work on making the new version backward compatible.

heckflosse:
@Lawrence37 
> If I can find the exact field of view used by the current tool, I think it is possible to match the current output using the new implementation.

That would be perfect :+1:

Thanatomanic:
@Lawrence37 Thanks for this contribution! I just tested your additions and I must say they look powerful. I'm sure this is a very good start to further development of the perspective tool. From a user pov I would always prefer to have some visual way to select straight lines, to set some parameters. We may still learn a little from darktable in that regard.

1. Setting the focal length and crop factor should ideally be taken from the image (if available). But I guess that's something for the future.
2. Vertical shift and Horizontal shift sliders should be limited to -50 and +50, otherwise you end up with a black frame.
3. How should I use the 'Recovery' tools?
4. The Post-correction 'Scale' tool doesn't seem functional.
5. Horizontal comes first, except for the 'Vertical' and 'Horizontal' sliders in the Correction block.

FWIW

Lawrence37:
Thanks for your feedback, @Thanatomanic.

> Setting the focal length and crop factor should ideally be taken from the image (if available). But I guess that's something for the future.

I agree. I will see if I can implement this soon.

> Vertical shift and Horizontal shift sliders should be limited to -50 and +50, otherwise you end up with a black frame.

This is with Auto-fill enabled. When disabled, values up to +-100 can be used before the image completely goes off the canvas. Values beyond +-50 or +-100 may also be useful depending on what the other parameters are set to.
Regardless, I'm not confident in the current limits for all adjusters. I've made them quite large with the intent of trying them out before choosing more sensible limits. I'm considering using +-60 for the vertical/horizontal adjusters and +-45 for rotation. I don't know about focal length, crop factor, shift, and scale.

> How should I use the 'Recovery' tools?

They introduce perspective distortion. I think the most common use case will be for making buildings "lean" slightly for a more natural look. It's different from backing off the correction in two ways. One is that recovery is the inverse of correction. If you look at how the image changes with Auto-fill disabled, you can see that correction stretches the image while recovery compresses it. The second is that recovery is applied after all other adjustments (you should only use them after you are finished with correction and post-correction adjustments). This produces more natural/expected results when horizontal perspective/shift is applied.
If only horizontal or vertical correction is applied, using the corresponding recovery adjuster is not much different from backing off the correction.

> The Post-correction 'Scale' tool doesn't seem functional.

Auto-fill undoes the scale. Perhaps the scale adjuster should be greyed out when Auto-fill is enabled?

> Horizontal comes first, except for the 'Vertical' and 'Horizontal' sliders in the Correction block.

The behavior of horizontal correction depends on the value of Vertical. It should be more intuitive to adjust Vertical before Horizontal. Since recovery is the inverse of correction, the order is reversed in the recovery section. After playing with the recovery adjusters, however, it seems more intuitive for me to change Vertical then Horizontal. I will switch their order in the GUI if others feel the same way. Finally, the adjusters for the simple method is ordered the same way as the original tool.

Thanatomanic:
@Lawrence37 I was processing a few architectural images recently and I found your modifications very useful. What would still be a really awesome addition is to have a more visual way of making corrections. For that I really like the way darktable handles the keystoning correction: there are four lines and corners that you can drag that determine what transformation is applied.
Would such a thing be possible with your tool?

Lawrence37:
I'm not familiar with such a tool in darktable. What you describe is similar to either GIMP's perspective transform or PhotoFlow's perspective correction. The first cannot be implemented using the new tool. The second one can. I would like this feature myself, but with lines instead of a cage: https://github.com/Beep6581/RawTherapee/issues/4505#issuecomment-380807285

Thanatomanic:
image

So you're saying something like this is impossible? Too bad :)

Lawrence37:
It is possible.

Lawrence37:
This pull request is ready for merging. I'm happy with current implementation and I don't plan on making further changes (aside from addressing bugs and code reviews). If I want to add additional features, I'll open a separate pull request.

Thanatomanic:
@Lawrence37 Since your modifications change the language file too, I think it is still blocked by #5664 for now. @Beep6581 can you comment?

Thanatomanic:
Since #5662 got merged and made changes to `rtdata/languages/default`, I suppose we can also safely merge this and ignore #5664 for now? @heckflosse, final opinion?

Lawrence37:
I'm working on a UI which allows you to draw lines and use them for perspective correction. I wonder if I should include it in this PR or open a new one after this is merged.

heckflosse:
I just tested again and found a bug. Using horiz. and vertic. shift together sometimes the preview is not updated while the navigator is updated. Count the number of windows on left and right in navigator and preview in this screenshot to verify ;-)
grafik

heckflosse:
> @heckflosse, final opinion?

:+1: for merge. Bugs can be fixed also when it's in dev. The earlier we merge the more tests we get.

heckflosse:
@Lawrence37 Feel free to add your name to contributors section of Authors.txt

Lawrence37:
I have no idea why the preview doesn't update. It seems to only happen with auto-fill enabled.

heckflosse:
@Lawrence37 I will try to reduce the case to the necessary steps to reproduce it. Maybe that helps.
At least, as now your branch is merged into dev, it might be easier to get feedback from other people ;-)

Lawrence37:
I can reliably reproduce the bug like this:
1. Set horizontal shift to 20 (or some other sufficiently large value).
2. Set vertical shift to any value between +-(horizontal shift value) by typing the number. -> Preview doesn't update.
3. Set vertical shift to some value outside the range. -> Preview updates correctly.

heckflosse:
Great, that you can reliably reproduce the bug. That's the first step to fix it :)

heckflosse:
@Lawrence37 
Maybe worth to think about why the navigator preview (improccoordinator.cc) is updated while the main preview (dcrop.cc) is not updated...

Issue #3198, opened 2016-03-06
Double-clicking on a folder should expand it in the treeview if it contains subfolders


Beep6581:
The treeview, at least in `gtk3` in Linux, has two active parts: the triangle, and the folder name with the folder icon.

The triangle only supports single-click events and it expands the folder if it contains subfolders. That's good. Could someone verify that it's working correctly in Windows? I ask because of https://discuss.pixls.us/t/need-some-help/963/7

The folder label and icon on the other hand support the double-click event, which loads the folder and shows the thumbnails in the file browser. What it should also do is the same as single-clicking on the triangle does - it should expand the folder in the treeview to show the subfolders (if the folder contains any). Currently I double-click on a folder to show the thumbs, but if the folder contains subfolders then there is no indication of that unless I then click on the triangle.

To demonstrate:
I double-clicked on this folder http://i.imgur.com/ikvG8aW.png
It loaded the thumbnails of the photos from that folder, but does it contain subfolders? Who knows? One must then click on the triangle to find out http://i.imgur.com/3pVZ3Eg.png

This bug or shortcoming is in `gtk3`, I haven't checked `master`.

gaaned92:
In master, double click on folder doesn't expand it. You must click on triangle.

Beep6581:
So double-click on the folder should also fire the same event as clicking on the triangle does, in both Gtk+ 2 and 3.

Beep6581:
This is a test, ignore this.

Beep6581:
Over the last months there has been a recurring packaging problem with the file browser arrows missing. If this feature was implemented (I tried and gave up, don't understand GTK+ signals yet) missing arrows would be less of a problem. I'm also using that as an excuse to bump the issue :)

gaaned92:
@Beep6581  if the folder contains raw images and subfolders,  presently double click permits to display the raw.
If now it expands the folder, how we can display the raw files?

Beep6581:
Still valid.

PR #5587, opened 2019-12-29
Expand subfolders in directory browser, see #3198

rfranke:
I was wondering why subfolders didn't expand when activating a folder. Then I saw #3198. This little extension should solve it (tested under Ubuntu 18.04).


rfranke:
The second commit removes the special treatment of expansions if subDirs.empty(). This lets the triangle disappear for expanded directories that have no subdirectories, like in other directory browsers.
I don't know the reason for the special treatment though. It had always been there according to the git history of dirbrowser.cc. Can someone check?

Beep6581:
@rfranke tested and works well in Linux, much better than before.

Would be great if RT probed all subfolders of the currently open folder, to automatically hide the arrows if those subfolders have no children.

Beep6581:
@rfranke I will merge this now to prepare for 5.8.

PR #5589, opened 2019-12-30
Applying geometric transformations leads to dark artifacts in combina…

### LINKED ISSUE #5588: Applying geometric transformations leads to dark artifacts in combination with capture sharpening
The bicubic interpolation we use for geometric transformations leads to artifacts in combination with capture sharpening especially in areas with high contrast. I will create a pr soon to avoid that.
Here's a first screenshot. Left with current code, right with new code:
grafik

heckflosse:
…tion with capture sharpening, fixes #5588


heckflosse:
Current implementation of logarithmic transform increases the processing time for geometric transformations by ~ factor 5.1
With next commit the factor will be ~ 1.7

heckflosse:
Just for information. Currently the log transform is fully backwards compatible to 5.7 as it's only used if Capture sharpening is enabled, which is not available for 5.7.

heckflosse:
Factor now is ~ 1.5

Thanatomanic:
I think I found a segfault when enabling profiled chromatic abberation correction... I'll investigate further tomorrow.

Thanatomanic:
To reproduce:
1. Load RAW file that has a profiled CA correction
2. Apply Neutral
3. Set to linear mode (should already be the case for Neutral)
4. Turn off all corrections
5. Go to 100% zoom mode (seems to matter...)
6. Set to logarithmic mode
7. Enable CA correction. SEGFAULT in `lut.h`

Here's a stacktrace: https://filebin.net/011vn748erxv8gx3

heckflosse:
@Thanatomanic I pushed a fix

Thanatomanic:
Fix confirmed 👍

heckflosse:
@Thanatomanic 
> Set to linear mode (should already be the case for Neutral)

Currently the default is `Logarithmic`. For existing pp3 files `Linear` is used for backwards compatibility.
I prefer `Logarithmic` as default because in my experience it gives less artifacts. The downside of `Logarithmic` is that it's slower.

What do you think?

Thanatomanic:
@heckflosse 
I agree with you observations: Logarithm gives better results overall and I like it much better, but not in darker regions near the black clipping point. You get nasty pixel artifacts then... 

Like this: 
image


But actually, in both cases it's very hard to see any difference when viewing at 100% or less.

heckflosse:
@Thanatomanic Can you provide an example file to test? Maybe there's a way to improve the loharithmic mode...

Thanatomanic:
@heckflosse Take a look at this picture for example: https://discuss.pixls.us/t/play-raw-nightscape-from-orange-free-state
Using `Logarithmic` is better for the stars, but gives blocky artifacts in the dark and noisy horizon.

heckflosse:
@Thanatomanic Thanks for the example. Confirmed.

heckflosse:
I would like to merge this for 5.8 as
1) it is backwards compatible
2) `logarithmic` gives better results overall
3) `linear` can still be used

Objections?

PR #5598, opened 2020-01-04
macOS: enable various package managers

Benitoite:
By using a cmake flag -DLOCAL_PREFIX= to set whether to use `/opt` or `/usr` for the gtk installation parent directory.

Example, if you use homebrew, the default is `/usr`, so the flag is not necessary.

If you use macports, the parent directory is `/opt`, so the flag should be used: `-DLOCAL_PREFIX="/opt"`


tools/osx/macosx_bundle.sh
```diff
@@ -141,45 +143,45 @@ rm -r "${LIB}"/gdk-pixbuf-2.0
...
+ditto {"${LOCAL_PREFIX}/local","${RESOURCES}"}/share/icons/Adwaita/index.theme
+"${LOCAL_PREFIX}/local/bin/gtk-update-icon-cache" "${RESOURCES}/share/icons/Adwaita"
 
 # Copy libjpeg-turbo into the app bundle
-cp /usr/local/lib/libjpeg.*.dylib "${CONTENTS}/Frameworks"
+cp ${LOCAL_PREFIX}/local/lib/libjpeg.*.dylib "${CONTENTS}/Frameworks"
 
 # Copy libexpat into the app bundle (which is keg-only)
-cp /usr/local/Cellar/expat/*/lib/libexpat.1.dylib "${CONTENTS}/Frameworks"
+#cp ${LOCAL_PREFIX}/local/Cellar/expat/*/lib/libexpat.1.dylib "${CONTENTS}/Frameworks"
```
Benitoite: `libexpat` handling needed here.

`pkgcfg_lib_EXPAT_expat` is available.  Maybe if there is no `/usr/local/Cellar/expat` dir we can use the `pkgcfg_lib`

PR #5610, opened 2020-01-17
review mlsharpen

heckflosse:
I reviewed mlsharpen (Sharpen edges).
It's about 30% faster now and the code is much cleaner.
There are differences in output because I fixed a bug in this line
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/ipsharpen.cc#L568


rtengine/ipsharpenedges.cc
```diff
@@ -0,0 +1,223 @@
...
+#include <memory>
+#include "improcfun.h"
+#include "labimage.h"
+#include "procparams.h"
+#include "rt_math.h"
+
+#define BENCHMARK
+#include "StopWatch.h"
+
+using namespace std;
```
Floessie: I don't think we should use `using namespace` in "new" code.

rtengine/ipsharpenedges.cc
```diff
@@ -0,0 +1,223 @@
...
+namespace {
+#ifdef __SSE2__
```
Floessie, line 27: Those implementations are generic. Are they really faster only on SSE2?

rtengine/ipsharpenedges.cc
```diff
@@ -0,0 +1,223 @@
...
+BENCHFUN
+
+    const int width = lab->W, height = lab->H;
+    constexpr float chmax[3] = {1.f / 8.f, 1.f / 3.f, 1.f / 3.f};
+    const int width2 = 2 * width;
+    constexpr float eps2 = 0.001f; //prevent divide by zero
+    const float amount = params->sharpenEdge.amount / 100.0f;
+    const float amountby3 = params->sharpenEdge.amount / 300.0f;
+
+    std::unique_ptr<float[]> L(new float[width * height]);
```
Floessie, line 81: Maybe a `std::vector<float>` is even easier to handle?

rtengine/ipsharpenedges.cc
```diff
@@ -0,0 +1,223 @@
...
+
+    const int channels = params->sharpenEdge.threechannels ? 1 : 3;
```
Floessie, line 83: Hm, by the name it looks swapped...

rtengine/ipsharpenedges.cc
```diff
@@ -0,0 +1,223 @@
...
+                            f2 = L[offset + 1 - width] - v;
+                            f3 = (L[offset + 1 - width] - L[offset + width + 1]) * (L[offset + 1 - width] - L[offset - width - 1]);
+                            f4 = std::sqrt(fabs((L[offset + 1 - width] - L[offset + width2 + 2]) * (L[offset + 1 - width] - L[offset - width2 - 2])));
+                            const float difRT = f1 * SQR(f2 * f3) * f4;
+                            if (difRT > 0.f) {
+                                lumD2 = (L[offset + 1 - width] * difLB + L[offset - 1 + width] * difRT) / (difLB + difRT);
+                                lumD2 = intp(contrast, lumD2, v);
+                            }
+                        }
+                    }
```
Floessie, line 198: Looks copied, should be factored out.

rtengine/ipsharpenedges.cc
heckflosse: I can only test with SSE. That's the reason I #ifdefed them.
Maybe you can test on different architectures?

heckflosse: @Floessie 

The fast check wether a value x is in range ]y;z[ could be moved to rt_math.h
What do you think?

Floessie: @heckflosse Positive.

> Maybe you can test on different architectures?

Can you give me a hint, how to most quickly test it, as the PPC emulation is awfully slow?

heckflosse: @Floessie Does a test on an emulation make sense at all?

Floessie: @heckflosse If the emulator gets the same share of CPU power each time, it might work out... And given enough runs, of course.

heckflosse: @Floessie Ok, use this raw and pp3
https://filebin.net/is2qv5my00wg5xk2

heckflosse: @Floessie How about adding this functions to rt_math.h ?

```
template <typename T>
bool inOpenIntervalUnordered(T a, T b, T c)
{
    // returns true if a is in ]b;c[ or in ]c;b[, else false
    return a < rtengine::max(b, c) && a > rtengine::min(b, c);
}

template <typename T>
bool inOpenIntervalOrdered(T a, T b, T c)
{
    // returns true if a is in ]b;c[, else false
    return a < c && a > b;
}

template <typename T>
bool inClosedIntervalUnordered(T a, T b, T c)
{
    // returns true if a is in [b;c] or in [c;b], else false
    return a == rtengine::max(rtengine::min(a, b), rtengine::min(c, rtengine::max(a, b)));
}

template <typename T>
bool inClosedIntervalOrdered(T a, T b, T c)
{
    // returns true if a is in [b;c], else false
    return a == rtengine::min(rtengine::max(a, b), c);
}
```

Beep6581:

PR #5613, opened 2020-01-20
Fix double promotion

heckflosse:
pr to get LGTM check
More fixes will follow


Beep6581:

heckflosse:
Though not being complete I will merge now. The remainig parts will be solved in further PRs

PR #5617, opened 2020-01-23
devirtualize method calls by declaring classes and methods final

heckflosse:


rtengine/iimage.h
```diff
@@ -658,7 +658,7 @@ class PlanarRGBData : virtual public ImageDatas
 
     /* If any of the required allocation fails, "width" and "height" are set to -1, and all remaining buffer are freed
      * Can be safely used to reallocate an existing image */
-    void allocate (int W, int H) override
+    void allocate (int W, int H) override final
```
Floessie: [AFAIK](https://stackoverflow.com/questions/29412412/does-final-imply-override), `final` should be enough.

Floessie (approved):
Great job, Ingo. Especially finding methods that really are functions. :+1:

heckflosse:
@Floessie `-Wsuggest-final-types` and `-Wsuggest-final-methods` are your friends, though with some false positives.

Issue #5552, opened 2019-12-03
macOS latency and performance


learmj:
Hi there. I've used RT for years on Linux. I think it's a fantastic tool that you have all developed and maintain. I'd be lost without it. I made the switch from Linux to a Mac last week and had a it of a shock when I tried using RT on my new macbook pro 16 (2019). It works, but there is so much latency (eg dragging sliders or moving curves, and the image taking a long time to update afterwards). It makes it pretty much unusable for me, which I'm really sad about. My first question is - is this normal for RT on mac os? The UI looks slightly odd too in that the fonts seem slightly blurry and out of alignment, so I am thinking there is something missing in my env. This was with the the latest pre-built 5.7 from your website. I then went to rawpedia, installed the dependencies using homebrew (making sure that my clang had support for openmp - which it does - I can compile a test program), compiled RT from source (5.7 release from git) but that build of RT doesn't run, and crashes with SIGABRT.

So I am wondering:
* If there is something in my env which makes the pre-built image look less than crisp?
* If the latency / lag is expected on mac os - Catalina in my case?
* If the instructions on rawpedia for building on mac os are up to date, and for building on Catalina?

The crash in my own build seems to be because of this:
 <pre>(rawtherapee-bin:687): GLib-GObject-WARNING **: 08:18:30.265: invalid cast from 'GtkMenuBar' to 'GtkWindow'

(rawtherapee-bin:687): Gtk-CRITICAL **: 08:18:30.265: gtk_window_add_accel_group: assertion 'GTK_IS_WINDOW (window)' failed

(rawtherapee-bin:687): Gtk-WARNING **: 08:18:32.744: Could not load a pixbuf from /org/gtk/libgtk/theme/Adwaita/assets/check-symbolic.svg.
This may indicate that pixbuf loaders or the mime database could not be found.

(rawtherapee-bin:687): Gtk-WARNING **: 08:18:33.747: Could not find the icon 'list-remove-symbolic-ltr'. The 'hicolor' theme
was not found either, perhaps you need to install it.
You can get a copy from:
	http://icon-theme.freedesktop.org/releases
**
Gtk:ERROR:../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): Failed to load /org/gtk/libgtk/icons/16x16/status/image-missing.png: Unrecognised image file format (gdk-pixbuf-error-quark, 3)
[...]

PR #5629, opened 2020-02-01
macOS bundle update

Benitoite:
Copies libraries using `ditto`, preserving binary resources
Codesigns each in Contents/Framework


rfranke:
You changed a couple of `cp *.dylib` to `ditto *.dylib`. Why are those operations needed at all? Shouldn't the recursive
```
CheckLink "${EXECUTABLE}"
```
just do it?

Beep6581:
Is this PR needed to get a 5.8 build for macOS?

Benitoite:
> Is this PR needed to get a 5.8 build for macOS?

@Beep6581 Yes...  will have a test build ready tonite of 5.7-dev, after work.

@rfranke  Yes...  Over the years some libraries had failed in that loop and band-aids were placed.

Benitoite:
Here is a build to test on macOS 10.9+ with latest pull from `gtk/gtk-3-24`

[outdated link removed]

```
Version: 5.7-438-gd6fb05938
Branch: dev
Commit: d6fb05938
Commit date: 2020-02-02
Compiler: clang 11.0.0.11000033
Processor: generic x86
System: Apple
Bit depth: 64 bits
Gtkmm: V3.24.2
Lensfun: V0.3.95.0
Build type: release
Build flags: -Wno-deprecated-register -Wno-unused-command-line-argument -std=c++11 -std=c++11 -mtune=generic -Werror=unused-label -Werror=delete-incomplete -mmacosx-version-min=10.9 -flto -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Xpreprocessor -fopenmp /opt/local/lib/libomp.dylib -I/opt/local/include -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:  -mtune=generic -headerpad_max_install_names -flto
OpenMP support: ON
MMAP support: ON

```

rfranke:
If you need these changes to build a functioning 5.8 app, then I'm of course fine with it.

Unfortunately it doesn't help me -- it adds a problem with libexpat, see #5552. I have never been able to package a functioning RT app on my system, even though the cmake process runs nicely and builds a fully functioning executable in the build directory. I hope that my problems with packaging an app are not intrinsic to macOS, but can somehow be solved by rigorously following `otool -L` links and cmake settings.

Another question: Why did you "harden app to /Applications"?
In my case, I have the officially released binaries under /Applications and test own bundles in the build directory -- this doesn't work anymore now.

rfranke:
OK, the libexpat problem disappears if I copy RawTherapee.app to /Applications, i.e. is related to the now hardcoded directories in the changed:
```
msg "Registering @rpath into the executable:"
echo "   install_name_tool -add_rpath /Applications/RawTherapee.app/Contents/Frameworks '${MACOS}/bin/rawtherapee-bin'" | bash -v
echo "   install_name_tool -add_rpath /Applications/RawTherapee.app/Contents/Frameworks '${EXECUTABLE}-cli'" | bash -v
for frameworklibs in ${CONTENTS}/Frameworks/* ; do
    echo "   install_name_tool -delete_rpath /opt/local/lib '${frameworklibs}'" | bash -v
    echo "   install_name_tool -add_rpath /Applications/RawTherapee.app/Contents/Frameworks '${frameworklibs}'" | bash -v
done
```
What is the reason for this hard coding, including also `/opt/local/lib`, i.e. MacPorts?

Three problems are remaining for my self built app package. They are not present after `make install`, but appear with `make macosx_bundle`. They are the same with or without this PR:

1. the directory browser does not show small triangles anymore
2. cannot access memory cards when starting the app (it works when starting from a console)
3. need to remove `export DYLD_LIBRARY_PATH="${lib}"` to avoid the error
```
dyld: Symbol not found: _iconv
  Referenced from: /usr/lib/libarchive.2.dylib
  Expected in: /Applications/RawTherapee.app/Contents/Frameworks/libiconv.2.dylib
 in /usr/lib/libarchive.2.dylib
Abort trap: 6
```

Benitoite:
@rfranke 
1. do share/icons make it over? 
2. How are you trying to access the memory card? Try clicking where the little triangle is supposed to be.
3. Changed to `DYLD_FALLBACK_LIBRARY_PATH`

I've put in hardcoded paths for sandboxing with hardened runtime.  It is a more secure paradigm and less prone to corruptions (only one signatory at a time / no simultaneities in the com.rawtherapee container).

rfranke:
1. Yes, Resources/share/icons contains `Adwaita` and `hicolor`.

2. Access to memory card has to do with System Preferences... -> Security & Privacy settings. Other apps trigger a popup dialog asking for permission and register themselves there. RT does not do this :-(
When starting from command line, then the terminal takes care about permissions.

3. `DYLD_FALLBACK_LIBRARY_PATH` works :-)

What is the motivation for new hardcoded paths?

Benitoite:
@rfranke The entitlement to utilize relative paths is not allowable in the `-o runtime` signature used on objects desitined for notarization.

rfranke:
The last commit efd28baa3a440e5b110c14e8cc3f9f629df79f3c looks good -- shall fix a remaining issue with liblensfun in my app package. I won't be able to do further tests the next three days.

Benitoite:
After much rebuilding, there is a new build to test (now on the 5.8 tag)

https://kd6kxr.keybase.pub/RawTherapee_OSX_10.9_64_5.8.zip

PR #5639, opened 2020-02-07
CMakeLists.txt LENSFUNDBDIR partial regress

Benitoite:
Restores a few lines in CMakeLists.txt 2cc8918 removed, applying to certain use cases on non-APPLE systems.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,6 +252,12 @@ if(NOT DEFINED APPDATADIR)
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


Beep6581:
Hey @Benitoite , what are those use cases?

Benitoite:
> Hey @Benitoite , what are those use cases?

That is a good question. According to the original commit by @agriggio these lines give the “possibility to use a private copy of the LF database”.

PR #5644, opened 2020-02-07
mac: simplifies bundle structure, bundle packaging fixes

Benitoite:
Simplifies bundle structure, eliminating the use of `/bin/sh` and the executable loader script.
Fixes building libpixbuf and immodules database by modifying install names later on.


tools/osx/macosx_bundle.sh
```diff
@@ -127,6 +127,19 @@ ditto ${LOCAL_PREFIX}/local/lib/libomp.dylib "${CONTENTS}/Frameworks"
 msg "Copying dependencies from ${GTK_PREFIX}:"
 CheckLink "${EXECUTABLE}"
 
+# dylib install names
+find -E "${CONTENTS}" -type f -regex '.*/(rawtherapee-cli|rawtherapee|.*\.(dylib))' | while read -r x; do
+    msg "Modifying dylib install names: ${x}"
+    {
+        # id
+        case ${x} in *.dylib) echo "   install_name_tool -id '@rpath/$(basename "${x}")' '${x}'";; esac
```
Beep6581: An `if` would be better suited than a `case`.

Benitoite: Thanks I will convert that tonite.

Benitoite: changed to `if` and made it a function.

Benitoite (commented):
@heckflosse @Beep6581 and other C++ folks, if you have a chance please review my C++ additions for `__APPLE__` to [`options.cc`](https://github.com/Benitoite/RawTherapee/blob/patch-22/rtgui/options.cc)

```
diff --git a/rtgui/options.cc b/rtgui/options.cc
index dcb16e1ea..5e9d48cc5 100644
--- a/rtgui/options.cc
+++ b/rtgui/options.cc
@@ -2270,8 +2270,18 @@ void Options::load(bool lightweight)
     const gchar* path;
     Glib::ustring dPath;
 
+#ifdef __APPLE__
+    // Build Application Support directory path for macOS.
+    const char* homedir = g_getenv("HOME"); // This returns the current container data dir in ~/Library
+    std::string homebuf{homedir};
+    int homelength = strlen(homebuf.c_str());
+    homebuf[homelength-44] = '\0'; // Terminate string after ${HOME}/Library
+    std::string homeconfig{homebuf};
+    std::strcat(&homeconfig[0], "/Application Support/RawTherapee/config");
+    path = homeconfig.c_str();
+#else
     path = g_getenv("RT_SETTINGS");
-
+#endif
     if (path != nullptr) {
         rtdir = Glib::ustring(path);
 
@@ -2280,6 +2290,7 @@ void Options::load(bool lightweight)
             throw Error(msg);
         }
     } else {
+
 #ifdef WIN32
         WCHAR pathW[MAX_PATH] = {0};
 
@@ -2312,9 +2323,14 @@ void Options::load(bool lightweight)
         rtdir = Glib::build_filename(argv0, "mysettings");
     }
 
-    // Modify the path of the cache folder to the one provided in RT_CACHE environment variable
+    // Modify the path of the cache folder to the one provided in RT_CACHE environment variable. Build the cache folder name in macOS.
+#ifdef __APPLE__
+    std::string homecache{homebuf};
+    std::strcat(&homecache[0], "/Application Support/RawTherapee/cache");
+    path = homecache.c_str();
+#else
     path = g_getenv("RT_CACHE");
-
+#endif
     if (path != nullptr) {
         cacheBaseDir = Glib::ustring(path);
```

It gets the value of HOME and puts a terminator 44 bytes from the end, resulting in the absolute path of `~/Library`.  This is concatenated with the proper macos app support directory suffix.  Because of the length of the removed portion from HOME is greater than the additional suffix, no unintended truncation occurs.  I don't *think* anything that I declared/initialized needs to be `free()`d…

Benitoite:
The `-cli` works in *Catalina* now. edit: By working I mean it loads now. I have a new error looking for the default profiles directory

```
Error: the user's processing profile path doesn't point to a directory or doesn't exist!

Could not open camera constants file "./../Resources/camconst.json": No such file or directory

The default profile for raw photos could not be found or is not set.
Please check your profiles' directory, it may be missing or damaged.
"${G}/Auto-Matched Curve - ISO Low" will be used instead.


The bundled profile "${G}/Auto-Matched Curve - ISO Low" could not be found!
Your installation could be damaged.
Default internal values will be used instead.
```
It should be a straightforward fix.  It's caused by *Catalina* not allowing a helper binary to live in `Contents/MacOS` next to the main executable.  So now it is located in `/Contents/Frameworks` from which it may execute.

Beep6581:
@Benitoite as I see new daily commits to this branch, let us know when this PR is ready for merging.

Benitoite:
I believe this concludes the work needed this PR.  `-cli` is running as normal now.
The only TODOs are to analyze and filter out the benign errors produced during the bundle script run, and build a command to call `create-dmg` to make a fancy `dmg`.

Edit: discovered the create-dmg repo is mostly not working, so I [forked](https://github.com/Benitoite/rt-create-dmg) and tailored it to RT.

UpdateInfo.cmake
```diff
@@ -107,4 +107,8 @@ endif(WIN32)
 # build version.h from template
 configure_file("${PROJECT_SOURCE_DIR}/rtgui/version.h.in" "${CMAKE_BINARY_DIR}/rtgui/version.h")
 # build AboutThisBuild.txt from template
-configure_file("${PROJECT_SOURCE_DIR}/AboutThisBuild.txt.in" "${CMAKE_BINARY_DIR}/AboutThisBuild.txt")
+if(APPLE)
+    configure_file("${PROJECT_SOURCE_DIR}/AboutThisBuild.txt.in" "${CMAKE_BINARY_DIR}/AboutThisBuild.txt")
```
Beep6581: @Benitoite these are identical.

tools/osx/macosx_bundle.sh
```diff
@@ -110,7 +110,9 @@ EXPATLIB="$(cmake .. -LA -N | grep "pkgcfg_lib_EXPAT_expat" | cut -d "=" -f2)"
 CODESIGNID="$(cmake .. -LA -N | grep "CODESIGNID" | cut -d "=" -f2)"
 NOTARY="$(cmake .. -LA -N | grep "NOTARY" | cut -d "=" -f2)"
 FANCY_DMG="$(cmake .. -LA -N | grep "FANCY_DMG" | cut -d "=" -f2)"
-echo ${FANCY_DMG}
+if [[ -n ${FANCY_DMG} ]] ; then
```
Beep6581: No need for curly braces.

tools/osx/macosx_bundle.sh
```diff
@@ -110,7 +110,9 @@ EXPATLIB="$(cmake .. -LA -N | grep "pkgcfg_lib_EXPAT_expat" | cut -d "=" -f2)"
...
+echo "Fancy .dmg build is ON."
```
Beep6581: Indentation.

tools/osx/macosx_bundle.sh
```diff
@@ -268,7 +270,7 @@ install_name_tool -delete_rpath RawTherapee.app/Contents/Frameworks "${EXECUTABL
 install_name_tool -add_rpath @executable_path "${EXECUTABLE}"-cli
 
 # Codesign the app
-if ! test -z "${CODESIGNID}" ; then
+if [ -n "${CODESIGNID}" ] ; then
```
Beep6581: @Benitoite if the script uses Bash, and it does, then it should stick to it, so use the `[[` shell keyword and not the `[` test command.

No need for curly braces.

No space between `]]` (or `]`) and `;`

tools/osx/macosx_bundle.sh
```diff
@@ -268,7 +270,7 @@ install_name_tool -delete_rpath RawTherapee.app/Contents/Frameworks "${EXECUTABL
...
 msg "Codesigning Application."
```
Beep6581: Indentation.

tools/osx/macosx_bundle.sh
```diff
@@ -281,7 +283,7 @@ spctl -a -vvvv "${APP}"
 fi
 
 # Notarize the app
-if ! test -z "$NOTARY" ; then
+if [ -n "$NOTARY" ] ; then
```
Beep6581: As above.

tools/osx/macosx_bundle.sh
```diff
@@ -330,10 +335,25 @@ dmg_name="${dmg_name}_${lower_build_type}"
 fi
 
 msg "Creating disk image:"
+if [ ! -z ${FANCY_DMG} ] ; then
```
Beep6581: As above.

tools/osx/macosx_bundle.sh
```diff
@@ -330,10 +335,25 @@ dmg_name="${dmg_name}_${lower_build_type}"
...
+--eula LICENSE.txt \
+--hdiutil-verbose \
+--rez /Library/Developer/CommandLineTools/usr/bin/Rez
+else
 hdiutil create -format UDBZ -fs HFS+ -srcdir "${srcDir}" -volname "${PROJECT_NAME}_${PROJECT_FULL_VERSION}" "${dmg_name}.dmg"
+fi
 
 # Sign disk image
-if ! test -z "$CODESIGNID" ; then
+if [ -n "$CODESIGNID" ] ; then
```
Beep6581: As above.

tools/osx/macosx_bundle.sh
```diff
@@ -110,7 +110,9 @@ EXPATLIB="$(cmake .. -LA -N | grep "pkgcfg_lib_EXPAT_expat" | cut -d "=" -f2)"
 CODESIGNID="$(cmake .. -LA -N | grep "CODESIGNID" | cut -d "=" -f2)"
```
Beep6581: Would be good if you added a comment showing the output of the first command (e.g. the output of `cmake .. -LA -N`) and then the expected final result (`$CODESIGNID`).

Benitoite: Actually the reason cmake is called is to show all the cached variables. Could there be a more proper way of getting cmake-cached variables?

Beep6581: @Benitoite looks like you could use `configure_file`:
https://cmake.org/cmake/help/v3.8/command/configure_file.html

So you could have a section in the Bash script with a list of `@VARIABLES@` as such:
```bash
codesign_id=@CODESIGNID@
```
and then in CMake run:
```cmake
configure_file(macosx_bundle.sh.in macosx_bundle.sh @ONLY)
```

Otherwise, if you prefer to `grep` then you could do it as you do now, with one small change. Since you're already using Bash then I would stick to Bash, so instead of calling `cut` (which is not part of Bash) I would use Bash's substring removal (here I will use `CMAKE_BUILD_TYPE` as an example):
```bash
# In: CMAKE_BUILD_TYPE:STRING=release
# Out: release
foo="$(cmake .. -L -N | grep CMAKE_BUILD_TYPE)"; foo="${foo#*=}"
```

UpdateInfo.cmake
Benitoite: Reverted it, thanks for catching.

Benitoite:
Sorry, having trouble with one file.  Tried a git revert and got a fatal error, then re-pasted it in via web, but somehow /tabs showed up.

tools/osx/macosx_bundle.sh
Benitoite: Thanks, substring removal is the best way with grep, which only needs to keep credentials in the cache.

tools/osx/macosx_bundle.sh
Benitoite: Curly braces and quotes removed from strings in [[ statements.

tools/osx/macosx_bundle.sh
Benitoite: Changed to `[[`, `{}` and `"` removed from strings in `[[` statements, space removed from some `;`

tools/osx/macosx_bundle.sh
Benitoite: Re-indented with bash-beautification
Screen Shot 2020-02-24 at 11 27 45 AM

tools/osx/macosx_bundle.sh
Benitoite: Converted statements to `[[`

tools/osx/macosx_bundle.sh
Benitoite: Converted statements to `[[`

Benitoite:
Ok, the last two commits fixed the problem.  Was able to copy from a fresh clone.

tools/osx/macosx_bundle.sh
Benitoite: Statements converted.

tools/osx/macosx_bundle.sh
Benitoite: Re-indented.

Beep6581:
@Benitoite ready for merge?

Benitoite:
@Beep6581 Yes.

PR #5662, opened 2020-02-24
Raw preprocessing WB

### LINKED ISSUE #5616: Allow usage of camera (or even user defined) white balance in raw preprocessing and demosaic
With the raw from [here](https://discuss.pixls.us/t/strange-green-artifacts-after-raw-conversion/15971)
automatic raw ca correction leads to green artifacts caused by auto white balance failing on this file.
Using camera white balance instead of auto white balance works fine.
Left without raw ca correction, middle with auto whitebalanced data and raw ca correction, right with camera whitebalanced data and raw ca correction (top left corner of the image):
grafik

Based on the first screenshot one could assume that raw ca correction is not needed for this image.
Top right corner of the image (same order as first screenshot) shows the difference:
grafik

Additionally the colour shift is much less in right image.

I propose to add an option to allow different than auto white balance for raw preprocessing and demosaic.

rom9:
Fixes #5616
New parameter to choose between Auto and Camera channel multipliers in preprocessing. It is also possible to apply custom compensation factors.


Thanatomanic:
So, is there any good reason (or an example) _not_ to use the camera multipliers?

rom9:
> So, is there any good reason (or an example) _not_ to use the camera multipliers?

If i try with "normal" images, i can't see any significant differences between the two modes.
Skimming through the old issue that was referenced [here](https://discuss.pixls.us/t/any-interest-in-a-film-negative-feature-in-rt/12569/108), it seems that the motivation for auto-balance was to support pictures where camera wb was way off. There were some links to example pictures in the issue, but they all seem to be expired from imgur/filebin, or disappeared from the `test_images/` directory of the site.
So, i've tried to create an example myself. I've set my in-camera wb to 9900K, then i took a test shot using a 2700K light source. Finally, i opened the file in RT and manually set white balance back to 2700K. This way, RT is compensating for the bad multipliers read from camera metadata.
- by setting preprocess WB to "Auto", this compensation happens _before_ demosaic & preprocess.
- by setting it to "Camera" instead, demosaic & preprocess are performed on badly-scaled channels, and the compensation happens _afterwards_ , in the usual WB tool.

Here is the difference:
image
notice how the bottom image (using "Camera" mode), looks a bit softer and has strong color deviation in the high-contrast pattern.
Capture sharpening was disabled and False Color Suppression Steps was set to zero. Setting FCSS to >0 fixes most of the color deviation in Camera mode, but Auto mode still looks better.

Anyway, with my camera (sony A7 mk I), i can only observe this difference in pictures where native camera multipliers are _completely_ off.
Testing with the sample raw file provided in issue #5616 , instead, i can confirm the exact same behavior reporder by @heckflosse , and Camera mode looks way better than Auto.

So, i guess each camera has its own taste for multipliers :-D

heckflosse:
> So, is there any good reason (or an example) not to use the camera multipliers?

Here an example where camera wb leads to artifacts (it's an uniwb shot)
left auto, right camera
grafik

Another one, also uniwb
grafik

heckflosse:
> So, i guess each camera has its own taste for multipliers :-D

No, it's just when auto wb fails to get the right values

rom9:
> No, it's just when auto wb fails to get the right values

ooh, ok so maybe i'm starting to get this. The reason why the initial reddish wall (or floor?) picture works better with Camera instead, is just because the entire picure is made of the same reddish shade, so finding the multipliers by assuming that the global channel average is neutral gray, will give a result that is far from reality. While UniWB shots on the other hand, will almost always work better with Auto, because they are natively (and purposely) off balance with camera mults.
Am i correct?

Unrelated: what about the sliders? I'm more and more convinced that they're a bit overkill... finding those values by hand it's tricky to say the least. I'd like to remove them, what do you think?

heckflosse:
@rom9 
imho we should keep the auto/camera choice and remove the sliders.

While the auto setting (as it is in RT now) often gives good results, your implementation of `camera` is welcome for the cases, where `auto` fails :+1:

heckflosse:
> The reason why the initial reddish wall (or floor?) picture works better with Camera instead, is just because the entire picure is made of the same reddish shade, so finding the multipliers by assuming that the global channel average is neutral gray, will give a result that is far from reality.

Exactly. Maybe itcwb would give better auto wb results here. But that's a different issue.

rom9:
Sliders removed ;-)
image

But now, the Mode combo looks a bit lonely in its own panel... i was wondering about moving it inside the existing "Preprocessing" panel, it's related after all... what do you think?

Thanatomanic:
@rom9 I would not put it in the dead/hot pixel panel. As a matter of fact, I think this panel's name is very confusing because there is also a Preprocessing sub-panel in the Demosaicing panel.
I think the dropdown would fit best without a specific sub-panel just below the Demosaicing sub-panel. The label could be 'White balance preprocess mode:'

Otherwise, this seems good to merge, right?

rom9:
@Thanatomanic you mean like this?
image
The problem with this layout is that the dropdown must actually be present in _two_ different parent containers: SensorBayer and SensorXTrans. These are just empty containers, and are populated by ToolPanelCoordinator.
I'm not sure about the best way to implement this: shoud i make both SensorBayer and SensorXTrans implement a common superclass, containing the dropdown? But in this case, should the dropdown component be laid out indirectly by ToolPanelCoordinator as well ?
Even from a conceptual point of view, this feature is not sensor-specific, so i'm a bit confused about having it inside both sensor panels.
Any suggestion is appreciated :-)

Thanatomanic:
You're right, it's more complicated than I thought and leaving the dropdown in its own panel for now is maybe for the best.
Quite honestly, what is the benefit of having a sensor-specific container at all? I'll give that some thought and maybe come up with a more agnostic layout in a separate PR...

heckflosse:
Here's another case where we need the functionality from this branch:
https://discuss.pixls.us/t/sometimes-1-pixel-artifacts-from-x-trans-demosaicing/18179/9

heckflosse:
@rom9 Could you solve the merge conflicts please? I will test then, so we can merge asap

rom9:
Hi @Thanatomanic , @heckflosse (don't know if this is the right place to discuss about this). Regarding the [example](https://discuss.pixls.us/t/white-aronskelk-arum/18504/3?u=rom9) on PlayRaw, i made some test and it seems that the color difference is only apparent when using Auto ITC white balance method.
That is: if you use Auto ITC WB method _and_ Auto CA correction is _disabled_, then there is a strong color difference between Camera and Auto preprocess WB.
If you use "Auto RGB grey" method, then the difference is not noticeable (you only can detect it using a lockable color picker).
If you use Spot WB, then there is no change at all, even with the color picker.
That said, i have no idea how this should be handled... do you think it is an issue ?

heckflosse:
@rom9 I tried to reproduce the difference (but failed, as I did not use Auto ITC in my tests). If the differences are only when using Auto ITC, then it's clearly not related to raw preprocessing WB.
Perhaps we should open an Issue concerning Auto ITC for this case

rom9:
@heckflosse it is also related to Auto CA correction, though, because with that _enabled_ you get the same results regardless of the preproc wb setting.
With Auto CA correction enabled, if you then increase the _Iterations_ to 3 or above, the color difference comes back, as if Auto CA was disabled.
It's a weird mix of things...

heckflosse:
@rom9 means it is not related to raw preprocessing wb, right?

rom9:
@heckflosse right, it does not seem to be related to preprocessing wb. I've also tried to comment the new preprocess wb code block at the start of `::preprocess()`, and move it back into `::load()` to be sure to replicate the old behavior, and still enabling or disabling Auto CA correction makes a difference in the computed WB temp and tint values. This only happens with Auto ITC WB, so yes, it must be only related to Auto ITC.
I'll open an issue right now :-)

PR #5672, opened 2020-02-27
Cat02preset - use simplified ciecam02 preset to generate chromatic adaptation cat02

Desmis:


Desmis:
I changed, compared to the "cat02wb" branch, the way of realizing cat02 automatic. Indeed the version in the "cat02wb" branch was at another place in the process. The algorithm is the same, but the results seem better now. The only consequence is at the GUI level where you have to look for the check box at the start of Ciecam,  instead of just after WB.

jacques

Desmis:
if there are no objections, I will merge this pull request tomorrow

Thanatomanic:
@desmis I don't have particular objections to merge.
I am not a user of the ciecam options in RT, because I don't understand their added value. Also, the number of options in the module is intimidating and I never understood the usage of the different sliders (some edits make things look really weird).
My conclusion: if something is now automatic due to cat02, and somehow improves the image, or gives a better starting point, then that's a good thing.

Desmis:
@Thanatomanic 
I think, perhaps maybe I'm wrong, that a lot of people are intimidated by certain module. in particular Ciecam, Wavelet, Retinex

This problem refers to the perceived complexity, in french we say : "une question n'est difficile que si on ne connait pas la réponse"

The organization of the RT project team is essentially organized around IT aspects (this is not a criticism, but an observation) - IT aspects are not my strong point.
after against I have a very strong social and technical culture (Human factor, management, Color management, etc.)
it is unfortunate that we do not have discussion zones on these subjects which are fundamental in the conduct of a project.

For example to explain this ignorance, I rely on the work of a Danish - Jes Rasmussen (1986) who modelize the 3 forms of action, and learning
S = skill, we do tasks automatically....we have learn by various channels (reading, TV, neigbours, etc.)
R =rules - we will look in our memory for things that we know, but which are not automated
K = knowledge - we learn by working, we search in books,, in encyclopedias, etc. and we understand if our level of knowledge and skills are sufficient

This taxonomies allows to understand why some people agree with some tolls and others not.

Imagine Leonard of Vinci arriving in our time. Despite his remarkable intelligence, he will not immediately understand what a smartphone is, what it is used for, etc.... when a 5 year old child knows how to use it. this shows the influence of the cognitive environment (collective learning).

Does that mean that de Vinci is less intelligent than a current child of 5 years, of course not...
De Vinci in in a few days will catch up by learning

But, I get lost, but reflected on this paradox, no discussion area between us.
In fact, when we understand, Ciecam is  simple, wavelet is smple

jacques

PR #5677, opened 2020-02-28
Use of thread-safe *time_r() functions

Benitoite:
Uses drop-in compatible thread-safe functions.
cf. https://lgtm.com/rules/2154840805/


Floessie:
@Benitoite How is that supposed to work if you don't provide a buffer to `gmtime_r()`? Shouldn't the second parameter be a `struct tm*`? Maybe I'm missing something, but I couldn't quickly find a positive example with `NULL` as buffer. Even if it allocated in case of `NULL`, that would surely introduce leaks...

Benitoite:
The memory referenced by the second parameter would be the same as the function output... so nullptr since there is no need for the second parameter in the current implementation.  Clang did not allow not sending the second parameter.

Floessie:
@Benitoite 
> Clang did not allow not sending the second parameter.

Sure, but `NULL` isn't a solution to the problem. Example:

```c++
#include <ctime>
#include <iostream>

int main()
{
    const std::time_t t = std::time(nullptr);

    std::cout << std::ctime(&t);
    std::cout << ctime_r(&t, nullptr);

    return 0;
}
```

```
Mon Mar  2 10:58:30 2020
Segmentation fault
```

Benitoite:
So it looks like the 26 char buffer is required.

Benitoite:
@Floessie Appreciate you having a look at the attempt.  Here is what I have made work in C so far...
```
#include <time.h>
#include <stdio.h>

int main()
{

    char buf[26];
    long t;
    time(&t);
    ctime_r (&t, buf);
    printf("%s\n", buf);

    return 0;
}

```

``` 
% ./ctimetest
Tue Mar  3 09:36:42 2020

```

Floessie:
All fine. The non-`_r` functions have a shared (static) internal buffer, thus cannot be used in multithreaded programs. So you have to provide an exclusive buffer to them.

Benitoite:
Note: For fun I tried out a result-handling variable, but it `valgrind`s the same with {0B,0B,72B} leaks.

```
#include <time.h>
#include <stdio.h>

int main()
{
    char buf[26]= "abcdefghijklmnopqrstuvwxy";
    long t;
    char *result;
    
    result = buf;
    time(&t);
    result = ctime_r (&t, buf);
    printf("%s\n", result);
    
    return 0;
}
```

```
==35025== 
==35025== LEAK SUMMARY:
==35025==    definitely lost: 0 bytes in 0 blocks
==35025==    indirectly lost: 0 bytes in 0 blocks
==35025==      possibly lost: 72 bytes in 3 blocks
==35025==    still reachable: 16,226 bytes in 28 blocks
==35025==         suppressed: 21,921 bytes in 154 blocks
==35025== 
```

Beep6581:
Hey @Benitoite , what is the status?

Benitoite:
@Beep6581 Still something I'd like to figure out or collaborate more on.  Should be able to test more once the new macOS version can build everything.

PR #5679, opened 2020-03-02
mac:fix passage of CACHE_NAME_SUFFIX

Benitoite:
Reformatting this command fixed a problem where the last variable dropped off and was reset with git info in a secondary script.


Benitoite:
Now part of #5703.

PR #5739, opened 2020-05-03
rtexif: Fix conversion from RATIONAL to int, double or string

pinchartl:
RATIONAL tags are defined as a ratio of two LONG, themselves defined as
32-bit unsigned integers. The value is misinterpreted when converting to
an int, double or string, as the numerator and the denominator are
interpreted as signed values. The problem has been noticed with the
ExposureTime tag generated by libtiff, which sets the denominator to
0xffffffff for exposure times lower than 1. Fix it.

Signed-off-by: Laurent Pinchart <laurent.pinchart@ideasonboard.com>

### DIFF
--- a/rtexif/rtexif.cc
+++ b/rtexif/rtexif.cc
@@ -1510,8 +1510,6 @@ int Tag::toInt (int ofs, TagType astype) const
         return attrib->interpreter->toInt (this, ofs, astype);
     }
 
-    int a;
-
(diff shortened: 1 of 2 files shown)


pinchartl:
Any comments on this ?

heckflosse:
Looks good to me. @Floessie @Thanatomanic

PR #5742, opened 2020-05-07
Do not set math errno

heckflosse:
In RT we never check [math errno](http://www.cplusplus.com/reference/cerrno/errno/).
Some functions (for example std::sqrt set errno and because they have to do that, loops calling such functions will not be vecorized by the compiler.

This example-code will not be vectorized:
```C++
#include <cmath>

void sqrtrow(float *data, int width) {

    for (int i = 0; i < width; ++i) {
        data[i] = std::sqrt(data[i]);
    }
}
```

If we use `-fno-math-errno`, the errno will not be set and the loop will be vectorized by the compiler

Any objections to use this flag?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -130,6 +130,7 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror=delete-incomplete")
 
 # Do net set math errno, as we never check its value.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-math-errno")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-math-errno")
 
 # Special treatment for x87 and x86-32 SSE (see GitHub issue #4324)
 include(FindX87Math)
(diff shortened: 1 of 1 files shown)


heckflosse:
From [here](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html)

```
-fno-math-errno

    Do not set errno after calling math functions that are executed with a single instruction, e.g., sqrt. A program that relies on IEEE exceptions for math error handling may want to use this flag for speed while maintaining IEEE arithmetic compatibility.

    This option is not turned on by any -O option since it can result in incorrect output for programs that depend on an exact implementation of IEEE or ISO rules/specifications for math functions. It may, however, yield faster code for programs that do not require the guarantees of these specifications.

    The default is -fmath-errno.

    On Darwin systems, the math library never sets errno. There is therefore no reason for the compiler to consider the possibility that it might, and -fno-math-errno is the default.
```

Thanatomanic:
Seems all right. So I take it we don't use "IEEE exceptions for math error handling"?

heckflosse:
> So I take it we don't use "IEEE exceptions for math error handling"?

Correct, that would not play well with crunching numbers in vetcorized code.

Floessie:
* Debian Testing AMD64, GCC 9.3: :heavy_check_mark: 
* Debian Testing AMD64, GCC 10.0: :white_check_mark: (there's an unrelated compiler bug)
* Debian Testing AMD64, Clang 10.0: :heavy_check_mark: 
* Debian Buster i386, GCC 8.3: :heavy_check_mark: 
* Debian Unstable PPC, GCC 9.3: :heavy_check_mark: 
* Debian Unstable PPC, GCC 10.1: :heavy_check_mark: 
* Debian Unstable PPC, Clang 10.0: :heavy_check_mark: 
* Debian Unstable PPC64, GCC 10.1: :heavy_check_mark: 

I don't see any reason for not adding the parameter also to `CMAKE_C_FLAGS`.

Floessie (changes requested):
`CMAKE_C_FLAGS` should also be tweaked.

Issue #5791, opened 2020-06-07
wavelet contrast: Levels 4,5,6 non-functional


ff2000:
I am on newlocallab2.

```
Version: 5.8-2065-g188f77804
Branch: 5.8-2065-g188f77804
Commit: 188f77804
Commit date: 2020-06-06
Compiler: gcc 10.1.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.1
Lensfun: V0.3.2.0
Build type: RelWithDebInfo
Build flags:  -std=c++11 -fno-tree-loop-vectorize  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O2 -g -DNDEBUG -ftree-vectorize
Link flags:   
OpenMP support: ON
MMAP support: ON
```

Changing wavelet contrast levels 4,5,6 has no effect. RT-5.8 (and also a recent commit some time ago) had no issues.
I tested with different RAWs from different cameras (SIGMA fp, Sony A7II, Olympus EM-1, Panasonic G85, Fuji XT-20) and all show the same behaviour.
Note that it doesn't matter how many levels you select. 4 Levels selected and 1,2,3 work, 4 doesn't. 9 levels selected and 1,2,3,7,8,9 work, 4,5,6 don't.

RT5.8 is OK:
![wavelet_contrast_RT5 8](https://user-images.githubusercontent.com/1973708/83971675-36e25080-a8e5-11ea-9197-b500030599af.jpg)

RT-git has issues:
![grafik](https://user-images.githubusercontent.com/1973708/83971698-727d1a80-a8e5-11ea-990b-3cf84089c3e9.png)

heckflosse:
@Desmis Jacques, shall I bisect this or do you already have an idea what causes that?

Desmis:
Perhaps it is a bug, but I don't think so

First
Try : replace "selective luminance range" by "whole luminance range"...works

reselect : "selective luminance range" 
and now move the double sliders "Finer luminance range" to the left....works (I think)
change "damper" and offset...works

go to the curve "Finer -Coarser levels"  - move the 2 grrens point to the right... works



@TechXavAL is writting  a documentation to explain

jacques

ff2000:
All right, works.
[...]

PR #5795, opened 2020-06-10
Improve wavelet contrast with sliders instead of curve issue 5791

Desmis:
I have change the behavior...
The main problem since  origin, is a GUI problem...
I think now it is solved

jacques


ff2000:
Thank you @Desmis, the sliders are functional now.
However there seems to be a bug with opening previously edited files.
I created a video: https://youtu.be/P_ZATVKy5q4
File first edited with RT-5.8, then opened in RT from this PR. You see that the gras area left of the tree doesn't show the effect of the wavelet tool, instead the trees are affected. I have to disable and reenable the contrast tool to get it displayed correctly.

Desmis:
@ff2000 

I think it will **be almost impossible to ensure compatibility** .... The operation is different...
To sum up, in 2014 I imagined an operation with the inverted "shadows threshold" slider...not intuitive at all

When @TechXavAL  begin to write documentation, he showed the difficulties of understanding.

After several stages **I had the wrong idea of the curve, which I fully assume**.

All these difficulties come from the GUI and my weak knowledge in this field.

Since one week, I got down to the GUI and found this solution which seems to work properly and is intuitive.

In the first step, I take into account the position of "wavelet level"
This value this sets the maximum of the 2 sliders
Then, the current value of "Finer levels" is passed to "Coarser levels" to set the minimum value

Now to sum up:
for "finer levels" : Only levels beyond the chosen value will be affected by the highlight luminance range. 
for "coarser levels": Only levels above the chosen value and selected 'wavelet levels' will be affected by the shadow luminance range.

ff2000:
> I think it will **be almost impossible to ensure compatibility** .... The operation is different...

I understand that, I don't mind if the effect of the tools in RT-git is slightly different from RT-5.8.
The problem I raised in the previous reply to this PR is on a different issue.
The original issue #5791 seems to be fixed. There is just one small issue left when loading a file previously edited in RT-5.8. I can workaround the issue by disabling and then enabling the wavelet contrast tool, as shown in the video.

Beep6581:
@Desmis could you delete your branches which have been merged and are no longer needed?

Desmis:
@Beep6581 
Done :)

PR #5797, opened 2020-06-13
Crash caused by Tone Mapping, fixes #5171

### LINKED ISSUE #5171: Division by zero in Tone Mapping
I was experimenting with Tone Mapping to see how it affected the tree in the supplied filebin and noticed that as I moved the Scale slider the preview darkened at certain specific points then returned to normal as I continued to move it.  This effect was sensitive - just one unit of movement was enough to make the difference.  After a while RT crashed out completely.  There were errors messages in the Terminal window, see filebin, I think one being produced each time the preview darkened.
Version: 5.5-98-g9210a56    Branch: dev    Commit: 9210a56    Commit date: 2019-02-04

https://filebin.net/mpv5nue16vf2mc79

Please delete copies of the CR2 once this issue is concluded/reproduced.

heckflosse:


Desmis:
@heckflosse 
Ingo

I tested...crash different, but always crash...you can set "edge stopping" to 4 , no crash, but if you decrease "edge stop"...crash
If I set in "local contrast" - darkness to 0...no crash

jacques

heckflosse:
@Desmis Jacques, thanks for testing. I will have a look...

Desmis:
Ok, thank you

Good night 
jacques

heckflosse:
I reverted the first commit and made another one, which avoids the segfault, but shows artefacts in the cases where it segfaulted before.

Desmis:
@heckflosse 
I tested

I confirme, no more crash
But
* differences between without patch and with patch, when "normal" settings
* strong artifacts when "edge stopping" = 4

But, it works :)
Jacques

heckflosse:
@Desmis Jacques, how large are the differences?

Desmis:
@heckflosse 

differences in TIF : Big, very big...


But perhaps...and silly remark .... isn't it easier to limit "edge stopping" when "local contrast" is used
For example , if no local contrast : range edge stopping 0==>4, if local contrast edge stopping 0==>3.2 (by calculation of course : 4 * 0.8)

Ok, it is with this image, but I try with very, very big values in local contrast and no artifacts, no crash

Jacques

heckflosse:
@Desmis Please test last commit.

Desmis:
When you increase "scale"... with big values "edge stopping"...and "local contrast"....also crash, with edge stopping about 3...

probably (perhaps ?)
in createblur
a[x + w * y] = Scale * pow_F(0.5f * sqrtf(gx * gx + gy * gy + sqreps), -EdgeStopping);
lead when Scale very high, or Edgestoping very high...to crash....due in some images to "bad" gradient ???
Have you tried to increase "sqreps = 0.0004f"
But perhaps it is another thing ?

jacques

Desmis:
@heckflosse 
I tested
No crash with "scale" normal
If I push scale to 10...crash
If we limit edge stopping with 2 parameters 
0..4 ; if locla contrsat desabled
0..3.2 : if scale < 8
0..3 : if scale > 8
and we can make a transition...or simply 0..3 in all cases when "local contrast" enabled

It is not a good solution... a work around !
what do you think ?

Jacques

heckflosse:
@Desmis 
> or simply 0..3 in all cases when "local contrast" enabled

I would prefer this one :+1:

heckflosse:
@Desmis Can you push the above mentioned change please? I am on newlocallab2 currently

heckflosse:
@Desmis Shall we merge this pr now?

Desmis:
@heckflosse 
Yes I think, no more crash

You do, or I do ?
jacques

heckflosse:
You ;-)

Desmis:
done :)

Desmis:
But we must make the same change... in 
EPDToneMaplocal
EPDToneMapCIE

but there is not fire :)

jacques

Issue #4261, opened 2017-12-27
Verbose feedback user warning if PP3 does not apply as expected


Beep6581:
RawTherapee should provide verbose user feedback if some element of the processing profile fails to load, for example:
- A flat-field image was not found - show the filename.
- A dark-frame image was not found - show the filename.
- The PP3 uses a tool which does not exist - show the keys.
- The HaldCLUT could not be found, e.g. #4388 - show the filename.
- The input/working/output profile is not found (e.g. #4478) - show the filename.
- The input/working/output profile is found but is not identical (PP3 could store the hash).
- An input profile could not be loaded, e.g. #4260 - show the filename.
- Raw decoding parameters changed in `camconst.json` (a PP3 is applicable to any camera make/model, so this is a difficult one).

This could be done for example using a Gtk::InfoBar, or a log textbox panel.

P.S. I thought we already had an issue for this but I can't find it...

Beep6581:
I again ran into an issue where I thought my PP3s were broken or I was somehow using the wrong PP3s, until I realized the HaldCLUT was not being found as specified in the sidecar PP3.
We should implement this one way or another for 5.5

heckflosse:
I agree that we should do something here. But I think we shouldn't do it for 5.5. Ok to set milestone to 5.6?

Beep6581:
Yes, it's time for a new release.

PR #5798, opened 2020-06-14
FilmNegative: non-raw file support and better accuracy

rom9:
Rework of the film negative tool. The processing code is now part of `ImProcFunctions` and works after input profile conversion. As a consequence, the white balance tool cannot be used to color balance the resulting image, but must be set to the color temperature of the backlight used to digitize the negative. I've added two more sliders for color balancing.
A single function now handles both raw _and_ non-raw files, as well as the thumbnail image. Inversion is performed on the dcrop image, so the interaction is much faster than before (no need to re-demosaic on each param change).
"Legacy mode" processing is still kept for backwards compatibility; it is triggered when reading an old processing profile _only if_ the film negative tool is enabled. In this case, i've added an "Upgrade" button and a warning message that should urge the user to switch to the new method, so we can get rid of the old code ASAP. Unfortunately, there is no way of automatically upgrade parameters while keeping the output unchanged, so manual re-adjustment is needed.
The "stable multipliers" concept is still valid in this version too, so users can simply adjust one picture and then copy/paste params to the whole film roll.


Thanatomanic:
That sounds great! You even took care of backwards compatibility. I'll give it a testrun.

Thanatomanic:
All right, something seems off. I had a few images that were processed with the old method. I opened the first one and got a nice notification asking to upgrade. After I did that, the colors were really off. So I started editing again from Neutral, picked a base color, picked an neutral spot. And I ended up with this:

image

Something is not right with the red channel, that was correct before.

Then I thought, let's look at another negative. Unfortunately, the update message didn't show up anymore. What's more, the thumbnails already seemed to have been reprocessed.

image

Thanatomanic:
Another observation is that when I apply Neutral and then enable Film Negative, I get a nicely bright image which is only slightly off in color. Then I select a base color and everything turns much darker. Is that normal behavior? If so, what causes it?

Neutral + enable module
image

Then pick base color (somewhere in the dark edge)
image

rom9:
> After I did that, the colors were really off

That is most likely because of White Balance, which now must be set to the temperature _of the backlight_ and not by spot-WB on an empty piece of film. For example, if the negative has been digitized with a xenon flash, WB must be set to 6000k or so. I know it's counterintuitive but i think it's the only way to get the output right. WB multipliers are applied _before_ input color matrix, so if you set those wrong, there is no way to compensate them after the (non-linear) inversion formula.

Regarding the upgrade, yes it is a one-shot operation, no way back. Do you think i should use a checkbox, letting the user go back? I wanted to force the migration a bit, maybe i've been too rough... :-D

rom9:
> Then I select a base color and everything turns much darker. Is that normal behavior? If so, what causes it?

Yes, that is normal. First, an attempt is made to estimate the film base level, based on image median. At this point you can see that the film base does not translate to black, but to mid-gray.  When you pick the true film base level, the output is scaled so that the film base will become 1/512th of MAXVALF (2^-9) which is almost black. This will keep scaling stable when copy/pasting settings on multiple images, regardless of the contents. You know that the film base will always be "anchored" to the same output level, and you can simply adjust levels by dragging the upper bound as needed. Also, changing the reference exponent will keep the film base level stable.

Thanatomanic:
> That is most likely because of White Balance, which now must be set to the temperature of the backlight and not by spot-WB on an empty piece of film. For example, if the negative has been digitized with a xenon flash, WB must be set to 6000k or so. I know it's counterintuitive but i think it's the only way to get the output right. WB multipliers are applied before input color matrix, so if you set those wrong, there is no way to compensate them after the (non-linear) inversion formula.

I understand. So after Neutral, the Camera WB is in use. As this is probably _not_ the correct WB for negative processing I need to adjust it. I can live with that. Maybe it is a good idea to put a little text (not tooltip) in the module explicitly stating that you have to set WB _first_?

> Regarding the upgrade, yes it is a one-shot operation, no way back. Do you think i should use a checkbox, letting the user go back? I wanted to force the migration a bit, maybe i've been too rough...

My expectation was that it's a per-image upgrade not a RT wide change. If it is intended to be the latter, then I would put this choice as a checkbox in Preferences.

Thanatomanic:
@rom9 When setting a reasonable value for the WB (the tint was _way_ too high), the problem with the Parisian bus is gone 😄

rom9:
> My expectation was that it's a per-image upgrade not a RT wide change.

No no, that IS a per-image upgrade, that must be a bug. So, you had a folder with more than one image in legacy mode, selected one image, did the upgrade, went back to the file browser, and now all the images were using the new method, is that right?

> Maybe it is a good idea to put a little text (not tooltip) in the module explicitly stating that you have to set WB _first_?

What if i put the entire tooltip text in the label? Would it be too ugly? Actually it would be handy to have all the steps visible at first glance

Thanatomanic:
> No no, that IS a per-image upgrade, that must be a bug. So, you had a folder with more than one image in legacy mode, selected one image, did the upgrade, went back to the file browser, and now all the images were using the new method, is that right?

My exact steps were: I had a folder with one image that I opened, then did the upgrade, then go back to file browser and navigate to a bigger folder with a few negatives. Those were already processed and opening them didn't show the upgrade option.

rom9:
> Those were already processed and opening them didn't show the upgrade option.

Wild guess: is it possible that those negatives in the second folder were already processed with the dev branch of RT ? The upgrade mechanism is only triggered by an old pp3 that was saved with 5.8 or 5.7 release (and only if the filmnegative tool was enabled).
I tried your exact steps, but i can't replicate the behavor you described i can happily jump between old and upgraded images (also in different folders) and the upgrade label and button appears and disappears as expected.

BTW: if testing goes well, do you think i'm still in time for this PR to be included in 5.9 ? I know i'm super late, but this would mean having only one previous version to be compatible with, instead of two :_(

> When setting a reasonable value for the WB (the tint was _way_ too high), the problem with the Parisian bus is gone

Yes, that's not a very critical value fortunately ;-)

rom9:
@Thanatomanic What about this label? Too much ?

image

Thanatomanic:
> Wild guess: is it possible that those negatives in the second folder were already processed with the dev branch of RT ? The upgrade mechanism is only triggered by an old pp3 that was saved with 5.8 or 5.7 release (and only if the filmnegative tool was enabled).
I tried your exact steps, but i can't replicate the behavor you described i can happily jump between old and upgraded images (also in different folders) and the upgrade label and button appears and disappears as expected.

Your guess is spot on. I've just tested going in 5.8 release, then so some editing, go to 5.8 dev, edit a few of the images, then to your branch. And voila, the images edited in 5.8 release still look fine. The ones edited in dev are crap and the upgrade message is not available. 
I'm not familiar with the mechanisms behind pp3 versions, but there is no way around this? Maybe @heckflosse can chime in?

As for the label: I think this is too much. How about "Upgrading changes the way the images need to be processed. Please refer to the manual for full instructions." Or something like that?

heckflosse:
> I'm not familiar with the mechanisms behind pp3 versions, but there is no way around this? Maybe @heckflosse can chime in?

Sure. in ppversion.h you can define a new version. And depending on the version you may be able to covert processing parameters (values from pp3 file) in procparams.cc

rom9:
> I'm not familiar with the mechanisms behind pp3 versions, but there is no way around this?

Problem is that i re-used a parameter name (greenBase) that i had already used in my previous PR (which is now in dev). I could use a different param name so i could detect 5.8 vs dev vs new version, but anyway if we also need to support backwards compatibility with the current (unreleased) dev, then i need to handle a third case (old method + dev method + new method).
My intent was to completely discard my previous PR by "overriding" it with this one, in the hope that dev users might be more tolerant to breakage since they're using an unstable version.

heckflosse:
> if we also need to support backwards compatibility with the current (unreleased) dev

We never cared about backwards compatibility to dev versions, only to releases

Thanatomanic:
@heckflosse @rom9 Sorry for the self-indulgent wishes for compatibility 😉 

I like this new version a lot. It's fast and accurate - as far as I have tested. 

> When you pick the true film base level, the output is scaled so that the film base will become 1/512th of MAXVALF (2^-9) which is almost black. This will keep scaling stable when copy/pasting settings on multiple images, regardless of the contents. You know that the film base will always be "anchored" to the same output level, and you can simply adjust levels by dragging the upper bound as needed. Also, changing the reference exponent will keep the film base level stable.

With regards to this: would it be possible and/or convenient to do some automatic scaling after setting the lower end? It is a little annoying that you _have_ to go back to the exposure slider after setting the base level. If the film negative module could silently scale the exposure up until clipping of the first pixels or so, that would give an image that is much more accessible to further editing immediately.

rom9:
> If the film negative module could silently scale the exposure up until clipping of the first pixels or so, that would give an image that is much more accessible to further editing immediately.

I did try in the old version, but it's not very easy to detect the max value, because of the film holder (see the white border at the bottom of the first picture, for example). The film holder is very dark, so it will always clip in the output, and it's hard to predict its size: some users take very "zoomed-out" shots where the holder constitutes the majority of the area.
The only solution i can think of is finding a maximum ratio given by the physical limits of the film such that the channel value of a fully exposed spot will never be less than 1/x of the film base value. Then i could increase the scaling, and subtract as necessary to take the film base output back to near-zero.
But i'm always afraid to hardcode some guesstimated value which leads to loss of information... i'll try to do a little survey on the film samples that i have.

> As for the label: I think this is too much. How about "Upgrading changes the way the images need to be processed. Please refer to the manual for full instructions." Or something like that?

i've updated the label, but this part: _"changes the way the images need to be processed"_ felt a little intimidating to me :-)
I used:
  _WARNING: some parameters will need to be adjusted. Please refer to the manual for full instructions._

How does that sound ?

Beep6581:
Hey

> In this case, i've added an "Upgrade" button and a warning message

Yay, sounds like work towards #4261 .

Compiling...

rom9:
Sorry, i have to freeze this PR for now... i got some problematic samples on the forum and i need to investigate more.
If i don't do it in time for 5.9, that's ok, i'll handle the backwards compat for both versions. Also, the upgrade button _is_ scary.
@Beep6581 agree on the idea of a common "platform" for warnings, i'll think about it.

heckflosse:
> Sorry, i have to freeze this PR for now... i got some problematic samples on the forum and i need to investigate more.

No need to apologize. Take your time to do it right :+1:

rom9:
Hi all, sorry for the huge delay, this commit should simplify things a lot. I've removed the scary upgrade button since the new upgrade mechanism should yieald almost identical result to the previous version. The colorspace used to perform the inversion is now user-selectable.
Some questions about coding conventions:
- i wanted a handy RGB triplet data type, so i made this nested `RGB` struct inside FilmNegativeParams
https://github.com/Beep6581/RawTherapee/blob/0f99e9be09bc053603383ca07eaf80ef9acddb33/rtengine/procparams.h#L2154
and used it throughout the code. Is this bad?
- also, i've added support for `float` data type and the `RGB` triplet itself to the `getFromKeyfile`/`putToKeyFile` functions:
https://github.com/Beep6581/RawTherapee/blob/0f99e9be09bc053603383ca07eaf80ef9acddb33/rtengine/procparams.cc#L119
https://github.com/Beep6581/RawTherapee/blob/0f99e9be09bc053603383ca07eaf80ef9acddb33/rtengine/procparams.cc#L170
https://github.com/Beep6581/RawTherapee/blob/0f99e9be09bc053603383ca07eaf80ef9acddb33/rtengine/procparams.cc#L250
Was it a bad idea?

Thanks :-)

Floessie:
@rom9 Having a `RGB` struct inside `FilmNegativeParams` is not a bad idea if we don't have a similar type somewhere else. Be aware that you can't forward declare an inner struct. The `RAWParams` just above do the same.

The `getFromKeyfile()` is also a good idea (make `v` const and maybe drop the `static_cast`s on assignment), but I'd expect to find a corresponding `putToKeyfile()` so that `saveToKeyfile<>()` can be used.

HTH,
Flössie

rom9:
Thanks @Floessie !

> is not a bad idea if we don't have a similar type somewhere else

i've looked everywhere but couldn't find any, so i decided to roll my own. I wanted something like `rtengine::Coord` is for coordinates, but it seems that there isn't an equivalent for RGB triplets.

> Be aware that you can't forward declare an inner struct.

`procparams.h` should already be included almost everywhere in the code, so this shouldn't be a problem?
One thing that could look weird, is the fact that i often create `RGB` data structures just to do some calculations and pass data around between internal functions, not as part of a `FilmNegativeParams` instance. Is this too confusing? Maybe `RGB` would make more sense as a top-level type? In this case, where should it be declared/defined? Any suggestion is appreciated, sorry again for my C++ ignorance :-)

> but I'd expect to find a corresponding putToKeyfile()

done, and removed the now redundant `RGB::toVector()` method ;-)

Floessie:
> Maybe RGB would make more sense as a top-level type? In this case, where should it be declared/defined?

That's a good question. @heckflosse Ingo, any opinion and/or idea?

heckflosse:
> Ingo, any opinion and/or idea?

In my opinion, if we have a use case outside filmnegative, we can move some elsewhere

heckflosse:
@rom9 
cppcheck gives 2 hints:
```
[H:/rt59/rtengine/filmnegativeproc.cc:236] (performance) Parameter 'params' is passed by value. It could be passed as a const reference which is usually faster and recommended in C++. [passedByValue]
[H:/rt59/rtengine/filmnegativeproc.cc:236] (performance) Parameter 'icmParams' is passed by value. It could be passed as a const reference which is usually faster and recommended in C++. [passedByValue]
```

Thanatomanic:
@rom9 I like your improvements! I think the Cool/Warm and Magenta/Green sliders are fine. I am a bit hazy on the _why_ of these sliders: is it because the White Balance module works before the Film Negative module and you shouldn't actually use it if you use Film Negative, therefore needing an alternative white balancer?

I'm going to nitpick on the GUI for a moment. Because of the two buttons and the line halfway the module, there are two distinct parts to the tool. This makes sense. However, I also get the impression that because the buttons are at the bottom, they do something to _all_ the sliders in their respective part. This is not so as far as I can tell.
Maybe it's better to move the 'Pick neutral spots' button above the two ratio sliders, keep the middle line and also move the 'Pick white balance spot' to above the Cool/Warm and Magenta/Green sliders. What do you think?

I also found a little bug: click 'Pick white balance spot', then the cursor changes and the mode in the toolbar changes (icons on top of the preview), go to another tool, e.g. Crop, and select a new crop. The cursor keeps showing the picker icon.

rom9:
@Thanatomanic 
> is it because the White Balance module works before the Film Negative module and you shouldn't actually use it if you use Film Negative, therefore needing an alternative white balancer?

yes, exactly. When performing the inversion in the working or the ad-hoc "FilmNeg" colorspaces, in theory you should set the main WB tool to balance the _backlight_, in order to get the most accurate image of the negative. This image is then converted to the desired colorspace, the inversion is performed, and finally you apply manual adjustments with the alternative white balancer.
If you set the main WB wrong, you won't be able to compensate in the final WB step, because the negative inversion is a non-linear operation: you are in fact exponentiating a _weighted sum_ of channels, instead of a single channel.
So, the only solution i've found was to add a second color balance step downstream. Which is confusing, i know :-(

> Maybe it's better to move the 'Pick neutral spots' button above the two ratio sliders, keep the middle line.

Ok, i see, so the button should be on top of the sliders it affects. Yes, this makes sense.

> and also move the 'Pick white balance spot' to above the Cool/Warm and Magenta/Green sliders.

Well, if you try picking spots of different luminance, you see that it actually affects _all three_ sliders, including the "Output Level", so maybe i should place the button above Output Level, just below the separation line ?
Here's how the sliders work: when you pick a spot, the backend returns two RGB triplets: the input (negative) color, and the corresponding output (positive) color, obtained after inversion with the current settings.
The "Output Level" slider is set to the luminance of the output color, and the "Cool/Warm" and "Magenta/Green" sliders are set to 1.0. This will inform the processing stage that the sampled input color must correspond to a neutral gray of that luminance in the output image. From that constraint, channel multipliers will be computed accordingly.
So, on the GUI this works backwards compared to the normal WB tool: picking a spot does not set the balance sliders to appropriate values, but instead always sets them to 1.0, and additionally, sets the Output Level to a value such that the overall luminance does not change after picking the new values.

> I also found a little bug: click 'Pick white balance spot', then the cursor changes and the mode in the toolbar changes (icons on top of the preview), go to another tool, e.g. Crop, and select a new crop. The cursor keeps showing the picker icon.

Thanks, nice catch! This also seems to affect Graduated Filter (from which i copied initially), Tone Curves, RGB Curves, and to some extent maybe all tools with an editor mode. Looks like it's caused by this line:
https://github.com/Beep6581/RawTherapee/blob/894947fce377e6dedaa8fc27f2acd274223ce446/rtgui/toolbar.cc#L112
`stopEdit` is never true unless the current tool is `TMHand`, so `switchOffEditMode` is never called on the `EditSubscriber`s.
I'm afraid to remove that TMHand condition :-) ... does anybody have any idea?

rtengine/improccoordinator.cc
```diff
@@ -613,8 +614,17 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
...
+                    procparams::FilmNegativeParams::RGB refIn, dummy;
+                    getFilmNegativeSpot(-1, -1, 32, refIn, dummy);
+                    changed = params->filmNegative.refInput != refIn;
+                    params->filmNegative.refInput = refIn;
+                }
+
+
                 // Perform negative inversion. If needed, upgrade filmNegative params for backwards compatibility with old profiles
-                if (ipf.filmNegativeProcess(orig_prev, orig_prev, params->filmNegative, params->raw, imgsrc, currWB) && filmNegListener) {
+                if ((changed | ipf.filmNegativeProcess(orig_prev, orig_prev, params->filmNegative, params->raw, imgsrc, currWB)) && filmNegListener) {
```
Floessie: @rom9 I know, this is just a POC, but I guess you meant `||`...

rom9: No, i really meant "|" because i want it to process in any case. Sorry for that, i know it's a super-dirty hack, i needed it just to let users on the forum make a quick evaluation. I'm going to squash this commit out of existance very soon :-)

Floessie: @rom9 Maybe `if ((ipf.filmNegativeProcess(...) || changed) && filmNegListener)` would have been more appropriate. Took me some time to realize the shortcut semantics workaround. :wink:

rom9: Dirty hack now removed ;-P

Thanatomanic:
@rom9 I tested your latest changes and they are a great improvement. I have very little difficulty to get very good results with a variety of negative samples I collected. Even without setting a proper 'backlight' whitebalance.

The only bug I see is the one I mentioned with the color picker. Other than that I would be :+1: for a merge!

Lawrence37:
I think the color picker issue is not a problem with this tool, but rather the crop tool or the on-canvas tools infrastructure.

rom9:
> I think the color picker issue is not a problem with this tool, but rather the crop tool or the on-canvas tools infrastructure.

Yes, exactly. If everybody agrees i can change the `stopEdit` condition described above, but that would affect many other tools as well... should i open an issue about it ?

Also, i have a question about the labels: in the combo box and the related tooltip, i used the term "Camera" color space... but since the tool now supports scanner tiffs too, should i call it "Input" color space instead ?
Note that i deliberately used the term "color _space_" and  not _profile_, since the actual profile is specified elsewhere (in the Color Management Tool).
Please pardon my ignorance, i'm not sure about the correct terms here :-)

Thanatomanic:
Ping @heckflosse for the `stopEdit` condition.

@rom9 I think "Input color space" describes the situation the best.

rom9:
Hi @Floessie , following your suggestion on pixls, i 've added a `Film Negative.pp3` file under the `rtdata/profiles` dir, next to `Unclipped.pp3`. Is that the correct path to store bundled profiles? Also, is there any other operation i need to do (like update some install script or whatever) in order for the new pp3 to be provided as a bundled profile?
Please feel free to RTFM me :-D

Floessie:
@rom9 Great!
> Also, is there any other operation i need to do (like update some install script or whatever) in order for the new pp3 to be provided as a bundled profile?

No, I don't think so:

https://github.com/Beep6581/RawTherapee/blob/dev/rtdata/CMakeLists.txt#L50

Thanks for this cool tool!

Best,
Flössie

rom9:
@Thanatomanic : please can you give me some advice on word choice? The first slider was always labeled "Reference Exponent _(contrast)_", also in previous versions, but since it indicates an _exponent_, maybe i should change it to "Reference Exponent _(gamma)_" ?
(as i understand it, "contrast" would imply an "S" curve, while in my case i'm applying a simple power curve...)
Or, should i leave it as it is, to avoid too much surprise for existing users ?
Sorry about my deep ignorance :-)

Thanatomanic:
@rom9 Hi, I forgot you asked for my input here. What about just leaving out any term between parentheses? "Reference exponent" should be enough to indicate we're dealing with a mathematical power curve. And I think it would be safe enough to assume people know that when an exponent increases, you can expect a brighter image.

rom9:
@Thanatomanic  Thanks, that's perfect ;-)

Regarding this PR, do you think it's good enough to be merged into dev? Is there a minimum quota of thumbs up that i have to reach, or can i merge it already ? :-)
There haven't been any big changes lately, and having this functionality into dev would expose it to even more testing, via the nightly builds.

@heckflosse regarding the `stopEdit` condition [above](https://github.com/Beep6581/RawTherapee/pull/5798#issuecomment-697174003), do you know why it's there?

Thanatomanic:
@rom9 I think you can merge if nobody objects within a few days or so. At least 👍  from me.
A fix for the stopEdit can be applied later.

rom9:
Ok thanks @Thanatomanic !
Should i ignore this macOS build fail on LGTM ? It keeps barking at me with:
```
-- Checking for module 'gtkmm-3.0>=3.16'
CMake Error at /usr/local/Cellar/cmake/3.18.4/share/cmake/Modules/FindPkgConfig.cmake:545 (message):
--   Package 'sigc++-2.0', required by 'glibmm-2.4', not found
  A required package was not found
```
but this PR doesn't require any particular dependency. Also, it seems that all macOS builds are failing since last week...

Thanatomanic:
I think that is safe to ignore, we should create a different issue to fix the Mac builds.

rom9:
Done. I decided to go with the "Squash and Merge" option based on [this](https://github.com/Beep6581/RawTherapee/pull/5887#issuecomment-711415087) discussion, since all this stuff is contained within the FilmNeg feature, and mostly replaces the previous implementation. Also, many commits contradict each other. I hope this is fine :-)
Now i'll open the issue about the macOS build.
Thanks everybody for your support! :-)

PR #5803, opened 2020-06-18
locallab review

heckflosse:
I started a new branch to cleanup, speedup and reduce memory usage of locallab.
I will make lot of small commits to simplify bisecting in case I make nonsense.

First commit is an SSE version for ImProcFunctions::Compresslevels()


Desmis:
@heckflosse 
Ingo
I just tested with and without patch
* on wavcontrast 4 : speedup about 30%

No differences on TIFF

Good job
jacques

heckflosse:
@Desmis Thanks for testing :+1:

Desmis:
@heckflosse 

Ingo

I tested
Works fine :)

jacques

heckflosse:
@Desmis Jacques, in next step I want to speedup and fix bugs for wavcont process == 2

About the bug: There is another race which can be easily reproduced by sending the same image into the queue twice, process the queue and then check for differences. Here's my output of the differences:
grafik

I can apply the same fix as for the other processings (1 and 3), means using beta from largest level and dir for all levels. Would that be ok for this case process == 2 as well?

Desmis:
@heckflosse 
Ingo

I tested your last commit, all works...

For process == 2 and process ==3... I think we must leave as it is (with beta which take into account "level" ans "dir" - the most important is "level")
It does take a little time, but the result will be better

thank you

Jacques

heckflosse:
@Desmis Jacques, with the latest changes to wavcont parallelization it is no problem to use per level/dir beta array. I con do that now if you like.
The problem with former parallelization was, that the beta array was written from multiple threads at the same time, which led to kind of random results.

Desmis:
@heckflosse 

Ingo

No problem you can :)

Thank you
Jacques

heckflosse:
@Desmis Pushed. Now all process use per level/dir beta. process == 2 also got a speedup. Memory usage is reduced as well.

Desmis:
@heckflosse 

I tested, works fine :)

Thank you

This afternoon... bathing and swim...  :)

jacques

Desmis:
@heckflosse 

Ingo

I just tested your new commits

All works fine :)

If I compare before after all patch this branch
with "levels dynamic range" + "levels directionnal contrast"
Speedup  about 50%

Very impressive :)

jacques

heckflosse:
@Desmis Jacques, thanks for testing :+1: I will merge now but keep the branch open for further work. I already have a cleanup and about 25% speedup for Pyramid, Graduated filter on hold ;-)

heckflosse:
@Desmis, I just found a major bug in locallab:

To reproduce:

1) Define a locallabspot in an image, change exposure compensation of the spot to 1.5
=> spot gets brighter, good
2) now set the spot to `excluding spot`
=> nothing changes

Can you confirm? I f you confirm, I will try to fix this bug now...

Desmis:
@heckflosse 

Ingo
No bug, it's the normal behavior :)

Each "excluding" can take all settings...
If you want to "exclude" you must add a new RT-spot

Others things
from tomorrow and for 2 or 3 days I work on the "common mask" (branch "localmask")
Tomorrow I think I will finish GUI part, then I will work on RTengine and on "iplocallab.cc" 

Could you avoid commits on iplocallab.cc...to avoid for me, big difficulties

jacques

Desmis:
But I understand the question...
I you have only one spot...and you change "normal"  - "excluding"  no differences...of course....

If you stack spots, with different settings, and then set "excluding" with "exposure", the result will be different than with "normal"

I verify...I think all works as expected.

Jacques

heckflosse:
@Desmis 
> But I understand the question...
I you have only one spot...and you change "normal" - "excluding" no differences...of course....

Ok, didn't know that. Then I will push only a speedup for `Exclude_Local` ;-)

heckflosse:
@Desmis 
> Could you avoid commits on iplocallab.cc...to avoid for me, big difficulties

I will commit only to my review branch ;-)

Desmis:
@heckflosse 

Ok

But many person thinks as you.... :)

The new "common mask" will be with a very different behavior
Instead the mask modify a tool (color and light, exposure,...) , to improve shape detection, now this new mask will be a new "tools", by itself without the intervention of another tool, it will allow you to modify the image, especialy "Colors / contraste"..it will have a "scope" settings...

I tested with a tampering with the code, and it works well...
 I think that in 2 or 3 days you will be able to test ... we can with this trick get "curious" results, but of course you can get "normal" results :)

jacques

Desmis:
@heckflosse 


I have finished (??) the first part of the job (branch "localmask"), for "common mask" (GUI, link with rtengine, iplocallab.cc...)...huge job :)
I asked to Pierre to control my work  (I work as a monkey...)

Tomorrow or this late afternoon, I will work on iplocallab.cc to achieve "common mask" (I hope ...)

@Floessie  So don't push commit... on iplocallab.cc 
Thank you  :)

Jacques

PR #5810, opened 2020-06-23
Common color Mask using scope 

Desmis:
Here a first try of a new feature...

Common color mask : it works as a tool..and not in enhancemet of a tool

to open : add tool to curren spot ==> Common color mask

Scope is used between "original image" and "mask"

You can get "special effects", but also of course "normal" effects

If blend = 0 , no action

jacques


Desmis:
@heckflosse  @Floessie 

I think now, the system "common mask" is ready for me :)

Just a "detail"...his name...if you have another idea.... "Color mask", "Mask as tool", "Color and structure mask"...etc.

 I took advantage of this "PR" to change certain parts of "settings"....

jacques

Desmis:
this afternoon...beach :)

Desmis:
@heckflosse 
Ingo

I see you "review"

I have forgotten to delete  "buftemp" and "blend2" they do nothing....it is a "copy and paste" error....(forgotten)
Of course you can change....

jacques

heckflosse:
@Desmis You can still change code. I set myself for review as a reminder.

rtgui/locallabtools2.cc
```diff
@@ -4723,3 +4723,740 @@ void LocallabLog::updateLogGUI()
...
+        mask_HCurveEditorG->set_tooltip_text(M("TP_LOCALLAB_HHMASK_TOOLTIP"));
+        radmask->set_tooltip_text(M("TP_LOCALLAB_LAPRAD_TOOLTIP"));
+        lapmask->set_tooltip_text(M("TP_LOCALLAB_LAPRAD_TOOLTIP"));
+        mask2CurveEditorG->set_tooltip_text(M("TP_LOCALLAB_WAVMASK_TOOLTIP"));
+        Lmask_shape->setTooltip(M("TP_LOCALLAB_LMASK_LL_TOOLTIP"));
+        mask2CurveEditorGwav->set_tooltip_text(M("TP_LOCALLAB_WAVMASK_TOOLTIP"));
+        LLmask_shapewav->setTooltip(M("TP_LOCALLAB_LMASK_LEVEL_TOOLTIP"));
+
+    } else {
+        exp->set_tooltip_text(M(""));
```
Pandagrapher, line 4946: @Desmis 
Should just write `exp->set_tooltip_text("")`: this avoid calling `M()` useless function here

rtgui/locallabtools2.cc
```diff
@@ -4723,3 +4723,740 @@ void LocallabLog::updateLogGUI()
...
+    blurmask->setValue(defSpot.blurmask);
+    contmask->setValue(defSpot.contmask);
+    HHhmask_shape->setCurve(defSpot.HHhmask_curve);
+//    Lmask_shape->setCurve(defSpot.Lmask_curve);
+    LLmask_shapewav->setCurve(defSpot.LLmask_curvewav);
+    csThresholdmask->setValue<int>(defSpot.csthresholdmask);
+
+    // Enable all listeners
+    enableListener();
+
```
Pandagrapher, line 5205: @Desmis 
`updatemaskGUI3` function shall be called here as `fftmask` has been set to a default value.

Pandagrapher:
@Desmis 
Just reviewed the GUI part. Just made two small remarks otherwise all is ok for me 👍 

Pierre

Desmis:
Merci,
Thank you
Jacques

Le jeu. 25 juin 2020 20:39, Pandagrapher <notifications@github.com> a
écrit :

Desmis:
@Pandagrapher 
Thank you for this review :)
I just push the 2 changes

jacques

Thanatomanic:
@Desmis I've tried out this branch, but I have a few questions, because I don't think I understand how it is supposed to work. My thought process:
1. I create a simple spot, default settings
2. I add 'Common Color Mask' and 'Color & Light' tool.
3. I modify the 'Common Color Mask' to something that is not uniform:
image
4. I enable 'Color & Light' and push the Lightness.
5. I expect that the 'Common Color Mask' somehow has an effect. Like, it blocks certain colors from being affected, or something. But it doesn't seem to do anything? Disabling the 'Common Color Mask' has no visible effect on the result of the 'Color & Light' tool.

What am I missing?

P.S. In 'Common Color Mask', when I change 'Smooth Radius Mask' even by 0.1 the mask changes _a lot_. Is that by design? It looks strange...
image

Desmis:
@Thanatomanic 
As it is said in **the tooltip, if blend=0 no effect**...I know it is different from others mask...but this is the reason for this new type which works differently from the others

For the "smooth radius mask"... as all the tools made with "guidefilter" (I am not the author of guidefilter), there is a "gap" since radius > 0....several users have already reported this phenomenon ... which is normal (depending on the algorithm) - even with a log function  for low values

jacques

Thanatomanic:
@Desmis Merci, that works. And I believe you if you say this is the normal behavior of the guided filter.

__About Blend__
I am still trying to wrap my head around the way Blend works. I can now get an effect by using Blend != 0, but even if Blend = 0, the "Show mask" options tells me something gets masked? That's a quite confusing.
Going from 0 to 1 also makes a huge difference on the image, but if I show the mask, the effect seems very small.

_With Blend = 0_
image

_With Blend = 1_
image

Can you help me understand what is going on?

Also: would it be helpful to set the default Blend value to 50 (?) or some other non-zero value? My expectation is that when I enable a tool, I see an effect. And that is not the case here.
And maybe 'Blend' is not the right word here. It's sort of an additive or subtractive mask. Maybe call it that way? "Add / subtract mask"?

__About Smooth radius mask__
Considering how the slider works, I think limits should be changed. Even with the most severe masking (i.e. Blend = 1), I see absolutely negligible difference for Smooth Radius Mask values >50.0. Also, the range is now -10 through +1000 which seems odd. Can I also request to be able to set this value with 2 decimal places? I think that could help with the sharp transitions in using the slider.
The 'Soft radius' slider suffers from exactly the same 'problem'. 

__A few more things__
1. What is the point of the "Enable mask" checkbox? Doesn't it do exactly the same as enabling the tool?
2. What is the exact difference between 'Soft radius' and 'Smooth Radius Mask'? Do we need them both?
2. For the Mask Curves, I think the grey transition 'line' should be an actual line, not a grey area. The area implies the whole region is important, whereas it is only the edge.
2. For the LC(H) Curve the default control points are not very useful: you have one at 0 and 255, but because of the hue circle, they are next to each other and give a very sharp gradient when you move one. In the L*a*b* adjustment tool there are two control points by default, one in the red and one in the blue. Maybe do that here as well?
3. Similarly for the C(C) and L(L) Curves: do we need three control points by default? Two could be enough as a starting point, similar to in Lab adjustments.

Desmis:
@Thanatomanic 
Of course we can put a default value non zero to blend....perhaps -10
If you look at the tooltip, I write "add or substact the mask   from the original image", its the same "principle" as in others mask...but different. I have no objection to change label...but is it necessary to do it also for the other mask. We can also use "opacity.."...
Semantics is an important problem, when I look at the time we spent Xavier and me on it for wavelet (and the time that Xavier and Wayne spend), it's impressive

should i wait for other user reviews

I can also change the range of all "soft radius"....they have the same settings for all usage...knowing that a year ago they were in the range à..100....

"Enable mask", allows to enable mask when you are using several masks boths..the system can only visualize one....suppress will product hazardous results

Soft radius mask : soft the mask...that is to say artifacts generates by the 3 curves C(C) L(L) ,LC(H)
Soft radius ; smooth the differences between "mask + blend" and "original image" (take also into account Scope)....I have add yesterday...

The grey transition is the same as others masks...since more than one year...it is a GUI problem

for LC(H) it's the same problem as all curves (H) in rawtherapee, I don't like that....its a GUI problem...For me curves (H) are very unusable...

I think L(L) and C(C) are necessary...

jacques

Desmis:
@Thanatomanic 

I will soon does a commit with changes above....easy :)

To test !

jacques

Desmis:
@Thanatomanic 
Done :)

I hope I forgot nothing... because "guidefilter soft" is used in others places as wavelet, retinex...but it  was in range 0..100 (clarity..)

jacques

rtdata/languages/default
```diff
@@ -1192,6 +1192,11 @@ HISTORY_MSG_950;Local - Mask Common GF strength
 HISTORY_MSG_951;Local - Mask Common GF angle
 HISTORY_MSG_952;Local - Mask Common soft radius
 HISTORY_MSG_953;Local - Mask Common blend chroma
+HISTORY_MSG_BLSHAPE;Blur by level
+HISTORY_MSG_BLURCWAV;Blur chroma
+HISTORY_MSG_BLURWAV;Blur luminance
+HISTORY_MSG_BLUWAV;Attenuation Response
+>>>>>>> dev
```
heckflosse: @Desmis Jacques, please check this relicts from merge conflict ;-)

Desmis:
Aaahhhh   :)
it's my old age...

jacques

heckflosse:
@Desmis Jacques, I fixed some more relicts from merge conflict...

Desmis:
Aiaiaihhhh
Thank you :)

Le sam. 27 juin 2020 15:23, Ingo Weyrich <notifications@github.com> a
écrit :

Desmis:
I have add "Common color mask" to Rawpedia "First steps"...."premiers pas", with a simple example
http://rawpedia.rawtherapee.com/Local_Adjustments/fr
http://rawpedia.rawtherapee.com/Local_Adjustments/fr#Comment_se_servir_de_Common_Color_Mask

jacques

Issue #5749, opened 2020-05-10
pink color cast when using input profile when rawtherapee is compiled with gcc-10 and -O3


casta:
Hello,

Here are 3 screenshots: https://cloud.xwing.info/s/zBMSRnrL2dXnAkx
I think it will speak by itself:
* _rawtherapee-O2-input-profile.png_: using an input profile with rawtherapee build with gcc-10 and -O2 -> it’s fine
* _rawtherapee-O3-input-profile.png_: **using an input profile with rawtherapee build with gcc-10 and -O3 -> bogus pink color cast**
* _rawtherapee-O3-no-input-profile.png_: using no input profile with rawtherapee build with gcc-10 and -O3 -> colors are a bit off since there is no input profile, but it’s OK

I can reproduce this as soon as input profile in enabled, with any choice (either the embedded one,  the one provide by RT, or any custom DCP), and with any raw file I could test (I tested Canon 400D, 5D, Fuji X20, X100s and X-T30).

I have not yet tested an old version that may work to bisect this, I just started using gcc-10 last week, so this is a new discover for me. This may hide some bug as gcc-10 seems a bit sensitive than gcc-9 when dealing with optimisations.

Version: 5.8-402-gb5af6eec6
Branch: dev
Commit: b5af6eec6
Commit date: 2020-05-09
Compiler: gcc 10.1.0
Processor: Intel(R)\ Core(TM)\ i5-4690\ CPU\ @\ 3.50GHz
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.2
Lensfun: V0.3.2.0
Build type: Gentoo
Build flags: -march=native -O3 -mtune=native -pipe -std=c++11  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -flto -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas  -ftree-vectorize
Link flags: -Wl,-O1 -Wl,--hash-style=gnu -Wl,--as-needed -Wl,--sort-common  -flto
OpenMP support: yes
MMAP support: ON
Build OS: Linux 5.6.11-gentoo x86_64
Build date: Sat, 09 May 2020 18:43:34 +0000 UTC
Build epoch: 1589049814
Build UUID: 9ae996ea-2840-4d76-8f25-1be0cb097757

gaaned92:
[...]

PR #5814, opened 2020-06-26
GUI improvements for perspective correction

Lawrence37:
I am experimenting with an interface which allows you to more easily correct perspective distortion in the case when automatic correction fails. After reading #4505, I started working on a tool where you can draw vertical and/or horizontal control lines on top of the image and have the perspective automatically corrected.

Here's what it looks like.
image

Current features:
- Draw, move, and delete individual lines
- Change line type between vertical and horizontal
- Delete all lines with one button
- Correct vertical only, horizontal only, or both
- Line type is automatically set based on its angle when first drawn (a7f3975)

Limitations:
- ~Lines are not saved in the PP3~ (changed with eb548f1)
- ~Lines are part of the perspective tool itself. All images share the same lines in single editor mode, and lines for an image are lost when you close an editor in multiple editor mode.~ (lines are part of processing parameters)
- Only works for camera-based perspective correction

I'm interested in collecting people's thoughts and suggestions while I go work on other things.


Thanatomanic:
Hi @Lawrence37, awesome work so far! I'm happy to try it out soon. It would help if you could merge the current `dev` branch with your own, so that (at the least) I don't need to manually fix #5749 to test things out (or have ugly red images 🙃 )

One question already: were you planning to do something about the current limitations?

Lawrence37:
Ok, I've merged from dev. What I will do with the limitations depends on the use cases. I envisioned the control lines being a one-time use tool. In this case, the limitations are not important. If there are good reasons for saving the lines, I will consider fixing the limitations.

Thanatomanic:
This works really well, thank you! A few quick remarks:
1. I think it would be really great if there was only one drawing mode and it would automatically detect horizontal or vertical lines. For example, anything drawn more upright than a 45° diagonal -> vertical, otherwise -> horizontal. These could still be overridden in case you want some seriously extreme correction.
2. The persistent parameters are something to change imo before we merge this with dev. It doesn't make much sense that you set your corrections for a particular image, then go to another image and the same corrections apply. Also, the sliders do not get reset when applying Neutral profile.
3. The third toggle button is a bit weird. It enables the perspective correction mode, but it shows a darkened tick mark. To me this implies something is done, which is not the case. I'm not sure yet what a good solution would be...

Thanatomanic:
Oh and one other thing, you now need to Ctrl+click to draw the lines. I think it would be better if this perspective correction becomes a tool similar to the crop and 'straight line' rotation tool. You activate it and then normal clicking is used for the tool, and not dragging the image.

Floessie (commented):
Concerning 6c59f0586fa930149fb0c2e069551f5451639070: I would feel much more comfortable, if we could do away with `new/delete` (or worse `malloc()/free()`) in new code and use modern C++ RAII features like `std::unique_ptr<>` (also with custom deleter) or `std::shared_ptr<>`. There are basically zero reasons to use raw allocation anymore.

Best,
Flössie

heckflosse:
@Lawrence37 Would you mind to create a branch in the official repository? That would simplify testing.

Lawrence37:
@Thanatomanic Thanks for testing.
1. What I can do is assign the line type after the line is drawn. Afterwards, the type can only be changed manually. Alternatively, the type can update automatically when the line is edited, overriding manual changes. I prefer the first option.
2. I agree that it doesn't make sense. I will fix this by either removing all lines when an image is loaded or saving the lines in the processing parameters. Also, how are you able to trigger the problem of the sliders not being reset?
3. I need an icon which means "Apply".
4. I can swap the behavior if you'd like. Dragging on the image will draw a line. Ctrl+dragging pans the image.

@Floessie I will change the code to use smart pointers.

@heckflosse Branch "perspective-lines" is now also in this repository.

Thanatomanic:
> 1. What I can do is assign the line type after the line is drawn. Afterwards, the type can only be changed manually. Alternatively, the type can update automatically when the line is edited, overriding manual changes. I prefer the first option.

I think either would be great. It may need some real-life test cases to determine whether manual switching is more commonly needed or that automatic switching would be more convenient.

> 2. I agree that it doesn't make sense. I will fix this by either removing all lines when an image is loaded or saving the lines in the processing parameters. Also, how are you able to trigger the problem of the sliders not being reset?

The lines are input for what the sliders values become, and the sliders can be overridden manually. So you would need to both the lines _and_ the slider values in the pp3 (ideally) in order to continue exactly where you left off.
~~If you pick Neutral here, all values should return to their default value. In case of the perspective correction, I assume everything needs to go back to 0, which currently is not the case (the values persist).~~
image


> 3. I can swap the behavior if you'd like. Dragging on the image will draw a line. Ctrl+dragging pans the image.

Let's wait for input from others too

Lawrence37:
That's strange. All values are properly reset on my build. The method returns to "Simple" and all values for "Simple" and "Camera-based" are reset to 0 except for focal length (default is 24) and crop factor (default is 1). I tried both the partial and full processing profile fill modes.

Thanatomanic:
@Lawrence37 Oh, you're absolutely right. I must have misclicked in my haste. Applying Neutral works ✔️

Beep6581:
> 3\. I need an icon which means "Apply".

I can help with that. Use any icon for now, I'll commit one after the weekend.

Lawrence37:
I'm going to make the line type automatically assigned when drawn. This involves a change to the buttons. There will be 3 buttons.
- "Edit" toggle button: Enable to enter line edit mode. Disable to exit without applying correction.
- "Apply" button: Available in edit mode only. When clicked, perspective is corrected using the current line positions. Also exits edit mode.
- "Delete" button: Available in edit mode only. Removes all lines. (Same as current behavior.)

Beep6581 (changes requested):
Hey

The icons need some tweaks. Icons should have 1px padding within the canvas (i.e. the image should not touch the edges), and should use the color #2A7FFF with a general opacity of 70% for outline/fill, with higher or lower opacity to accent certain elements as needed. See e.g. `edit-point.svg`

Beep6581 (commented):
The unlabeled buttons could use simple tooltips, e.g. "Edit", "Apply", "Delete", "Vertical", "Horizontal", "Both".
For such reusable strings, you could use keys prefixed [`GENERAL_`](https://github.com/Beep6581/RawTherapee/blob/ef405eeb1dbe7f5d49a41e1b42346d4300fd20a7/rtdata/languages/default#L214).

Beep6581:
First-time user of this cool new branch. When I press the "Draw" button in the Perspective tool, the "Hand tool" icon in the upper toolbar changes to `crosshair-adjust.svg`. It took me a while to notice this. However, even though the "Hand tool" button disappeared and was replaced with a different button, I can still pan the preview, and when I do, the on-preview cursor is still the hand, so it feels wrong that the "Hand tool" button is replaced with something else. Seems to me it would be more intuitive and consistent with how the other buttons behave if the "Hand tool" button was left alone, and a new "Perspective correction" button was added. That way not only does the user notice a new feature, but when I open a new image I can click that "Perspective correction" button and be taken right into the Perspective Correction tool. Seems to me the only advantage of this button-switching is that we save a tiny amount of space in the toolbar, but I don't think the price of that is worth it. Though I may be wrong - first time user.

Beep6581:
After more use, two changes to my previous comments.

Regarding my previous icon color comment - on-canvas icons need a high-contrast look and it's fine that you used orange. Though the sizes still need to be tweaked to be consistent with the other icons.

Regarding the "Hand tool" button being replaced, I see now that this is also how it works for other on-canvas tools, such as the Graduated Filter. Never noticed that before. So it's fine for this feature to use that mechanism (even though I still think my points against the workings of this mechanism are valid - even the tooltip still says "Hand tool" though the icon is no longer "Hand tool").

Beep6581:
When I'm in the mode where you draw lines (hold the Ctrl key and move the cursor), the cursor is changed to a tiny gray crosshair. This is very difficult to see in some cases.

screenshot_20200720_185950

Seems to be done here:
https://github.com/Beep6581/RawTherapee/blob/ef405eeb1dbe7f5d49a41e1b42346d4300fd20a7/rtgui/controllines.cc#L288

Better to use one of the `-hicontrast` cursors.

Lawrence37:
@Beep6581 A separate perspective correction button instead of the hand tool would be nice. I will look into it. In the mean time, I will fix the icons, cursor, and tooltips.

Lawrence37:
I've changed the icon padding, added tooltips, and used different cursor shapes. In panning mode, the cursor is an open hand. In drawing mode, the cursor is a crosshair. Side note: I find it strange that the hand tool uses the crosshair icon instead of the open hand when hovering over the image.

Lawrence37:
Unless anyone wants additional changes, this is ready for merging.

Lawrence37:
I will merge this weekend if there are no objections.

heckflosse:
@Lawrence37 
> I will merge this weekend if there are no objections.

I already gave you a :+1: 

Just merge! The sooner, the better.. :+1:

Thanatomanic:
Agreed to merge!

Lawrence37:
Ok, I will merge now.

PR #5816, opened 2020-06-28
Locallab engine review

heckflosse:
@Desmis Jacques, could you again test this branch please? I would like to merge it before I continue locallab review.


Desmis:
@heckflosse 
Ingo

I tested, tested...and found a difference in "graduated filter local contrast"


But after verification, it is your last commit which is good :)

All others functions, no differences


Thank you  very much

jacques

PR #5834, opened 2020-07-05
Improve pyramids

heckflosse:
I would like to merge this into dev before I continue work on the remaining pyramids.

@Desmis Jacques, please test if all works correctly


Desmis:
@heckflosse 
Ingo

I tested all functions "wav pyramid"...with your last commits

All seems to works fine...

jacques

heckflosse:
@Desmis Jacques, thanks for testing :+1:

PR #5836, opened 2020-07-06
improvements for wavelets and also for other use cases of pyramids

heckflosse:
This is a follow-up of my latest work on improving usage of pyramids.
As pyramids are used in wavelets quite often, I start with a commit which improves the first usage of a pyramid in wavelet code, which is `Wavelet Blur levels`

This reduces processing time and memory usage of this use case while (hopefully) giving more continuous result.
I used a 45.6MP file for my test and the attached .pp3 file

```
before : 3230 ms
after  : 1530 ms
```
[wavelet_blur.zip](https://github.com/Beep6581/RawTherapee/files/4879211/wavelet_blur.zip)

More commits will follow


heckflosse:
I want to add some more details about the last commit.

In this use case (attached pp3) 7 levels have been used for wavelets. My machine has 8 cores.

That led to a memory usage of 8 (number of cores at my machine) x 3 * width * height bytes on top of the already used memory by wavelets.

This memory allocations werde done by 7 (number of wavelet levels) * 3 * 3 = 63 memory allocations/deallocations of size width * height bytes.

New code only needs 8 (number of cores) * width * height bytes, and the number of allocations/deallocations is also only 8 (number of cores) instead of 63

Desmis:
@heckflosse 
 Ingo

I will have a look, tomorrow...or in late afternoon

Jacques

heckflosse:
@Desmis Jacques, no hurry ;-)

Desmis:
@heckflosse 

Ingo

I just tested

Speedup as you 50%

Small differences in TIF, but it is obvious
All seems to work well ;)

Good job

jacques

heckflosse:
I attaached a  .pp3 to test the last commit.
I measured in `WaveletcontAllAB`

```
before : 2315 ms
after:    665 ms
```

Same reductions of memory usage and allocations as in first commit

[wavelet_blur_ab.zip](https://github.com/Beep6581/RawTherapee/files/4884431/wavelet_blur_ab.zip)

Desmis:
@heckflosse 

Ingo
I tested your commit

Speedup about 75% (as you)
Small obvious differences in TIFF

Good  job  :)

jacques

heckflosse:
Let me give some more details about the optimizations of memory usage of the last two patches:

Let's assume number of wavelet levels is 7
Let's also assume we process an 24 MP image of size 6000x4000 pixels in wavelets.

In wavelet processing the individual levels are reduced by factor 4, so we can just use 3000x2000 pixels for our calculations or we use 6000x4000 pixels but do not multiply by 4 (sizeof(float)) ;-)
I will use the latter for my explanations...

Wavelets are processed in 3 directions and we have a loop which processes one combination of level and direction per thread in parallel

```
// maxlvl == 7 in this example
#ifdef _OPENMP
        #pragma omp for schedule(dynamic) collapse(2)
#endif

        for (int dir = 1; dir < 4; dir++) {
            for (int lvl = 0; lvl < maxlvl; lvl++) {
```

In old code each of this iterations allocated 3 buffers of size 6000x4000 bytes, means each parallel thread allocated 72 MB, means on an 8 core machine memory usage was 576 MB.
On a 24 core machine it would have used 21 (because 3 x 7 = 21) * 72 MB = 1512 MB for this 24 MP file.
On a 24 core machine using a 100MP file it would have used 6300 MB only for this step (on top of the memory already used for base wavelet usage and so on)

The new code reduces this `on top of memory usage` by 2/3

Desmis:
@heckflosse 

Ingo

I knew that by programming like I did, that the memory and speed was not good. 
But, I knew you were going to take it back. 
The improvements are impressive, for an average user 50% gain in memory and speed, or even 60 to 75% depending on the case. 
Well done  :)

Thank you very much

jacques

heckflosse:
@Desmis Jacques

> I knew that by programming like I did, that the memory and speed was not good.

I didn't want to offend you at all. My only intention was, to tell about my improvements, how they were done, to keep that in mind for future use

heckflosse:
@Desmis Jacques

> But, I knew you were going to take it back.

You know me quite well :smiley:

Desmis:
You don't offend me at all...
Your work is very good :)

Le mar. 7 juil. 2020 16:45, Ingo Weyrich <notifications@github.com> a
écrit :

Desmis:
@heckflosse 

Ingo

What offends me are the authoritarian positions ... very critical ... without management

jacques

heckflosse:
@Desmis Jacques

> What offends me are the authoritarian positions ... very critical ... without management 
I guess (hope) you don't mean me...

Desmis:
Of course no :)
you are a true friend whom i trust

Desmis:
@heckflosse 
Ingo

I tested, no differences, works fine :)

jacques

heckflosse:
@Desmis Jacques, I have a question about this part of ipwavelet.cc, especially about maxkoeLi

https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/ipwavelet.cc#L2810

In the first Iteration of this loop maxkeoLi is 0, in the second iteration it is max(0, value of second iteration) and so on.

Wouldn't it be more correct to calculate maxkoeLi a priory?

If so, I can do that with my next commit

Desmis:
@heckflosse 

Ingo

I think you can test, to verify :)

jacques

PR #5840, opened 2020-07-09
Make libcanberra optional

ya-isakov:
It is the only X11 dep in RawTherapee, so here is an option to
compile without libcanberra. This will completely disables sounds
for long operations on Linux, but canberra is enabled by default.


heckflosse:
@Floessie Ok to merge?

Thanatomanic:
Seems like we can merge this without much trouble?

Floessie:
Looks good to me.

PR #5850, opened 2020-07-15
make some files cppcheck clean

heckflosse:
@Desmis Jacques, can you please test my changes to color.cc?

### DIFF
--- a/rtengine/FTblockDN.cc
+++ b/rtengine/FTblockDN.cc
@@ -2108,17 +2108,17 @@ float ImProcFunctions::Mad(const float * DataList, const int datalen)
     }
 
     //find median of histogram
-    int median = 0, count = 0;
+    int lmedian = 0, count = 0;
 
     while (count < datalen / 2) {
-        count += histo[median];
-        ++median;
+        count += histo[lmedian];
+        ++lmedian;
     }
 
-    int count_ = count - histo[median - 1];
+    int count_ = count - histo[lmedian - 1];
 
     // interpolate
-    return (((median - 1) + (datalen / 2 - count_) / (static_cast<float>(count - count_))) / 0.6745f);
+    return ((lmedian - 1) + (datalen / 2 - count_) / (static_cast<float>(count - count_))) / 0.6745f;
 }
 
 float ImProcFunctions::MadRgb(const float * DataList, const int datalen)
@@ -2141,18 +2141,18 @@ float ImProcFunctions::MadRgb(const float * DataList, const int datalen)
     }
 
     //find median of histogram
-    int median = 0, count = 0;
+    int lmedian = 0, count = 0;
 
     while (count < datalen / 2) {
-        count += histo[median];
-        ++median;
+        count += histo[lmedian];
+        ++lmedian;
     }
 
-    int count_ = count - histo[median - 1];
+    int count_ = count - histo[lmedian - 1];
 
     // interpolate
     delete[] histo;
-    return (((median - 1) + (datalen / 2 - count_) / (static_cast<float>(count - count_))) / 0.6745f);
+    return ((lmedian - 1) + (datalen / 2 - count_) / (static_cast<float>(count - count_))) / 0.6745f;
 }
 
 
--- a/rtengine/ashift_lsd.c
+++ b/rtengine/ashift_lsd.c
@@ -1770,10 +1770,10 @@ static void region_grow( int x, int y, image_double angles, struct point * reg,
   int xx,yy,i;
 
   /* check parameters */
-  if( x < 0 || y < 0 || x >= (int) angles->xsize || y >= (int) angles->ysize )
-    error("region_grow: (x,y) out of the image.");
   if( angles == NULL || angles->data == NULL )
     error("region_grow: invalid image 'angles'.");
+  if( x < 0 || y < 0 || x >= (int) angles->xsize || y >= (int) angles->ysize )
+    error("region_grow: (x,y) out of the image.");
   if( reg == NULL ) error("region_grow: invalid 'reg'.");
   if( reg_size == NULL ) error("region_grow: invalid pointer 'reg_size'.");
   if( reg_angle == NULL ) error("region_grow: invalid pointer 'reg_angle'.");
--- a/rtengine/color.cc
+++ b/rtengine/color.cc
@@ -1189,48 +1189,26 @@ void Color::computeBWMixerConstants (const Glib::ustring &setting, const Glib::u
     float koymcp = 0.f;
 
     if(setting == "ROYGCBPM-Abs" || setting == "ROYGCBPM-Rel") {
-        float obM = 0.f;
-        float ogM = 0.f;
-        float orM = 0.f;
-
-        float ybM = 0.f;
-        float yrM = 0.f;
-        float ygM = 0.f;
-
-        float mgM = 0.f;
-        float mrM = 0.f;
-        float mbM = 0.f;
-
-        float pgM = 0.f;
-        float prM = 0.f;
-        float pbM = 0.f;
-
-        float crM = 0.f;
-        float cgM = 0.f;
-        float cbM = 0.f;
-        //printf("mixred=%f\n",mixerRed);
-
         float fcompl = 1.f;
 
-        if(complement && algo == "SP") {
+        if (complement && algo == "SP") {
             fcompl = 3.f;    //special
-        } else if(complement && algo == "LI") {
+        } else if (complement && algo == "LI") {
             fcompl = 1.5f;    //linear
         }
 
         // ponderate filters: report to R=G=B=33
         // I ponder RGB channel, not only orange or yellow or cyan, etc...it's my choice !
-        if(mixerOrange != 33) {
-            if (algo == "SP") { //special
-                if (mixerOrange >= 33) {
-                    orM = fcompl * (mixerOrange * 0.67f - 22.11f) / 100.f;
-                } else {
-                    orM = fcompl * (-0.3f * mixerOrange + 9.9f) / 100.f;
-                }
+        if (mixerOrange != 33.f) {
+            float ogM = 0.f;
+            float orM = 0.f;
 
-                if (mixerOrange >= 33) {
+            if (algo == "SP") { //special
+                if (mixerOrange > 33.f) {
+                    orM = fcompl * (mixerOrange * 0.67f - 22.11f) / 100.f;
                     ogM = fcompl * (-0.164f * mixerOrange + 5.412f) / 100.f;
                 } else {
+                    orM = fcompl * (-0.3f * mixerOrange + 9.9f) / 100.f;
                     ogM = fcompl * (0.4f * mixerOrange - 13.2f) / 100.f;
                 }
             } else if (algo == "LI") { //linear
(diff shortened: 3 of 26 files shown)


rtengine/FTblockDN.cc
```diff
@@ -2141,18 +2141,18 @@ float ImProcFunctions::MadRgb(const float * DataList, const int datalen)
...
+        ++lmedian;
     }
 
-    int count_ = count - histo[median - 1];
+    int count_ = count - histo[lmedian - 1];
 
     // interpolate
     delete[] histo;
-    return (((median - 1) + (datalen / 2 - count_) / (static_cast<float>(count - count_))) / 0.6745f);
+    return ((lmedian - 1) + (datalen / 2 - count_) / (static_cast<float>(count - count_))) / 0.6745f;
```
Floessie, line 2155: Shouldn't this be factored out (anon namespace)? It's mostly a duplicate of the above...

rtengine/ashift_lsd.c
```diff
@@ -1770,10 +1770,10 @@ static void region_grow( int x, int y, image_double angles, struct point * reg,
   int xx,yy,i;
 
   /* check parameters */
-  if( x < 0 || y < 0 || x >= (int) angles->xsize || y >= (int) angles->ysize )
-    error("region_grow: (x,y) out of the image.");
   if( angles == NULL || angles->data == NULL )
     error("region_grow: invalid image 'angles'.");
+  if( x < 0 || y < 0 || x >= (int) angles->xsize || y >= (int) angles->ysize )
```
Floessie, line 1775: Good catch! :+1:

rtengine/color.cc
```diff
@@ -1189,145 +1189,109 @@ void Color::computeBWMixerConstants (const Glib::ustring &setting, const Glib::u
...
+            const float pgM = complement ? (-0.492f * mixerPurple + 16.236f) / 100.f : 0.f;
 
             mixerRed   += prM;
             mixerGreen += pgM;
             mixerBlue  += pbM;
             koymcp += (prM + pgM + pbM);
         }
 
-        if(mixerCyan != 33) {
+        if(mixerCyan != 33.f) {
```
Floessie: Space.

rtengine/color.cc
```diff
@@ -1496,100 +1460,78 @@ void Color::interpolateRGBColor (float realL, float iplow, float iphigh, int alg
...
-                b_1 = b_1 + (b_2 - b_1) * calan * balance;
-            } else if(metchrom == 2) {
-                a_1 = a_1 + (a_2 - a_1) * calan * balance;
-                b_1 = b_1 + (b_2 - b_1) * calby * balance;
+            } else if (metchrom == 1) {
+                a_1 = a_1 + (a_2 - a_1) * chromat * balance;
+                b_1 = b_1 + (b_2 - b_1) * chromat * balance;
+            } else if (metchrom == 2) {
+                a_1 = a_1 + (a_2 - a_1) * chromat * balance;
+                b_1 = b_1 + (b_2 - b_1) * luma * balance;
```
Floessie, line 1522: As always, having proper `enum`s would ease reading the code and would also enable `switch` without `default:` to warn about missing `case`s. :expressionless:

rtengine/color.cc
```diff
@@ -1496,100 +1460,78 @@ void Color::interpolateRGBColor (float realL, float iplow, float iphigh, int alg
                                  const float xl, const float yl, const float zl, const float x2, const float y2, const float z2,
                                  const double xyz_rgb[3][3], const double rgb_xyz[3][3], float &ro, float &go, float &bo)
 {
-    float X1, Y1, Z1, X2, Y2, Z2, X, Y, Z, XL, YL, ZL;
-    float L1 = 0.f, L2, LL, a_1 = 0.f, b_1 = 0.f, a_2 = 0.f, b_2 = 0.f, a_L, b_L;
+    float L1 = 0.f, a_1 = 0.f, b_1 = 0.f, a_2 = 0.f, b_2 = 0.f, a_L, b_L;
```
Floessie: Is it worth it to omit the `a_L` and `b_L` initialization?

rtengine/color.cc
```diff
@@ -1496,100 +1460,78 @@ void Color::interpolateRGBColor (float realL, float iplow, float iphigh, int alg
...
     Color::Lab2XYZ(L1, a_1, b_1, X, Y, Z);
-
     Color::xyz2rgb(X, Y, Z, ro, go, bo, rgb_xyz);// ro go bo in gamut
 }
 
 void Color::calcGamma (double pwr, double ts, int mode, GammaValues &gamma)
 {
     //from Dcraw (D.Coffin)
-    int i;
     double g[6], bnd[2] = {0., 0.};
```
Floessie: ` = {}`?

rtengine/color.cc
```diff
@@ -1621,7 +1563,7 @@ void Color::calcGamma (double pwr, double ts, int mode, GammaValues &gamma)
         g[5] = 1. / (g[1] * SQR(g[3]) / 2. + 1. - g[2] - g[3] - g[2] * g[3] * (log(g[3]) - 1.)) - 1.;
     }
 
-    if (!mode--) {
+    if (!mode) {
```
Floessie: Why? If `mode != 0` there is no result?!

rtengine/color.cc
```diff
@@ -2078,15 +2020,11 @@ void Color::gamutmap(float &X, float &Y, float &Z, const double p[3][3])
...
-    float scale = 100.0f / 100.1f; //reduction in normal zone
-    float scaleext = 1.0f; //reduction in transition zone
-    float deltaHH = 0.3f; //HH value transition : I have choice 0.3 radians
+
     float HH;
     bool doskin = false;
-
     //rough correspondence between h (JC) and H (lab) that has relatively little importance because transitions that blur the correspondence is not linear
-    if     ((float)h > 8.6f  && (float)h <= 74.f ) {
+    if ((float)h > 8.6f  && (float)h <= 74.f ) {
```
Floessie: I'm pretty sure the cast is implicit.

rtengine/color.cc
```diff
@@ -2117,22 +2058,23 @@ void Color::skinred ( double J, double h, double sres, double Sp, float dred, fl
...
-        if(chromapro > 1.0) {
-            interm = (chromapro - 1.0f) * 100.0f;
+        float factorskinext;
+        if(chromapro > 1.f) {
+            const float interm = (chromapro - 1.0f) * 100.0f;
             factorskin = 1.0f + (interm * scale) / 100.0f;
             factorskinext = 1.0f + (interm * scaleext) / 100.0f;
         } else {
             factorskin = chromapro ;
             factorskinext = chromapro ;
```
Floessie: This needs to be run through a linter...

rtengine/color.cc
```diff
@@ -2296,26 +2227,19 @@ void Color::AllMunsellLch(bool lumaMuns, float Lprov1, float Loldd, float HH, fl
...
-
-                        if(contin1 && contin2) {
-                            correctlum = correctlumprov2 - correctlumprov;
+                            if(contin1) {
+                                correctlum = correctionHueLum - correctlumprov;
+                            }
                         }
                     }
                 }
+                break;
```
Floessie, line 2169: Optimization?

rtengine/color.h
```diff
@@ -1805,7 +1805,7 @@ class Color
     * @param Z Z channel input value and corrected output value [0 ; 65535]
     * @param p working profile
     */
-    static void gamutmap(float &X, float &Y, float &Z, const double p[3][3]);
+    static void gamutmap(float &X, float Y, float &Z, const double p[3][3]);
```
Floessie, line 1804: You should adjust the doxygen description.

rtengine/imagefloat.h
```diff
@@ -109,35 +109,35 @@ class Imagefloat final : public IImagefloat, public ImageIO
         } tmp;
 
         tmp.f = f;
-        int32_t sign     =  (tmp.i >> 16) & 0x00008000;
+        const int32_t lsign = (tmp.i >> 16) & 0x00008000;
```
Floessie, line 112: `std::int32_t`

rtengine/imagefloat.h
```diff
@@ -214,7 +214,7 @@ class Imagefloat final : public IImagefloat, public ImageIO
         exponent += (128 - 64);
         mantissa <<= 7;
         // Assemble sign, exponent and mantissa.
-        return (uint32_t) ((sign << 31) | (exponent << 23) | mantissa);
+        return (uint32_t) ((lsign << 31) | (exponent << 23) | mantissa);
```
Floessie, line 217: Does this really have to be inline in the (or *this*) header?

Floessie (commented):
Just a few comments.

Desmis:
@heckflosse 

I tested with various change in BW, Color toning....
No differences in TIF

@Floessie 
i am still impressed :)

Jacques

rtengine/color.cc
heckflosse: Fixed

rtengine/color.cc
heckflosse: Fixed

rtengine/color.cc
heckflosse: Fixed

rtengine/color.cc
heckflosse: I removed this (unused) function.

rtengine/color.cc
heckflosse: Function removed

rtengine/color.cc
heckflosse: Yes

rtengine/color.h
heckflosse: Fixed

rtengine/imagefloat.h
heckflosse: We should refactor this stuff, as it's duplicated code. Have a look:
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/dcraw.cc#L10709

rtengine/imagefloat.h
heckflosse: As above

rtengine/color.cc
heckflosse: Fixed

heckflosse:
@Desmis Jacques, 
here
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/dirpyr_equalizer.cc#L272
cppcheck correctly detected that the condition `level == 5` can never be true because above there is the condition `level != 5`.

Is this a typo? How should I fix this?

Desmis:
@heckflosse 

The good choice is in   "if(blurcb > 0.f && choice == 0  && level != 5) {" to suppress level != 5

But I disabled this function by setting "blurcb" to zero (and also for GUI , I suppress slider "blur"), because it brought artifacts as soon as the radius was not very small and only for low levels.
The same function in wavelets works  (local and main) fine with much more functionality, for all levels and for high radius.

If we compare "wavelet" with "cbdl", in all cases (whether local or main) wavelet is clearly superior to cbdl ... 
The only reasons to maintain "cbdl" are the habit and an apparent simplicity (wavelet seems complex ...)

jacques

PR #5866, opened 2020-07-29
array2D.h: cleanup

heckflosse:


Floessie:
@heckflosse Doesn't build here, as `ssize_t` is POSIX and needs an `#include <sys/types.h>`. With the include it builds and runs fine on Debian Testing PPC.

Thanatomanic:
Notwithstanding the useful cleanup, but why are we using this class anyway instead of STL arrays?

heckflosse:
> why are we using this class anyway instead of STL arrays?

For STL array (std::array), we have to know the size at compile time.

heckflosse:
@Floessie 
>  Doesn't build here, as ssize_t is POSIX

I will change to size_t and push soon.

Floessie:
@heckflosse 
> For STL array (std::array), we have to know the size at compile time.

I think Roel means STL containers. It would indeed be nice, if `array2D` used `std::vector<>` internally instead of relying upon raw pointers.

heckflosse:
> I think Roel means STL containers. It would indeed be nice, if array2D used std::vector<> internally instead of relying upon raw pointers.

Ok, misunderstanding then. I will change the code to std::vector usage with one of the next commits.

Edit: Though I still think, we should keep this class, but change it internally to use std::vector. Throwing away the class (as suggested by Roel) and using native std::vector constructions everywhere we use array2D atm, is much error prone imho.

Floessie:
With the `std::size_t` change, I get myriads of warnings like this on PPC32 (it's the bitness, I suppose):

```
/home/user/src/rawtherapee/rtengine/cfa_linedn_RT.cc: In member function ‘void rtengine::RawImageSource::cfa_linedn(float, bool, bool, const rtengine::RawImageSource::CFALineDenoiseRowBlender&)’:
/home/user/src/rawtherapee/rtengine/cfa_linedn_RT.cc:86:54: warning: ISO C++ says that these are ambiguous, even though the worst conversion for the first is better than the worst conversion for the second:
   86 |                 RawDataTmp[i * width + j] = rawData[i][j];
      |                                                      ^
In file included from /home/user/src/rawtherapee/rtengine/rawimagesource.h:25,
                 from /home/user/src/rawtherapee/rtengine/cfa_linedn_RT.cc:27:
/home/user/src/rawtherapee/rtengine/array2D.h:175:9: note: candidate 1: ‘T* array2D<T>::operator[](size_t) [with T = float; size_t = unsigned int]’
  175 |     T * operator[](size_t index)
      |         ^~~~~~~~
/home/user/src/rawtherapee/rtengine/cfa_linedn_RT.cc:86:54: note: candidate 2: ‘operator[](const float* const*, int)’ (built-in)
   86 |                 RawDataTmp[i * width + j] = rawData[i][j];
      |                                                      ^
```

heckflosse:
@Floessie 
> With the std::size_t change, I get myriads of warnings like this on PPC32 (it's the bitness, I suppose)

Shall I revert to ssize_t or do you have an idea how to fix this?

Floessie:
@heckflosse I can think of two ways to solve it:

- Make indexing unsigned: Clean, but a lot of work.
- Use `using Index = std::make_signed<std::size_t>::type`: Drop-in for `ssize_t`.

HTH,
Flössie

Issue #5867, opened 2020-07-29
Full-screen inspector: several issues


Thanatomanic:
@rfranke Thank you for the work on the full-screen inspector in RawTherapee/RawTherapee#5593. I like the idea very much, and the execution gets a thumbs up as well!

I have (only) just tested it, and I do have a few points of criticism, that I hope you can address:
1. ~~Serious performance issue: you seem to update the preview as soon as you hover over an image in the file browser (preloading?). This takes up a lot of CPU time and even makes file selection noticeably slower. Since not all users use the preview, they should not experience any performance loss because of it. I would much rather have a preview screen that shows immediately when I press 'f' and then wait a few milliseconds before the image actually shows.~~ Fixed in RawTherapee/RawTherapee#5872
~~Side-note: didn't the 'old' inspector show the embedded JPEG? Maybe I missed it somewhere along the line, but why is the new inspector doing (partial) raw processing instead?~~ My mistake, the JPEG is shown, as can be easily seen when looking at a UniWB file for example.
2. ~~I suppose the sudden lack of the 'Inspector' tab will raise a fair amount of questions to people who casually update RT and have been long time users. What is the best way to ease the transition in your opinion?~~ You can now choose between the old mode and the new mode in Preferences.
3. There seems to be no indication in the GUI of the existence of the full-screen inspector (I find this a downside in darktable as well). Unless I am overlooking something... I think it is bad to have hidden functionality. Could we have a button (per thumbnail I guess?) to open the inspector? This is not without problems of its own...
4. ~~There is a bug/feature: I expect to see a preview of the _selected_ image, but now it shows the preview of the _image over which the cursor is positioned_ when I press 'f'.~~ Intended behavior, conclusion pending. See (8).
5. The behavior of hold-f-to-show and release-f-to-hide unexpectedly changes when you scroll or switch zoom-levels while in full-screen mode: if you do either of those, you enter a permanent full-screen preview and releasing your f or shift-f keys won't make the preview disappear. I don't find this intuitive. I would prefer to either make the preview always disappear as soon as you release the f key, or make the preview state a toggle (this is already mentioned in the PR). I can imagine the latter is also a much more accessibility friendly option.
> [from the PR] Panning is possible with a second monitor for the inspector window by moving the mouse over the thumbnail on the first monitor.
[...]

PR #5872, opened 2020-08-04
Extensions of inspector window (#5867)

rfranke:
This reduces CPU load if the inspector window is not used.


rtgui/inspector.cc
```diff
@@ -166,6 +171,25 @@ bool Inspector::on_button_press_event(GdkEventButton *event)
     return false;
 }
 
+bool Inspector::on_motion_notify_event(GdkEventMotion *event)
+{
```
Lawrence37, line 218: Make sure `currImage` is not null here (and also in `Inspector::on_zoom_begin`, I think).

rfranke: Thank you for the code review. Added two checks to 947f3dc.

Thanatomanic:
@rfranke What is the status of this? In particular the commit that fixes the CPU usage would be very good to merge.

rfranke:
Some of the other commits might be important as well, like panning with mouse drag and F11 to toggle the fullscreen status. Moreover, the extension of the filebrowser context menu makes the inspector window accessible through the GUI (cat mode ;) ).

Shall I revert the last commit as you don't like the either .. or decision?

heckflosse:
@rfranke btw: please excuse that I introduced a merge conflict with my latest commit :frowning_face:

ff2000:
> Shall I revert the last commit as you don't like the either .. or decision?

I would vote for "keep it". Having the context menu entry (that IMO is important to discover the feature) and a visible "Inspector" tab nearby might totally confuse people when a new window opens instead of the image being shown in the tab. IMO it really is an either-or thing.

heckflosse:
What's the status of this PR?
In my opinion we have to merge this for 5.9 because 5.9 filebrowser will become unusable when working with raw files from smartphones, which do not have an embedded jpg, if we do not merge.

Issue #5813, opened 2020-06-25
Relatively high CPU usage for reading out navigator and histogram


Thanatomanic:
See here https://github.com/Beep6581/RawTherapee/issues/5809#issuecomment-649494251

Comment by @heckflosse 
>> In fact, just moving my mouse over the image adds page faults, pushes my CPU to 10% and causes fluctuations in allocated memory
>
> I want to come back to this. It looks like the reason for CPU usage being ~10% when moving the mouse, are the updates to histogram and navigator. When I hide both, mouse movements don't push cpu load. Can you confirm?
> 
> Perhaps we should use a delay as we do for adjusters...

I confirm that when the histogram and navigator are hidden, there is virtually no CPU usage and no page faults happening. When they are shown and I hover over the image, my CPU is consistently at 10% and many thousands of page faults happen.

I don't know what a delay would do. I don't understand why this is so CPU intensive? Surely if the processed preview is cached, it's trivial to obtain a readout?

heckflosse:
Thanks for testing and confirming.
The fact that there is no CPU usage when histogram and navigator are hidden, means that the load when navigator and histogram are not hidden, is caused by updating the gui (the values for the gui are also readout if navigator and histogram are hidden, so this can not be the culprit).

A delay would mean that we update histogram and navigator not every time the mouse is moved, but only after a configurable delay (for example 50 ms after mouse movement stopped).

We could try to adapt the code from [here](https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/adjuster.cc#L378)...

heckflosse:
About the delay: Currently it's only used for adjusters. Curves for example don't have a delay unfortunately. Each change to a curve leads to an immediate reprocessing. By implementing the delay-functionality a bit more generic, we could use it for curves, mouse movements etc. as well.

Maybe @Floessie can help with that, as iirc he also made the nice `IdleRegister` in guiutils.*

heckflosse:
To clarify, what causes the cpu-load when moving the mouse are the ui updates to this two regions
![grafik](https://user-images.githubusercontent.com/13733487/85768887-c7a89100-b719-11ea-8fe5-16ea758981f0.png)

PR #5878, opened 2020-08-08
Delayed GUI updates

Floessie:
Hello fellow RawTherapists,

I've added a little facility in `rtgui/delayed.h` as a generic solution to #5813. I started with something for `adjuster.h`, that's why the branch is called `delayed_connection`.

The interesting thing was to come up with an idea on how to store and especially replay a function call. While it was relatively clear that a `std::tuple<>` was the right "container" to store the parameters, a bit of research was necessary for the replaying. I've added the references in the code. Should we ever switch to C++17, the solution is already there, namely [`std::apply()`](https://en.cppreference.com/w/cpp/utility/apply).

Hope you like it,
Flössie

### DIFF
--- a/rtgui/adjuster.cc
+++ b/rtgui/adjuster.cc
@@ -36,7 +36,7 @@ double one2one(double val)
 }
 }
 
-Adjuster::Adjuster (
+Adjuster::Adjuster(
     Glib::ustring vlabel,
     double vmin,
     double vmax,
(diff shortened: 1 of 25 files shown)


Thanatomanic (approved):
I see no real downsides. The CPU usage is measurably lower, which is nice.

Pandagrapher (approved):
Really nice work 👍 I haven't seen side effect while testing.

Just a general question:
Should we generalize it to `adjuster` class by default instead of only using it for heavy algorithms (maybe with a smaller delay compared to heavy algorithms)?

heckflosse:
@Pandagrapher 

> Should we generalize it to adjuster class by default instead of only using it for heavy algorithms (maybe with a smaller delay compared to heavy algorithms)?

There already are two delays (one for the heavier and one for the not so heavy algorithms)

Floessie:
@Thanatomanic @Pandagrapher @heckflosse Thanks for the review. :+1: I'll merge the branch to `dev`.

Issue #5748, opened 2020-05-09
Fast bayer demosaicer for flat regions


heckflosse:
When using dual-demosaic, currently the very slow VNG4 demosaic is used for flat regions of the image.

I just coded a very fast (vectorized and multithreaded by compiler), bilinear bayer demosaicer, which can be used instead of VNG4 for flat regions.
Here's the code, PR will follow later:

```
////////////////////////////////////////////////////////////////
//
//  Bilinear bayer demosaic, optimized for speed, intended use is for flat regions of dual-demosaic
//
//  copyright (c) 2020  Ingo Weyrich <heckflosse67@gmx.de>
//
//
//  code dated: May 09, 2020
//
//  bayer_bilinear_demosaic.cc is free software: you can redistribute it and/or modify
//  it under the terms of the GNU General Public License as published by
//  the Free Software Foundation, either version 3 of the License, or
//  (at your option) any later version.
//
//  This program is distributed in the hope that it will be useful,
//  but WITHOUT ANY WARRANTY; without even the implied warranty of
//  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
//  GNU General Public License for more details.
//
//  You should have received a copy of the GNU General Public License
//  along with this program.  If not, see <https://www.gnu.org/licenses/>.
//
////////////////////////////////////////////////////////////////

#include "rawimagesource.h"
#define BENCHMARK
#include "StopWatch.h"

using namespace rtengine;

void RawImageSource::bayer_bilinear_demosaic()
{
BENCHFUN
#ifdef _OPENMP
    #pragma omp parallel for
#endif
    for (int i = 1; i < H - 1; ++i) {
        float **nonGreen1 = red;
        float **nonGreen2 = blue;
        if (FC(i, 0) == 2 || FC(i, 1) == 2) { // blue row => swap pointers
            std::swap(nonGreen1, nonGreen2);
        }
#if defined(__clang__)
        #pragma clang loop vectorize(assume_safety)
#elif defined(__GNUC__)
        #pragma GCC ivdep
#endif
        for (int j = 2 - (FC(i, 1) & 1); j < W - 2; j += 2) { // always begin with a green pixel
            green[i][j] = rawData[i][j];
            nonGreen1[i][j] = (rawData[i][j - 1] + rawData[i][j + 1]) * 0.5f;
            nonGreen2[i][j] = (rawData[i - 1][j] + rawData[i + 1][j]) * 0.5f;
[...]

PR #5885, opened 2020-08-14
Bayer bilinear

heckflosse:
#5748


Thanatomanic:
I tested the branch and noticed only that the Contrast threshold slider does not seem to appear for AMaZE+Bilinear and DCB+Bilinear, but it does for RCD+Bilinear.

heckflosse:
@Thanatomanic Thanks for feedback :+1:  I will fix that.

Thanatomanic (approved):
I have no further comments, except that it would still be better to have a different GUI because of the rather long list of options. But that's for another time.

Issue #4763, opened 2018-08-31
Add more scopes


eprovst:
Scopes help us to quickly see what is wrong with an image and also give us the ability to do certain corrections 'by the numbers' that are hard to judge on a screen.

Potential scopes include:

**Vectorscope**  
Currently RawTherapee only has histograms which is great to see the overall lightness, redness, etc. of the image. However, it's often useful to have an idea of the color balance in terms of hue and saturation -- eg. to check skintones; this is exactly what a vectorscope is. For the skintones use case a skintone line is also a must have on a vectorscope.

A possible alternative to a vectorscope is a chrominance chart which fits easier within the current interface.

**More to be added.**

Thanatomanic:
I agree that a vectorscope and color wheels could be nice additions. I have done some color grading for film on a hobby project, and both the vectorscope and color wheels can be very convenient to use.
Personally, I would not call this a high priority thing though, because I believe the proposed color wheel operations can already be achieved using other tools available in RT. Maybe not as convenient ...

We might get some inspiration from the way darktable has implemented their ~~vector scope~~ (my bad, darktable only has a waveform tool).

eprovst:
I agree that it isn't a high priority however it's a really nice to have :smile: But it's true anything you can do with color wheels can also be done with levels although color wheels make it a lot easier to get the color you want.

I can't recall Darktable having a vectorscope, however when it comes to design I think something similar to the one in Nuke (with less lines), Final Cut or Speedgrade would work. As long as it has the skintone line and a reasonable mapping of color and saturation it will be ideal for the purpose of photography. I don't think there is a need to see the NTSC color targets for example.

**Addition:** 0 saturation should also be marked on the vectorscope of course.

heckflosse:
@ElecProg Some screenshots or a video would help more than 1000 words for this issue!

Thanatomanic:
![image](https://user-images.githubusercontent.com/6567747/44935880-bf2bdd00-ad72-11e8-9a9d-d97d9bda55ad.png)
This is how the ~~scope~~ waveform looks like in darktable.

![image](https://user-images.githubusercontent.com/6567747/44935958-13cf5800-ad73-11e8-8a47-fc8ac264594a.png)
This is what DaVinci Resolve can give you in terms of color gauges.

eprovst:
Third Davinci scope in @Thanatomanic s comment. You wont need all the lines and boxes on that scope for photography. So you could simplify the design quite a bit.

[...]

PR #5887, opened 2020-08-15
Waveform and vectorscopes

Lawrence37:
Related to #4763. The three new scopes are R G B L* waveform, HSL vectorscope, and LCH (L* a* b*) vectorscope. There are many moving parts to this: color profiles, performance, GTK, Cairo, and others. I hope there aren't many problems.


heckflosse:
I can reproduce a crash when switching to `Hue-Chroma vectorscope`.
backtrace follows

heckflosse:
This fixes the crash

```diff
diff --git a/rtengine/improccoordinator.cc b/rtengine/improccoordinator.cc
index 7ae814f8c..5920e6bc6 100644
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -1863,10 +1863,14 @@ void ImProcCoordinator::updateVectorscope()
     memset((int*)vectorscope, 0, size * size * sizeof(vectorscope[0][0]));
 
     const int lab_img_size = (hListener->vectorscopeType() == 1) ? (x2 - x1) * (y2 - y1) : 0;
-    float a[lab_img_size], b[lab_img_size];
+
+    std::unique_ptr<float[]> a;
+    std::unique_ptr<float[]> b;
     if (lab_img_size) {
-        float L[lab_img_size];
-        ipf.rgb2lab(*workimg, x1, y1, x2 - x1, y2 - y1, L, a, b, params->icm);
+        a.reset(new float[lab_img_size]);
+        b.reset(new float[lab_img_size]);
+        std::unique_ptr<float []> L(new float[lab_img_size]);
+        ipf.rgb2lab(*workimg, x1, y1, x2 - x1, y2 - y1, L.get(), a.get(), b.get(), params->icm);
     }
 
     int ofs_lab = 0;

```

Floessie:
[Too much on stack will kill you...](https://www.youtube.com/watch?v=rNa5kRX45tg) :notes:

heckflosse:
@Lawrence37 May I push the fix and also make some speedups for `ImProcCoordinator::updateVectorscope()` ?

Lawrence37:
> @Lawrence37 May I push the fix and also make some speedups for `ImProcCoordinator::updateVectorscope()` ?

Yes

heckflosse:
@Lawrence37 

I measured on my FX8350 using a 96.3 MP file

```
        before    after
HS:     170 ms     66 ms
CH:      41 ms     36 ms
```

I let the BENCHFUN in code because I will try to vectorize and parallelize the loops...

heckflosse:
@Lawrence37 I would like to push some more stuff to your branch at the weekend (until tomorrow evening). Can I just push from now on to tomorrow evening without asking you again?

Lawrence37:
@heckflosse Sure.

Thanatomanic:
@Lawrence37 This works really great! I like that you kept the bar for readout and made a crosshair for the scopes. The ability to change the intensity by click-drag is also great.

I was a bit confused by the layout of the vectorscopes though, because I expected a 90° rotation of what you have now: https://en.wikipedia.org/wiki/Vectorscope#/media/File:Vectorscope_graticule.png
Do you think it may be a good idea to add labels to the different 'spokes' of the wheel? That helps orientation.

At some point we may need to think about reorganizing the histogram buttons though, because it's getting quite crowded - not your fault, just an observation.

Thanatomanic:
Tiny other thing I noticed: by using double click you can still change the option for the histogram (linear, log, log-log), even though you are in another mode. Not a bad thing per se, but not totally logical either.

Now that I think of it, maybe the double click should become bound to rotate between the different representations. The histogram button should then be used to change the display mode. What do you think?

Lawrence37:
@Thanatomanic Adding labels is a good idea. Regarding the double clicking, I'm concerned that it will require too many clicks to cycle through the scopes. It might work better if the two vectorscopes are combined, and the mode button used to change the vectorscope type. This would also be good if more waveform modes (RGB parade, vertical waveform) get added in the future.

heckflosse:
@Lawrence37 I will try to parellelize and vectorize the two loops tomorrow (not trivial, but doable imho)

Floessie:
@heckflosse I was going to mend the LGTM warnings, when I stumbled upon this:

https://github.com/Beep6581/RawTherapee/blob/waveform/rtgui/histogrampanel.cc#L1027

Doesn't seem like a legitimate use of `array2D<>` to me, as pointers are copied. That was the original reason to derive it from `NonCopyable`, and the `memcpy()` abuses it.

Shall we provide a move c'tor and assignment? Or even a copy c'tor and assignment?

Floessie:
@heckflosse I'll provide copy c'tor and assignment first. It's needed here.

heckflosse:
@Floessie Ok

heckflosse:
While @Floessie is working on this branch, I already parallelized the loops in `ImProcCoordinator::updateVectorscope()`

New timings:

```
        before    after
HS:     170 ms     27 ms
CH:      41 ms     32 ms
```

Edit: @Floessie please ping me when I can push

Floessie:
I'll push ASAP. Stay tuned.

Floessie:
Pushed. I'm AFK now. CU later.

heckflosse:
Something broke the vectorscopes. Bisecting...

Floessie:
Most probably one of my commits... :grimacing:

heckflosse:
@Floessie this one
3af822b6f7627773ec009861123f59054b200cea

Floessie:
Hm, I must be blind...

Floessie:
Found it. Pushing...

heckflosse:
@Lawrence37 I'm done with speedups and cleanups for this pr. Imho the current speed especially for vectorscope HS is much improved now. There might be a possibility to get even more speed by using the vectorized xsincosf function. But the data layout RGBRGBRGB does not fit really well to vectorization. Perhaps I will make another try next weekend.

Lawrence37:
Thank you @heckflosse for the improvements. Everything works well.

Floessie:
Tested on Debian Testing AMD64 and Unstable PPC with GCC 9.3, GCC 10.1/10.2, and Clang 10. :+1:

heckflosse:
Switching from waveform to raw histogram requires 4 clicks (3 clicks from waveform to standard histogram and 1 click to switch to raw histogram). Wouldn't it be more userfriendly to allow switching to raw histogram by just 1 click?

rtgui/histogrampanel.cc
```diff
@@ -33,7 +33,36 @@ using namespace rtengine;
...
+                update_hist_area = histogramArea->updatePointer(r, g, b, profile, profileW);
+            }
+            if (histogramRGBArea) {
+                histogramRGBArea->queue_draw();
+            }
+            if (update_hist_area) {
+                histogramArea->queue_draw();
+            }
+        },
+        50,
```
heckflosse, line 66: @Floessie Are this hardcoded delays?

Floessie: @heckflosse Yes. Feel free to change them or make them configurable (in `dev`, not on this branch).

Lawrence37:
> Wouldn't it be more userfriendly to allow switching to raw histogram by just 1 click?

Yes, but how? We can make the raw histogram button always available. In this case, does pressing the scope type button toggle off the raw histogram? How about a pop up menu? This allows you to change to any mode in two clicks.

heckflosse:
>In this case, does pressing the scope type button toggle off the raw histogram?

Just set the scope button to insensitive if raw histogram is active ;-)

heckflosse:
> This allows you to change to any mode in two clicks.

Why not? Opinions?

Lawrence37:
> Just set the scope button to insensitive if raw histogram is active ;-)

It will take up to 4 clicks to switch from the raw histogram to another scope, although the current implementation still requires up to 3 :)

heckflosse:
> It will take up to 4 clicks to switch from the raw histogram to another scope, although the current implementation still requires up to 3 :)

Ok. Let's start from non-raw histogram mode and count:

currently :

from waveform to raw and back to waveform : 5 clicks
from HS vectorscope to raw and back to HS vestorscope: 5 clicks
from HC vectorscope to raw and back to HS vestorscope: 5 clicks

Could be done in 2 clicks by just using the histogram raw button...

Thanatomanic:
I have an opinion from a different angle: in video editing and color grading it is very common to have multiple scopes available simultaneously. I wouldn't mind a slightly bigger redesign with collapsible panels for the histogram/parade and another for the vectorscopes.

Lawrence37:
@heckflosse's idea makes switching between the raw histogram and another scope fast, but why should the raw histogram be special? Ideally, a user should be able to switch between any two scope quickly. Cycling between two scopes still takes 4 clicks with a pop up menu. Having all scopes on separate toggle buttons is better, but may clutter the interface. Multiple scopes eliminates the need to switch, but I'll have to think about how to best design that.

heckflosse:
> but why should the raw histogram be special?

It became special by adding the new scopes (which I really appreciate, I mean the new scopes, not that it became special).

heckflosse:
@Lawrence37 Let me phrase it simple: In past I could switch to the raw histogram with one click. Another idea: how about a raw waveform histogram? That would at least allow to visualize where (from left to right) raw values are clipped.

Lawrence37:
@heckflosse That makes sense (although you can still do that if you don't use the new scopes). Interesting that you mentioned the raw waveform. I kept the raw "histogram" button separate for the possibility of adding a raw waveform.

Lawrence37:
I'd like opinions on labeling the vectorscopes. Text labels can become tricky due to font faces, size, localization, visibility, etc., but those can be ignored for now. Color labels are simpler and can be implemented in many ways, but may not be ideal for color blind people. Here's an example of text labels and colored lines (together).
image

Thanatomanic:
What kind of representations do we actually want and need? Let's first try to list them all.
- **Processed histogram** (Very common, very basic. Currently can show separate channels, including luminance and chroma)
- **Raw histogram** (What is it actually good for apart from checking raw-clipping? A specialist tool imo)
- **Waveform** (Also gives spatial information, much more insightful than a regular histogram imo)
- **Raw waveform** (Similar advantage, though, how useful is it?)
- **Vectorscope** (Useful for color representation in non-RGB models. 'Skin tone' line. However, we already have two different color models, are they both needed?)
- **Parade** (Similar to Waveform, but separate channels adjoined)
- **Raw parade** (Because, why not?)

As for the necessary options:
- The non-raw histogram, waveform and parade should/could toggle channels: R, G, B, L and c.
- The raw histogram, waveform and parade should/could toggle channels: R, G and B. Luckily non-RGB sensors are not supported in RT...
- The vectorscope should/could toggle modes: LCH, HSL, whatnot.

Am I missing anything? More to add? Edit: yes, one thing to add. How do we handle (un)clipping?
Edit 2: The option to switch to linear, log-linear and log-log graphs is also that can be beneficial, in particular for the 'regular' histogram. For the waveform I'm not sure.

@Lawrence37 I really like the labeled vectorscope like this. Do the lines increase and decrease brightness along with the rest?

Lawrence37:
The vectorscope lines have fixed brightness.

Raw HS vectorscope is possible. Not very useful IMO. The waveforms and RGB parades can have vertical and horizontal variants. Don't forget the option to change between working and output color profile. This affects values for RGB histogram/waveform and HS vectorscope, and clipping of CH vectorscope.

Lawrence37:
> **Vectorscope** (Useful for color representation in non-RGB models. 'Skin tone' line. However, we already have two different color models, are they both needed?)

The CH vectorscope is essentially L* a* b* and has all the benefits of it. The HS vectorscope is useful to prevent clipping colors and not much else.

> How do we handle (un)clipping?

Just a note: All processed RGB values are currently clipped at 0% and 100%. L and C values are not clipped (intentional?), except for out of gamut colors in the CH vectorscope.

Thanatomanic:
I feel like we are heading toward a separate dialog or preference page where you can set everything to your liking, considering the sheer number of options. 🙄 

> Raw HS vectorscope is possible. Not very useful IMO.

Raw values don't have color information, so how can you extract a hue and saturation?

heckflosse:
> Raw values don't have color information, so how can you extract a hue and saturation?

though a raw waveform would be possible (r, g, b, and even L/Y)

@Lawrence37 please don't get me wrong, I really like the new scopes. I just want to make them accesible with less clicks ;-)

Thanatomanic:
> though a raw waveform would be possible (r, g, b, and even L/Y)

After conversion to XYZ, yes, but do we do that in the raw histogram? (atm too lazy to check, sorry)

> @Lawrence37 please don't get me wrong, I really like the new scopes. I just want to make them accesible with less clicks ;-)

+1 for me!

Lawrence37:
> Raw values don't have color information, so how can you extract a hue and saturation?

Whoops. I was thinking about demosaiced data.

> @Lawrence37 please don't get me wrong, I really like the new scopes. I just want to make them accesible with less clicks ;-)

I know. :)

Floessie:
While not perceptible on a fast machine, switching to the HC vectorscope (the last one) takes more time than every other histogram options on an emulated PPC system.

If you're going to add more options, please consider `enum`s from the start. The only place to deal with numbers should be when loading and saving the settings to disk. Then `switch` on the `enum` value without `default:`, and the compiler will warn us about new and unhandled options. It's really as easy as this.

heckflosse:
> While not perceptible on a fast machine, switching to the HC vectorscope (the last one) takes more time than every other histogram options on an emulated PPC system.

good catch @Floessie  :+1: 
It is even perceptible on a fast machine when the output-profile is used to render the histogram, because then lcms kicks in
grafik

For the image I used above it takes ~200 ms while in working-profile mode it takes only ~32 ms

Imho this can be optimized, because here `workimg` is created from the Lab image
https://github.com/Beep6581/RawTherapee/blob/waveform/rtengine/improccoordinator.cc#L1627
and then here it is converted back using the same settings.
https://github.com/Beep6581/RawTherapee/blob/waveform/rtengine/improccoordinator.cc#L1897

Imho we could just use `nprevl` instead to at least avoid the backwards conversion. That should give a nice speedup for the HC mode.

Edit: I expect processing time for my above test image in HC mode to be in the range of ~ 2 ms when using `nprevl`

Floessie:
@heckflosse Much better! :+1:

heckflosse:
@Thanatomanic 
> Raw histogram (What is it actually good for apart from checking raw-clipping? A specialist tool imo)

It's also useful to check your images for too much underexposure. For example this one I took is at least one whole stop underexposed. Next time in such a haze I will set +1 on camera to get some more bits and so less noise ;-)
grafik

heckflosse:
@Lawrence37 
> Just a note: All processed RGB values are currently clipped at 0% and 100%. L and C values are not clipped (intentional?), except for out of gamut colors in the CH vectorscope.

For the CH vectorscope after my latest commits the values are not clipped anymore because I removed the conversion Lab->RGB and RGB-> Lab. At first that looks like kind of a regression, but I think it's not, because now we could draw lines where there would be clipping in the vectorscope, so we could visualize if there are out of gamut values for the working/output profile.

Lawrence37:
Unfortunately, seeing out of gamut colors in the CH vectorscope is not a simple as drawing lines. Here's an example using colors at the edges of the gamut.

Visualization of the gamut boundaries:
image

A cropped region with lightness adjustment:
image
The two bright lines radiating from the center of the vectorscope clearly shows pixels getting crushed to cyan and magenta.

Same as above, but without clipping:
image
The vectorscope looks like this regardless of the lightness even though the out of gamut colors change with lightness.

The good news is that the HS vectorscope is excellent for judging out of gamut values values. With the scope brightness turned down, it's clear many pixels are fully saturated cyan and magenta:
image

In summary, the latest commit makes it difficult to judge out of gamut colors with the CH vectorscope, but you probably want to use the HS vectorscope for this purpose anyway. Is the performance gain worth it? Maybe. Thoughts?

heckflosse:
If I am not totally wrong the CH vectorscope shows a/b values from Lab space. This values don't depend on the RGB values of working/output profile. The choice of RGB working/output profile determines, wheter the values will be clipped or not clipped.
Therefore it would be a better approach to visualize this clipping by showing all values instead of clipping them before visualizing them. Means, the a/b values should be visualized the same independend on working/output space, but the visualization of the clipping in the vectorscope should depend on the working/output space.

Please correct me if I miss someting here.

About the performance: A gain from 200 ms to 2 ms is not so bad.

heckflosse:
@Lawrence37 

Let's assume this is the vectorscope in ProPhoto RGB
grafik

In sRGB it would have the same shape except for the clipping boundaries (simplified by a smaller circle in the screenshot)
grafik

Lawrence37:
Marking the profile's boundaries with a "cage" is not sufficient as the clipping boundaries change for different values of L. One could always use the HS vectorscope instead (in fact, I encourage this to visualize clipping), but some users may find themselves flipping between the two vectorscopes. Which reminds me, we still need to decide on a faster way to change scopes.

heckflosse:
@Lawrence37 I am aware that is not always easy to follow my ideas (espacially as I am often not good in explaining them :frowning_face:)

> Marking the profile's boundaries with a "cage" is not sufficient as the clipping boundaries change for different values of L.

Can you give me a pointer to this dependency for CH vectorscope?

Lawrence37:
Sure. For sRGB: https://en.wikipedia.org/wiki/CIELAB_color_space#/media/File:Lab_color_space.png

heckflosse:
>Sure.

I meant in current RT waveform code

Lawrence37:
There's nothing in the code. What I wrote refers to the properties of Lab.

Can you show me what you mean by "visualization of the clipping"?

heckflosse:
@Lawrence37 I had something in mind, but it did not work. I reverted my last commit.

Lawrence37:
For scope switching, I'm thinking of a hybrid between a popup menu and having buttons for each scope. Clicking the scope type button reveals or hides a second column of buttons used to select the scope type.

Lawrence37:
My latest commit contains the new way of switching scopes. It only takes one click to set a new scope. The extra buttons can be hidden for more space.
image image

Thanatomanic:
Looks good @Lawrence37, I'll test it asap!

heckflosse:
I get 3 cppcheck warnings:
```
[H:/rt59/rtgui/histogrampanel.cc:1050] (warning) Member variable 'HistogramArea::needPointer' is not initialized in the constructor. [uninitMemberVar]
[H:/rt59/rtgui/histogrampanel.cc:1050] (warning) Member variable 'HistogramArea::pointer_a' is not initialized in the constructor. [uninitMemberVar]
[H:/rt59/rtgui/histogrampanel.cc:1050] (warning) Member variable 'HistogramArea::pointer_b' is not initialized in the constructor. [uninitMemberVar]
```

Lawrence37:
Thank you @heckflosse. I fixed the warnings and some other ones too.

heckflosse:
@Lawrence37 I really like the vector scopes :+1: 

May I suggest a really small improvement by asking a question? Why do you not draw the lines to match the circle?

grafik

Lawrence37:
It does look nicer without the gap (at the outside circle and the center). I think I will change it.
image

Thanatomanic:
@Lawrence37 I like your submenu idea, but in my mind it would make more sense to switch this around: first pick your mode of representation, then present some options as a secondary menu. What do you think?

Lawrence37:
That's an interesting idea. Do people use the options often? If they do, that would be my only argument against switching.

heckflosse:
Do people use the options often?

Imho, No.

Thanatomanic:
@heckflosse :+1: 
I can also imagine that turning individual channels on/off is not something you do too often, same goes for activating the RGB readout bar. You pick a mode and setup and stick with it. Choosing linear vs. log modes might be something a little more common (at least for myself), but I wouldn't mind if that option was in the submenu.

Still, before you change it, maybe it's good to get some more opinions? You could start a topic on the Pixls forum if you like.

Floessie:
@heckflosse This fails due to a5a44ce9b2a3702dd8b0b52401dfd9b0d85ff93c.

heckflosse:
@Floessie It turned out to be good that I removed that function, else we would not have detected this case, which I will fix now.
@Lawrence37 Can I push a fix to your branch soon?

heckflosse:
@Lawrence37 May I suggest to change the brightness (or the opacity) of the horizontal lines in waveform when click and drag?

currently with dark view they are visible
grafik

but with brighter view they are not
grafik

Lawrence37:
@heckflosse You can push the fix. I will try to increase the visibility of the horizontal lines.

heckflosse:
@Lawrence37 I pushed

heckflosse:
Btw: I really think, this work should find its way into RT 5.9 :+1:

Lawrence37:
Improving the visibility of the lines was harder than I expected. The best results came from solid lines over a darkened waveform.
Before/After:
image image

How does it look?

heckflosse:
@Lawrence37 There is still some transparency for the horizontal lines. Could you try increasing the opacity of the hor. lines when brightening the waveform.
Edit: attached an upsized screenshot to show the transparency (opacity)
grafik

Lawrence37:
With the opacity increased:
image
I hope the lines don't interfere too much with the waveform.

Thanatomanic:
@Lawrence37 Hmm, I can't say I like your last suggestion too much. Maybe clamp the opacity of the gridlines at (idk) 80%?

Also, have you checked how the waveform and scopes look like when using the different RT themes?

Lawrence37:
The opacity is 75% (original is 25%). At 50%, it looks like this:
image

I just checked the scopes with different themes. They look fine.

heckflosse:
> Maybe clamp the opacity of the gridlines at (idk) 80%?

Or make them dashed, as they were before, but with increased opacity?

Lawrence37:
Dashed and high opacity:
image
I'm not a big fan of high opacity, but that's just my opinion.

heckflosse:
@Lawrence37 Can you point me to the code, which draws the horizontal lines? I would like to try something...

Lawrence37:
@heckflosse https://github.com/Beep6581/RawTherapee/blob/8d2754575a35e8ad84377440e160a79de9930c3d/rtgui/histogrampanel.cc#L1289

heckflosse:
@Lawrence37 
Here's a patch which changes opacity of the hor. lines dynamically (currently only for waveform). Just as a test

```diff
diff --git a/rtgui/histogrampanel.cc b/rtgui/histogrampanel.cc
index 0df6496c7..a6a2fb0d1 100644
--- a/rtgui/histogrampanel.cc
+++ b/rtgui/histogrampanel.cc
@@ -1326,12 +1326,15 @@ void HistogramArea::updateBackBuffer ()
 
     // draw horizontal gridlines
     if (options.histogramScopeType == ScopeType::PARADE || options.histogramScopeType == ScopeType::WAVEFORM) {
+        const double opacity = std::min(trace_brightness / 3.0, 0.75) + 0.25;
+        cr->set_source_rgba (1., 1., 1., opacity);
         for (int i = 0; i <= nrOfVGridPartitions; i++) {
             const double ypos = h - padding - (pow(2.0,i) - 1) * (h - 2 * padding - 1) / 255.0;
             cr->move_to(0, ypos);
             cr->line_to(w, ypos);
             cr->stroke();
         }
+        cr->set_source_rgba (1., 1., 1., 0.25);
     } else if (options.histogramScopeType == ScopeType::VECTORSCOPE_HC || options.histogramScopeType == ScopeType::VECTORSCOPE_HS) {
         // Vectorscope has no gridlines.
     } else if (options.histogramDrawMode == 0) {

```

heckflosse:
@Lawrence37 I can reproduce a segfault in waveform branch (without my patch from above) when activating before/after view
Here's the bt
```
Thread 130 received signal SIGSEGV, Segmentation fault.
[Switching to Thread 20984.0x175c]
rtengine::ImProcCoordinator::updatePreviewImage (this=0x2e26fbc0, todo=4095,
    panningRelatedChange=true) at Z:/H2/rt59/rtengine/improccoordinator.cc:367
367                 hist_raw_dirty = !hListener->updateHistogramRaw();
(gdb) bt full
#0  rtengine::ImProcCoordinator::updatePreviewImage (this=0x2e26fbc0,
    todo=4095, panningRelatedChange=true)
    at Z:/H2/rt59/rtengine/improccoordinator.cc:367
        rp = {bayersensor = {method = {static npos = 18446744073709551615,
              string_ = "amaze"}, border = 4, imageNum = 0, ccSteps = 0,
            black0 = 0, black1 = 0, black2 = 0, black3 = 0, twogreen = true,
            linenoise = 0,
            linenoiseDirection = rtengine::procparams::RAWParams::BayerSensor::LineNoiseDirection::BOTH, greenthresh = 0, dcb_iterations = 2,
            lmmse_iterations = 2, dualDemosaicAutoContrast = true,
            dualDemosaicContrast = 20,
            pixelShiftMotionCorrectionMethod = rtengine::procparams::RAWParams::BayerSensor::PSMotionCorrectionMethod::AUTO, pixelShiftEperIso = 0,
            pixelShiftSigma = 1, pixelShiftShowMotion = false,
            pixelShiftShowMotionMaskOnly = false, pixelShiftHoleFill = true,
            pixelShiftMedian = false, pixelShiftGreen = true,
            pixelShiftBlur = true,
            pixelShiftSmoothFactor = 0.69999999999999996,
            pixelShiftEqualBright = false,
            pixelShiftEqualBrightChannel = false,
            pixelShiftNonGreenCross = true, pixelShiftDemosaicMethod = {
              static npos = 18446744073709551615, string_ = "amaze"},
            dcb_enhance = true, pdafLinesFilter = false}, xtranssensor = {
            method = {static npos = 18446744073709551615,
              string_ = "3-pass (best)"}, dualDemosaicAutoContrast = true,
            dualDemosaicContrast = 20, border = 7, ccSteps = 0, blackred = 0,
            blackgreen = 0, blackblue = 0}, dark_frame = {
            static npos = 18446744073709551615, string_ = ""},
          df_autoselect = false, ff_file = {
            static npos = 18446744073709551615, string_ = ""},
          ff_AutoSelect = false, ff_BlurRadius = 32, ff_BlurType = {
            static npos = 18446744073709551615, string_ = "Area Flatfield"},
          ff_AutoClipControl = false, ff_clipControl = 0,
          ca_autocorrect = true, ca_avoidcolourshift = true,
          caautoiterations = 2, cared = 0, cablue = 0, expos = 1,
          preprocessWB = {
            mode = rtengine::procparams::RAWParams::PreprocessWB::Mode::AUTO}, hotPixelFilter = false, deadPixelFilter = false, hotdeadpix_thresh = 100}
        lcur = {enabled = false,
          lcurve = std::vector of length 1, capacity 1 = {0},
          acurve = std::vector of length 1, capacity 1 = {0},
          bcurve = std::vector of length 1, capacity 1 = {0},
          cccurve = std::vector of length 1, capacity 1 = {0},
          chcurve = std::vector of length 1, capacity 1 = {0},
          lhcurve = std::vector of length 1, capacity 1 = {0},
          hhcurve = std::vector of length 1, capacity 1 = {0},
          lccurve = std::vector of length 1, capacity 1 = {0},
          clcurve = std::vector of length 1, capacity 1 = {0},
          brightness = 0, contrast = 0, chromaticity = 0,
          avoidcolorshift = false, rstprotection = 0, lcredsk = true}
        autowb = false
        cmp = {inputProfile = {static npos = 18446744073709551615,
            string_ = "(cameraICC)"}, toneCurve = false,
          applyLookTable = true, applyBaselineExposureOffset = true,
          applyHueSatMap = true, dcpIlluminant = 0, workingProfile = {
            static npos = 18446744073709551615, string_ = "ProPhoto"},
          workingTRC = {static npos = 18446744073709551615,
            string_ = "none"}, workingTRCGamma = 2.3999999999999999,
          workingTRCSlope = 12.9231, outputProfile = {
            static npos = 18446744073709551615, string_ = "RTv4_sRGB"},
          outputIntent = rtengine::RI_RELATIVE, outputBPC = true,
          static NoICMString = {static npos = 18446744073709551615,
            string_ = "No ICM: sRGB output"}}
        needstransform = false
        processingLock = {<rtengine::NonCopyable> = {<No data fields>},
          mutex = @0x2e271e98, locked = true}
        highDetailNeeded = true
```

heckflosse:
@Lawrence37 
This fixes the segfault
```diff
diff --git a/rtengine/improccoordinator.cc b/rtengine/improccoordinator.cc
index 72dfb0cd3..e9e424ea8 100644
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -364,7 +364,7 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
             }
 
             imgsrc->getRAWHistogram(histRedRaw, histGreenRaw, histBlueRaw);
-            hist_raw_dirty = !hListener->updateHistogramRaw();
+            hist_raw_dirty = hListener && !hListener->updateHistogramRaw();
 
             highDetailPreprocessComputed = highDetailNeeded;
 
```

Lawrence37:
Good catch. It should be `hist_raw_dirty = !(hListener && hListener->updateHistogramRaw());`

Lawrence37:
What do you think of this for the horizontal lines?
```diff
diff --git a/rtgui/histogrampanel.cc b/rtgui/histogrampanel.cc
index 0df6496c7..b4bd8683e 100644
--- a/rtgui/histogrampanel.cc
+++ b/rtgui/histogrampanel.cc
@@ -1326,12 +1326,7 @@ void HistogramArea::updateBackBuffer ()
 
     // draw horizontal gridlines
     if (options.histogramScopeType == ScopeType::PARADE || options.histogramScopeType == ScopeType::WAVEFORM) {
-        for (int i = 0; i <= nrOfVGridPartitions; i++) {
-            const double ypos = h - padding - (pow(2.0,i) - 1) * (h - 2 * padding - 1) / 255.0;
-            cr->move_to(0, ypos);
-            cr->line_to(w, ypos);
-            cr->stroke();
-        }
+        // Draw gridlines on top of the waveform trace instead of here.
     } else if (options.histogramScopeType == ScopeType::VECTORSCOPE_HC || options.histogramScopeType == ScopeType::VECTORSCOPE_HS) {
         // Vectorscope has no gridlines.
     } else if (options.histogramDrawMode == 0) {
@@ -1467,6 +1462,26 @@ void HistogramArea::updateBackBuffer ()
         drawVectorscope(cr, w, h);
     }
 
+    if (options.histogramScopeType == ScopeType::PARADE || options.histogramScopeType == ScopeType::WAVEFORM) {
+        // Draw horizontal lines over parade and waveform.
+        const double opacity = rtengine::LIM(0.25 * trace_brightness, 0.25, 1.0);
+        cr->set_dash (ch_ds, 0);
+        cr->set_source_rgba(1.0, 1.0, 1.0, opacity);
+        for (int i = 1; i < nrOfVGridPartitions; i++) {
+            const double ypos = h - padding - (pow(2.0,i) - 1) * (h - 2 * padding - 1) / 255.0;
+            cr->move_to(0, ypos);
+            cr->line_to(w, ypos);
+            cr->stroke();
+        }
+        cr->unset_dash();
+        cr->move_to(0, h - padding);
+        cr->line_to(w, h - padding);
+        cr->move_to(0, padding + 1);
+        cr->line_to(w, padding + 1);
+        cr->stroke();
+        cr->set_source_rgba(1.0, 1.0, 1.0, 0.25);
+    }
+
     // Draw the frame's border
     style->render_frame(cr, 0, 0, surface->get_width(), surface->get_height());
 
@@ -1903,7 +1918,7 @@ void HistogramArea::drawWaveform(Cairo::RefPtr<Cairo::Context> &cr, int w, int h
             wave_buffer_luma.data(), Cairo::FORMAT_ARGB32, wave_width, wave_height, cairo_stride);
         cr->set_source(surface, 0, 0);
         cr->set_operator(Cairo::OPERATOR_OVER);
-        cr->paint();
+        cr->paint_with_alpha(0.85);
         surface->finish();
     }
     if (needRed || needGreen || needBlue) {
@@ -1911,7 +1926,7 @@ void HistogramArea::drawWaveform(Cairo::RefPtr<Cairo::Context> &cr, int w, int h
             wave_buffer.data(), Cairo::FORMAT_ARGB32, wave_width, wave_height, cairo_stride);
         cr->set_source(surface, 0, 0);
         cr->set_operator(Cairo::OPERATOR_OVER);
-        cr->paint();
+        cr->paint_with_alpha(0.85);
         surface->finish();
     }
     cr->set_matrix(orig_matrix);
```

Thanatomanic:
@Lawrence37 I'm really happy with your recent changes, looks great and works great!
Two minor things I don't really like:
1. The configuration icon at the bottom is too far away from the rest of the buttons, particularly when you work with a larger area. There is a separator line, so wouldn't it be better to just stick it to the immediate end of the column of icons?
2. The icons for the different modes are quite small making it very hard to click and to know what you're clicking. I think I would prefer a different way to indicate the currently active button, but I'm not sure what would be better.

Lawrence37:
I will remove the spacing between the scope buttons and the bottom button. It seems better thinking about it now. I have been thinking about the tiny icons for a while. Now that the scope buttons are in a separate column, I can break away from the styling of the other histogram buttons and try a different approach for indicating the current scope type.

Lawrence37:
I think the best option is to set the button background in the themes' CSS. Here's the RawTherapee theme with the style copied from the flat button style.
image

Thanatomanic:
Hi @Lawrence37, I get a minor compilation warning in the current situation

```
[ 69%] Building CXX object rtgui/CMakeFiles/rth.dir/histogrampanel.cc.obj
C:/msys64/home/Roel/rtdev/rtgui/histogrampanel.cc: In member function 'void HistogramArea::drawParade(Cairo::RefPtr<Cairo::Context>&, int, int)':
C:/msys64/home/Roel/rtdev/rtgui/histogrampanel.cc:1603:23: warning: comparison of integer expressions of different signedness: 'int' and 'std::vector<unsigned char*>::size_type' {aka 'long long unsigned int'} [-Wsign-compare]
 1603 |     for (int i = 0; i < buffers.size(); i++) {
      |                     ~~^~~~~~~~~~~~~~~~
```

Other than that, do you think things are ready for a merge to get more people to test? After that I will also continue to work on my PR for improved (raw) histogram binning and drawing.

Lawrence37:
I think it can be safely merged as it is. There are some potential features such as labels and dynamic horizontal line brightness, but of course they can be added separately.

Thanatomanic:
In that case I propose we wait a few days for final comments and then you can merge whenever you feel like pushing the button. 😃 

Might I propose to squash and merge so that your changes show up as a single, easily traceable entry in the commit log?

dafrasaga:
Hi, I'll be very happy if waveform will be into dev but this morning I tried to compile git version but I have this errors:

`[100%] Linking CXX executable rawtherapee
/root/git/RawTherapee/rtengine/dcraw.cc: In member function 'nikon_yuv_load_raw':
/root/git/RawTherapee/rtengine/dcraw.cc:1412:33: warning: 'yuv[3]' may be used uninitialized in this function [-Wmaybe-uninitialized]
 1412 |       rgb[0] = yuv[b] + 1.370705*yuv[3];
      |                                 ^
/root/git/RawTherapee/rtengine/dcraw.cc:1402:17: note: 'yuv[3]' was declared here
 1402 |   int row, col, yuv[4], rgb[3], b, c;
      |                 ^
/root/git/RawTherapee/rtengine/dcraw.cc:1413:33: warning: 'yuv[2]' may be used uninitialized in this function [-Wmaybe-uninitialized]
 1413 |       rgb[1] = yuv[b] - 0.337633*yuv[2] - 0.698001*yuv[3];
      |                                 ^
/root/git/RawTherapee/rtengine/dcraw.cc:1402:17: note: 'yuv[2]' was declared here
 1402 |   int row, col, yuv[4], rgb[3], b, c;
      |                 ^
/root/git/RawTherapee/rtengine/imageio.cc: In member function 'getAutoMatchedToneCurve':
/root/git/RawTherapee/rtengine/histmatching.cc:296:105: warning: 'h' may be used uninitialized in this function [-Wmaybe-uninitialized]
  296 |                 std::cout << "histogram matching: the embedded thumbnail is too small: " << w << "x" << h << std::endl;
      |                                                                                                         ^
/root/git/RawTherapee/rtengine/histmatching.cc:282:16: note: 'h' was declared here
  282 |         int w, h;
      |                ^
/root/git/RawTherapee/rtengine/histmatching.cc:291:22: warning: 'w' may be used uninitialized in this function [-Wmaybe-uninitialized]
  291 |         } else if (w * 33 < fw || w * h < 19200) {
      |                      ^
/root/git/RawTherapee/rtengine/histmatching.cc:282:13: note: 'w' was declared here
  282 |         int w, h;
      |             ^
[100%] Built target rth-cli
/root/git/RawTherapee/rtengine/dcraw.cc: In member function 'nikon_yuv_load_raw':
/root/git/RawTherapee/rtengine/dcraw.cc:1412:33: warning: 'yuv[3]' may be used uninitialized in this function [-Wmaybe-uninitialized]
 1412 |       rgb[0] = yuv[b] + 1.370705*yuv[3];
      |                                 ^
/root/git/RawTherapee/rtengine/dcraw.cc:1402:17: note: 'yuv[3]' was declared here
 1402 |   int row, col, yuv[4], rgb[3], b, c;
      |                 ^
/root/git/RawTherapee/rtengine/dcraw.cc:1413:33: warning: 'yuv[2]' may be used uninitialized in this function [-Wmaybe-uninitialized]
 1413 |       rgb[1] = yuv[b] - 0.337633*yuv[2] - 0.698001*yuv[3];
      |                                 ^
/root/git/RawTherapee/rtengine/dcraw.cc:1402:17: note: 'yuv[2]' was declared here
 1402 |   int row, col, yuv[4], rgb[3], b, c;
      |                 ^
/usr/bin/ld: /tmp/rawtherapee.0MZCzs.ltrans108.ltrans.o: in function `HistogramArea::setBrightness(float)':
<artificial>:(.text+0xbf45): undefined reference to `HistogramArea::MAX_BRIGHT'
/usr/bin/ld: <artificial>:(.text+0xbf5c): undefined reference to `HistogramArea::MIN_BRIGHT'
/usr/bin/ld: /tmp/rawtherapee.0MZCzs.ltrans108.ltrans.o: in function `HistogramArea::on_motion_notify_event(_GdkEventMotion*)':
<artificial>:(.text+0xc1a3): undefined reference to `HistogramArea::MAX_BRIGHT'
/usr/bin/ld: <artificial>:(.text+0xc1cd): undefined reference to `HistogramArea::MIN_BRIGHT'
collect2: error: ld returned 1 exit status
make[2]: *** [rtgui/CMakeFiles/rth.dir/build.make:2442: rtgui/rawtherapee] Error 1
make[1]: *** [CMakeFiles/Makefile2:264: rtgui/CMakeFiles/rth.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
`

Any hints??

Thanatomanic:
@dafrasaga I have no trouble to compile this branch. Can you try to `rm -rf *` in your build directory and try compiling again?

dafrasaga:
Hi , I delete all into build dir but I have this error

`[100%] Linking CXX executable rawtherapee
/usr/bin/ld: CMakeFiles/rth.dir/histogrampanel.cc.o: in function `HistogramArea::on_motion_notify_event(_GdkEventMotion*)':
/root/git/RawTherapee/rtgui/histogrampanel.cc:1956: undefined reference to `HistogramArea::MAX_BRIGHT'
/usr/bin/ld: /root/git/RawTherapee/rtgui/histogrampanel.cc:1956: undefined reference to `HistogramArea::MIN_BRIGHT'
/usr/bin/ld: CMakeFiles/rth.dir/histogrampanel.cc.o: in function `HistogramArea::setBrightness(float)':
/root/git/RawTherapee/rtgui/histogrampanel.cc:1971: undefined reference to `HistogramArea::MAX_BRIGHT'
/usr/bin/ld: /root/git/RawTherapee/rtgui/histogrampanel.cc:1971: undefined reference to `HistogramArea::MIN_BRIGHT'
collect2: error: ld returned 1 exit status
make[2]: *** [rtgui/CMakeFiles/rth.dir/build.make:2442: rtgui/rawtherapee] Error 1
make[1]: *** [CMakeFiles/Makefile2:264: rtgui/CMakeFiles/rth.dir/all] Error 2
make: *** [Makefile:149: all] Error 2
`

I deleted all RT git dir and pull from the start but the issue is again there
PS I have no problem to compile dev branch

dafrasaga:
If I select -DCMAKE_BUILD_TYPE="release" other than -DCMAKE_BUILD_TYPE="debug" now it works.

heckflosse:
I confirm the debug build is broken.

heckflosse:
This fixes the debug build linkage error:

```diff
diff --git a/rtgui/histogrampanel.cc b/rtgui/histogrampanel.cc
index bbad4ecf3..d7e9bbc69 100644
--- a/rtgui/histogrampanel.cc
+++ b/rtgui/histogrampanel.cc
@@ -29,6 +29,9 @@
 
 using namespace rtengine;
 
+constexpr float HistogramArea::MAX_BRIGHT;
+constexpr float HistogramArea::MIN_BRIGHT;
+
 using ScopeType = Options::ScopeType;
 
 //

```

heckflosse:
@Lawrence37 

https://github.com/Beep6581/RawTherapee/blob/waveform/rtgui/histogrampanel.cc#L1784

```
constexpr float ab_factor = 1.f / 256.f;
```

if this is to convert from [0;65535] to [0;255] wouldn't it be more correct to divide by 257?

Lawrence37:
Is 257 the correct conversion factor in RT?

heckflosse:
@Lawrence37 

> Is 257 the correct conversion factor in RT?

65535 / 257 = 255

heckflosse:
@Lawrence37 

Please push the merge button. My comment was not meant to hinder you. I was just nitpicking.

heckflosse:
@Thanatomanic 

> Might I propose to squash and merge so that your changes show up as a single, easily traceable entry in the commit log?

I am ambivalent on this because lots of seperate code changes make it easier to git bisect in case...

Lawrence37:
I ask because in some places, values are divided by powers of two. In other places, they are not...

Thanatomanic:
@heckflosse I know, it's a head scratcher. I think I prefer having one commit that says: here's new functionality, this is all of it. That makes it easier to bisect feature-wise: either it was there, or it wasn't. Of course it makes bisecting _within_ a feature development harder. In this particular case I believe it is warranted to squash, because the changes are rather contained and mostly introduce new stuff instead of changing existing stuff (as it stands 2,617 lines added, only 266 removed/replaced).

@Lawrence37 The division by 255, 256 or 257 is something I had encountered before when working on the histogram. I think it really depends on what you want to achieve (division in equal parts or just scaling/normalizing something), but I can check again when proceeding with the other merge for the histogram.

Lawrence37:
Since I rarely look at the commit history in dev, I have no opinion on whether to squash or not. I will wait for a little bit to see if there is a consensus. Otherwise, I will just perform a normal merge as that appears to be the norm here.

Issue #5855, opened 2020-07-18
Regression in Sony lens identification


Beep6581:
There is a regression in Sony lens identification.

I bisected the bad commit to 24684e5cb which shows "Unknown (70-200mm f/4.0)" and also lensfun identification fails.

The previous commit 227c2fbba is good and shows "Sony FE 70-200mm f/4 G OSS".

Test image:
http://rawtherapee.com/shared/test_images/chicken.arw

Beep6581:
Upstream issue opened:
https://github.com/exiftool/exiftool/issues/54

Floflo81:
Nice to see this is being worked on.
I think it is a duplicate of issue #5670
Check this other issue for other examples

Beep6581:
Hi @Floflo81 
Yes, it is a duplicate, thanks.

PR #5888, opened 2020-08-15
Some updates / changes to lens data in rtexif

Thanatomanic:
Added explanations for syncing metadata, updated some lenses based on current ExifTool and exiv2, aligned naming for some lenses with the same IDs


Floessie:
@Thanatomanic Any reason why you removed the Tamron model numbers? The model names are so long and only differ slightly that I always look for the model number first to find my lenses...

rtexif/olympusattribs.cc
```diff
@@ -77,6 +77,24 @@ class OLLensTypeInterpreter : public Interpreter
...
+        //            present in lensfun.
+        // "01 05 00" exiv2 does not mention EX DC HSM and there is also a second
+        //            entry for this lens "01 13 00". Uncertain what is correct.
+        //
+        // There are several lenses with identical names but two different ID's:
+        //  - Leica D Summilux 25mm f/1.4 Asph.
+        //  - Leica D Vario Elmarit 14-50mm f/2.8-3.5 Asph.
+        //  - Sigma 18-125mm f/3.5-5.6 DC
+        //  - Sigma 30mm f/1.4 EX DC HSM
+        //  - Sigma APO 50-500mm f/4.0-6.3 EX DG HSM
```
Beep6581, line 96: Hey

Model-specific comments such as these do not belong here, as they will become obsolete during future updates and will themselves need to be maintained.

Beep6581 (commented):
Hey

I see no changes to the script which is used to extract the data from ExifTool, so how are these updates generated? Which version of ExifTool was used? How is Exiv2 involved? How are upstream changes to existing entries handled?

Thanatomanic:
@Floessie I see your point. I would certainly not be against returning the model names for convenience sake. The reason I took them out is that exiv2 doesn't use them, and neither does lensfun (which is the primary reason why lens recognition is necessary).

@Beep6581 There is a script? 🙈 
I did manual checking against the latest release versions of ExifTool (12.04) and exiv2 (0.27.3).
Summary of changes:
- Removal of Tamron identifiers (easily restored)
- Improved consistency of labeling with 'or ### Lens' on multiple occasions
- Added a few lenses: [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-fe9bbaf79b7a919302eee1620bb6b1a3R729), [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-fe9bbaf79b7a919302eee1620bb6b1a3R809), [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-fe9bbaf79b7a919302eee1620bb6b1a3R938), [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-a6beb8c2c5054baf32df9ef27d4e0df2R727), [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-a6beb8c2c5054baf32df9ef27d4e0df2R783), [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-a6beb8c2c5054baf32df9ef27d4e0df2R945), and [here](https://github.com/Beep6581/RawTherapee/pull/5888/files#diff-a6beb8c2c5054baf32df9ef27d4e0df2R968)
- Added comments to each list as some sort of documentation, which I thought was lacking (case in point: otherwise I could have seen that there was a script that does this work)

If you feel that any of the changes should be undone, I have no objections to that. I was merely looking for a way to procrastinate by doing this.

rtexif/olympusattribs.cc
Thanatomanic: Hmm, maintenance is tricky indeed. However, I thought the fact that these duplicates exist is interesting to mention, but that should probably be sent upstream. I couldn't find any mention of why these are duplicates in ExifTool nor exiv2.

Beep6581:
@Thanatomanic see https://github.com/Beep6581/RawTherapee/blob/dev/tools/generateRtexifUpdates

I've been maintaining rtexif for a number of years. I originally intended the update of data used in rtexif to be a one-time thing, but it ended up being a regular thing for every release. As you know, but I'll repeat here for other readers' benefit, rtexif is a home-made solution to handle metadata, and it has not been a viable solution for many years - there are two other mature projects who specialize in handling metadata and we should leave that to them and focus on raw. See #3801. Continuing down our current path will just get worse over time, as metadata-related issues pile up and go unfixed, metadata is incorrect or missing from RawTherapee's saved files, and upstream diverges from what RawTherapee can use - the recent Sony upstream cleanup (#5855) is one example. So while I don't mind re-running the script to synchronize rtexif with ExifTool for every release, everyone must be aware that this is a temporary hack. Any effort that goes into updating rtexif (instead of replacing it) should at least not increase the difficulty of maintaining it. And any manual changes to rtexif data do just that - make maintaining it more difficult. Syncing rtexif for RawTherapee 5.9 will remove manually-made changes.

And the good thing is that the work to implement Exiv2 support has already been done in branch [`metadata-exiv2`](https://github.com/Beep6581/RawTherapee/tree/metadata-exiv2). It's a pity we hadn't merged it sooner, as there are now conflicts to be resolved. I opened a PR now anyway (PR #5889). Yes it might cause a regression in some corner cases, such as multi-frame raw formats, but IMHO those regressions:
1. affect few people,
1. are temporary, as upstream is likely to implement/fix them, and
1. are overwhelmed by the advantages of doing so.

Thanatomanic:
I was aware of the custom implementation and the trouble it sometimes causes. The script was new, but is obvious in hindsight. I was a bit overzealous here (though I didn't find the maintenance I did manually too cumbersome).
Of course adopting a dedicated library is the right way forward, and any inconsistencies we find should be propagated upstream. I'm all up for that 😄 

Even though Robin Mills will soon retire, the `exiv2` library seems to have regular commits by others, so it's more or less safe to assume updates will happen. So... let's try to get the merger underway. I'll close this PR for now.


P.S. @Floessie After the move to exiv2, you will lose your Tamron model numbers...

Floessie:
@Thanatomanic That shouldn't be a blocker. :wink: I had the slight feeling that @heckflosse considers the missing multi-frame support as a no-go, that's why the `metadata-exiv2` came to nothing (yet).

heckflosse:
@Floessie 
>  I had the slight feeling that @heckflosse considers the missing multi-frame support as a no-go, that's why the metadata-exiv2 came to nothing (yet).

I would not say, it's a no-go, but a regression.

PR #5894, opened 2020-08-28
Wavelet levels: denoise and guided filter with hue curve and local contrast

Desmis:
This pull-request developed with 3 team members consist of

For wavelet levels:
* "denoise luminance" : addition of a hue curve, and take into account "local contrast" to differentiate the action between solids and structures
* "final touchup" : addition of "Guided filter" with 2 features added : curve Hue, and "Local contrast"


Desmis:
If nobody objects I will merge tomorrow :)

jacques

ff2000:
Just wondering: there are commits "merge with Dev" and "merge with DEV" - the correct development branch is called " dev". Do you have "Dev" and "DEV" locally? Can this cause issues when merging?

Desmis:
@ff2000 
It doesn't matter, it's just a label. I do this a bit for fun  :)


jacques

PR #5896, opened 2020-08-29
Fixed bug in Linux mask Blur denoise - rationalise code

Desmis:


Thanatomanic:
Hi @Desmis , just out of curiosity, what was the bug and why did it only affect Linux builds?

Desmis:
This bug was found by  Wayne Sutton...In windows all works fine (I don't have Linux).

I spent a lot of time with Wayne to find out what was wrong. 
it concerned the masks "Denoise" / "Blur" with random chaotic colors....Now it's fixed :)


Jacques

Desmis:
If nobody objects, I will merge this afternoon :)

jacques

Issue #5899, opened 2020-08-31
Raw histogram very different and less insightful than e.g. RawDigger


Thanatomanic:
Related thread: https://discuss.pixls.us/t/puzzling-rawtherapee-raw-histogram/20006

As Alberto mentions, there is some scaling going on which may not be necessary at all. This is issue is here to revisit the (raw) histogram plotting.

Thanatomanic:
The log-transformation of the histogram happens after the binning of the pixels. It should be the other way around for a properly detailed histogram on the left-hand side.

There is a difficulty with implementing that, because click-drag currently changes the compression factor for the histogram (i.e. the actual transformation that needs to be applied). This means that every change would require a complete recalculation of the histogram, which will probably be slow. The best (only?) alternative seems to be to always _calculate_ a log-histogram in base 2, and still _show_ it with a modifiable compression.

heckflosse:
Isn't the real problem that currently the raw histogram has only 256 bins?

Thanatomanic:
@heckflosse I don't think so. Here's a comparison of `amsterdam.pef` with three histograms, all with 256 bins. First linear-linear, then log2-linear (first transformed, then binned), then log2-linear (first binned, then transformed - current RT situation).

![image](https://user-images.githubusercontent.com/6567747/91970175-8b943c80-ed17-11ea-81b0-54098eaf1949.png)

Clearly the second option is better than the third.

heckflosse:
@Thanatomanic I meant the following:

Why do we squeeze the data into 256 bins a priori (in getRAWHistogram()) at all? Wouldn't it be simpler and clearer to use 65536 bins and pass these to the presentation layer, which then could do with this values what it needs to do?

Thanatomanic:
Ah, that would probably work. Should be plenty of bins for floating point images as well.

heckflosse:
@Thanatomanic Here's a float raw in [0;1] range for testing
[D700_20111013_8870-8872.zip](https://github.com/Beep6581/RawTherapee/files/5162091/D700_20111013_8870-8872.zip)

Thanatomanic:
@heckflosse Thanks, but I'm afraid my toolbox doesn't support floating point dng's, so I cannot analyse it 🙃

heckflosse:
@Thanatomanic It was meant for testing your upcoming improvements to RT histogram ;-)

heckflosse:
[...]

PR #5904, opened 2020-09-07
Improvements to change main histogram to 65535 bins

Thanatomanic:
Following the discussion here (#5899), I made a branch `histogram-improved` to work on implementing a more accurate main histogram. The initial commit simply upscales the number of bins naively and spreads out values over all bins. The three main consequences of this:
- The RAW histogram now looks much more like the one in RawDigger 👍 
- Viewing the histogram is dead slow 👎 
- There are visual artifacts due to many empty bins 👎 

About these visual artifacts. For example, the histogram of the working image is constructed from an 8-bit `Image8` source. I now simply spread out the 265 pixel values over all 65536 bins, so many bins remain empty. This leaves a lot of empty room between lines. I already changed the drawing mode to single vertical lines, instead of connected lines, but I am still thinking about how this could be improved. 

Expect more changes in the upcoming week(s). Suggestions are highly appreciated.


Thanatomanic:
Sample shot of RAW histogram:
image

Same shot, but now the log-linear regular histogram:
image

Lawrence37:
The new histogram looks much more useful.

One think I immediately noticed is that the green histogram covers the red, and the blue covers red and green. Would it be possible to compose the individual histograms to make all of them more visible?

Thanatomanic:
@Lawrence37 What do you mean exactly by 'to compose'? Put the separate RGB histograms next to each other, or on top of each other, instead of overlapping? I'm not sure that would be very useful considering the relatively small scale of the histogram...
One thing that I could do is to reduce the opacity of the lines, but then you get mix colors as well.

This is one of the points that are currently less than optimal. It will probably only really improve if there is a better way to connect the points, instead of drawing lines from the axis. For that I need a way to skip empty bins or something smarter, like dynamic binning depending on the width of the histogram.

afr-e:
An alternative to opacity is to introduce a pattern instead of a solid fill.

Lawrence37:
I mean blending the histograms. darktable, for example, uses additive blending. Some people find it confusing though. Using the old line style avoids the issue, but I have a feeling that it's much harder to implement than it sounds.

Thanatomanic:
@afr-e What kind of pattern would you propose?

Side-note: currently, drawing 65536 lines per histogram is slow, but I believe that drawing a polygon and filling it, is virtually unworkable (the L channel used to use a filling mode, but it took 20 seconds when you enabled it, and then it didn't draw anything).

Thanatomanic:
@Lawrence37 I experimented once with additive blending [here](https://github.com/Beep6581/RawTherapee/pull/4618), but there was some opposition.

heckflosse:
> Viewing the histogram is dead slow

I am willing to help on this

heckflosse:
@Thanatomanic Can you point me to the code which causes the dead slowdown?

rtgui/histogrampanel.cc
```diff
@@ -869,105 +869,72 @@ void HistogramArea::updateBackBuffer ()
...
-
-        unsigned int histheight = 0;
+        unsigned int rgbmax = 0;
+        unsigned int lumamax = 0;
+        unsigned int chromamax = 0;
 
-        for (int i = 1; i < 255; i++) {
-            if (needLuma && lhisttemp[i] > histheight) {
-                histheight = lhisttemp[i];
+        for (int i = 1; i < 65535; i++) { // Values at far left and right end are handled differently
```
Thanatomanic: @heckflosse This loop increased in size quite by a factor of 255, which might be the slowdown.

rtgui/histogrampanel.cc
```diff
@@ -869,105 +869,72 @@ void HistogramArea::updateBackBuffer ()
         LUTu& gh = rawMode ? ghistRaw : ghist;
         LUTu& bh = rawMode ? bhistRaw : bhist;
 
-        // make double copies of LUT, one for faster access, another one to scale down the raw histos
```
Thanatomanic, line 1324: @heckflosse I didn't see the need for these additional arrays, but maybe they are really useful in getting a speedup? At least the intended 'scaling' is no longer necessary because of my alternative histogram scaling.

rtgui/histogrampanel.cc
```diff
@@ -994,10 +961,10 @@ void HistogramArea::drawCurve(Cairo::RefPtr<Cairo::Context> &cr,
     double s = RTScalable::getScale();
 
     cr->set_line_width(s);
-    cr->move_to (padding, vsize - 1);
+    //cr->move_to (padding, vsize - 1);
     scale = scale <= 0.0 ? 0.001 : scale; // avoid division by zero and negative values
 
-    for (int i = 0; i < 256; i++) {
+    for (int i = 0; i < 65536; i++) {
```
Thanatomanic: @heckflosse This could be an additional bottleneck, but since it's related to GUI drawing, it's probably not parallellizable.

Lawrence37: If each line averages 50 pixels long, that's 65536 x 50 x 3 = about 10 million pixels of line for the raw histogram. That's a lot of line...

jackho9:
Excellent Roel, I like the top log-log raw histogram, I can clearly see if/how much specular clipping there is (a little bit), where reflected exposure actually begins (less than a stop from clipping) and if/how many blocked pixels there are (a few).

What are the units of the two axes? Ideally gradation every stop, with 1/3EV just in the highlights, some indication of where mid-gray could be (EV0, -3EV from clipping?).  I personally do not mind a toothy histogram, though you could widen the bars in the deep shadows if you felt that was more readable.  There is a lot of information there so readability is important even when the histogram is small, I find it a bit hard to see the blocked/clipped histogram 'bars' against the dark background.  

One of the linearX-logY histogram's main uses for me is to nail down how close To The Right reflected Exposure ended up, so it would be useful to have gradation every thousand or two on the X axis.   I personally practically never use the linear-linear histogram view.

As for vertical log axis units, If we take something of the order of 0.01% as a typical acceptable number of clipped/blocked pixels, the axis should show clearly whether a few, a few hundred, thousand or tens of thousand pixels are clipped/blocked.  I see that RD shows gradation at 1, 10, 50, 200, 1k, 5k, 20k, 100k, 500k etc., not bad.

As for speed, the raw histogram(s) needs to be computed once only, and it doesn't necessarily need to be the first thing to come up.  Anyways, just one user's usage, I hope others will contribute theirs.  Thanks a lot for doing this!

Jack

rtgui/histogrampanel.cc
heckflosse: I placed a `BENCHFUN` in `HistogramArea::drawCurve`
Each time I change for example the exposure adjuster I get 3 timings of ~28 ms, means changing exposure adjuster needs 3 * 28 ms to update the histogram.

Questions:
1) why 3 timings instead of one? EDIT: Ok, 3 colors ...
2) I was in RAW histogram mode. Why is the histogram redrawn when I change the exposure adjuster?

I made a quick hack (still without using SSE), which reduces the time to 3 * 7 ms. Will try with SSE now...

Thanatomanic: @heckflosse Feel free to push to this branch whenever you feel it's useful!

heckflosse:
@Thanatomanic 
Here's the `quick hack` patch (still without using vectorized xlogf)

```diff
diff --git a/rtgui/histogrampanel.cc b/rtgui/histogrampanel.cc
index 6331e8364..77d13670d 100644
--- a/rtgui/histogrampanel.cc
+++ b/rtgui/histogrampanel.cc
@@ -25,6 +25,9 @@
 #include "../rtengine/LUT.h"
 #include "rtimage.h"
 #include "../rtengine/color.h"
+#include "../rtengine/sleef.h"
+#define BENCHMARK
+#include "../rtengine/StopWatch.h"
 
 using namespace rtengine;
 
@@ -365,10 +368,16 @@ void HistogramPanel::toggleButtonMode ()
 //
 //
 // HistogramScaling
-double HistogramScaling::log(double vsize, double val)
+float HistogramScaling::log(float vsize, float val) const
 {
     //double factor = 10.0; // can be tuned if necessary - higher is flatter curve
-    return vsize * std::log(factor / (factor + val)) / std::log(factor / (factor + vsize));
+    return vsize * xlogf(factor / (factor + val)) / xlogf(factor / (factor + vsize));
+}
+
+float HistogramScaling::log_test(float vsize, float val, float mult) const
+{
+    //double factor = 10.0; // can be tuned if necessary - higher is flatter curve
+    return vsize * xlogf(factor / (factor + val)) * mult;
 }
 
 //
@@ -958,22 +967,25 @@ void HistogramArea::on_realize ()
 void HistogramArea::drawCurve(Cairo::RefPtr<Cairo::Context> &cr,
                               const LUTu & data, double scale, int hsize, int vsize)
 {
+    BENCHFUNMICRO
     double s = RTScalable::getScale();
 
     cr->set_line_width(s);
     //cr->move_to (padding, vsize - 1);
     scale = scale <= 0.0 ? 0.001 : scale; // avoid division by zero and negative values
 
+    const float temp1 = 1.f / xlogf(HistogramScaling::factor / (HistogramScaling::factor + vsize));
+    const float temp2 = 1.f / xlogf(HistogramScaling::factor / (HistogramScaling::factor + 65535.0));
     for (int i = 0; i < 65536; i++) {
         double val = data[i] * (double)vsize / scale;
 
         if (drawMode > 0) { // scale y for single and double log-scale
-            val = HistogramScaling::log ((double)vsize, val);
+            val = HistogramScaling::log_test (vsize, val, temp1);
         }
 
         double iscaled = i;
         if (drawMode == 2) { // scale x for double log-scale
-            iscaled = HistogramScaling::log (65535.0, (double)i);
+            iscaled = HistogramScaling::log_test (65535.f, i, temp2);
         }
 
         double posX = padding + iscaled * (hsize - padding * 2.0) / 65535.0;
diff --git a/rtgui/histogrampanel.h b/rtgui/histogrampanel.h
index 740b0a12c..497a0c3aa 100644
--- a/rtgui/histogrampanel.h
+++ b/rtgui/histogrampanel.h
@@ -49,9 +49,10 @@ struct HistogramRGBAreaIdleHelper {
 class HistogramScaling
 {
 public:
-    double factor;
-    HistogramScaling() : factor(10.0) {}
-    double log (double vsize, double val);
+    float factor;
+    HistogramScaling() : factor(10.f) {}
+    float log (float vsize, float val) const;
+    float log_test (float vsize, float val, float mult) const;
 };
 
 class HistogramRGBArea final : public Gtk::DrawingArea, public BackBuffer, private HistogramScaling, public rtengine::NonCopyable

```

heckflosse:
@Thanatomanic 
I pushed the speedup for log-Histogram. Currently I kept the timing code. I will remove it later, when the SSE speedup is done.

afr-e:
> @afr-e What kind of pattern would you propose?

I was thinking of the classic Morse code for lines, and stippling and hatching for area under the curve, a different pattern for each channel. However, care must be taken so as not to disturb or distract the user.
___
General thoughts

Histograms are meant to help the user glean as much useful info as possible. In my opinion, that can (and often should) mean that you don't necessarily have to represent the data exactly; i.e. it should be represented symbolically and emphasize important elements. E.g. the outlier lines are not only difficult to read because of the dark background as @jackho9 noted but because they are thin. Simply putting a dot on top of those lines and highlighting them (persistently or on hover) would make them more apparent and recognizable.

I tend to be brimming with ideas, so feel free to ask me for more feedback. 😉

heckflosse:
@Thanatomanic With the SSE code for the log mode I'm now at ~5ms per color from which ~4.5ms are for drawing the lines. I think, that's fine now.

Interestingly, latest clang auto vectorizes the two loops, but latest gcc does not...

Thanatomanic:
@heckflosse Thank you, these speed ups make the performance return to how it was before the change. Only when you show the RAW histogram for huge files (100 MP), and drag to change the factor, you have a very slow response. But that's perfectly fine for me.

heckflosse:
@Thanatomanic 
> Only when you show the RAW histogram for huge files (100 MP), and drag to change the factor, you have a very slow response.

I do not understand why the MP count of a file has an influence on the response time when dragging.
Edit: But I can confirm the behaviour...

Can you elaborate on this?

jackho9:
> @heckflosse Thank you, these speed ups make the performance return to how it was before the change. Only when you show the RAW histogram for huge files (100 MP), and drag to change the factor, you have a very slow response. But that's perfectly fine for me.

Yes, thank you.  Question: is the factor necessary?

Thanatomanic:
@heckflosse I've timed both `drawCurve` and `updateBackBuffer`. The first remains mostly constant no matter what the file and hovers at 3-5 ms. This makes sense, since it needs to process the same number of bins.
The second function is different. It should probably take at least 3×5 ms because of its calls to `drawCurve`, but actually it takes anything between 40 ms an 500 ms depending on the bit-depth and resolution of the image. I see no obvious clue as to why...
I'll think about this.

@jackho9 I think I implemented the flexible factor a few years back to have some form of 'zoom' functionality in the histogram and allow you to choose which area to look for details. Maybe that is no longer necessary because the histogram now actually renders as it should and more details are visible anyhow.

As for the axes: the x-axis divisions need some work, fully agreed on that. I'll see what I can do for a proper EV scale.
The y-axis is simply a linear division into equal parts. If you see three lines, they are at 25%, 50% and 75%. If you see 15 lines, they are at 6,25%, 12,5%, etc.  I don't intend to make this any more complicated, unless there is a very good reason to.
I think that checking for clipping for 0,01% is pretty arbitrary, but maybe it would be cool to have a movable line so you can set clipping levels yourself? (wild idea.. )

The more pressing matter for me is to get a nicer non-raw histogram that looks smoother than it does now. And I just got an idea on how to do that 😄

afr-e:
One thing that I suggested for PhotoFlow that may be applicable here is that it would be nice to have labels for the axes and lines.

> I think that checking for clipping for 0,01% is pretty arbitrary, but maybe it would be cool to have a movable line so you can set clipping levels yourself? (wild idea.. )

Not wild. At the very least, if there is an indication of the clipping on the graph, I would be happy.

Thanatomanic:
@heckflosse I think that `drawCurve` does not actually draw the curve, but only prepares the lines to draw. The calls to `cr->stroke()` in `updateBackBuffer` are the actual Cairo command to start drawing. This explains the timing difference between the two functions.

It does not explain why the drawing scales with the resolution of the image. What does factor in, is what @Lawrence37 mentioned before:

> If each line averages 50 pixels long, that's 65536 x 50 x 3 = about 10 million pixels of line for the raw histogram. That's a lot of line...

So, the more the lines need drawing, the slower it gets. A histogram with a lot of variation would be slower than a flat one. That still doesn't quite explain why the time scales with megapixels...

afr-e:
Would it have to do with the calculations that generate the histogram?

Thanatomanic:
@afr-e You would think so. Unfortunately, the bins for the raw histogram are filled once, when you open the file. The non-raw bins are only filled whenever you change something in the processing parameters. Modifying the look of the histogram itself does not cause a recalculation afaik - ~~I can check quickly~~. Edit: No it really doesn't trigger a full recalculation.

afr-e:
I am just thinking that it might be redoing or re-accessing something.

heckflosse:
@Thanatomanic 
> I see no obvious clue as to why...
I'll think about this.

:+1:  Also thinking about this

Thanatomanic:
@heckflosse I pushed a change. Now it avoids drawing lines that are zero. For me this gives a nice speed boost, but still the 100 MP 16 bit files are a bit slow. It's all down to the performance of Cairo I think...

heckflosse:
@Thanatomanic Great! I will use the same condition to further reduce the log calculations

TechXavAL:
May I suggest a new feature in the histogram? As it is going to give more information, maybe you could add some sort of visual indication about where are the min and max values of the curve. That is, where are located the min and max bins in the graph. And perhaps where is located EV0.

Take this image as an example:

histmaxmin

The leftmost small grey triangle on the lower part of the histogram would mean where is located the absolute minimum bin (in this case at 0, meaning it's clipped. Or perhaps it could indicate where falls the Threshold for clipped shadows as set in Preferences.

The center triangle would mean where is EV0 (in this image I have placed it randomly), and a vertical grey bar would be a visual aid.

The rightmost triangle (and grey bar) would be the absolute max of bins, or the Threshold for clipped highlights. If in the image it was the absolute max bin, then it would mean that such image would have 1/4-1/5 of a stop of room before clipping.

And the left and right triangles may even change in color when they reach the clipping point (e.g., they may become white).

Just another suggestion, though.

afr-e:
When I asked to be able to see the clipping and middle, PhotoFlow's answer was to provide lines and toggles. Its histograms are still sparse and have much to be desired but useful as a working example.

image

image

I know this is my second reference to PhotoFlow. I throw most of my feature requests @aferrero2707 because RT folks are so busy. But many of my requests are relevant to RT as well.

Thanatomanic:
@TechXavAL and @afr-e Good suggestions, and should be easy to implement. I was under the impression that @jackho9  actually wanted a horizontal mark or line or reference to have a limit on which fraction of pixels are clipped (anything below x% is worth ignoring).

@afr-e  What does the 'clip range' button do in the second screenshot?

heckflosse:
@Thanatomanic Can I commit a further small speedup for `drawCurve` ?

Thanatomanic:
@heckflosse Go ahead!

heckflosse:
@Thanatomanic 

I thought about another condition. In case you want to try it, just remove the `// remove me` strings in `drawCurve`.
For a 100 MP IIQ file, the condition gives another 10% speedup for `updateBackBuffer`

Thanatomanic:
@heckflosse I think I understand you change: am I correct in saying that this condition makes sure that only lines longer than 1px will be drawn?

@afr-e @TechXavAL I implemented your suggestion of min/max levels with a very simple light grey line. Can you test this if possible? There seems to be something off with the coordinates somewhere, because sometimes some histogram bins are drawn before the grey line. That will need some investigation.

Side-note: I don't have much time to work on this PR until the beginning of next week.

heckflosse:
@Thanatomanic 
> am I correct in saying that this condition makes sure that only lines longer than 1px will be drawn?

exactly.

Edit: longer ore equal to 1px

heckflosse:
Just for the record: With the raw I used for the intial testing of this branch, processing time for `drawCurve()`  for one color channel was ~28 ms. Now it's ~0.33ms for the same file.

TechXavAL:
@Thanatomanic 

The min/max lines are there but if you have an image with both shadows and highlights clipped, then they merge with the side of the graph. It would be good that when the line gets to the limits of the graph, there would be some visual indication telling us that there's some clipping going on.

And yes, there are bins placed outside the min/max limits. In fact the chromaticity histogram seems to be completely to the left of the min line.

Anyway, it seems to be feasible. :+1:  Thanks for taking the time to implement it.

Thanatomanic:
@techxaval I specifically left out the chromaticity and luminosity graphs, because they measure something on a different scale than RGB pixel levels. But even some RGB lines are outside the clipping lines, and I'm not sure yet why.

As for the clipping at 0 and max, there are already indicators for that. The little squares at the top of the histogram.
Currently however, I think they don't work on the upper end of the non-raw histogram because of the way the histogram is constructed - bit too complex to type from mobile. Happy to elaborate at a a later stage.

heckflosse:
@Thanatomanic 

> am I correct in saying that this condition makes sure that only lines longer than 1px will be drawn?

With that condition the drawing is not the same as before. But using a threshold of 0.5px gives same drawing in my tests while still giving a small speedup.

TechXavAL:
@Thanatomanic 

> As for the clipping at 0 and max, there are already indicators for that

Didn't knew that! So the little squares show when each channel is clipping. Interesting (and undocumented, I think).

afr-e:
@Thanatomanic 
> What does the 'clip range' button do in the second screenshot?

Checked, we don't see clipped data; e.g. counts of -ve values. Unchecked, we see the great beyond.

TechXavAL:
I wonder if adding a warning icon on top of those colored squares would be useful. At least it would be more descriptive.

The warning icon could be something like this:
warning_8px

Which would translate into a histogram like this:
clipping_warning
It seems to me that this way you will instinctively know that something is happening there.

Desmis:
@Thanatomanic 

I can only approve of this kind of initiative. Until then, to look at the histogram in detail, especially in low light, I used "Histogrammar" by Guillermo Luijk (2011, but no longer updated)

:+1: 

jacques

Desmis:
Example "Histogrammar"

histogrammar

jacques

PhilipClarke:
There is an argument for the removal of the vertical and horizontal lines from rtgui/histogrampanel.cc lines 832 to 864 (newbie - no idea how to cross-reference the code). 

Representing "8 stops" appears to be slightly misleading, as are discussions about EV values and clipping unless the user is going to map their input and output. I'm not explaining this well, but I arrived after seeing a YouTuber counting "stops" on the histogram. 

The curves graphs throughout RT have a grid divided by ten probably stemming from Ansel Adams, so there's a mismatch with the majority of representations in RT.

Photoshop, lightroom, Darktable have no lines. Gimp has five vertical lines.

heckflosse:
@Thanatomanic I just had a look at current version of ART and detected this tooltip, which I personally find quite useful
grafik

Would be nice to have something similar for the raw histo in RT.

afr-e:
I would love an overlay toggle (with obviously less info for it to fit).

TechXavAL:
It seems that in commit `dd95f6b43` if you set exposure all the way down or all the way up to underexpose/burn the image, the colored squares warning about clipping don't appear in histogram.

If you play with the black slider and underexpose the image, the colored squares do appear.

Thanatomanic:
@TechXavAL Hi Xavier! I've started working on this today so expect some more changes coming in the next few days!
As for the clipping squares: the lower ones should definitely appear, because values of 0 should trigger them. However, the upper ones will probably not show (for non-RAW histogram), because of the way the values are currently binned. The non-RAW histogram data is 8-bit, but maps to a 16-bit array. This currently leaves the 65535th value empty so it will not trigger the clipping indicator. I still need a smart solution for that...

Edit: I see your comment was more subtle. Only when the lowest bin contains more values than all other bins, the indicator appears. In other words: only when the bin exceeds the plot range, it shows the square. Apparently due to the way the Exposure slider works, this does not happen when underexposing.

Thanatomanic:
@Floessie @heckflosse If you have some time, could you point me to a 'best practice' way to expose the bitdepth of the image to the `HistogramArea` class?

TechXavAL:
> This currently leaves the 65535th value empty so it will not trigger the clipping indicator

Is there any way that the indicator will show when values reach 65534?

> In other words: only when the bin exceeds the plot range, it shows the square. Apparently due to the way the Exposure slider works, this does not happen when underexposing.

I'm afraid I'm not really following you. But let me show you an example about something I think is not desirable. This is a black feather I shot for testing purposes:

feather_test

After setting the _Threshold for clipped shadows_ at _0_, lowering _Exposure_ to _-5_ and _Lightness_ to _-43_, the leftmost bin is almost at the top of the histogram, meaning there are quite a lot of clipped shadows (or more precisely, shadows with values at _0_), BUT the histogram doesn't show clip warnings. If I go ahead and set _Lightness_ to e.g. _-45_, then the little squares do appear, but the bin bar must reach the top of the graph to show them (I think this is the behavior you were referring to).

Notice that the _Clipped shadow indication_ does show lots of clipped areas:

feather_test_clipblk

To me, not showing a clipping alert when the first pixel gets to _0_ is a mistake, or bug, or however it should be called. We can even argue about the user being able to set a minimum number of clipped pixels to show a warning, but not so many that they have to pile up until they reach the top of the graph.

Now I must admit that those images were screen-grabbed from dev branch. This is how it looks in histogram-improved branch (exactly the same settings as before):

feather_test_clipblk_HstImpd

As you can see the histogram is not behaving correctly.

A different image showing that half the histogram graph is not used/plotted (it shows clipping, but no bins in the left half of the histogram):

crop_test_clipblk_HstImpd

And finally, I've made a quick test with the feather image, using the Local Contrast tool, and even though there is a nice amount of  pixels clipped in their green channel (and some amount in the blue channel, too), the histogram doesn't show the warning. If I go ahead and set the _Darkness level_ to _1.03_, then the green channel clip warning does show.

Note that all sliders in the Exposure tool where reset to default value.

feather_test_clipgreen_HstImpd

Beep6581:
Hey @Thanatomanic 

If this PR is not ready for merging, could you convert to "draft" please?

Thanatomanic:
This makes sense, thanks @Beep6581 

>@Floessie @heckflosse If you have some time, could you point me to a 'best practice' way to expose the bitdepth of the image to the HistogramArea class?

I still hope for someone who can help me out with this. It is important for the histogram class to know about the actual bitdepth of the image because it changes how the data needs to be visualized.

heckflosse:
@Thanatomanic I would start here
https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/toolpanelcoord.cc#L344

and look where it is called from

Thanatomanic:
@heckflosse First step fixed I believe. I'm quite happy I figured this out on my own eventually, because there is so much going on with derived classes and object types going around that I found it quite difficult to pinpoint where the changes were needed... This exposes the `tiff_bps` property from `dcraw` nicely. It is 0 when we're dealing with a non-raw file, so that still needs fixing to handle higher bitdepth non-raw images.
Next step is to elegantly forward this to the histogram class.

```diff
diff --git a/rtengine/imagesource.h b/rtengine/imagesource.h
index 25c024ed2..d92fd56bf 100644
--- a/rtengine/imagesource.h
+++ b/rtengine/imagesource.h
@@ -123,6 +123,10 @@ public:
     virtual void        WBauto(double &tempref, double &greenref, array2D<float> &redloc, array2D<float> &greenloc, array2D<float> &blueloc, int bfw, int bfh, double &avg_rm, double &avg_gm, double &avg_bm, double &tempitc, double &greenitc, float &studgood, bool &twotimes, const procparams::WBParams & wbpar, int begx, int begy, int yEn, int xEn, int cx, int cy, const procparams::ColorManagementParams &cmp, const procparams::RAWParams &raw) = 0;
     virtual void        getrgbloc(int begx, int begy, int yEn, int xEn, int cx, int cy, int bf_h, int bf_w) = 0;
 
+    virtual unsigned int getBitdepth() const
+    {
+        return 0;
+    }
     virtual double      getDefGain  () const
     {
         return 1.0;
diff --git a/rtengine/rawimage.h b/rtengine/rawimage.h
index 871267dac..41b32f960 100644
--- a/rtengine/rawimage.h
+++ b/rtengine/rawimage.h
@@ -109,6 +109,11 @@ public:
         return fuji_width;
     }
 
+    unsigned int get_bitdepth() const
+    {
+        return tiff_bps;
+    }
+
     float const * get_FloatRawImage() const
     {
         return float_raw_image;
diff --git a/rtengine/rawimagesource.cc b/rtengine/rawimagesource.cc
index 7ae919e0a..9c1dd496f 100644
--- a/rtengine/rawimagesource.cc
+++ b/rtengine/rawimagesource.cc
@@ -6091,6 +6091,11 @@ ColorTemp RawImageSource::getSpotWB (std::vector<Coord2D> &red, std::vector<Coor
     }
 }
 
+unsigned int RawImageSource::getBitdepth() const
+{
+    return ri->get_bitdepth();
+}
+
 void RawImageSource::transformPosition (int x, int y, int tran, int& ttx, int& tty)
 {
 
diff --git a/rtengine/rawimagesource.h b/rtengine/rawimagesource.h
index 16677b1da..064e76da2 100644
--- a/rtengine/rawimagesource.h
+++ b/rtengine/rawimagesource.h
@@ -158,6 +158,7 @@ public:
     {
         return rawData;
     }
+    unsigned int getBitdepth() const override;
 
     double      getDefGain  () const override
     {
```

heckflosse:
> It is 0 when we're deadline with a non-raw file, so that still needs fixing to handle higher bitdepth non-raw images.

ImageIO::getTIFFSampleFormat

Thanatomanic:
@heckflosse @Lawrence37 @afr-e @Desmis @Beep6581 Finally I have made good progress today, up to the point that I believe this is ready for finetuning, testing and a merge. Your feedback would be much appreciated, in particular on whether you like the look or how it could be improved.

Biggest changes since the last time I worked on this, is that the histogram now knows what the bit depth of the raw image is. The gridlines and histogram are adjusted accordingly. You can easily check the difference by comparing raw images with e.g. 12 bit and 16 bit. The normal histogram is always 8 bit and is now scaled properly again (it was stretched into 65536 bins before...).
You will see three vertical white lines: one for the lowest value, one for the ([arbitrary](https://discuss.pixls.us/t/ev0-in-raw-histograms/25415/5)) 0EV line and one for the highest value.

Thanks for taking a look!

Thanatomanic:
I just spotted that the RGB readout bar is not working in all cases. I'll take a look in a few days when I have some time again.

Lawrence37:
Some observations:
- Horizontal lines in the histograms, RGB parade, and waveform are gone.
- Switching to a vectorscope with no image open causes a crash.
- The fact that blue hides green and green hides red makes the histogram harder to use. We've discussed this before and I think it is important to resolve before merging.
- I agree the 0 EV line is arbitrary. What is the goal of this line? Is it for middle grey? If so, do we use 12%, 12.5% (-3 EV), 18%, or 18.4% (Lab L=50)?

Thanatomanic:
@Lawrence37
* The RGB bar readout should be working now
* The horizontal gridlines should be back, but need work. They don't always work correctly depending on the mode. TODO
* ~~I'll investigate the crash, not sure what caused it...~~ Inadvertently fixed below
* ~~The scopes scale weirdly compared to `dev`. Try enlaring the histogram to see for yourself.~~ Fixed below
* Any suggestion for the visual? For the regular histogram, I could reintroduce the lineplot instead of the filled plot.
* A middle-grey line was suggested before, I took the RawDigger approach and put it at -3 stops from maximum. I would actually prefer the L = 50% approach and put it at 18,42% of the maximum raw value. But opinions are welcome.

Lawrence37:
@Thanatomanic
Was there a reason for not using the old style lines for the histograms? I though I read an argument against it but I couldn't find it. 
It occurred to me that the middle gray line for the normal histogram is not correct because there is a hard-coded sRGB gamma. The correct placement should be closer to the middle using a linear x-axis. I also think using L=50 is a reasonable choice.

Thanatomanic:
@Lawrence37 The primary reason for not using the lines was that prior to my bitdepth changes, there could be _a lot_ of empty bins, making the line go up and down a lot and messing up the visualization. I will try out how it will look now.
Another, more fundamental, argument is that connecting the dots implies something is known about the values in between points. For the regular histogram this is fine, for the RAW histogram it is not.

Good point about the sRGB middle gray.

Lawrence37:
Another idea is to only plot points. The down side is that they will be hard to see on the left side because they are spaced apart.

Thanatomanic:
I have tried to re-introduce the old line mode and it works well for the normal histogram, but for the raw histogram things look pretty terrible:

image

The problem is that the _really raw_ sensor data are rescaled according to the black and white levels. Because the histogram has integer bins, this leads to quantization errors. I see three possible ways out of this:
1. We provide the _really raw_ data in the histogram without scaling. Downside is that you would need to understand black and white levels in order to interpret the histogram.
2. We do the scaling and adapt the histogram width to the actual width of the data (instead of mapping the data to the available width). This would give you a histogram that grows and shrinks whenever you play with the raw black and white level controls. Getting the gridlines on the right locations is going to be tricky.
3. We leave things as they are, but don't use the old line graph mode, but the newer 'bar chart' mode.

I think the 2nd option is the nicest, but also the most work (I think). The 1st option is not very useful, except to analyse the actual sensor data. The 3rd option is easy to implement but the quantization errors remain.

Thoughts? Perhaps I should pose this question on Pixls for some wider feedback.

Lawrence37:
I don't quite understand what the second option means so I apologize if this is the same as your idea. My approach would be this:
Bin the raw data after black point adjustment. Draw the histogram such that the white point touches the right side. The gridlines stay where they are now.

Thanatomanic:
So, let's assume you have a 16 bit file, with a black point of 512 and a white point of 65400. Currently we subtract 512 from all sensor values and then rescale so that (65400 - 512) becomes 65535.
Edit: reworded parts...

How would your approach differ?

Lawrence37:
After black point subtraction, the binning occurs. There are 64889 bins that may have values (bins number 0 to 64888) so we plot a line using 64889 points. On the linear histogram, bin 0 is on the left edge and bin 64888 is on the right edge. On the log histogram, bin 64888 is on the right edge. Bin 32444 is on the first EV line from the right.

Beep6581:
> The 1st option is not very useful, except to analyse the actual sensor data

That makes it very useful to me. I need to see two things in my work: 
1. the raw histogram, to help find the right values for `camconst.json`, 
2. and the final histogram (end of pipeline).
3. Additionally, anyone who uses the raw black and white point correction sliders would probably want to see a histogram of the data just after those corrections. Maybe this does not need a histogram of its own though, because maybe its enough to apply the neutral profile and then to do those corrections while viewing the final histogram.

Lawrence37:
Converting this to a draft because the way the histogram is drawn is up for discussion.

PR #5911, opened 2020-09-19
Local adjustments - Various improvment to GUI - labels - tooltip

Desmis:
@heckflosse 

Ingo

I create this PR to notify you of changes.

This PR aims to make labels and tooltip compatible with the interface (and vice versa), as well as with Rawpedia

I would like to thank Wayne Sutton for a cooperative work on Labels, Tooltip and GUI for several weeks which leads to a simplification of the interface in particular the creation of the "basic" level, simpler and more explicit labels and tooltips.

There are no (or few) interactions with iplocallab.cc, otherwise the GUI changes (combobox replace by a checkbox)

On the other hand that will bring some additional mismatches of merge with the branch you work (denoise - localgflc)

If you don't object (or someone else) I will merge tonight or tomorrow morning...

jacques


Pandagrapher:
@Desmis 
In `locallabtool.cc` and `locallabtool2.cc`, in function `updateAdviceTooltips`, tooltips of all widgets shall be set to empty string in the else case (otherwise tooltips won’t be removed according to user preference).

Pierre

Desmis:
@Pandagrapher 

Changes done, thank you :)

jacques

PR #5926, opened 2020-10-03
Fix wrong minimum calculation in ImProcFunctions::Aver()

heckflosse:
This fixes a wrong minimum calculation (probably introduced by me long time ago) in the function `ImProcFunctions::Aver()`
The old code never gave a min > 0
@Desmis Jacques, please confirm


heckflosse:
Scratch that. That was nonsense...

Desmis:
OK :)

Jacques

PR #5929, opened 2020-10-03
GUI changes: highlight enabled modules, RawTherapee theme CSS clean-up, minor tweaks

Thanatomanic:
After the recent bout of comments on the 'bad' GUI for RawTherapee, in particular the fact that people are very confused that modules need to be enabled, I have spent a good part of the day fiddling with this issue. I came up with a solution where active modules are highlighted. This reinforces the visual feedback that modules might not be active yet.
While diving into the CSS and GUI code, I made several other (minor) changes. In particular I found that the CSS was quite a spaghetti. I think I have simultaneously simplified the styling process _and_ improved the look of several elements (padding and margins mostly, some coloring). 

Let me know what you think!

**Active modules are now highlighted:**
image

**No unnecessary frames for the Navigator, History and Snapshots:**
image

**Better alignment of labels in frames:**
image

**Less cramped interface with nested elements, and more distinct shading and padding of these elements:**
image


RichardRegal:
Certainly helps. I think it would be necessary to make sure that the tools that are active automatically (for example on the exposure tab the exposure tool) are also highlighted.

I am not sure about using colour as the highlighting though. I like a neutral background to avoid changing the way I interpret the photo in the middle. That is why I would prefer “lowlighting” inactive tools (possibly by not printing them in bold) rather than highlighting the active one.

TechXavAL:
I also don't think many people would like highlighting in blue (or any other color). It wouldn't be a problem to me, though.

Perhaps «lowlighting» (as RichardRegal said) the titles of tools (same as the power icon) may be a good idea. And it may be a good idea too having «turned-off power icons» just a tiny bit smaller than their counterparts, but with the same shape (i.e., not a simple circle).

More suggestions:
* the box containing the controls of each tool may have the top and left borders with a dark gray, while right and bottom with a lighter gray. That should make it look as an engraved box that will make it easier to the eye knowing the extent of the tool
* each box containing controls would benefit of rounded corners of 1-2px (not larger)
* the background of such boxes should have a shade clearly different from its surroundings. And given the amount of nested boxes, some testing may be needed to get the proper shade for each box level (to me, in the screenshots there's not enough separation from one nested block of controls to its surroundings). And perhaps the main box of each tools should have the darkest shadow

Thanatomanic:
@RichardRegal 
> Certainly helps. I think it would be necessary to make sure that the tools that are active automatically (for example on the exposure tab the exposure tool) are also highlighted.

While this is technically possible, I doubt this is very useful. A tool that cannot be turned off has no real need for a highlight that it is active imo. Also, this would light up nearly the whole RAW tab. This effect may be less when I get rid of the blue highlight - that is not decided.

> I like a neutral background to avoid changing the way I interpret the photo in the middle.

I see you point. If things are color critical, keeping things as grey as possible is worthwhile. This is also an approach darktable uses. My main gripe from an UX pov, is that it becomes much, much harder to have clear visual cues on what is going on in your interface. Perhaps we can make a grey version of the main theme, similar to the two versions of TooWaBoo's theme.

@TechXavAL If I get your meaning well, all your three suggestions seem to want to go back to the old situation. I specifically went with a 'flatter' approach with fewer edges and more acute corners. Only the 'framed' tools still have a slightly rounded edge. As for the shading: I choose 'more nested = darker', but I will try out 'more nested = lighter' to see if that works better.

Lawrence37:
Could you add a separator between the navigator and history?

afr-e:
I agree with @Lawrence37.

While I don't dislike @Thanatomanic changes, visually the separation isn't as clear among the elements as before and more padding means less info density in smaller (tinier) screens.

That said, cleaner CSS is almost always an improvement.

In my mind an on / off or active / inactive text is way better than highlighting the title bars. Thinking of

image

A child knows that it is a switch meant for a hired mustachioed plumber to hit with their head. Many won't need a translation either.

TechXavAL:
@Thanatomanic 

I think the old situation uses 3px rounded edges. Try with one pixel, so the boxes are square yet, but a tiny bit more polished. It may seem silly, but it works with webpages. Completely square boxes seems to «poke» in the eye. But I know it's just a matter of taste.

About the borders in different shades, they help isolating things that share something in common.

And about the backgrounds: try darkest, lighter, darker. That way it may be easier to tell the difference between areas.

Or you may point to me the lines in the css file and I will edit them for you with my suggestions. But in the end it's you who decide.

RichardRegal:
> > Certainly helps. I think it would be necessary to make sure that the tools that are active automatically (for example on the exposure tab the exposure tool) are also highlighted.
> 
> While this is technically possible, I doubt this is very useful. A tool that cannot be turned off has no real need for a highlight that it is active imo. Also, this would light up nearly the whole RAW tab. This effect may be less when I get rid of the blue highlight - that is not decided.

My reasoning is that doing so makes a clear distinction between tools that are active and tools that are not. It provides a visual cue that you need to do something to activate the tool.

SoupyGit:
I'm not sure you need the highlighting, the line separation you have included between modules is enough. That's a good change. I would also either a) extend those lines to the very edges and remove the grey outside border - given it is the same shade as the module header background, and it extends down to the next module, those two modules become connected - but what you want is the opposite - each different module separated [it would also help the sidebar take up less precious screen real estate], or b) keep the grey outside border, but add a thin black border all the way around each module.

I would change the off symbol from a little circle (which doesn't immediately make clear what it's purpose is) to a more muted version of the on power button. Of lesser importance, I would also remove the little arrow some modules have, and just give everything a power button. 

Removing unnecessary frames around Navigator/History was good, but you at least need a line separation at the bottom of Navigator, before History. And again, as per (a) above, you could remove the thin grey borders on the sides, which IMO needlessly take up space, and make things less clear.

Better alignment of labels in frames was a good change.

I like square borders better, if only because I think the issue becomes redundant if you get rid of the unnecessary grey borders around the outside.

ff2000:
> In my mind an on / off or active / inactive text is way better than highlighting the title bars. Thinking of
> 
> image
> 
> A child knows that it is a switch meant for a hired mustachioed plumber to hit with their head. Many won't need a translation either.

This isn't a computergame, unfortunately.
The issue with simple buttons labeled "ON" and "OFF" in a gui is, that you are used to buttons "turn on", "turn off", so the text tells which action will be performed. Labels that show the state confuse users. gtk.Switch would be a labeled option that works better.
Another option would be a styled checkbox that's visually more pleasing than the default gtk.CheckButton (in case the current icons can't be made easier recognizable)

Thanatomanic:
@SoupyGit 
> I would also either a) extend those lines to the very edges and remove the grey outside border - given it is the same shade as the module header background, and it extends down to the next module, those two modules become connected - but what you want is the opposite - each different module separated [it would also help the sidebar take up less precious screen real estate], or b) keep the grey outside border, but add a thin black border all the way around each module.

The space on the right side is there to make room for the scrollbar whenever it is needed. At the moment, there is no mechanism in place that checks whether the scrollbar should be visible or not, and therefore whether the space could be otherwise used. If you want to reduce the space, please go to Preferences and check 'Hide vertical scrollbar'.

SoupyGit:
@Thanatomanic 

> The space on the right side is there to make room for the scrollbar whenever it is needed. At the moment, there is no mechanism in place that checks whether the scrollbar should be visible or not, and therefore whether the space could be otherwise used.

Ah yes I see. In that case, I would make the scroll bar visible at all times on the right, and remove the space on the left where there is no scroll bar. The problem is not really the space it takes up (that's very minor), it's more an idea to try achieve better separation between each module.

TechXavAL:
I have tried to make sense of your new css stylesheet, but I couldn't, sorry. I really don't know what is what in RT GUI.

If the scroll bar doesn't increase size when hovering it and keeps being a thin bar, then the modules may have a few extra pixels wide. Not exactly what @SoupyGit requested, but close.

I think that the line separating the module title and its contents may be placed on top of the title: as I see it you have added a bottom border on the module title box, so it separates the title from its contents, but not the modules. If you add a top border instead, then you will effectively separate modules when they are expanded.

And perhaps if the power on/off button is located at the very left edge of the module (that is, no padding), it will seem that the whole module is a block. It would be better if the module title was at the very left of the box and the icon outside it, but for several reasons I don't think that would be possible or desirable.

Thanatomanic:
> I have tried to make sense of your new css stylesheet, but I couldn't, sorry. I really don't know what is what in RT GUI.

I know, the GUI and CSS was messier than I thought and a lot was unclear to me as well. It helped to enable the GTK inspector. My intention is to further clean that up in due time.

Your suggestion for a top border probably works better.

afr-e:
@ff2000 You are quite right. Rather than humour, I should have been more specific.

fatso83:
Good job, @Thanatomanic! I think it is _much_ clearer what is going on, with regards to active states and not. Not totally sure about the _amount_ of colour used, though. The biggest issue today is understanding the state of the various categories. I think the suggestion of using two versions of the same icon, but using colour to toggle the states would help. Not sure if blue+grey or white+grey would work the best - the white version in the official version is a bit ambiguous, but probably because the inactive state does not seem to relate to it in any way.

TechXavAL:
@Thanatomanic 

The GTK inspector hint was helpful :smiley: , but I can't really say I understand how it points to the css rules...

Anyway, I've managed to add a few of my suggestions into those rules, and here is what I was talking about:

csssuggestion

And these are the changes I suggested and a couple more (in no way means that they should be accepted):
* the power button is closer (next) to the left edge
* the powered on button lines are a little wider than on the powered off button
* label text is a bit larger
* all the sliders and controls that belong to a tool are enclosed by an engraved box
* the engraved box has 2px rounded corners (to me that looks a bit more polished than a pure square)

There are a thousand small changes that can be done, but to me it's quite hard going almost blind trying to find which rules have any effect to the area I'm trying to change.

What do you think? Is it useful? I can upload the relevant css and svg files if needed.

Thanatomanic:
@TechXavAL Good suggestions. The CSS is a bit flying in the dark sometimes and you need decent to good knowledge of CSS selectors to get stuff done. It would be a lot better if more GUI elements have specific names, which would make styling a lot less obfuscated.

- The inset box is not as bad as I thought it would be! How does it look for nested tool, e.g. the Resize > Post-resize sharpening?
- I don't like the power button so close to the edge... not in the blue-highlighted state anyway. I have an idea that I'll try to work on in the next few days.
- ~~"the powered on button lines are a little wider than on the powered off button" I don't see what you mean by that~~
- Edit: I hadn't even noticed the biggest change with the greyed-out power icon 🙄 . Looks pretty nice!

@SoupyGit Having the scrollbar visible at all times is something I will try out.

TechXavAL:
> you need decent to good knowledge of CSS selectors to get stuff done

I thought being the author of the new responsive look of RawPedia was enough, but it clearly isn't :wink: 

>The inset box is not as bad as I thought it would be! How does it look for nested tool, e.g. the Resize > Post-resize sharpening?

csssuggestion2

I haven't found yet the rule to make that P-RS box less rounded, or giving it a different look...

Here are the files, just in case:
[RT-gui-css-improvements_suggst.zip](https://github.com/Beep6581/RawTherapee/files/5329896/RT-gui-css-improvements_suggst.zip)

As I see it the changes may need to be subtle, and it's quite easy to go too far. Trial and error is a must...

SoupyGit:
The greyed out power buttons by xaval look good.

Part of me likes the look of the engraved box, part of me thinks extra lines makes things look even busier and is the opposite of whats required. In some parts there are already numerous borders, so what I think is required is less. As I mentioned earlier, getting rid of left hand side strip of empty space entirely. But it is perfectly useable to me as is anyway, so I'm happy, and will say no more.

Edit: Looking now on a bigger screen, the engraved box is certainly an improvement on what was. And I agree with everything Afre says in comment below - icon is too far left (has mis-alignment issues with the blue highlight) - and that the blue highlight is off putting, and that greyed out power buttons could still be more distinct.

afr-e:
I agree with @Thanatomanic  that the power icon is too far to the left in @TechXavAL's take. The greying is okay but still not distinct enough. I suggest giving it an inset (on) and outset (off) box, so it behaves like a button. This will also bring attention to it. As for the highlighting, the blue is definitely off-putting. Make it grey with enough DE2000 to make it sensible for most user's screens.

Thanatomanic:
@TechXavAL I meant no disrespect to your CSS foo! It's just that even with the GTK inspector you need to apply quite a lot of selector magic to style the proper elements. 

I've made a few changes:
- Included the requested separator line below the Navigator
- Increased the width of the major separators (this will be followed by a reduction of the margins around these separators)
- Changed from nesting increasingly darker to nesting increasingly lighter, with the additional lightening of checkboxes, sliders, etc.
- Applied the greyed power-off icon, supplied by @TechXavAL and made ever so slightly smaller
- Removed the completely blue headers when a tool is enabled, and replaced it with a small blue box around the power icon
- Made a variant of the inset appearance of (nested) tools, still shying away from rounded corners and a full inset
- Removed the borders below/above the tool headers, because I don't think they are necessary anymore with the above changes
- I'm not sure if I actually made the whole look more compact in the process, so that can still be tweaked further if necessary

image

Next up:
- Completely eliminate `size.css`, it's hard enough to maintain a single css with understandable styling
- Remove more borders where they are not functional (this will probably gain 10px vertical or horizontal, but hey, a pixel is a pixel!)
- ...
- Profit!?

Thanatomanic:
If you want to try out the latest version, please make sure to empty the `cache/svg2png` folder in your RawTherapee cache.

TechXavAL:
I like how it's looking :wink: 

Just one more suggestion: as already said quite a few times, whenever a tool or sub-tool is turned-off (grayed power icon), the UX would benefit quite a lot if the label was grayed too. Sadly, if I remember correctly from some attempts I made months ago, there's no disabled state for the text label, so it would mean coding a bit the user interface. To me, that would be a bigger change than the disabled power icon.

@SoupyGit : as already stated, tweaking the css is complicated and is easy to go too far for everybody's taste (like I did), so the more comments the better for @Thanatomanic (I think). If we can't tweak the code ourselves, at least we can say what to change and what to keep.

Thanatomanic:
> Just one more suggestion: as already said quite a few times, whenever a tool or sub-tool is turned-off (grayed power icon), the UX would benefit quite a lot if the label was grayed too. Sadly, if I remember correctly from some attempts I made months ago, there's no disabled state for the text label, so it would mean coding a bit the user interface. To me, that would be a bigger change than the disabled power icon.

That was quite easy to do actually 😄 I am a little worried about the legibility of the greyed text though. Can be tweaked (very easily now).

image

ff2000:
> image

While better IMO the big box pulls the attention too much on it.
Would it be possible to just paint the power symbol in the highlight color, without the surrounding box?

SoupyGit:
@TechXavAL 

> as already stated, tweaking the css is complicated and is easy to go too far for everybody's taste (like I did), so the more comments the better for @Thanatomanic (I think). If we can't tweak the code ourselves, at least we can say what to change and what to keep.

I hope it didn't come off as too critical. I like seeing various ideas being put forth. Unfortunately I am one who can't tweak the code, but am trying to contribute how I can, along with:

@Thanatomanic 

> Profit!?

Yes ;) I'm getting a few things sorted with my accounts at the moment, but in a few weeks or so will start making donations.

> Included the requested separator line below the Navigator

Good job. Any reason it is different width to the other separators in that panel?
    

> Changed from nesting increasingly darker to nesting increasingly lighter, with the additional lightening of checkboxes, sliders, etc.

Looks good!

> Removed the completely blue headers when a tool is enabled, and replaced it with a small blue box around the power icon"

Looks better. I would still prefer no colour, but one thing I noticed is you have blue sliders, thus a different theme from me - as all my sliders are grey. I use 'TooWaGrey - Average Surround' theme. So perhaps the blue colour could be related to a theme? Leave it grey in the TooWaGrey theme, but make it blue in whatever theme you are using?
    

> Made a variant of the inset appearance of (nested) tools, still shying away from rounded corners and a full inset"

Looks good! Makes it pop out.

> Removed the borders below/above the tool headers, because I don't think they are necessary anymore with the above changes

I liked it better with borders (but of course haven't seen borders along with the new changes).

> That was quite easy to do actually 😄 I am a little worried about the legibility of the greyed text though. Can be tweaked (very easily now).

I think greyed out power button (off), greyed out text (off), blue power button (on) is going a bit overkill now. For me, greyed out power button (off), normal text (all), and line separators (top) would be enough. That said, I like greyed out text (off) more than blue button (on).

With all these changes, it should definitely help solve the problem, and look nicer as well.

TechXavAL:
Look at this new edit:

csssuggestion3

There are a couple of changes here:
* the disabled label is now a bit lighter and not bold, so perhaps legibility improves a bit (as seen in the Shadows/Highlights tool). Strangely, there are a lot of tools that don't follow the rule of «text NOT being bold» (in this screenshot: Tone Mapping and DRC)
* the icon background has different shades or hue, so maybe people will be able to choose the one that best fits their needs (NOTE: in the final CSS rule there will only be one background. Here you have several of them thanks to some editing with GIMP)

And the CSS code **changed** is:

```
#MyExpanderTitle label {
    color: #AAAAAA;
    font-weight: normal;
}

#MyExpander.Fold > #MyExpanderTitle label,
#MyExpander.OnOff.enabledTool > #MyExpanderTitle label {
    font-weight: bold;
}


#MyExpander.enabledTool > #MyExpanderTitle #MyExpanderStatus image {
    background: rgba(200,200,200,.3); /* for the light gray background */
    background: rgba(200,200,200,.1); /* for the middle gray background */
    background: #002d59; /* for the dark blue background */
    background: black; /* for the black background */
}
```

Hope it somehow helps

afr-e:
What I mean before was to have the borders change to make it look like it is depressed or pressed. But I prefer the actual surface to be unfilled.

TechXavAL:
I'm just posting suggestions :wink: 

Does this look better? @afr-e  @ff2000 ?

csssuggestion4
The blue color of the icon can be lighter, if needed.

To note that I have also changed a bit the turned off icon.

Thanatomanic:
@TechXavAL How did you manage to make the icon blue? The SVG looked blue, but in RT it didn't. I'm curious :-)
Also, I think the blue needs a white shadow to stand out more on the darker background and to be more in line with the 'white' label.

Thanatomanic:
@afr-e 
> What I mean before was to have the borders change to make it look like it is depressed or pressed. But I prefer the actual surface to be unfilled. 

I have tried several variations of buttonized toggles, but I found none of them very pleasing in the bigger picture. Maybe I'm just particularly bad in making fake buttons, but my attempts always felt like the buttons were out of place. I can try again tonight and maybe post an example.

TechXavAL:
@Thanatomanic 

I have spent several hours trying to give some eye-catching effect to the power on button (glows below and above the lines of the icon, radial backgrounds, different borders for the icon box), but none was good enough.

In particular the white border makes the icon look fuzzy, but here you have both icons (the disabled icon too, in case you want to try it): 
[buttons.zip](https://github.com/Beep6581/RawTherapee/files/5340605/buttons.zip)

Regarding the blue filling, I think it's nothing else than a different hue of blue, so the script converting from svg to png doesn't «translate» it to semitransparent gray.

And the thickness of the icon strokes is important, too. Not every thickness look good.

Thanatomanic:
So basically anything we do about the button itself becomes tacky or a carnaval? I think this attempt with a minor gradient is kinda all right, but it's also... not...

image

RichardRegal:
> 
> 
> I'm just posting suggestions 😉
> 
> Does this look better? @afr-e @ff2000 ?
> 
> csssuggestion4
> The blue color of the icon can be lighter, if needed.
> 
> To note that I have also changed a bit the turned off icon.

That certainly is more what I had in mind.

afr-e:
@Thanatomanic Much improved. Would be even better if it were grey.
(Sorry for being an armchair critic. Might give you an example if I had more time and less drain.)

TechXavAL:
> is kinda all right, but it's also... not...

Maybe with a shorter gradation and stronger contrast? Something similar to this?:
csssuggestion5

RichardRegal:
> I have spent several hours trying to give some eye-catching effect to the power on button

I don’t think it has to be particularly eye-catching. It is just that the design of the button should be the same for both on and off and the emphasis for the on state needs to be just enough to distinguish the two. I like the fact that the lables for the active tools are also bolder than the labels for the inactive tools.

TechXavAL:
> the on state needs to be just enough to distinguish the two

So maybe just an icon filled in blue, but with a lighter hue? But remember that we are talking here about suggestions, and @Thanatomanic is not forced to accept what we say.

Same goes for the active/inactive labels. The distinction with a bolder typeface is just a suggestion I made, and it's currently not accepted. The only 2 ways I see that a suggestion would be accepted is if @Thanatomanic just likes it, or if many of us like a suggestion (and say so).

As an example: the version of the powered-on icon I like the most is the one with the black background, but that's just me, as many of you have already said it's better without a colored background. And that's fine. So in the end the icon won't have a black background...

SoupyGit:
FWIW, I like blue power button (with subtle gradient) better than blue background.

ff2000:
After looking at the change (I proposed myself...) I think the power symbol itself should have the same color as the text.
How do you like this one instead?
image

fatso83:
@ff2000 Having a blue fill color seems inconsistent: I don't see that 3-color (background, fill, stroke) pattern recurring anywhere. But neither does the rest, so ... 🤷 

So far, I think @TechXavAL and @Thanatomanic are the closest to something that looks crisp while making the turned on and off states clearly distinguishable, while also being similar.

Thanatomanic:
Thanks for all the suggestions, they are most welcome and I really like to see this kind of involvement! 😀  That said, I want to try and wrap up this PR. The most important change will be an improved way to highlight active modules, with some minor tweaks for visibility along the way. A follow-up can deal with more fundamental stuff that will improve the future ease of styling of the application itself.

Changes since last time:
- Reintroduced separator lines between tools, also when collapsed. This plays very nicely with the outset-look of the tool contents itself.
- Adopted an alternative blue gradient for the power on icon as a suggestion (@TechXavAL thanks!)
- Minor tweaks to disabled font color
- Fixed: some tool titles showed in bold font and some not, now they can all be styled properly

I seem to have an issue with font sizes. The custom `font-size: 1.1em` in `#MyExpanderTitle > label` is not picked up by RawTherapee. It used to be though. I can get the proper scaling when I change theme and change back, but this scaling is lost again after restart. Not sure what is going on...

@SoupyGit 
> >Included the requested separator line below the Navigator
>
> Good job. Any reason it is different width to the other separators in that panel?

The thin line is fixed, the thicker line is movable. This might change again in the future.

> >Removed the completely blue headers when a tool is enabled, and replaced it with a small blue box around the power icon"
>
> Looks better. I would still prefer no colour, but one thing I noticed is you have blue sliders, thus a different theme from me - as all my sliders are grey. I use 'TooWaGrey - Average Surround' theme. So perhaps the blue colour could be related to a theme? Leave it grey in the TooWaGrey theme, but make it blue in whatever theme you are using?

This is actually a very good point. Changing the icon to a blue version will break the intention of the TooWaGrey theme (see below). I need to check if it is possible to have a themed override of a certain icon...
@TooWaBoo Maybe you have some idea yourself? Love to hear your opinion of my main theme suggestions as well.

I have now the following options. Option A is currently implemented. Please give your final opinions and preference 🥇 
image

RichardRegal:
> Please give your final opinions and preference

All three are better than the current. In order my preference is B then C then A.

I think that a big difference would be distracted and possibly confusing. I don’t think that the difference needs to be huge, just big enough to be noticeable.

TechXavAL:
@Thanatomanic: you're welcome! :smiley: 

> I seem to have an issue with font sizes

Confirmed here. Even with older css files.

I like the separator lines.

My preferences are: C > A > B, but if you fill the powered on icon with pure white (instead of that blue color that makes it translucent) then my choices would be C > B > A

SoupyGit:
B > A > C 
(and keep white power button in toowagrey)
But there is an inconsistency in  A -  Soft Light is an open module with blackish background for the module header, whereas White Balance is an open module with grey background. The blackish background certainly helps the open module stand out, I like it. Can't decide if I like it better or not. I think it works because of the new engraved borders which makes the open module pop out - that brings the header along with it.

Thanatomanic:
@SoupyGit the black is because of the mouse hover. I don't intend to change the title background.

ghost:
If there are going to be any changes that add colors/squares etc., then perhaps it would be a good idea to ask a bigger sample of users (on pixls.us) what they think. I for one prefer a neutral interface without too many distractions and would not like to see that option removed.

TechXavAL:
@pseudopanax 

Although I agree with you about asking more users about their opinion (which may be a nightmare when trying to decide which option is best, by the way), if I'm not wrong you will still have a TooWaBoo neutral version available. So if you don't like the final look, you will be free to change to another «skin».

And by the way, when finished, I don't think it would be so difficult to remove any added color and create a new gray version, but keeping all the other tweaks.

afr-e:
The power icon should not change size, only colour. I choose C but prefer a lighter smaller box.
I still prefer a more compressed appearance for smaller screens to prevent the need for scrolling as much and a better at a glance view.
The right edge has space compared to @TooWaBoo's. Is that for the scroll bar?

TechXavAL:
Which is the current state of the GUI changes?

@Thanatomanic : may you tell me which css rule controls the blue line below the tabs?

Thanatomanic:
Thanks for the bump. I haven't made up my mind yet (and there were some pending CSS changes from the histogram/waveform update _and_ busy irl). Neither does there seem to be consensus here for those who contributed to the discussion, which doesn't make a decision easier ;-) What I can say is that after using it a for while now, the blue power icon is not really doing it for me, so I think I'll drop that in any case. Sorry Xavier.

As for the CSS for the blue line below the tabs, this is an override:
```
notebook tab {
box-shadow: none;
}
```

TechXavAL:
> the blue power icon is not really doing it for me, so I think I'll drop that in any case. Sorry Xavier.

No problem :smiley: I wasn't for nor against that. I just made suggestions based on other people comments.

And for the blue line, I can completely remove it, but not style it. Where is that style coming from? I would like to change the appearance like in Capture One (a gradient below the tab), just for my own use at least, but I can't find a way to point to the right rule.

Also, I made a suggestion about making the «on» icon pure white, so it stands out and makes it clear it is on. Maybe you wish to try that for a while.

Thanatomanic:
>  I would like to change the appearance like in Capture One (a gradient below the tab), just for my own use at least, but I can't find a way to point to the right rule.

The blue line is not a line, it is a fake, made by abusing the `box-shadow` [property](https://www.w3schools.com/cssref/css3_pr_box-shadow.asp). This also limits what you can do with it, so I'm not sure if you can get a proper gradient.

I'll give the pure white icon a try soon.

TechXavAL:
Well, it short of works:

shadow
but it's applied to all and every tab.

I would further play with it with different thicknesses and colors, but only if I can focus the changes to the active tab.

If you find this worthless, just say so. I'm writing it here to give ideas. It's not that I'm asking it to be in RT release.

Thanatomanic:
This should work (with bogus parameters):

```
#ToolPanelNotebook tab {
box-shadow: -1px 6px 3px 0px rgba(255,0,0,0.75);
}
```

I'm not sure if you can manage to pull off something nice with a box shadow though. The big challenge is to prevent the shadow from spilling (too much) to all sides of the box. But maybe if you also play a little with the margins of the boxes you can get something elegant. Good luck 😄

TechXavAL:
I kind of like this attempt:

shadow2

but as you can see, I haven't found a way to apply it only to the tab actually selected. The code in your previous message is in use here, and with it the shadow is only applied to the tools tabs, but all of them at the same time. Any suggestion?

Thanatomanic:
@TechXavAL Give `tab:checked` a try for the selected tab only.

TechXavAL:
That's it!!! Thank you! :smiley: 

shadow3

Thanatomanic:
I've pushed some final changes that I would like to incorporate in this PR. Further GUI/CSS fixes will be put in a different PR to keep things manageable.

If no one objects, I will commit this on Saturday.

ghost:
Can we please retain the existing theme as an option and rename it Rawtherapee-legacy or something similar.

Thanatomanic:
@pseudopanax I have little intention of keeping the old CSS, because that increases the burden of maintenance. Is there a particular change in the style that you really don't like?

TechXavAL:
@Thanatomanic 

I think it won't hurt anybody having an extra tiny css file in v.5.9

I can see why people may wish to keep using the current default style: it's compact, there is a bit more preview area space, and most importantly they are used to it. With time most probably many of those users will change to the new style, but I hate that they have to be forced to change.

You can always leave the current css file and change it's name from «**RawTherapee-GTK3-20_**» to «**RawTherapee-legacy_unmaintained-GTK3-20_**». I think that may be pretty shelf explanatory. *(No need to rename the «**size.css**» file, as it is never shown in GUI)*

By the way: I like the new style and I'm going to use it :+1:

Thanatomanic:
All right, I guess I can leave it in for 5.9 as a legacy option. But I wish to continue working on improving (= simplifying) the layout and styling of RT, so in the future I do not want to keep and maintain two working versions of the main style, but perhaps rather work towards a base theme with a few flavors like TooWaBoo also has.

ghost:
> Is there a particular change in the style that you really don't like?
On my system (Ubuntu 18.04):
1. The interface looks cramped and the labels are harder to distinguish, especially the way they are greyed out when not activated
2. The lighter grey background in the slider panels makes it harder to distinguish the sliders themselves, especially for the modules with lots of sliders.
3. Changing the small circle alongside the module names to a greyed out power button is just as confusing, if not more so than the existing interface because you get the impression that the function is not available for some reason. I doubt whether it really addresses the problem you initially set out to fix.
I have used it for the last week to see whether I could 'get used to it', but I have come to the conclusion that visually this is a step backwards.

TechXavAL:
@pseudopanax 

WOW! That was not the idea I had in my previous post!

I have a linux mint os, based on ubuntu 18.04, and it looks like this (I doubt you'll find a tool with more sliders than the Wavelet Levels tool):

sliders

Does your GUI look like this? If not, a screenshot would help here.

ghost:
> Does your GUI look like this?


Yes it does. And for comparison purposes, here is what it currently looks like:
Existing theme

TechXavAL:
Now I understand it better, I think.

Let me tell my opinion about it:
* «*The interface looks cramped and the labels are harder to distinguish...  The lighter grey background in the slider panels makes it harder to distinguish the sliders themselves*». There's definitely less contrast in the current new style, but I wouldn't say the labels and sliders are harder to distinguish (to me). But this is not something that can't be addressed in the future with those *flavors* @Thanatomanic talked about. There may be a *more contrasty* variation of the main style.
* «*Changing the small circle alongside the module names to a greyed out power button is just as confusing, if not more so than the existing interface because you get the impression that the function is not available for some reason. I doubt whether it really addresses the problem you initially set out to fix*» Well, it seems to me that this point precisely is one of the strong points of the new style. I'm pretty sure you have seen (or even own) a Hi-Fi appliance which has exactly that *icon* in its power button. There are plenty of computers that have that kind of button, too. And when they are turned off, you wouldn't say that the power button is *not available*, would you?

Anyway, this is a work in progress, and I'm eager to see the end results.

In the meantime, you will have your legacy style, as it seems it's more appropriate for you. So everybody will be happy! :smile:

heckflosse:
@Thanatomanic I would like to test this. I'm on Windows and using latest libjpeg-turbo, which causes a linker error when building your branch. Currently I locally merged dev into your branch. But wouldn't it make sense to push this change (the merge of dev into your branch) to simplify testing of your branch for other users which build RT on a windows system?

SoupyGit:
Picking up on the comments by @pseudopanax 

It does have less contrast than before. I can see how this would be harder for some to read.

Perhaps the reason some objects look 'not available' is because both power button and module text are greyed. It might be worth keeping all module text white, and just the power button grey. Perhaps also the 'off' power button being smaller than the 'on' contributes to this. But I don't think the issue is with the icon itself - the power button is definitely an improvement over little circle.

The drop down boxes do look cramped, like they are sitting right on top of each other without space in between.

However in general I think its a nice layout that looks neater.

Thanatomanic:
@pseudopanax 
>The interface looks cramped and the labels are harder to distinguish, especially the way they are greyed out when not activated
> Changing the small circle alongside the module names to a greyed out power button is just as confusing, if not more so than the existing interface because you get the impression that the function is not available for some reason. I doubt whether it really addresses the problem you initially set out to fix.

I've just made the titles a little brighter. I believe the titles are readable and there is still enough difference between the active an inactive state (creating a better distinction between the two was the whole point of this PR). The inactive power icon is a necessary change as well. Many new users were very confused about having to turn a tool on. The fact that there is a power icon now hopefully triggers people to press it to see what happens. When the title and icon light up as a result, this is a confirmation that whatever the user did, it activated something.
I will not change this behavior anymore in this PR, so if you really do not like this mechanism, you will need to use the legacy version of the theme.
And as a final note: there is also plenty of difference between a disabled piece of GUI and the inactive titles. Disabled is much darker still.

> The lighter grey background in the slider panels makes it harder to distinguish the sliders themselves, especially for the modules with lots of sliders.

I actually thought the lighter background made the sliders stand out better... To be honest, this is one of the design parts I am struggling with the most and I'm not 100% happy with the end result yet. For now though, I think I will keep it and finetune later if needed.

@SoupyGit 
> The drop down boxes do look cramped, like they are sitting right on top of each other without space in between.

I've added a little more space to hopefully address this issue.
There is an underlying problem though, and that is that there are big differences in the way different modules are built using GTK elements. Since this is not uniformly done, the styling is messy and error prone as well. Combine this with the fact that some tools have a huge number of necessary bells and whistles, it becomes a big headache to style everything in a pleasant manner. Certainly a work in progress.

@heckflosse I've merged with `dev` so I hope you can test! Looking forward to your opinion.

ghost:
@Thanatomanic 
Thank you for making the changes, they look much better to my eyes.

> I actually thought the lighter background made the sliders stand out better..

The darker background in the latest version looks about right to me. Rather than keeping a hard to maintain legacy version, perhaps it would be easier to have a variant for those that prefer a lighter background.

heckflosse:
@Thanatomanic I used your branch now for a couple of days. I really like it on my 1920x1080 screen, but I also appreciate, that the legacy theme is still available for my old low res laptop screen.

afr-e:
I think it is fine right now. The rest is nitpicking. To help with the low contrast and shapes problem, would it be a good idea to increase contrast upon hover? Is hover even supported in GTK?

ghost:
> The darker background in the latest version looks about right to me.

I must have compiled the wrong version when I made this remark so I withdraw it.

Issue #3794, opened 2017-04-01
PNG files cause crash


teplit:
The attached PNG files cause reproducible crash on three different Windows systems.

Steps to reproduce the problem:
1. Place the PNG files in an empty folder
2. Configure RawTherapee to parse PNG files
3. Open the folder from the first step with RawTherapee

RawTherapee versions: 5.0-r1-gtk3 (64 and 32 bit)
Screenshot of the error message: ![rawtherapee crash](https://cloud.githubusercontent.com/assets/5978882/24576991/5b5ff19e-16ce-11e7-9807-44efdeac5fe1.png)

The problematic PNG files:
[gradient45](https://cloud.githubusercontent.com/assets/5978882/24576934/4afecd6c-16cd-11e7-929b-b48317bcb591.png)
[gradient46](https://cloud.githubusercontent.com/assets/5978882/24576935/4b027c46-16cd-11e7-9007-1d1092fe80d4.png)

Beep6581:
Reproduced in Linux.

heckflosse:
It crashes at thumbnail generation because rt tries to generate a thumbnail according to the thumb height set in preferences (and maybe zoom factor of thumb too). In my test it tries to generate a thumbnail of size 80640 x 84 (width x height). 
Even If that would work correctly it would explode the filebrowser afterwards.
@Beep6581 @Hombre57 @Desmis @Floessie @agriggio : We have to decide how to handle this kind of images (for example skip them based on certain conditions or whatever)

agriggio:
would make sense to crop thumbnails that have a too big w:h ratio (say e.g >= 10:1) and of course signal this to the user in some way? what do you
think?

heckflosse:
@agriggio Currently I'm not sure about a good solution.

TooWaBoo:
A "Bounding Box" like in the "Resize" tool would be the best solution.

Hombre57:
Yes, a customable bounding box large enough to display panoramas without going into extreme width.

Beep6581:
There is already a discussion about File Browser thumbnail improvements with ideas and mockups over at issue #1239.

Beep6581:
Still valid.

agriggio:
If we want a hotfix, we can simply reject images that have too unbalanced aspect ratios. It's not a proper solution but it's better than crashing I'd say. In that case, something like this should do:

```diff
[...]

PR #5930, opened 2020-10-04
Backport fix for #3794 from ART

Floessie:
Hi,

On behalf of @heckflosse I backported ART's solution for #3794 to RT. Please test.

Best,
Flössie

### DIFF
--- a/rtengine/rtthumbnail.cc
+++ b/rtengine/rtthumbnail.cc
@@ -205,11 +205,6 @@ Thumbnail* Thumbnail::loadFromImage (const Glib::ustring& fname, int &w, int &h,
 
     ImageIO* img = imgSrc.getImageIO();
 
-    // agriggio -- hotfix for #3794, to be revised once a proper solution is implemented
-    if (std::max(img->getWidth(), img->getHeight()) / std::min(img->getWidth(), img->getHeight()) >= 10) {
-        return nullptr;
-    }
-    
     Thumbnail* tpp = new Thumbnail ();
 
     unsigned char* data;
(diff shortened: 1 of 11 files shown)


Thanatomanic:
@Floessie For when you're back from vacation: I've tested the branch but managed to get easy crashes while opening a folder that contains the two gradient files from the OP in #3794. There seem to be a few division by zero issues. Looking...

Thanatomanic:
I pushed two fixes. Thumbnail generation for extremely thin images (1px) now works as well.

Thanatomanic:
Ping @brussell1 @teplit, since you commented on the original issue #3794. If you can, please test this PR locally.

@agriggio You way be interested to apply the division by zero prevention fixes to ART as well.

ff2000:
There are several integer operations with implied rounding errors.
What do you think about enforcing floating point arithmetics (convert one value to double instead of size_t)?
And always round the resulting value up (using cmath.ceil) before converting back to int to avoid 0 as a result?

Thanatomanic:
@ff2000 Could work as well. I have no strong preference, also not in terms of computational efficiency (though I think integer operations and a simple _if_ could be faster than a double conversion and rounding - not that you would probably notice in this case). There may also be cases where a `ceil` is not the desired outcome, so as a general solution I don't think it's desirable.

Another alternative solution would be to always enforce a minimum image size of 1×1 in the constructor of an image class.

Floessie:
@Thanatomanic Thanks for testing and the fix! I would have used `x = std::max(1, x)` but that's a matter of taste. Hit the merge button, if you feel it's ready. :smile:

PR #5934, opened 2020-10-05
Revert "Suppress TRC working profile"

heckflosse:
This reverts commit 21758da234871b53e7f77cb03e6c34dfa0e9a7e3.

This is not meant as a PR, but as a discussion whether we should revert this commit.

### DIFF
--- a/rtengine/dcrop.cc
+++ b/rtengine/dcrop.cc
@@ -804,7 +804,7 @@ void Crop::update(int todo)
 
     if (todo & M_RGBCURVE) {
         Imagefloat *workingCrop = baseCrop;
-/*
+
         if (params.icm.workingTRC == "Custom") { //exec TRC IN free
             const Glib::ustring profile = params.icm.workingProfile;
 
@@ -818,7 +818,7 @@ void Crop::update(int todo)
                 parent->ipf.workingtrc(workingCrop, workingCrop, cw, ch, 5, params.icm.workingProfile, params.icm.workingTRCGamma, params.icm.workingTRCSlope, parent->getCustomTransformOut(), false, true, true);
             }
         }
-*/
+
         double rrm, ggm, bbm;
         DCPProfileApplyState as;
         DCPProfile *dcpProf = parent->imgsrc->getDCP(params.icm, as);
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -819,7 +819,7 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
         }
 
         if (todo & (M_AUTOEXP | M_RGBCURVE)) {
-          /*  if (params->icm.workingTRC == "Custom") { //exec TRC IN free
+            if (params->icm.workingTRC == "Custom") { //exec TRC IN free
                 if (oprevi == orig_prev) {
                     oprevi = new Imagefloat(pW, pH);
                     orig_prev->copyData(oprevi);
@@ -848,7 +848,6 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
                     ipf.workingtrc(oprevi, oprevi, cw, ch, 5, params->icm.workingProfile, params->icm.workingTRCGamma, params->icm.workingTRCSlope, customTransformOut, false, true, true);
                 }
             }
-            */
         }
 
 
--- a/rtengine/simpleprocess.cc
+++ b/rtengine/simpleprocess.cc
@@ -918,7 +918,7 @@ private:
             ipf.lab2rgb(labcbdl, *baseImg, params.icm.workingProfile);
         }
 
-/*        //gamma TRC working
+        //gamma TRC working
         if (params.icm.workingTRC == "Custom") { //exec TRC IN free
             const Glib::ustring profile = params.icm.workingProfile;
 
@@ -932,7 +932,7 @@ private:
                 ipf.workingtrc(baseImg, baseImg, cw, ch, 5, params.icm.workingProfile, params.icm.workingTRCGamma, params.icm.workingTRCSlope, dummyTransForm, false, true, false);
             }
         }
-*/
+
         // RGB processing
 
         curve1(65536);
--- a/rtgui/icmpanel.cc
+++ b/rtgui/icmpanel.cc
@@ -194,21 +194,21 @@ ICMPanel::ICMPanel() : FoldableToolPanel(this, "icm", M("TP_ICM_LABEL")), iuncha
     wTRCHBox->pack_start(*wtrclab, Gtk::PACK_SHRINK);
     wTRC = Gtk::manage(new MyComboBoxText());
     wTRCHBox->pack_start(*wTRC, Gtk::PACK_EXPAND_WIDGET);
- //   wProfVBox->pack_start(*wTRCHBox, Gtk::PACK_EXPAND_WIDGET);
+    wProfVBox->pack_start(*wTRCHBox, Gtk::PACK_EXPAND_WIDGET);
     wTRC->append(M("TP_ICM_WORKING_TRC_NONE"));
     wTRC->append(M("TP_ICM_WORKING_TRC_CUSTOM"));
 
-//    wTRC->set_active(0);
-//    wTRC->set_tooltip_text(M("TP_ICM_WORKING_TRC_TOOLTIP"));
+    wTRC->set_active(0);
+    wTRC->set_tooltip_text(M("TP_ICM_WORKING_TRC_TOOLTIP"));
 
 
     wGamma = Gtk::manage(new Adjuster(M("TP_ICM_WORKING_TRC_GAMMA"), 0.40, 15.0, 0.001, 2.4));
     wSlope = Gtk::manage(new Adjuster(M("TP_ICM_WORKING_TRC_SLOPE"), 0., 150., 0.01, 12.92310));
-//    wProfVBox->pack_start(*wGamma, Gtk::PACK_SHRINK);
-//    wGamma->show();
+    wProfVBox->pack_start(*wGamma, Gtk::PACK_SHRINK);
+    wGamma->show();
 
-//    wProfVBox->pack_start(*wSlope, Gtk::PACK_SHRINK);
-//    wSlope->show();
+    wProfVBox->pack_start(*wSlope, Gtk::PACK_SHRINK);
+    wSlope->show();
 
 
     wGamma->setAdjusterListener(this);


Desmis:
It's the second time, we ask me to suppress...so I do :)

gaaned92:
As I already expressed in pixls, I think that the TRC working profile option is not useful and unsafe at least for two reasons:
- It is said in rawpedia that the processing is done in a linear WS
- there is a consensus about linear processing, at least when reading pixls post.

moreover, if this option is kept, perhaps some dev should correct the fact that TRC with gamma=1 is not equivalent to no TRC.

And going on , I wish the gamma=1.8 used 
to display WS histogram, color values and clipping would be suppressed. The display of histogram in log scale should be sufficient.
It would avoid also relevant remarks from some meticulous users.

A solution for backward compatibility is to keep the option working for old pp3.

heckflosse:
@Desmis 

> It's the second time, we ask me to suppress...so I do :)

Jacques, if I ask you twice to jump into the river rhone from a bridge, would you do that? ;-)

Desmis:
Probably no :)

Desmis:
I probably went a little fast, because this code located here, in some cases (??) causes malfunctions.
On the other hand, the same functionality in iplocallac.cc ( shadows light - Tone equalizer - TRC)  works correctly hence my hasty deletion

:)

Thanatomanic:
@gaaned92 You claim a custom TRC for the WS is not useful. Is there really no possible use-case?
@Desmis It seems you added the code a few years ago with this commit https://github.com/Beep6581/RawTherapee/commit/19b7a942af2b5e8520bf0ce0d81b7036569e4881 Could you perhaps explain the original intention? Under what circumstances would it be useful to have a custom TRC?

Desmis:
@Thanatomanic 

Excuse my bad english, but it is the "same principle" as in "Ufraw"

It's allow you, to change TRC to lighten the shadows (slope), or act on the highlights (reduce gamma) without artefacts and with continuity....Try with very underexposed image and set slope about 100...Try also with gamma=1... In this case your output (TIF..jpg) is with gamma =1 and you see the result on your screen

Desmis:
@Thanatomanic 
@heckflosse 

To complete...

RT’s display profile is either sRGB (g=2.4, s=12,92), or a custom profile, which may have slightly different values.

RT’s working profile is g=1, s=0

I modify the code, just after the working profile (i.e. at the beginning of the RGB code and before Lab) to change the values that will be applied to the output data (display). The default is g=2.4 s=12.92.

The modification allows the user to see the effect of the TRC on the values applied to the output data (display)
• If he wants to darken the output with a TRC, he can lower the gamma.
• If he wants to lighten the shadows he can increase the slope.
• If he wants a gamma of 1, he puts gamma = 1

The disadvantage of locating the modification at the beginning of the RGB code, is that the input profile is interfered with and in some cases may lead to erroneous results.
This is not the case by locating it in iplocallab.cc

Jacques

RichardRegal:
> 
> 
> @Thanatomanic
> 
> Excuse my bad english, but it is the "same principle" as in "Ufraw"
> 
> It's allow you, to change TRC to lighten the shadows (slope), or act on the highlights (reduce gamma) without artefacts and with continuity....Try with very underexposed image and set slope about 100...Try also with gamma=1... In this case your output (TIF..jpg) is with gamma =1 and you see the result on your screen

I think then it would be useful to have something indicating what different slope values do compared with what different gamma do. I did ask once and received no reply.

Desmis:
@RichardRegal
There is no answer..to that :)


But, for each gamma and each slope the TRC is calculated to have no discontinuities..no artifacts

Eg : 
* for a gamma = 2.12 and slope = 4.6 it works
* for a gamma = 1.8 and slope = 0 ii works (gamma for Prophoto)
* for a gamma = 2.2 and slope =4.5 you have BT709
You can choose what you want (ig gamma > 1) between gamma and slope

I wrote this in rawpedia some years ago (in french)
https://rawpedia.rawtherapee.com/Color_Management/fr#Gamma_-_TRC_-_Courbe_de_r.C3.A9ponse_tonale

heckflosse:
Just for the record: I will not comment on the nature of this issue (Jacques, Roel and Flössie know why) . My intention was only to give a starting point for discussion. I hope it turnes out well :+1:

gaaned92:
@Thanatomanic 
It seems to be used as a sort of tone curve to enhance the shadows or reduce the highlights.
I agree that can be useful but not if it is applied at the very beginning of the pipeline as the location of the tool suggest. in this case it is not useful as it can induce artefacts in the downstream tools. Now I am ready to change my mind with some examples.
If an other tonemapping tool is required it is better to locate it somewhere toward the end of pipeline.

And last, it is beyond my understanding why you have to select a WS TRC ( (g=2.4, s=12,92) to obtain about the same  display as with no WS TRC.

That said, it doesn't really bother me if that WS  custom TRC selection remains.

What bother me is that the WS histogram is wrong as the displayed WS  RGB values have a 1.8 gamma applied.

SoupyGit:
The functionality - as a tool to lift shadows or reduce highlights without artifacts, sounds very desirable. I had been contemplating the last day or two whether to make a feature request for something like this. I came from darktable, which has filmic, and is the only thing I really miss in raw therapee. I see ART has log encoding which is the same as filmic but without the s curve (even better imo, as we can make our own curve with other tools) and with hard coded chrominance preservation (unfortunate, as I did like playing with that). So perhaps that can be taken from ART, and/or perhaps this custom tone curve can be used in a similar manner placed later in the pipe (can it also pull back highlights boosted beyond 255 in exposure module like log encoding/filmic?)

However, to ensure correct functionality I think the two points raised by gaaned need to be answered, 1) Why does trc = 1 give different results to default pipeline which has gamma = 1?, 2) why does WS histogram display gamma 1.8? What purpose does it serve by applying wrong values?

Desmis:
hello all
There is 4 ways to solve
1) suppress "TRC" in working profile, and use TRC in "Local adjustments (shadows highlight & tone equalizer" (standard)
2) keep TRC in working profile as before suppress - it seems is not a good solution...
3) Keep the GUI of "TRC in working profile", but move the code further in the process , with this option we keep the compatibility of pp3
4) move GUI and code further inthe process - in this case we lose the compatibility of pp3, and we duplicate "Local adjustment"
For 2) 3) 4) I think we must add a tooltip to explain 

@SoupyGit 
I you look at "Local adjustments" you can find
* Log encoding
* tone equalizer
* TRC
* etc
All this tools can work in full image

For the behavior with another gamma, this is the very principle of TRC curves. Here you change R, G, and B and not only Luminance channel. if the histogram did not change, there would be a bug
Traditionally there are 5 types of TRC
* sRGB gamma=2.4 s=12.92 - the one implanted by default in RT 
* BT709 gamma=2.2 s=4.5 - this TRC is "near" Prophoto 1.8, but with a linear part at the beginning
* gamma=2.2
* gamma =1.8
* gamma = 1.0
But with this tool you can make the TRC you want

But as I say before, this feature in now at the beginning of the process. When the default settings are g=2.4 s=12.92 this is to match the TRC of the image that appears to the user. This is not directly related to the gamma of 1 of RT. If you change TRC , you change RGB values, you change histogram...it is the goal of the tool.

Jacques

Desmis:
I create a new branch "trcnew" https://github.com/Beep6581/RawTherapee/tree/trcnew  where 
1) I change the location of "Tone response curve"...Now it's near the end of the process (after Lab adjustments). No more interference with "working profile"
2) I keep the location of the GUI in Icmpanel, but I change the groups
3) I add a tooltip (in bad english) to try to explain

jacques

SoupyGit:
@Desmis Thanks for the explanation, is more clear now.
I'm using RT 5.8 on Mac at the moment and don't have access to local adjustments, but great to know this work has been done! New Linux machine will be ordered in coming months.

Desmis:
has anyone tried the "trcnew" branch with the GUI and location code (without interference with working profile) changes ?

jacques

Thanatomanic:
@Desmis I have not had the time, but I will try and comment in the weekend. Thanks for taking action on this.

TechXavAL:
@Desmis 

I think if TRC is set to «None», the 2 sliders should be grayed out, or turned off. Currently on loading a raw image, the Slope slider is active, even if it does nothing to the image or the histogram. If you set Custom and then back to None, both sliders are turned off.

If you set Custom from None (g=2.4 s=12.92) without touching anything else, there's a clear change in histogram, although I can't see any difference in the Preview area.

I don't understand the tooltip: **_«Allows you to change the 'Tone response curve' which by default in RT is with a gamma sRGB (g=2.4 s=12.92). This TRC modifies the tones of the image - RGB values and histogram are changed»_**: where is that default TRC applied? If I set my display profile, it shouldn't be applied in the Preview area. If I set a linear working profile, it shouldn't be applied in the pipeline (let's not talk about custom TRCs for each tool). If I set a proper output profile, it shouldn't be applied while exporting the image. So where exactly is that default TRC applied?

And the label may point out where that TRC belongs to: if it changes the working profile TRC, it may be labeled «Working Profile TRC». If not, it would be nice knowing where it is applied straight away from the label.

Desmis:
@TechXavAL 

Hello Xavier

Happy to meet you :)

For the 2 gray.... probably a GUI problem..I will look..

If you change "none" to default...and there is difference, it is probably because you have a screen profile, with a TRC different form sRGB (perhaps BT709 or another...)

In RT all procees is done with linear . At the end of the process, either you have no screen profile and a TRC sRGB is applied (g=2.4 s=12.92), either you have a screen profile and it is this TRC which is applied.  Otherwise the image that would appear in the preview would be linear (very dark)

Before this version "newtrc", TRC was changing the working profile, but it leeds to bad behavior. Now It works in the same manner as in "Local adjustments" and it change the Lab datas at the end of the process. 

jacques

TechXavAL:
@Desmis 

According to DisplayCAL, my display custom profile is ≃sRGB (Δ 0.59%). So not exactly sRGB, but should I see these differences?
hist-trcnew
Top histogram, TRC set to None. Bottom histogram, TRC set to Custom, but without changing default values.
I have highlighted 2 of the most clear differences, but when viewed *live* all the histogram changes.

It's not that this can make any difference in the final image, as it's only a histogram. But I'm wondering if the image gets changed too, even if I can't see the difference.

> At the end of the process, either you have no screen profile and a TRC sRGB is applied (g=2.4 s=12.92), either you have a screen profile and it is this TRC which is applied

Now I get it, thanks.

SoupyGit:
>At the end of the process, either you have no screen profile and a TRC sRGB is applied (g=2.4 s=12.92), either you have a screen profile and it is this TRC which is applied. Otherwise the image that would appear in the preview would be linear (very dark)

@Desmis this is the first I understood the custom TRC we have been talking about adjusted screen profile. The GUI places it with working profile which is very misleading. I know you are working on finding a solution to that. Therefore to me, the proper place to put it in GUI would be with screen profile. However, this is currently in settings, not grouped with modules. So perhaps to implement this custom TRC properly, screen profile should be removed from settings, and placed in colour management section? Or, perhaps you can simply give another sub-heading in colour management section called Screen Profile, with the custom TRC.

Desmis:
@TechXavAL 
The differences between histograms are only due (in this case) to calculation problem... ; I changed the default TRC with BT709 (g=2.2 s=4.5) - another vey usual TRC 
I change the GUI

@SoupyGit 
It is not only "Screen profile"  that is changed, but the entire output (screen, TIF, JPG...) because datas RGB and Lab are changed

In the "trcnew" branch, TRC is a block different from "working profile" and "Output profile"

jacques
jacques

Thanatomanic:
@Desmis Would you consider opening a PR for your new branch so we can follow the changes there?
I am inclined to merge this PR and revert the original change, then we can work on finding the better solution in the new branch. Do you agree?

Desmis:
@Thanatomanic 

I will create a new PR this afternoon

Jacques

Desmis:
@Thanatomanic 
We can suppress, or keep open in memory, this PR, but no merge

jacques

RichardRegal:
@Desmis 

> But, for each gamma and each slope the TRC is calculated to have no discontinuities..no artefacts
> Eg :
>     * for a gamma = 2.12 and slope = 4.6 it works
>     * for a gamma = 1.8 and slope = 0 ii works (gamma for Prophoto)
>     * for a gamma = 2.2 and slope =4.5 you have BT709
>       You can choose what you want (ig gamma > 1) between gamma and slope

So if I choose the gamma then surely I should just have the slope that gives no discontinuities. Why would I choose a gamma of 2,2 and a slope of (for example) 4,0? That would give me discontinuities. Why should the program permit me to do so?

Desmis:
@RichardRegal 

If gamma > 1, all couples of values "gamma" and "slope" give a curve without discontinuity, the linear part and the parabolic part connecting perfectly (to the precision of the floating point calculations)

Thanatomanic:
@Beep6581 This is not a draft at all, neither is information missing imo. If we want to release 5.9 and  https://github.com/Beep6581/RawTherapee/pull/5949 is not merged into `dev`, we should definitely merge this PR to avoid breaking backwards compatibility.

Beep6581:
@Thanatomanic "draft" indicates that this PR should not be merged yet, "missing info" means we're not sure what to do and there is an ongoing discussion. I just read through the whole PR and that is still the case.

Issue #5693, opened 2020-03-26
manage custom crops profiles


artofit:
Hi,
It is possible to define custom crops ratio, for instance 16:7
However, I do not see the possibility (https://rawpedia.rawtherapee.com/Crop) to
1/ add/store a bunch of custom crop profiles, so that it select and applied even when closing and reopened Rawtherapee.
2/ delete prefined crop profiles, for example 6:17

Aka, user should be allowed to add or delete crop profiles

thanks

shapirus:
I will chime in on this.

When preparing pictures for printing, especially if there are many of them to be printed on a specific type of paper, it is important to crop the picture exactly so that it matches the actual dimensions of the paper, and doing that using the current approach for setting up a custom crop aspect ratio is tedious.

My vision of it is as follows. It may be implemented in two steps.

1. [basic functionality] The user-defined crop profiles are read from a local config file and become available in the drop-down list in GUI, maybe below a separator line after the hardcoded ones. The ratio should be defined as rational fraction in the config file, e.g. "[16, 7]" to maintain both precision and human readability.
2. [bells & whistles] There is support for saving new custom crop profiles in GUI. For example, "right click -> add new aspect ratio". These new profiles added via GUI are saved to the config file mentioned in step 1.

If done as described, it will not be necessary to implement both steps at the same time.

Having just the step 1 implemented (and documented) will already be a huge improvement for most people, because you don't need to add new crop profiles very often. Having to edit a config file occasionaly in such cases should be okay.

Having the step 2 implemented as well will further improve usability and add to the professional look of RT.

PR #5939, opened 2020-10-06
Add a 13:18 photo paper crop aspect ratio

shapirus:
There is a standard photo paper format 13x18 cm. This commit makes it easier to process pictures to be printed on this paper.


Thanatomanic:
I see no reason why we couldn't add this, except maybe that the list gets longer.

shapirus:
This time it sounds okay, because it's a standard paper format, but ideally, this should be user-configurable, because some people will have their custom, non-standard, not useful for other users, aspect ratios that they will use more than once, and doing this every time via unchecking the 'lock ratio' option and then entering numbers and selecting 'current ratio' is tedious.

Probably a new feature such as "right click -> add custom aspect ratio" could be implemented, or at least a possibility to define new custom ratios via a config file. Do you think it's a good idea to open a feature request issue about this?

Thanatomanic:
@shapirus This issue has already been filed here https://github.com/Beep6581/RawTherapee/issues/5693, feel free to chime in on a good solution for this.

shapirus:
Nice, thank you.

Also, I have just realized that I will probably need to change the actual aspect ratio calculation a little, because it's a question of the actual dimensions of the paper. It's called both 13x18 cm and 5"x7" on the market as though it were interchangeable, while mathematically 13/18 != 5/7, so the actual ratio is a question. I will receive some paper (marketed as 13x18) later this week and make a change, if necessary.

It may be possible that we will have to add support for both formats, if they are slightly different, for example, in EU and US.

p.s. so far it sounds like the actual paper size is:

1) based on imperial units (I have just measured some 10x15 cm paper that I have, which is really more like 4"x6", but not precisely);
2) may vary

which really suggests that the list of supported crop aspect ratios should be customizable. We may still add some suitable-for-most defaults, however.

PR #5997, opened 2020-11-18
Improve error logging and memory handling when reading tiff files

heckflosse:
Trying to read corrupted tiff files, RT (at least on Windows) flushed the error message from libtiff at closing RT.
I changed that to flush the error message when it occurs and also now the filename is shown.

### DIFF
--- a/rtengine/imageio.cc
+++ b/rtengine/imageio.cc
@@ -25,6 +25,7 @@
 #include <cstring>
 #include <fcntl.h>
 #include <libiptcdata/iptc-jpeg.h>
+#include <memory>
 #include "rt_math.h"
 #include "procparams.h"
 #include "utils.h"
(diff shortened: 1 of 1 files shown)


Floessie:
I see no difference in output when switching to this branch:

```
TIFFFetchDirectory: Can not read TIFF directory count.
TIFFReadDirectory: Failed to read directory at offset 108705840.
```

is printed immediately when I enter a directory with a trimmed TIFF, which is expected, as `stderr` is unbuffered on Unix-like OSes. But I would have expected your new messages to appear with this branch. :confused:

heckflosse:
Here's the tif I used for my tests. It has only one channel, which is not supported by RT.

[Mars_001419_pss_f2090_p50_b34_ap18.zip](https://github.com/Beep6581/RawTherapee/files/5566353/Mars_001419_pss_f2090_p50_b34_ap18.zip)

Output with this file on Windows is
```
TIFFFillStrip: Invalid strip byte count 0, strip 0.
Error 3 loading Z:\Eigene Bilder\Mars_001419_pss_f2090_p50_b34_ap18.tif
```

heckflosse:
> It has only one channel, which is not supported by RT.

This was partly wrong. It's true that my example file has only one channel. It's not true, that this is not supported by RT. The example file indeed seems to be corrupted, which is, why I made this pr.

PR #6017, opened 2020-12-06
Ciecam - added cam16 to cam02

Desmis:
I added Ciecam16 to Ciecam02 - I hope no error in the code....Because the 2 options are now usable

* I think Ciecam02 is (slightly) better for colorimetry...but despite the corrections made by M.H.Brill S.Susstrunk and the RT team (Ingo), there are still in some rare cases possible artifacts
* Ciecam16 should in theory generate less artifacts, the code is "simpler".... Which is not the case here, because it was necessary to make compatible both Ciecam02 and Ciecam16, and ensure compatability with 5.8

Jacques

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -437,38 +437,38 @@ HISTORY_MSG_171;L*a*b* - LC curve
 HISTORY_MSG_172;L*a*b* - Restrict LC
 HISTORY_MSG_173;NR - Detail recovery
 HISTORY_MSG_174;CIECAM02
-HISTORY_MSG_175;CAM02 - CAT02/16 adaptation
-HISTORY_MSG_176;CAM02 - Viewing surround
-HISTORY_MSG_177;CAM02 - Scene luminosity
-HISTORY_MSG_178;CAM02 - Viewing luminosity
-HISTORY_MSG_179;CAM02 - White-point model
-HISTORY_MSG_180;CAM02 - Lightness (J)
-HISTORY_MSG_181;CAM02 - Chroma (C)
-HISTORY_MSG_182;CAM02 - Automatic CAT02/16
-HISTORY_MSG_183;CAM02 - Contrast (J)
-HISTORY_MSG_184;CAM02 - Scene surround
-HISTORY_MSG_185;CAM02 - Gamut control
-HISTORY_MSG_186;CAM02 - Algorithm
-HISTORY_MSG_187;CAM02 - Red/skin prot.
-HISTORY_MSG_188;CAM02 - Brightness (Q)
-HISTORY_MSG_189;CAM02 - Contrast (Q)
-HISTORY_MSG_190;CAM02 - Saturation (S)
-HISTORY_MSG_191;CAM02 - Colorfulness (M)
-HISTORY_MSG_192;CAM02 - Hue (h)
-HISTORY_MSG_193;CAM02 - Tone curve 1
-HISTORY_MSG_194;CAM02 - Tone curve 2
-HISTORY_MSG_195;CAM02 - Tone curve 1
-HISTORY_MSG_196;CAM02 - Tone curve 2
-HISTORY_MSG_197;CAM02 - Color curve
-HISTORY_MSG_198;CAM02 - Color curve
-HISTORY_MSG_199;CAM02 - Output histograms
-HISTORY_MSG_200;CAM02 - Tone mapping
+HISTORY_MSG_175;CAM02/16 - CAT02/16 adaptation
+HISTORY_MSG_176;CAM02/16 - Viewing surround
+HISTORY_MSG_177;CAM02/16 - Scene luminosity
+HISTORY_MSG_178;CAM02/16 - Viewing luminosity
+HISTORY_MSG_179;CAM02/16 - White-point model
+HISTORY_MSG_180;CAM02/16 - Lightness (J)
+HISTORY_MSG_181;CAM02/16 - Chroma (C)
+HISTORY_MSG_182;CAM02/16 - Automatic CAT02/16
+HISTORY_MSG_183;CAM02/16 - Contrast (J)
+HISTORY_MSG_184;CAM02/16 - Scene surround
+HISTORY_MSG_185;CAM02/16 - Gamut control
+HISTORY_MSG_186;CAM02/16 - Algorithm
+HISTORY_MSG_187;CAM02/16 - Red/skin prot.
+HISTORY_MSG_188;CAM02/16 - Brightness (Q)
+HISTORY_MSG_189;CAM02/16 - Contrast (Q)
+HISTORY_MSG_190;CAM02/16 - Saturation (S)
+HISTORY_MSG_191;CAM02/16 - Colorfulness (M)
+HISTORY_MSG_192;CAM02/16 - Hue (h)
+HISTORY_MSG_193;CAM02/16 - Tone curve 1
+HISTORY_MSG_194;CAM02/16 - Tone curve 2
+HISTORY_MSG_195;CAM02/16 - Tone curve 1
+HISTORY_MSG_196;CAM02/16 - Tone curve 2
+HISTORY_MSG_197;CAM02/16 - Color curve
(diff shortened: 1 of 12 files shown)


Thanatomanic:
@Desmis I'm not sure why you added a function argument to many of the functions. Wouldn't it be more elegant to set the switch for 02 or 16 as a variable in the class that you can simply refer to in the functions?

Desmis:
@Thanatomanic 
As a reminder, I am not a computer scientist, I have no training in this field and what you are telling me looks like Chinese to me. 
I am a scientist, an engineer and a social scientist at the same time and in the "small world" of RAW software, I am known for my (modest) skills in colorimetry.
I implement the code "my way" (which is probably not elegant...).

But why did it come to this? Simply by trial and error (many...) , because :
* the Ciecam02 code has been modified many times, in particular to introduce the changes made by M.H.Brill S.Susstrunk.
* Ciecam16 code is not "official"...It is only a recommandation of CIE...and there is not an "official code".
* In fact the current code of Ciecam02 should probably not be called Ciecam02, because it does not correspond to the initial principles (Hunt, Fairchild...) notably in the functions... (cat02_to_xyzfloat, hpe_to_xyzfloat,cat02_to_hpefloat, xyz_to_cat02float..) all these functions are not "conformed"..and the solution depends on what you want to do
* I'm not sure I've made all the required changes (I think I have), because the literature on Ciecam02 and the switch to Ciecam16 is more than succinct and in any case nothing exists (??) that makes the 3 solutions work "together". (Ciecam02 legacy, Ciecam02 RT with M.H.Brill S.Susstrunk, Ciecam16...adapted to RT.)

I finally chose to propose only 2 choices (CAM02 RT and CAM16 RT), but why now that there can be a choice not to put "CAMO2 legacy" ?

I still have a few "Labels" to change, I would do that during the day.

Now if it's about changing the code to make it more elegant...Why not, but show me how to do, and step by step because for me, what you wrote is Chinese

But beyond the remarks on the "elegance" of the code, what do you think of Ciecam? Especially the contribution of CAM16 ?

jacques

Thanatomanic:
> As a reminder, I am not a computer scientist, I have no training in this field 

Neither am I, and my suggestions are also only based on my very limited experience 🙂
What I meant is that you add a variable to the Ciecam02 class. This variable could be called `version`. Then, when you create an Ciecam02 instance in `improcfun.cc` you could set this variable to something based on `params->colorappearance.modelmethod`. In all the functions within the `Ciecam02` class you can then simply use the value of the `version` variable to make the distinction between the 02 or 16 operations. No need anymore for a function argument.

Maybe @Floessie or @heckflosse could chime with their best practice?

@Desmis As I have not yet had the need to use CIECAM02 in my usual photography workflow, I am not the best person to ask for their opinion on the benefit of CIECAM16. Maybe you can point me to a website that explains why the CIE developed CIECAM16 as a 'better' version of CIECAM02?

Desmis:
Actually I don't see much interest, apart from 'having fun' with a nice code...Nevertheless why not?

But it should be taken into account that Ciecam is not only called by "Improcfun.cc" for general use, but also by "Local adjustments" (for "Warm-cool", and as a complement to "Log Encoding" - I forced to CAM6), as well as other potential uses, and also by "Curves" (in Exposure tab) - I forced to CAM02 (compatibility)

Here a link CIE
http://cie.co.at/technicalcommittees/new-colour-appearance-model-colour-management-systems-ciecam16

It is a pity that Ciecam is so little known, whereas it is one of the particularities (with Wavelet-levels, "Local Adjustments", "Capture sharpening", etc.) that marks the difference with the other products of the market

Recall
https://discuss.pixls.us/t/ciecam-and-rawtherapee/21368

jacques

Desmis:
@Thanatomanic 
Yesterday I have tried your solution....with my limited understanding...not possible to compile

I tried another this morning
Here what I do
1) in cieacam02.h , I added a line like this
 public:
    Ciecam02 () {}
    static void curvecolorfloat (float satind, float satval, float &sres, float parsat);
    static void curveJfloat (float br, float contr, const LUTu & histogram, LUTf & outCurve ) ;
    **static int versio;**
    /**
     * Inverse transform from CIECAM02 JCh to XYZ.
     */

2) in improcfun.cc, I added Ciecam02::versio
        int c16 = 1;
        Ciecam02::versio = 1;
        
        if (params->colorappearance.modelmethod == "02") {
            c16 = 1;
           Ciecam02::versio = 1;
        }else if (params->colorappearance.modelmethod == "16") {
            c16 = 16;
           Ciecam02::versio = 16;
        }

        Ciecam02::initcam1float (yb, pilot, f, la, xw, yw, zw, n, d, nbb, ncb, cz, aw, wh, pfl, fl, c, c16);

3) impossible to compile
and here a copy Msys2
ciecambug

4) I probably missed something ? Can you help me ?

Thank you

Jacques

Thanatomanic:
@Desmis I've just glanced at what is actually going on in `improcfun.cc` with regards to ciecam code and I see that the `Ciecam02` class only contains static methods. So in fact, it is simply a bunch of related functions that you can call, instead of a fully functional class in the object-oriented sense (I hope this is not complete gibberish to you).
My suggestion for a variable that you can set will not work here without a much bigger rewrite I think.

Desmis:
@Thanatomanic 
Yes it is complete gibbersih....as I have said, for me all these things are near of chineese...

So what do I do? I'm not the one who's going to change something I don't understand...

I merge? or am I waiting for someone to do that rewrite?

Jacques

Desmis:
If nobody objects, I will merge tomorrow 

Jacques

heckflosse:
Are the new default settings backwards compatible?

Desmis:
@heckflosse

Yes, of course....and that was the difficulty  :)



Jacques

Desmis:
if a user had previous settings, the system switches to Ciecam02, free to him then to choose Ciecam16. but in general the differences are minimal....

ValZapod:
This was not final standard. It is finally **CIE**CAM16, standard [CIE 248:2022](https://cie.co.at/publications/cie-2016-colour-appearance-model-colour-management-systems-ciecam16).

Changes are very small: https://github.com/colour-science/colour/pull/1015

Thanatomanic:
Thanks for commenting on various things, but what is your point here exactly? This branch has been merged nearly 2 years ago. If there is an issue, please create a ticket.

Desmis:
@ValZapod 

I don't understand at all what you mean.
Ciecam comes from the French "CIE" "Commission Internationale de l'Eclairage"

She has over time used several CAM models (Color Appearance Model) CAM97, CAM02, CAM16

The latest one solves a lot of bugs present in the CAM02 version, and is now really very good quality. Currently I think in terms of colorimetry (SDR) it's the best.
And for chromatic adaptation CAT16 replace CAT02

But why this question, a bug, semantics, or? 

Jacques

ValZapod:
CAM16 uses two matrices, CIECAM16 uses one. CIE is a standards body and at the time you implemented this @Desmis CIECAM16 did not exist.

Thanatomanic:
I see now what you mean. I have created an issue (https://github.com/Beep6581/RawTherapee/issues/6581) to address this.

@ValZapod For future reference, if you find anything else, please [create an issue yourself](https://github.com/Beep6581/RawTherapee/issues/new/choose) and simply reference this PR. It will do no good to add a comment to a thread that is long-dead.

Desmis:
@Thanatomanic  @ValZapod 

I continue not to understand anything, excuse me, but I am an old Frenchman (75 years old), very ill.
I am the author of the Ciecam02 version in 2012, there are indeed 2 conversion matrices including in Ciecam16. But the "16" version is much simpler with less subcode than "02" and less bugs (black dots, narrow gamut,...)

Version 16 is also my code, which can be found in a simplified way in "Local adjustments". I even made personal modifications to adapt CAM16 partly to HDR.

But, to simplify, these modifications, if there are any to be done, will be done without me.

jacques

ValZapod:
Oh, so it is not even pure CAM16, but with HDR magic. Understood. The problem it was not a final version.

PR #6062, opened 2021-01-15
Local adjustments - Retinex - Last Recovery based on luminance mask

Desmis:
Same process and algo than others "Recovery"
But "adapted" to Retinex Code - more complex.
It is the last change "Recovery with mask"...all tools with mask have now "Recovery"


Thanatomanic:
> It is the last change "Recovery with mask"...all tools with mask have now "Recovery"

🎉  Good job @Desmis, your output the last weeks is quite impressive!

Desmis:
@Thanatomanic 

Thank you :)

jacques

PR #6063, opened 2021-01-16
Local adjustments - GUI 'full image' - allows you  to use the local adjustment tools on the whole image

Desmis:


waynesutton50:
This is a welcome addition and is particularly useful with the Log Encoding tool. Combining an RT Spot in full-image mode with Log Encoding provides some interesting possibilities, not the least of which is the ability to fine tune the result by simply moving the spot centre. That along with the recent addition of Ciecam to Log Encoding makes this a significant enhancement.

PR #6072, opened 2021-01-21
Update issue templates

Thanatomanic:
In an attempt to enhance/professionalize our issue tracker, I've worked on some templates for issue submission. Please provide feedback.


Thanatomanic:
@Beep6581 As owner of the repo and with the experience of handling many issues in the past, could you comment on this idea?

Beep6581:
Hey @Thanatomanic 

We could try using a template. It could be good, but it should be short and to the point, else nobody reads it. I would do away with all the faux niceties like "thank you for reporting" type of stuff, no one needs a machine to thank them. I like the default GitHub template, it's straight to the point.

Your templates include formatting I don't see when editing a real template, for instance the real one has no comment sections `<!-- (...) -->`, how do I preview your templates correctly?

Thanatomanic:
Hi @Beep6581, I'll make things more to the point, that's fine with me too. It's a little less personal imo, but it works either way.

As for the commenting syntax, I took that idea from our colleagues at darktable. This ensures that the 'hints' we provide in the template never show up in the actual post. This prevents leftovers from people that forget to remove/replace the info. So this is intentional.

Thanatomanic:
@Beep6581 What do you think of these updates?

Beep6581:
@Thanatomanic let's go ahead and try them out.

Thanatomanic:
@Beep6581 Clearly it doesn't work for everyone, see https://github.com/Beep6581/RawTherapee/issues/6121 😉

syyrmb:
Adding an entry for translation and localization stuff would be nice, or should we just use "Feature request / enhancements" for these things?

Thanatomanic:
@syyrmb You have a point, but I think this can be filed under 'Feature request / enhancements' for sure

Beep6581:
@syyrmb I don't think localization howto belongs in an issue template.

syyrmb:
@Beep6581 Well, what my "localization" means is more about submitting translated files/documents, and currently no template is explicitly in charge of it, so I asked if we should add a new template or use the "Feature request / enhancements" template.

Issue #6082, opened 2021-01-31
LA: Changing Spot method between Normal and Excluding resets shape outlines


ff2000:
This is with current master 6b1e5d.

To reproduce:
Enable Local Adjustments, Add a spot, change the shape points (e.g. make the area larger), then change from "normal" to "excluding" - the shape gets reset.
Unfortunately the changes aren't recorded in the history so one has to manually adjust the shape again.

Desmis:
@ff2000 
Hello

Thank you fo testing :)

it is the "normal" behavior.... Sorry, very difficult to change

Jacques

ff2000:
But AFAIR this hasn't been the behaviour until recently, probably since addition of "full image" method. (I'll checkk ASAP)
Update: Yes, went back to commit d36545 and changing the method does not alter the shape outlines.

Desmis:
It s due to the 3 choices "normal" "excluding" , and "full image"...This last mode "auto" for "full image" leeds to this behavior...

I don't know how to to do otherwise :)

jacques

ff2000:
If it's hard to keep the outline at least register the change properly so that going back in history restores the previously (carefully crafted ;)) shape.

Where in the code do you handle the mode change? I could have a look if I have an idea (likely not... but at least I want to try to help :))

Desmis:
@ff2000 
The code is only GUI....And I am not at all a specialist
The main code is in "controlspotpanel.cc", (spotMethod_) but also to save / restore with pp3 , locallab.cc, procparams.cc, paramedited.cc and of course for "events" in procevents.h and refreshmap.cc

The problem is for me the same as for others variables when you change "basic" / "standard" / "advanced"...reset are "default"

The system cannot know what you want to do, and (I think) cannot keep en memory more than the last change.
For example, if no reset to 'default',  if you choose "Full image"...and after "excluding"....Excluding spot will have a very very big size....3000*3000 totaly unusable...

But, if you have a solution, I take

jacques

Pandagrapher:
@ff2000 @Desmis 
The problem is due to a bug in `ControlSpotPanel::spotMethodChanged()` method. I will provide a patch this WE.

Pierre

Pandagrapher:
Fixed with pull request #6093

PR #6093, opened 2021-02-03
Local Adjustments - Fix for GUI issue #6082

Pandagrapher:
@ff2000 
Fixes the issue #6082

Pierre

### DIFF
--- a/rtgui/controlspotpanel.cc
+++ b/rtgui/controlspotpanel.cc
@@ -954,6 +954,7 @@ void ControlSpotPanel::spotMethodChanged()
     const auto iter = s->get_selected();
     Gtk::TreeModel::Row row = *iter;
 
+    const int oldSpotMethod = row[spots_.spotMethod];
     row[spots_.spotMethod] = spotMethod_->get_active_row_number();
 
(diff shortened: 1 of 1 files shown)


ff2000:
Thank you very much!
I confirm the fix. I assume there can't be done anything going Normal -> full image -> normal to restore the previous Normal shape. Which is fine as the shape correctly is registered in the history now.
:+1:  for merging.

Desmis:
@Pandagrapher 
I confirm the fix :)

Thank you Pierre.
Same remark as @ff2000 for full image ?

I think you can merge :)

Jacques

Pandagrapher:
@ff2000 @Desmis 

> I assume there can't be done anything going Normal -> full image -> normal to restore the previous Normal shape. Which is fine as the shape correctly is registered in the history now.

That would be complicated because full image replace previous shape params by its own so I can't retrieve previous shape.
Moreover, in new GUI, I have planned to consider full image as a dedicated shape (with ellipse, rectangle, and polygone ?) where I am not sure to be able to add conversion from one shape to another (for instance, how to transform a polygone to something else ?!).

Pierre

PR #6096, opened 2021-02-05
added german comment to rawtherapee.desktop.in

freddii:


Thanatomanic:
@freddii Thanks for an easy addition :-)

Issue #6092, opened 2021-02-02
Local Adjustments: preview DeltaE does not work with Black-and-White tool enabled


chaimav:
The preview DeltaE function does not work when Color: Black-and-White tool is enabled. The image remains B & W and there is no green overlay.
RT 5.8-2745, Win 10, i3-8100.

Desmis:
@chaimav 
Thank you for testing :)

for me I don't think I can find a solution. 

When the PR will be merged (BW after LA) we see a little better the incidences of the BW settings...
https://github.com/Beep6581/RawTherapee/pull/6080

jacques

chaimav:
Can you at least write in the tooltip to warn of the issue? It took me a bit of time to narrow down why preview DeltaE was not working.

Desmis:
@chaimav 
I push this change with the PR 
https://github.com/Beep6581/RawTherapee/pull/6080

Thank you

Jacques

Desmis:
@chaimav 
With the new placement LA in the pipeline. Now it is before "Black and white", I think it is possible to do...
Before with LA after it was totally impossible because BW neutralized all scope deltaE actions.

I push a commit - 856d445802 - in branche "lanewlocation". I hope if fixed this behavior

Jacques

PR #6099, opened 2021-02-08
Local adjustments - Neutralize process after LA when when the user uses preview deltaE

Desmis:
Up to now only BW was concerned by this behavior

Now, process after "LA" are neutralize (provisory) when the user uses one of the functions "Preview deltaE".
So we can see the impact of LA without the action of downstream processes

Jacques


Thanatomanic:
I guess this was prompted by my comments on the forum? Seems like a good move.

Desmis:
@Thanatomanic 
in part, yes... :)

But I should have done it in the same modification as for issue  Black and White
https://github.com/Beep6581/RawTherapee/issues/6092

jacques

PR #6109, opened 2021-02-10
Fix new color propagation method

Thanatomanic:
In the beginning of January a new color propagation inpaint method was ported from ART. A bug was found (https://discuss.pixls.us/t/issue-with-black-blotches-help/23378) and a fix was suggested by @agriggio and implemented by @Desmis. This PR is made for easy testing.


Thanatomanic:
I would like to comment on the primary modifications to the method, because I am not convinced they do a better job than the old method. I made a few comparisons using the image provided by Stefan on Pixls.

1. Start from Neutral
2. Enable 'Highlight reconstruction' and set to 'Color propagation'.
3. Set 'Highlight compression' to 100.

**Old Color Propagation method** (dev at e8da58ad2)
S7_09850-colorprop-old

**New Color Propagation method** (this branch/PR)
S7_09850-colorprop-new

The old method is by no means perfect, and you can clearly see some jagged edges. Strong CA and high contrast aggravate this.
There are two problems I spot with the new method. First, there is noticeable haloing/light leaking. Compare the top part of the images around the dark bars and you'll see the overexposed areas have a 'glow'. I don't think you want any HR method to touch the correctly exposed parts of your image.
Second, there is loss in texture. You can see that on the fabric in the lower left corner. The new method gives those areas a smoothened, slightly pinkish look, whereas the old method managed to keep a lot more detail visible.

I am not sure if this is just a single case that performs worse with the new method, or that these are general features. I could try to make some more examples. Combined with the fact that @heckflosse mentioned the method is slower, I actually don't think we should keep these changes at all, and revert back to the old method.
@Desmis, you mentioned there has been discussion before about this new method with people from darktable and Alberto. Can you point me to it?

Desmis:
@Thanatomanic 

I think it is here
https://discuss.pixls.us/t/3-4-scene-referred-workflow-bad-bokeh-highlights-reconstruction/22135/37

And, on the image A7C02846.ARW, the improvment is important
To note, I have add a rubric in Rawpedai with this example, 1 mounth ago
https://rawpedia.rawtherapee.com/Local_Adjustments#Log_Encoding_and_Highlight_Recovery

Jacques

heckflosse:
If the new method significantly improves the result for some images while it degrades the result for other images, why not keep both methods, try to classify the use cases for both methods (we already have 3 methods in combobox, so a 4th wouldn't matter much). Coding that would be quite easy, though I would like to see a before/after example where the new method improves the result.

Desmis:
@heckflosse 
For example try on the image A7C02846.ARW....link with darktable...and also in Rawpedia

Jacques

Thanatomanic:
I can provide some more examples:

**Before**
A7C02846-colorprop-old

**After**
A7C02846-colorprop-new

**Before**
Nikon - D80 - 16bit (3_2)-colorprop-old

**After**
Nikon - D80 - 16bit (3_2)-colorprop-new

**Before**
Nikon - D800 - 12bit uncompressed (3_2)-colorprop-old

**After**
Nikon - D800 - 12bit uncompressed (3_2)-colorprop-new

For all these edits:
1. Start from Neutral
2. Enable 'Highlight reconstruction' and set to 'Color propagation'.
3. Set 'Highlight compression' to some reasonable value to visualise effect.

I'm really not a fan of the blurring in the new method.

heckflosse:
@Thanatomanic for your first example, new has indeed much less artifacts than old. Old was pretty much unusable because of too many artifacts for this file. For the other files I agree, that new blurs too much compared to old

Thanatomanic:
@heckflosse The first one is a tough one: lots of very high contrast, artificial lighting. As a matter of fact, the purple pixels are not (only) due to HR but happen with AMaZE, and not RCD for example.

heckflosse:
> but happen with AMaZE, and not RCD for example.

Good to know, so we should not count this example valid imho.

heckflosse:
I vote for reverting the change. Then we can try to start from scratch based on the old code. That are my 2 cents,

Desmis:
No problem for me for reverting the change.

Even if, since I ported the code of ART, I prefer this version...It's like for the noise a matter of taste. :)

Good night

Jacques

Desmis:
I just added a slider to "Color propagation". It allows to set the initially constant blur level in ART....and find the good settings.

I act on the 4 guidedfilters
* to blur mask
* and the 3 RGB components

The values I set are arbitrary, but show that the tool can be set up as you wish

At blur= 1 - practically no blur, 
With blur = 5 - settings as ART

just for information, but for me, no problem we can "revert" the change  :)

Jacques

heckflosse:
@Desmis Jacques, can I push some changes to your branch today?

Desmis:
@heckflosse 

Ingo

No problem you can :)

Jacques

heckflosse:
@Desmis Please test :-)

Desmis:
@heckflosse 

Good idea, with one slider whe have the 2 algorithms :)


If I look for image A7C02846.ARW, when we go from 0 to 1... Quasi no differences in images structure, less artiacts with 1, but 50% slower...

Thank you

Jacques

heckflosse:
This ugliness (height of combobox) needs to be fixed
grafik

heckflosse:
@Desmis Jacques, can I push again?

Desmis:
No problem, i am not at home :)

Le jeu. 11 févr. 2021 14:02, Ingo Weyrich <notifications@github.com> a
écrit :

heckflosse:
@Desmis Jacques, I didn't refer to the Label of the combobox, but to its height, which is too large. I reverted `Met` to `Method` ...

Desmis:
Ok
Jacques

Le jeu. 11 févr. 2021 15:04, Ingo Weyrich <notifications@github.com> a
écrit :

Desmis:
@heckflosse 
Ingo
I just come back home...
All works fine :)

Thank you
Jacques

heckflosse:
@Desmis Ok to push a speedup today?

Desmis:
@heckflosse 

Ingo

No problem you can :)

jacques

heckflosse:
@Desmis Jacques, please test :)

Desmis:
@heckflosse 
Ingo
I just tested,...
Very, very small differences...in TIF 16 bits...
No visible differences

:)

jacques

Desmis:
@heckflosse 

Ingo
I did other tests on several images, the differences are very small, but I think the transition zones are a little better (AT 400%) with the new one

Jacques

heckflosse:
@Desmis 
> with the new one

You mean with my last commit?

Desmis:
Yes with the last commit...I think it's better.

With my test, small differences in speed-up, but I think better transitions... But small differences...

Jacques

heckflosse:
> With my test, small differences in speed-up

The speedup can be really huge when the clipped region is small (for example only the sun)

Desmis:
Ok. Thank you
Jacques

Le ven. 12 févr. 2021 19:04, Ingo Weyrich <notifications@github.com> a
écrit :

heckflosse:
@Desmis Jacques, if you don't mind I will make some more cleanups and minor speedups at the weekend.

Desmis:
@heckflosse 
Ingo

 I made a modification for blur=1 (radius )...In this case I don't call GuidedFilter for RGB. Very few differences, but speedup about 30%
I also clean GUI code

No problem, you can make some more cleanups and minor speedups.

:)

jacques

Desmis:
@heckflosse 
Ingo
I push a change to harmonize "events"...I change placement of "EvHlbl" from procevent.h and refreshmap.cc to Tonecurce.cc
Of course no change in performances...

I think I have finished all changes... :)

Jacques

heckflosse:
Imho now the engine code is good. The slowness of the old code is fixed. The code is backwards compatible. Ok, we have one more adjuster, but as it's for less artifacts, it's also good.

Remaining point is the ugliness (height) of the combobox in gui.

@Thanatomanic Could you help on this?

Desmis:
@heckflosse  @Thanatomanic 
A possible solution with this patch....In this case , the slider is under the "method"

[gui.txt](https://github.com/Beep6581/RawTherapee/files/5980573/gui.txt)

jacques

heckflosse:
@Desmis Jacques, I tested your patch. If you apply neutral profile from file browser and then open that file, it does not work correctly

grafik

grafik

Thanatomanic:
@heckflosse @Desmis I'll take a look at the GUI tonight and see if I can help out

Desmis:
@Thanatomanic  @heckflosse 
The same patch, but modified....
But GUI is not at all my cup of tea :)
[gui1.txt](https://github.com/Beep6581/RawTherapee/files/5981709/gui1.txt)

jacques

Desmis:
Another fix GUI bad behavior

[gui2.txt](https://github.com/Beep6581/RawTherapee/files/5982207/gui2.txt)

jacques

Desmis:
@heckflosse  @Thanatomanic 
My precedent patch...was not the good one...

Now 
[gui3.txt](https://github.com/Beep6581/RawTherapee/files/5982323/gui3.txt)

and more
[gui4.txt](https://github.com/Beep6581/RawTherapee/files/5982760/gui4.txt)

heckflosse:
@Desmis Jacques, I pushed your gui4.patch with some additional changes. Please test.

rtgui/tonecurve.cc
```diff
@@ -91,22 +91,31 @@ ToneCurve::ToneCurve() : FoldableToolPanel(this, "tonecurve", M("TP_EXPOSURE_LAB
...
     hrenabled->set_tooltip_markup(M("TP_HLREC_ENA_TOOLTIP"));
-    pack_start(*hrenabled);
 
     method = Gtk::manage(new MyComboBoxText());
     method->append(M("TP_HLREC_LUMINANCE"));
     method->append(M("TP_HLREC_CIELAB"));
     method->append(M("TP_HLREC_COLOR"));
     method->append(M("TP_HLREC_BLEND"));
+    Gtk::VBox *hrVBox;
+    hrVBox = Gtk::manage ( new Gtk::VBox());
```
Thanatomanic: Please replace this with `Gtk::Box` and `new Gtk::Box(Gtk::ORIENTATION_VERTICAL)`

heckflosse: @Thanatomanic Ok, sorry for missing that :(

Thanatomanic: No worries, that's what reviews are for 😄

heckflosse:
I will try to solve the merge conflicts now...

Desmis:
@heckflosse  @Thanatomanic 

I tested, all seems to work fine :)

Jacques

heckflosse:
@Thanatomanic @Desmis please review

Desmis:
@heckflosse 

Ingo
Compile fine :)

All seems to work well :)

Jacques

Thanatomanic:
@Desmis @heckflosse I think the solution for the GUI is fine for a merge, but it does clutter the Exposure panel a bit more. Maybe it's worthwhile to move things into a separate module at some point, seeing as the tool has gained some functionality.

There does seem to be a bug though where the slider moving from 1 to 2 doesn't so anything. Left is with Blur = 1, Right is Blur = 2
image

But there is difference between 2 and 3.
image

Is it by design that the slider only works for integer values, but you can actually set the slider in between?

heckflosse:
@Thanatomanic 

> Is it by design that the slider only works for integer values, but you can actually set the slider in between?

How can you set the slider in between integer values?

Do you mean, that the slider can be set between for example value 1 and value 2 but represents the value closer to one of this values? Then it's the same behaviour as the iterations slider in raw ca correction...

Desmis:
@Thanatomanic  @heckflosse 

There is a difference between 1 and 2 but it is small, generally not always visible (unless you use layers...and in all cases it's visible)
I have just removed the value "2" and limited the slider to 4 

Difficult to replace integer by float....because algorithm has several variable stages (4 Guidedfilters - one on mask and others on RGB values...)

Jacques

Thanatomanic:
In multiple images I found no visible differences between 1 and 2 (but you are right, there were pixel differences), so reducing the levels is a good move.

As for the 'float' vs 'int' it is as Ingo says, I was confused by the intermediate positions of the slider and integer values, but I see now that this is actually expected behavior.

@Desmis for me this would be good for a merge once the benchmark code is removed or put in a verbose block. :+1:

Desmis:
@Thanatomanic 
Benchmak code, removed :)

Jacques

Desmis:
Who merges this PR? if everyone agrees...

jacques

Thanatomanic:
Let's wait for a final :+1: from Ingo, then you can go ahead Jacques! I would again advocate for a 'squash and merge', because the number of changes is quite manageable - and the chances of introducing bugs elsewhere are very small.

heckflosse:
:+1: for merge or squash and merge

Desmis:
@heckflosse  @Thanatomanic 
I just "squash and merge"

jacques

Issue #6119, opened 2021-02-19
make iplocallab.cc a cppcheck and also -Wdouble-promotion clean


heckflosse:
I would like to make iplocallab.cc a bit more cppcheck clean at the weekend.
@Desmis Jacques, can you avoid changes to iplocallab.cc until monday?

Desmis:
@heckflosse 
Ingo

No problem at all 

In addition, I would not be available for a week, because my family is coming to visit us in Fréjus.

jacques

heckflosse:
@Desmis 
> my family is coming to visit us in Fréjus.

lucky you. In Germany that's not allowed atm

Desmis:
@heckflosse 
Ingo

Yes, but before,  all my family do an anti covid test
jacques

heckflosse:
I changed the topic title because I will try now to make iplocallab.cc also -Wdouble-promotion clean

PR #6122, opened 2021-02-20
make iplocallab.cc cppcheck and also -Wdouble-promotion clean #6119

heckflosse:


Desmis:
@heckflosse 
Ingo

Just before my grandchildren arrive, I just tested on several images with several processes (color and light, masks, log encoding, ciecam, wavelet, denoise...).
Compiles fine.
Run fine
No visible differences
All seems to work well.

Very  very small differences in 16 bits TIFF
:)
jacques

Thanatomanic:
Sorry for not having reviewed, but :+1: for the cleaning action.

> Very very small differences in 16 bits TIFF

This is a little disconcerting though.

heckflosse:
> This is a little disconcerting though.

I changed a std::pow in a loop to a pow_F. That explains very very small differences

PR #6126, opened 2021-02-22
Fix GCC warning about comparing signed and unsigned integers.

adamreichold:
GCC 10.2 seems somewhat unhappy about
```
../rtengine/array2D.h:187:39: warning: comparison of integer expressions of different signedness: 'int' and 'std::vector<float*, std::allocator<float*> >::size_type' {aka 'long unsigned int'} [-Wsign-compare]
  187 |         assert((index >= 0) && (index < rows.size()));
      |                                ~~~~~~~^~~~~~~~~~~~~~
```
which triggers a lot of warnings as `array2D.h` is transitively included in many translation units.

Casting `index` to a wider unsigned type should work as a fix since we check for non-negativity first.

### DIFF
--- a/rtengine/array2D.h
+++ b/rtengine/array2D.h
@@ -184,13 +184,13 @@ public:
     // use with indices
     T * operator[](int index)
     {
-        assert((index >= 0) && (index < rows.size()));
+        assert((index >= 0) && (std::size_t(index) < rows.size()));
         return rows[index];
     }
 
     const T * operator[](int index) const
     {
-        assert((index >= 0) && (index < rows.size()));
+        assert((index >= 0) && (std::size_t(index) < rows.size()));
         return rows[index];
     }
 


heckflosse:
@adamreichold Thanks for the pr. This warnings went under my radar because they only occur in debug builds. :+1: for merge

adamreichold:
> +1 for merge

@heckflosse What is the current procedure for merging? Should I press the button myself or is there someone who does all the merging? (I am somewhat surprised to still have a commit bit set in any case.)

Thanatomanic:
@adamreichold I have not had the pleasure of being around when you were actively contributing to RT. Welcome back! I wonder the same about your commit rights, but I guess you can press the button, seeing the very limited scope of your PR :+1:

heckflosse:
@adamreichold As @Thanatomanic wrote, just press the button. Afaik RT has no rules, when (after which time of inactivity) to remove commit rights. For this reason you still have them.

PR #6127, opened 2021-02-22
A few more minor fixes geared towards lgtm.com alerts

adamreichold:
The lgtm.com alert on `loadedProfileData` is actually a false positive AFAIU as the allocation function is indirectly tracked via `loadedProfileDataJpg`, but code that is easier analyse statically should also be easier to understand and maintain for human readers. (Using `operator new` in `read_icc_profile` might be considered controversial and there should be no issue with resolving it the other way around, i.e. allocate `loadedProfileData` using `malloc` in the other two cases.)

Concerning the leaky catch, I am not really sure it is catching anything at all at this point as the only two instances of `throw "some-error-message-as-a-string-literal";` left in the code base are in the constructor `LCPProfile::LCPProfile` but I am not completely certain that `ImProcFunctions::lab2*` does not call that constructor indirectly somehow.

### DIFF
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -1646,8 +1646,10 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
 
                 // Computing the internal image for analysis, i.e. conversion from WCS->Output profile
                 delete workimg;
+                workimg = nullptr;
+
                 workimg = ipf.lab2rgb(nprevl, 0, 0, pW, pH, params->icm);
-            } catch (char * str) {
+            } catch (std::exception&) {
                 return;
             }
         }
@@ -1722,6 +1724,7 @@ void ImProcCoordinator::freeAll()
         }
 
         delete workimg;
+        workimg = nullptr;
 
     }
 
--- a/rtengine/canon_cr3_decoder.cc
+++ b/rtengine/canon_cr3_decoder.cc
@@ -843,7 +843,7 @@ inline void crxFillBuffer(CrxBitstream* bitStrm)
 #endif
 
             if (bitStrm->curBufSize < 1) {  // nothing read
-                throw std::exception();
+                throw std::runtime_error("Unexpected end of file in CRX bitstream");
             }
 
             bitStrm->mdatSize -= bitStrm->curBufSize;
--- a/rtengine/iccjpeg.cc
+++ b/rtengine/iccjpeg.cc
@@ -18,7 +18,7 @@
  */
 
 #include "iccjpeg.h"
-#include <cstdlib>          /* define malloc() */
+#include <new>
 
 
 /*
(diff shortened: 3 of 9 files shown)


heckflosse:
@adamreichold Thanks for the pr. Really great to see you contributing again!

adamreichold:
> Concerning the leaky catch, I am not really sure it is catching anything at all at this point as the only two instances of throw "some-error-message-as-a-string-literal"; left in the code base are in the constructor LCPProfile::LCPProfile but I am not completely certain that ImProcFunctions::lab2* does not call that constructor indirectly somehow.

The fix also did not work out as lgtm.com does not seem to care about the type of pointer used. Considering the type of things we throw elsewhere: `Glib::FileError`, `Options::Error` and `std::exception` which all derive from `std::exception`, I think I go for throwing `std::runtime_error` in `LCPProfile::LCPProfile` and catching `std::exception&` here as the language gods intended...

Thanatomanic:
@Floessie Could you review and/or approve?

rtengine/improccoordinator.cc
```diff
@@ -1646,8 +1646,10 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
 
                 // Computing the internal image for analysis, i.e. conversion from WCS->Output profile
                 delete workimg;
+                workimg = nullptr;
```
Floessie, line 1649: Maybe it's possible to convert `workimg` to a `std::unique_ptr<>` to get rid of the `delete`?

Floessie (approved):
Sorry, I completely forgot about @heckflosse's request. Thanks for nudging, @Thanatomanic.

All fine, @adamreichold. :+1: Great to see you again!

Floessie:
Also compile-tested on Debian Testing i386 with GCC 10.2 and Clang 11.0.

rtengine/improccoordinator.cc
adamreichold: Certainly, but I decided against doing it here to keep the PR focussed on certain classes of lgtm.com alerts. There are a bunch of alerts concerning possibly missing deletions which would suggest using `unique_ptr` and friends which I plan to look into in a separate PR.

adamreichold:
@Thanatomanic Could you merge? I dropped my commit bit because I don't think I am versed enough in the current code base to have that kind of responsibility. Thanks!

Thanatomanic:
@adamreichold With pleasure!

Floessie:
Maybe for another PR:

```patch
index 9bd4f1962..a1e778e45 100644
@@ -60,7 +60,6 @@ ImProcCoordinator::ImProcCoordinator() :
     nprevl(nullptr),
     fattal_11_dcrop_cache(nullptr),
     previmg(nullptr),
-    workimg(nullptr),
     ncie(nullptr),
     imgsrc(nullptr),
     lastAwbEqual(0.),
@@ -1645,10 +1644,7 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
                 ipf.lab2monitorRgb(nprevl, previmg);
 
                 // Computing the internal image for analysis, i.e. conversion from WCS->Output profile
-                delete workimg;
-                workimg = nullptr;
-
-                workimg = ipf.lab2rgb(nprevl, 0, 0, pW, pH, params->icm);
+                workimg.reset(ipf.lab2rgb(nprevl, 0, 0, pW, pH, params->icm));
             } catch (std::exception&) {
                 return;
             }
@@ -1723,7 +1719,6 @@ void ImProcCoordinator::freeAll()
             delete previmg;
         }
 
-        delete workimg;
         workimg = nullptr;
 
     }
@@ -1767,7 +1762,7 @@ void ImProcCoordinator::setScale(int prevscale)
 
         //ncie is only used in ImProcCoordinator::updatePreviewImage, it will be allocated on first use and deleted if not used anymore
         previmg = new Image8(pW, pH);
-        workimg = new Image8(pW, pH);
+        workimg.reset(new Image8(pW, pH));
 
         allocated = true;
     }
diff --git a/rtengine/improccoordinator.h b/rtengine/improccoordinator.h
index 5c17f36b8..a4934e564 100644
--- a/rtengine/improccoordinator.h
+++ b/rtengine/improccoordinator.h
@@ -66,7 +66,7 @@ protected:
     LabImage *nprevl;
     Imagefloat *fattal_11_dcrop_cache; // global cache for ToneMapFattal02 used in 1:1 detail windows (except when denoise is active)
     Image8 *previmg;  // displayed image in monitor color space, showing the output profile as well (soft-proofing enabled, which then correspond to workimg) or not
-    Image8 *workimg;  // internal image in output color space for analysis
+    std::unique_ptr<Image8> workimg;  // internal image in output color space for analysis
     CieImage *ncie;
 
     ImageSource* imgsrc;
```

HTH,
Flössie

Issue #4338, opened 2018-01-27
Improve accuracy of Y to L conversion


heckflosse:
For low Y values the conversion from Y to L is inaccurate.

Here's a patch with test code. It converts L to Y and back again 100 times and shows before after values.
To get the output, just open a raw file in editor.

```diff
diff --git a/rtengine/rawimagesource.cc b/rtengine/rawimagesource.cc
index 04066d49..7299169f 100644
--- a/rtengine/rawimagesource.cc
+++ b/rtengine/rawimagesource.cc
@@ -4623,6 +4623,30 @@ void RawImageSource::getAutoExpHistogram (LUTu & histogram, int& histcompr)
 // Histogram MUST be 256 in size; gamma is applied, blackpoint and gain also
 void RawImageSource::getRAWHistogram (LUTu & histRedRaw, LUTu & histGreenRaw, LUTu & histBlueRaw)
 {
+    float a = 0;
+    float b = 0;
+    
+    for (float L = 0; L < 2.f; L += 0.1f) {
+        float newL = L;
+        float x,y,z;
+        for (int i = 0; i < 100; ++i) {
+            Color::Lab2XYZ(newL,a,b,x,y,z);
+            Color::XYZ2Lab(x,y,z,newL,a,b);
+        }
+        printf("Lold = %.9f\tLnew = %.9f\n",L,newL);
+    }
+
+    for (float L = 2.f; L < 33000.f; L *= 2.f) {
+        float newL = L;
+        float x,y,z;
+        for (int i = 0; i < 100; ++i) {
+            Color::Lab2XYZ(newL,a,b,x,y,z);
+            Color::XYZ2Lab(x,y,z,newL,a,b);
+        }
+        printf("Lold = %.9f\tLnew = %.9f\n",L,newL);
+    }
+
+    std::cout << std::endl;
 //    BENCHFUN
     histRedRaw.clear();
     histGreenRaw.clear();

```

And here's the output:

```
Lold = 0.000000000      Lnew = 0.044082642
Lold = 0.100000001      Lnew = 0.144088745
Lold = 0.200000003      Lnew = 0.244094849
Lold = 0.300000012      Lnew = 0.344100952
Lold = 0.400000006      Lnew = 0.444107056
Lold = 0.500000000      Lnew = 0.544113159
Lold = 0.600000024      Lnew = 0.644119263
Lold = 0.700000048      Lnew = 0.700317383
Lold = 0.800000072      Lnew = 0.800323486
[...]

PR #6128, opened 2021-02-22
make improcfun.cc cppcheck and -Wextra clean

heckflosse:


adamreichold:
Changes look good to me. (I did wonder whether pulling out `mult` could have effects on the accuracy but did not see an obvious difference to the previous implementation.)

heckflosse:
We could even rewrite this
```
    const float mult = gamm / maxL;
#ifdef _OPENMP
    #pragma omp parallel for
#endif
    for (std::size_t i = 0; i < N; i++)
    {
        L[i] = (L[i] - minL) * mult;
    }
```

to this:

```
  const float mult = gamm / maxL;
  const float add = -minL * mult;
#ifdef _OPENMP
    #pragma omp parallel for
#endif
    for (std::size_t i = 0; i < N; i++)
    {
        L[i] = L[i] * mult + add;
    }
```
That would turn the calculation inside the loop into an FMA instruction (on archs which support FMA)

Thanatomanic:
You can always count on @heckflosse to scrape off one more instruction 😉 

image

adamreichold:
:thinking: We have

> Do not use -ffast-math, it will not make RawTherapee faster but will introduce artifacts.

in `RELEASE_NOTES.txt` and

```c++
#error Using the -ffast-math CFLAG is known to lead to problems. Disable it to compile RawTherapee.
```
in `main.cc` and I think this is good advice as floating-point arithmetic is for example not associative or distributive which is why these transformation are not trivially correct. Does the compiler explorer yield the same result without that flag?

In the case of manually massaging this into a multiply-add, I would worry about cancellation effects, at least without a better understanding of the range of `L[i]`.

Thanatomanic:
Without `-mfma -ffast-math` but keeping `-O3` I get a very different result, but something which is actually _the same for both cases_. So the compiler knows how to handle this, and we don't need to optimize in code if we don't rely on `-ffast-math`.

```asm
fun(float*):
        movaps  xmm2, XMMWORD PTR .LC0[rip]
        movaps  xmm1, XMMWORD PTR .LC1[rip]
        lea     rax, [rdi+2000]
.L2:
        movups  xmm0, XMMWORD PTR [rdi]
        add     rdi, 16
        mulps   xmm0, xmm2
        addps   xmm0, xmm1
        movups  XMMWORD PTR [rdi-16], xmm0
        cmp     rax, rdi
        jne     .L2
        ret
.LC0:
        .long   1072858812
        .long   1072858812
        .long   1072858812
        .long   1072858812
.LC1:
        .long   -1130678398
        .long   -1130678398
        .long   -1130678398
        .long   -1130678398
```

adamreichold:
> Without -mfma -ffast-math but keeping -O3 I get a very different result, but something which is actually the same for both cases.

But shouldn't we keep the `-mfma` flag to allow the compiler to use that instruction and see if it is able to use it in the second but no the first case? If we drop the flag, it seems understandable that codegen is the same as the instruction is not on the table so to speak.

heckflosse:
Neither using `-mfma` nor using `-ffast-math` is what we want.
Either we make generic builds, where `-mfma` is not set, or native builds, where the compiler sets the correct flags for the target architecture.

For my personal usage, I always make native builds, which in my case equals to -march=bdver2.

Here's the assembly for that case.
grafik

The compiler uses fma instrucions, knows, that using avx on a Bulldozer is not good, unrolls the loop and even inserts a prefetch instruction.

Conclusion: Trust the compiler.

adamreichold:
> Either we make generic builds, where -mfma is not set, or native builds, where the compiler sets the correct flags for the target architecture.

I think the discussion was purely within the context of using the compiler explorer to look into the resulting codegen, not which flags to use to create release builds. In this particular context, using only `-mfma` instead of `-march` makes it possible to avoid changing all the other machine parameters which influence codegen to allow a more targeted comparison.

In any case, I am glad we are sticking to `mult` without `add`...

heckflosse:
> In any case, I am glad we are sticking to mult without add...

I'm curious why...

adamreichold:
> I'm curious why...

As written in https://github.com/Beep6581/RawTherapee/pull/6128#issuecomment-783963759

> In the case of manually massaging this into a multiply-add, I would worry about cancellation effects, at least without a better understanding of the range of L[i].

Thanatomanic:
@adamreichold `L[i]` should be limited to 0..1 floats afaik.

adamreichold:
> L[i] should be limited to 0..1 floats afaik.

I think this could still be affected: If `maxL` is close to zero, this means that `mult = gamma / maxL` would be very large and could loose precision. `add = -minL * mult` would then mirror that loss precision into the negative direction so `L[i] * mult + add` could have cancellation artefacts.

I am not completely certain that this will actually happen because `minL <= maxL` should also hold and maybe this prevents any issues. Maybe keeping in line with what was written above, my default conclusion is to _not_ trust algebraic manipulations of floating-point arithmetic.

heckflosse:
> my default conclusion is to not trust algebraic manipulations of floating-point arithmetic.

your conclusion would require to define what's correct in first step...

e.g. : is (a + b) * c correct or is a * c + b * c correct in a certain use case?

In my opinion, for simple calculations like this, it does not matter.
For other cases (large summations for example) we have to care about cancelation effects for sure.

heckflosse:
Anyway, any objections to merge this?

Thanatomanic:
Not from me :+1:

heckflosse:
@adamreichold Adam, to show you, that I also care about accuracy, have a look at this old issue ;-)

https://github.com/Beep6581/RawTherapee/issues/4338

adamreichold:
> your conclusion would require to define what's correct in first step...

I think that when changing a shipping code base, we need to prove that our changes are improvements. Either by showing the current implementation is defective somehow. Or by showing how the new implementation is superior. Or more succinct, the current implementation is considered correct until we can reasonably argue that it is not.

> Adam, to show you, that I also care about accuracy, have a look at this old issue ;-)

I think that I have never said anything to the contrary.

> Anyway, any objections to merge this?

No.

heckflosse:
> I think that I have never said anything to the contrary.

I agree. But I also think, that small arithmetic changes (replacing a division with a mult or for example replacing a divisison by 65535 with a division by 65536 (which turns the division into a multiplication at -O3) can be worth it for faster processing without getting worse results.

heckflosse:
@adamreichold A good example for changing calculations is the current RCD code, which is much faster than the old code. This was a collaboartion of the original rcd author (Luis Sanz Rodríguez), Hanno from darktable, and me. We worked that out in a private communication without discussing the details on RT github because not only RT was involved, but also darktable, and the new code is now also in Siril and librtprocess. We didn't write a paper why the new code is simpler and faster. We simply agreed about it while collaborating...

PR #6133, opened 2021-02-24
Make some more files cppcheck clean

heckflosse:


rtengine/dcrop.cc
```diff
@@ -1128,53 +1128,17 @@ void Crop::update(int todo)
     }
 
     if (todo & M_RGBCURVE) {
-        Imagefloat *workingCrop = baseCrop;
```
Thanatomanic: Please see https://github.com/Beep6581/RawTherapee/pull/5934. I don't think removing this is wise atm.

heckflosse: Ok, I will revert the change to dcrop.cc

rtengine/ipwavelet.cc
```diff
@@ -2265,11 +2263,7 @@ void ImProcFunctions::ip_wavelet(LabImage * lab, LabImage * dst, int kall, const
                                 absciss = asig * std::fabs(tempwav) + bsig;
                             } else {
                                 absciss = amean * std::fabs(tempwav);
-                                float k = sig;
-                                if(sig > 1.f) {
-                                    k = SQR(sig);
-                                }
-                                float abs = pow(2.f * absciss, (1.f / k));
+                                float abs = pow(2.f * absciss, (1.f / SQR(sig)));
```
Thanatomanic, line 2266: It looks as if the value of sig is always 2.0f... so what is going on anyway?

heckflosse: good catch, looking

heckflosse: @Thanatomanic Not sure, that my change (sqrt(sqrt(...))) is what you had in mind, but it's fast ;-)

heckflosse: @Thanatomanic My intention was to get the files cppcheck clean. It we want them to get perfectly clean (not only cppcheck...) we will get from 100th to 1000th issue and will not be able to do this for 5.9 ;-)

rtengine/ipwavelet.cc
```diff
@@ -2205,7 +2203,6 @@ void ImProcFunctions::ip_wavelet(LabImage * lab, LabImage * dst, int kall, const
 
             Evaluate2(*LdecompLL, meang, meanNg, sigmag, sigmaNg, MaxPg, MaxNg, wavNestedLevels);
             Evaluate2(*Ldecompdst, mean, meanN, sigma, sigmaN, MaxP, MaxN, wavNestedLevels);
-            float sig = 2.f;
```
adamreichold: While it does not change the resulting values, I think readability is reduced by inlining this instead of e.g. making it a `constexpr`. I am now unable to see that this is always the "same two" in the following places where it was inlined.

I understand this would prevent replacing `pow` by `sqrt(sqrt(..))` but I think it is not obvious that the gain in speed is wort the loss of readability. (IIRC, compilers even replace `pow(x, y)` by optimized variants for specific `y` even though I am not sure if `y=2^{-n}` would qualify.)

adamreichold: It seems like this computation is not done in a loop, e.g. per-pixel, so speed gains are probably inconsequential.

heckflosse: > It seems like this computation is not done in a loop, e.g. per-pixel, so speed gains are probably inconsequential.

The pow() was inside a pixel loop

```
                        for (int i = 0; i < Wlvl_L * Hlvl_L; i++) {
```

adamreichold: > The pow() was inside a pixel loop

Yes, you are right. So it is not inconsequential. Just a trade-off.

heckflosse: > IIRC, compilers even replace pow(x, y) by optimized variants for specific y

I just checked using godbolt. At least gcc does a call to std::pow in this case

heckflosse: while a sqrt(sqrt( is just two branch free instructions

adamreichold: > while a sqrt(sqrt( is just two branch free instructions

I am not arguing that the current code will reliably be as a fast as the proposed change. I am questioning whether loosing the information that `sig` is a parameter of the algorithm that is used in multiple places is worth the gain. (That both of us do not seem to have measured ATM.)

heckflosse: grafik

heckflosse: > (That both of us do not seem to have measured ATM.)

I will measure it tomorrow. I guess it will speedup a little bit as the new code may be auto-vectorized by the compiler while the old code (calling std::pow()) is not.

We'll see.

Also, my first cppcheck fix for this file didn't contain this `optimization` at all. I just wanted to make the file cppcheck clean...

adamreichold: I am not sure which point the screenshot is making? You mean that there is a `call` in there? I already agreed that this can happen even if `sig` is a `constexpr`. We are still guessing as to how much faster this ends up being on actual images.

It is also not the point I am personally trying to make. I think we should acknowledge that information is lost with this code change and that changing this algorithm in the future will be harder. I would be fine with saying that the two instructions versus the call is definitely worth it, but a screenshot does not do that.

heckflosse: > I am not sure which point the screenshot is making?

The point is that a call to std::pow will not be vectorized by -O3

But the std::sqrt can be vecorized (as you can see in the `sqrtps` instructions in assembly, which processes 4 floats at once...

heckflosse: we should simply ask @Desmis about the meaning of sig = 2.f in this code...

adamreichold: > But the std::sqrt can be vecorized (as you can see in the sqrtps instructions in assembly, which processes 4 floats at once...

So is this the point where you would say that the loss of information is definitely worth it?

Thanatomanic: @adamreichold @heckflosse You/we could start a more general discussion on coding standards and when to split hairs. Could be worthwhile. There is even a tool for that in GitHub https://github.com/Beep6581/RawTherapee/discussions

heckflosse: > So is this the point where you would say that the loss of information is definitely worth it?

Without knowing about the informational value of `sig` we don't know if there is a loss of information at all.
Let's wait for @Desmis reply and not split hairs...

adamreichold: Personally, I would be most happy with continuing along the way of

> Also, my first cppcheck fix for this file didn't contain this optimization at all. I just wanted to make the file cppcheck clean...

Making the file cppcheck-clean does not require dropping `sig`, inquiring the original author, performing benchmarks or changing coding standards. I think we could merge this quickly.

adamreichold: > Without knowing about the informational value of sig we don't know if there is a loss of information at all.
Let's wait for @Desmis reply and not split hairs...

Waiting is fine with me. Concerning hair, I actually did not see that second comment before replying to its predecessor.

heckflosse: > Making the file cppcheck-clean does not require dropping sig, inquiring the original author, performing benchmarks or changing coding standards. I think we could merge this quickly.

I agree. I didn't even get rid of sig in my first post...

Desmis:
@heckflosse @adamreichold  @Thanatomanic 

float sig = 2, has the same role as "attenation response" in many parts of Wavalet Levels , but also of iplocallab.cc... 
iI changes the distribution of the decomposition by taking into account the "mean" of the signal and its standard deviation (which are not Gaussian...it's a model)
https://rawpedia.rawtherapee.com/Wavelet_Levels#The_attenuation_curve

in this case, to simplify the interface I set a constant value of 2 - normally with a slider it varies between 0.05 and 3.
Of course you can change the code, but it is important to keep track of the choice sig=2, for example with a "define" declaration, or something else.

jacques

heckflosse:
@adamreichold 

> I think it is not obvious that the gain in speed is wort the loss of readability.

I measured using a 45.6 MP file.
The loop is executed 18 times in my testcase.
Original code: 18 times ~290 ms = 5.52 seconds
Code using pow_F instead of std::pow: 18 times ~100 ms = 1.8 seconds
Code using sqrt(sqrt(x)): 18 times ~60 ms = 1.08 seconds

adamreichold:
> I measured using a 45.6 MP file.

Considering your measurements and what Jacques wrote above, the second option seems to be a good compromise?

heckflosse:
I will create an issue for further discussion on optimizing this part of th wavelet code

Issue #6132, opened 2021-02-24
Local Adjustments + Exposure = Local Contrast display issue


chaimav:
Toggling local contrast compounds the contrast on the displayed image.
To reproduce, use the raw in this post here https://www.dpreview.com/forums/thread/4546372 
-Add a full image local spot to the haze below the fireworks
-Add Dynamic Range & Exposure tool
-Increase exposure of haze
-Toggle local contrast on and off (the more it is toggled, the more severe the issue)
PP3: https://drive.google.com/file/d/1wtoq4lRb1vobLwML-BwObLRRbqikHGVL/view?usp=sharing

RawTherapee 5.8-2773. Windows 10, i3 8100

![local contrast bug](https://user-images.githubusercontent.com/47765782/109004670-0e6ad400-7677-11eb-8261-a7396fdec296.JPG)

chaimav:
Let me know when the fix is available in the nightly so I can recheck it.

chaimav:
RT 5.8-2804
I have found the same issue with checking and un-checking 'Look table' under Color Management >Auto matched camera profile > DCP
![look table](https://user-images.githubusercontent.com/47765782/110006117-3e554f80-7ce7-11eb-99b8-d760d7c8c353.JPG)
I can produce the bug using the raw available here https://discuss.pixls.us/t/hazy-snowy-bridge/23605

Thanatomanic:
@chaimav Do you mean that you start from Neutral and only toggle this checkbox? I cannot seem to reproduce that with the raw supplied.

chaimav:
No. You need to add a full image Local Adjustment and change the exposure first. Change the extension of this file to pp3
[look table bug.txt](https://github.com/Beep6581/RawTherapee/files/6086076/look.table.bug.txt)

Lawrence37:
I can confirm the pull request fixes the issue for local contrast but not for the look table.

Thanatomanic:
@chaimav Could you try out the changes that have been made in `dev`? Is this issue resolved?

chaimav:
> @chaimav Could you try out the changes that have been made in `dev`? Is this issue resolved?

The issue with _local contrast_ has been resolved with Dev 5.8-2816. However, the bug with _look table_ has not. I also noticed that the Color Management >Auto matched camera profile > DCP > [] Tone curve checkbox has the same issue as look table.

chaimav:
I have located yet another place where what I believe is the same bug shows up.
[...]

PR #6134, opened 2021-02-25
Local contrast  - Try to solve bug issue #6132

Desmis:
This bug seems very old in principle. Some time ago, Ingo @heckflosse  and I tried to solve the incompatibilities between tone-mapping and "local contrast" 
https://github.com/Beep6581/RawTherapee/issues/6132
Now, it appears with "locallab"...in exposure, but also with all modules when Luminance change

It seems that it's due to placement of "Local contrast" which was in "rgbproc"
I change the placement, now, "Local contrast" is in Improccoordinator, dcop, simpleprocees, rtthumbnail just after Colorregion.
Of course I change the "events"

It seems to work... To verify

@heckflosse  @Thanatomanic  could you review :)

Thank you

Jacques


Desmis:
If nobody objects, I will merge this PR tomorrow :)

jacques

Issue #6137, opened 2021-02-25
Speedup for wavelets/final touchup/local contrast


heckflosse:
There is a nice speedup possible for wavelets/final touchup/local contrast.

I will use http://rawtherapee.com/shared/test_images/clematis_vitalba.nef for my tests using the settings of this screenshot
![grafik](https://user-images.githubusercontent.com/13733487/109216819-52fc8980-77b5-11eb-8196-b878080db617.png)

I will make a draft to discuss changes and compare processing times.

heckflosse:
I added some stopwatches for initial measure of processing times.
Tomorrow I will add optimized code.

heckflosse:
On my machine, with current code, processing times are

```
Evaluate2 took 189 ms
level loops took 1453 ms
final touchup took 2117 ms
```

where `final touchup` includes `Evaluate2` and `level loops`

heckflosse:
Using optimized code https://github.com/Beep6581/RawTherapee/pull/6138/commits/10e0ab4eaaa784ec86437aae00d88d030d4ecea9:

```
evaluate took 191 ms
level loops took 257 ms
Final touchup took 923 ms
```

heckflosse:
Imho the optimized code readability is not worse than the original code readability, maybe even better because of the `intp`...

Desmis:
:)

jacques

PR #6138, opened 2021-02-25
Speedup for wavelets/final touchup/local contrast

heckflosse:
This is a draft for #6137

### DIFF
--- a/rtengine/ipwavelet.cc
+++ b/rtengine/ipwavelet.cc
@@ -2203,12 +2203,13 @@ void ImProcFunctions::ip_wavelet(LabImage * lab, LabImage * dst, int kall, const
             Evaluate2(*LdecompLL, meang, meanNg, sigmag, sigmaNg, MaxPg, MaxNg, wavNestedLevels);
             Evaluate2(*Ldecompdst, mean, meanN, sigma, sigmaN, MaxP, MaxN, wavNestedLevels);
             constexpr float sig = 2.f;
-            /* original code for variable sig
+
+            // original code for variable sig
             float k = sig;
             if (sig > 1.f) {
                 k = SQR(sig);
             }
-            */
+
             float thr = 0.f;
             if (thrend < 0.02f) thr = 0.5f;
             else if (thrend < 0.1f) thr = 0.2f;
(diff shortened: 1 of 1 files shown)


rtengine/ipwavelet.cc
```diff
@@ -2216,67 +2221,61 @@ void ImProcFunctions::ip_wavelet(LabImage * lab, LabImage * dst, int kall, const
...
-                                float abs = pow(2.f * absciss, (1.f / SQR(sig)));
-                                absciss = 0.5f * abs;
-                            }
-                            float kc = wavguid.getVal(absciss) -1.f;
-
-                            if(kc < 0) {
-                                kc = -SQR(kc);//approximation to simulate sliders denoise
+                                absciss = amean * tempwav;
+                                /*
+                                if (sig == 2.f) { // for sig = 2.f we can use a faster calculation because the exponent in this case is 0.25
```
adamreichold, line 2265: If we go for commented-out code, I would prefer to have a static assert here to check the value of `sig`.

adamreichold: Thinking about this some, we really should not do this: The comments are not compiled and will just rot away. Having the slow path will not affect the fast path at all as it should be removed by the optimizer. I think the the initial approach posted was superior.

Thanatomanic: I agree about comment-rot, we have a lot of that tbh. If anyone says/proves that the old, more verbose, way does not result in a significant slowdown, it should be kept. My argument would be that a lot of algorithms we use are not so well documented. Doing 'smart' optimizations in the code might lead to semantic-rot, which is worse imo. 
Edit: if there is a meaningful speed-up, I would actually prefer to keep the comments in. Development is currently done incrementally, tackling and improving little pieces here and there. Having some comments with old or alternative code might actually be a helpful reminder that something could still improve.

heckflosse: a static assert of a constexpr which is defined some lines before?

> If anyone says/proves that the old, more verbose, way does not result in a significant slowdown, it should be kept.

So I will keep it. The only downside is, that the file now is not cppcheck clean.

heckflosse:
Any objections to merge this?

rtengine/ipwavelet.cc
```diff
@@ -2216,67 +2223,57 @@ void ImProcFunctions::ip_wavelet(LabImage * lab, LabImage * dst, int kall, const
...
-                            }
-                            float kc = wavguid.getVal(absciss) -1.f;
-
-                            if(kc < 0) {
-                                kc = -SQR(kc);//approximation to simulate sliders denoise
+                                absciss = amean * tempwav;
+                                if (sig == 2.f) { // for sig = 2.f we can use a faster calculation because the exponent in this case is 0.25
+                                    absciss = 0.5f * std::sqrt(std::sqrt(absciss));
+                                } else { // original code for variable sig
+                                    absciss = 0.5f * pow_F(absciss, 1.f / k);
```
adamreichold, line 2268: I may not be following the transformation correctly, but isn't there a factor two missing before the calls to `pow_F` and `sqrt` that was in the old code:
```c++
absciss = amean * std::fabs(tempwav);
float abs = pow(2.f * absciss, (1.f / SQR(sig)));
absciss = 0.5f * abs;
```

heckflosse: That factor of two went into an outside of loop calculation

Before it was
```
float amean = 0.5f / (tempmean);
```
Now it is
```
const float amean = 1.f / (tempmean);
```

adamreichold: Thanks! No objections then.

Thanatomanic:
Go ahead for a merge! :+1:

PR #6145, opened 2021-02-28
speedup for a loop in wavelet code, fixes #6143

### LINKED ISSUE #6143: Possible speedup for a loop in wavelet code
In wavelet code there is still one pixel-loop which uses std::pow instead of the much faster pow_F
https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/ipwavelet.cc#L1393

@Desmis Jacques, can you provide a pp3 which calls this loop? I will test and optimize then.

heckflosse:


adamreichold:
@heckflosse It has nothing to do with the code, but it would be nice if you could put the issue ID into the PR description instead of the commit message using something like "Fixes #XYZ" or "Closes #XYZ". The has the benefit that GitHub will automatically link the PR instead of the individual commits on the issue page. Thereby, all comments are collected here instead of accidentally ending up on the commits. It also implies that the issue is not pinged repeatedly if you force-push the branch. Traceability is still maintained by including the issue number in the merge commit message. It also has the additional benefit of automatically closing the issue on merge. (I learned about this recommendation at rust-lang/rust and found that it really helps keeping busy threads aligned.)

Thanatomanic:
@adamreichold Are you sure this works if you just put it in the title? Normally GitHub convert references to issues to links, but that does not work in titles here? I think that when you merge (or merge-squash) and make the commit message "fixes #...." it does close the corresponding issue.

rtengine/ipwavelet.cc
```diff
@@ -1360,131 +1358,112 @@ void ImProcFunctions::ip_wavelet(LabImage * lab, LabImage * dst, int kall, const
...
+                                                } else if(cp.denmet == 3) {
+                                                    if(level == 0 || level == 1) {
+                                                        kcFactor = 1.7f;
+                                                    }
+                                                } else if(cp.denmet == 4) {
+                                                    if(level == 0 || level == 1) {
+                                                        kcFactor = 0.3f;
+                                                    }
+                                                }
+                                                const float k = 1.f / siglh[level] > 1.f ? SQR(siglh[level]) : siglh[level];
```
adamreichold: Any reason for not checking `sighl[level] < 1.f` here?

adamreichold: Wasn't that comparison the other way around in
```c++
float k = siglh[level];
if(siglh[level] > 1.f) {
    k = SQR(siglh[level]);
}
```

adamreichold: Ah, maybe add parentheses like
```c++
const float k = 1.f / (siglh[level] > 1.f ? SQR(siglh[level]) : siglh[level]);
```
(I have no idea how operator precedence works in this case...)

adamreichold:
> Are you sure this works if you just put it in the title? 

Not in the title, it should be part of the pull request _description_ which can be used as the commit message of the merge commit automatically.

rtengine/ipwavelet.cc
adamreichold: Out of curiosity, I tried this and it appears to work out as
```c++
const float k = (1.f / siglh[level] > 1.f) ? SQR(siglh[level]) : siglh[level];
```
without the parentheses. (Same precedence, right-to-left evaluation.)

heckflosse:
Any objections to merge this?

PR #6149, opened 2021-02-28
Local adjustments - Added Guidedfilter to avoid color shift

Desmis:
All is in the titlle

I add a guidefilter on "L", "a", and "b" with a very small radius, to reduce artfiacts in transitions - due to Munsell and Relative colorimetry

@heckflosse 
Ingo, could you review this PR

Thank you :)

Jacques

### DIFF
--- a/rtengine/iplocallab.cc
+++ b/rtengine/iplocallab.cc
@@ -10701,7 +10701,17 @@ void ImProcFunctions::avoidcolshi(struct local_params& lp, int sp, LabImage * or
 #endif
             for (int y = 0; y < bh; y++) {
                 for (int x = 0; x < bw; x++) {
-                    float2 sincosval = xsincosf(xatan2f(transformed->b[y][x], transformed->a[y][x]));
+                    const float Chprov1 = sqrtf(SQR(transformed->a[y][x]) + SQR(transformed->b[y][x]));
+                    float2  sincosval;
+
+                    if (Chprov1 == 0.0f) {
+                        sincosval.y = 1.f;
+                        sincosval.x = 0.0f;
+                    } else {
+                        sincosval.y = transformed->a[y][x] / Chprov1;
+                        sincosval.x = transformed->b[y][x] / Chprov1;
+                    }
+
                     transformed->a[y][x] = 32768.f * blechro[y][x] * sincosval.y;
                     transformed->b[y][x] = 32768.f * blechro[y][x] * sincosval.x;
                 }
--- a/rtengine/iplocallab.cc
+++ b/rtengine/iplocallab.cc
@@ -10498,6 +10498,7 @@ void ImProcFunctions::avoidcolshi(struct local_params& lp, int sp, LabImage * or
         };
 
         float softr = params->locallab.spots.at(sp).avoidrad;//max softr = 30
+        bool muns = params->locallab.spots.at(sp).avoidmun;//Munsell control with 200 LUT
         //improve precision with mint and maxt
         float tr = std::min(4.f, softr);
         float mint = 0.15f - 0.03f * tr;//between 0.15f and 0.03f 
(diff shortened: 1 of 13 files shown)


heckflosse:
@Desmis Jacques, I will test (and vectorize) in the evening

Desmis:
@heckflosse 

Ingo

Thank you
Jacques

heckflosse:
@Desmis Jacques, for the png from above the result is better now.
But for this file there are still artifacts in the yellow regions

raw and pp3
[_DSC0488_nef.zip](https://github.com/Beep6581/RawTherapee/files/6057412/_DSC0488_nef.zip)
[_DSC0488.NEF.pp3.zip](https://github.com/Beep6581/RawTherapee/files/6057423/_DSC0488.NEF.pp3.zip)




left without, right with avoid color shift

grafik

Desmis:
@heckflosse 
I tested with your image and pp3...and no white point as you..curious  ?

But be aware, artefacts are on transitions between  2 areas.. (munsell).  A point (a small area)  with a color desaturated shows that "avoid color shift" works by reducing chroma...if area out of gamut...

Could you test and change the value of "softr = 0.7" for example 2.f

Good night 

Jacques

heckflosse:
> Could you test and change the value of "softr = 0.7" for example 2.f

where is softr in gui?

Desmis:
@heckflosse 
Ingo

I will push a commit this morning with various improvments :)

Jacques

Desmis:
@heckflosse 
Ingo
I just push a change
* I add a slider "soft radius" in GUI and change a bit the GUI
* I change the Color::gamutLchonly for the one of Lab adjustments (best precision)
* I change also parameters "mint" and "maxt"  for this function to best result
Jacques

Desmis:
@heckflosse 
Ingo

I think , I have finished improvments

Jacques

heckflosse:
@Desmis Jacques

> I tested with your image and pp3...and no white point as you..curious ?

I tested with latest code and softrad = 2.0. Still white artifacts in the yellows
grafik

Desmis:
@heckflosse 
Ingo

It's why I add a slider...Yesterday, I say "2" without trying...

In this case, the white gradually fades and from 5, it has practically disappeared....

But, there are no miracles - and it's not a bug - when the colors are out  of gamut, the "soft radius" will make it disappear by "blurring" the neighboring pixels. It's only a question of point of view.


And for me, always the right question. do you want to keep colors out of gamut? Certainly with zero settings, it is not serious, but then we can end up with anything ...

:)

jacques

heckflosse:
@Desmis Jacques, please review my changes

Desmis:
@heckflosse 
Ingo

I looked at your change, they are good :)

I tested on various images : works fine.

Thank you

Jacques

heckflosse:
@Desmis Jacques

:+1: for merge from me. Bet let's wait a day or two for other contributors' opinions

rtengine/iplocallab.cc
```diff
@@ -10636,6 +10645,78 @@ void ImProcFunctions::avoidcolshi(struct local_params& lp, int sp, LabImage * or
...
+            }
+
+            array2D<float> &blechro = ble; // reuse buffer
+#ifdef _OPENMP
+            #pragma omp parallel for schedule(dynamic,16) if (multiThread)
+#endif
+
+            for (int y = 0; y < bh ; y++) {
+                for (int x = 0; x < bw; x++) {
+                    blechro[y][x] = std::sqrt(SQR(transformed->b[y][x]) + SQR(transformed->a[y][x])) / 32768.f;
```
adamreichold, line 10696: Maybe using [`std::hypot`](https://en.cppreference.com/w/cpp/numeric/math/hypot) here and for `Chprov1` helps? (The description suggests better accuracy, but performance could be worse if the square root instruction isn't used.)

rtengine/iplocallab.cc
```diff
@@ -10636,6 +10645,78 @@ void ImProcFunctions::avoidcolshi(struct local_params& lp, int sp, LabImage * or
...
+            constexpr float bepsil = epsilmin;
+            const float epsil = softr < 0.f ? 0.001f : aepsil * softr + bepsil;
+
+            const int bw = transformed->W;
+            const int bh = transformed->H;
+            array2D<float> ble(bw, bh);
+            array2D<float> guid(bw, bh);
+
+#ifdef _OPENMP
+            #pragma omp parallel for schedule(dynamic,16) if (multiThread)
```
adamreichold, line 10670: Are these loops really good candidates for dynamic scheduling? The probability of load imbalance seems low without any control flow.

Desmis:
I have added a checkbox to "avoid color shift". This checkbox allows to use "Munsell control" only...i.e. without doing a gamut control. Which means that in this case, if the box "Munsell correction only" is checked  and "avoid color shift" is checked then only a "Uniform Perceptual Lab" is realized. 
As a reminder, this feature provides a linear response to Lab "drifts", especially in Blue-Purple and Orange, especially when the user increases the chromaticity settings (and of course to a lesser degree the other settings and characteristics of the image)

@adamreichold 
My C++ skills are not enough to make a judgment

jacques

Thanatomanic:
I cannot really comment on either the correctness or the usefulness of this, sorry. So @Desmis and @heckflosse if you're okay, please (squash and) merge.

Desmis:
@Thanatomanic 

All these corrections are small...

Done :)

jacques

Issue #6153, opened 2021-03-04
Local Adjustments: Horizontal field size too large


rebio:
**Short description**
The horizontal size of some elements of the local adjustments is too large and requires horizontal scrolling even there is no need for it (unless sliders have a fixed size)

**Steps to reproduce**
1. Open any image
2. Activate Local Adjustments 
3. Add a new area of effect
4. Add a "Shadows/Highlights & Tone equalizer" section (others like "Common Color Mask" will also work; some like "Color & Light" or "Sharpening" will not cause this)

Side Notes:
1. When you start RT and you do not select an image and go right to the LA-tab the size of the panel is also too large and needs to be scrolled
2. Is it maybe because of the name of the single setting? Maybe the width of the panel is set by the width of the text of the single setting. I'd rather have a small box where I do not have to scroll and maybe have to hover over the text to see the full name.

**Expected behavior**
I would expect the panel to keep the original size. The problem is not only that you have to scroll but also that it's much harder to set the values using the slider because at some point the slider will _jump_. Therefore fine tuning by using the sliders is pretty much impossible.

**Additional information**
RT:
> Version: 5.8-2802-g84751ad26 (2021-03-01 - win64)
> Branch: dev
> Commit: 84751ad26
> Commit date: 2021-03-01
> Compiler: x86_64-w64-mingw32-gcc\ 9.3.0
> Processor: generic x86
> System: Windows
> Bit depth: 64 bits
> Gtkmm: V3.24.3
> Lensfun: V0.3.2.0
> Build type: Release
> Build flags: -mwin32\ -m64\ -mthreads\ -msse2\ -std=c++11\ -mtune=generic\ -Werror=unused-label\ -Werror=delete-incomplete\ -fno-math-errno\ -Wno-attributes\ -Wall\ -Wuninitialized\ -Wcast-qual\ -Wno-deprecated-declarations\ -Wno-unused-result\ -Wunused-macros\ -fopenmp\ -Werror=unknown-pragmas\ -Wno-aggressive-loop-optimizations\ -DNDEBUG\ -O3\ -ftree-vectorize
> Link flags: -m64\ -mthreads\ -static-libgcc\ -mtune=generic\ -s\ -O3
> OpenMP support: ON
> MMAP support: ON
> Build OS: Linux 4.4.0-101-generic x86_64
> Build date: Mon, 01 Mar 2021 23:36:34 +0000 UTC
> Build epoch: 1614641794
[...]

PR #6160, opened 2021-03-07
Local adjustemnts - adapt various parameters to fit labels in right tool panel issue #6153

Desmis:


heckflosse:
@Desmis Jacques, I'm testing your patch atm. And from a first view I'm not glad about what I see (reducing preview area size). Will test a bit more now...

heckflosse:
@Desmis Jacques, here a comparison.

Max preview area size with dev on my machine

grafik

and with your branch

grafik

Thanatomanic:
@Desmis @heckflosse When you set both left and right panels to their minimal sizes, the preview perfectly contains a 3:2 image on a 1920x1080 monitor. _However_ I also think that the right panel is now too wide. It takes up ~25% of the screen.
Jacques, can you indicate which GUI elements are the widest at the moment? 

image

heckflosse:
Reducing the width of this combobox(es) (or moving them to a different position) probably would solve the issue:
grafik

Desmis:
@heckflosse  @Thanatomanic 
it is an almost insoluble problem, except if we reduce the fontsize to 8, or change the GUI...

I tried to reduce to the maximum, all labels, pack_shrinks, etc. and with sizefont = 9, that leads to that...
Or to accept in LA to have GUI problems...

I have adapted to the minimum all the labels and the line in mask where the 3 curves are aligned  (C(C), L(L), LC(H))
if you have a solution I am interested

In fact the minimum size for the right panel is about 465 with a fontsize of 9, and 413 with a fontsize of 8
But if we change (a little) the GUI in curveditorgroup.cc by suppressing  the ":" and if we suppress "curves" in labels for the 3 curves "mask" (above) we can adpoted a size about 430 with a size font of 9, and 375 with a size font of 8

I will push soon a solution with ";" and "curves" suppresed to test

Jacques

Thanatomanic:
@Desmis I think you are already maximizing your effort to get a compact GUI :+1: I am still much too unfamiliar with the whole LA GUI to have any good suggestions for further improvement.

Desmis:
@Thanatomanic  @heckflosse 

Since one week I have tried many solutions...

In the last commit, I made the changes...":" and suppress "curves"...With that 
now with sizefont = 9, minimum  right panel is 429 and with sizefont=8 right panel is 382

I think the solution without ":" for curves (except to add to the label of the others curves..if need), and suppress "curves" for the 3 curves "mask" is acceptable...The gain is about 35 in size

Jacques

heckflosse:
@Desmis Jacques

> it is an almost insoluble problem

I will concentrate my work on the `almost` this weekend ;-)

Ingo

Desmis:
@heckflosse  @Thanatomanic 

I have added much ":" to labels with curves...I hope I haven't forgotten too much

Ingo , the box with "basic" have also a label "advanced"....
I have "adapted" the size by (in locallabtools.cc")        " complexity->setPreferredWidth(options.fontSize * 11.7, -1);"

After many test, it seems  - except we change the GUI for mask with the 3 curves - that this part dimensions the minimum of the right panel - about 423 - Of course others labels, box  also contribute to....



Jacques

Desmis:
@Thanatomanic  @heckflosse 
Of course I have tried others things as replace C(C) and LC(h) by c(c) and Lc(h) gain about 3 pixels
We can also change some labels...

To note a big work with @waynesutton50  Wayne Sutton on all labels and tooltips during several mounths....to make labels and tooltips understandable and usable... For example we have replaced "expert" by "advanced"...but 2 chars more...We can also admit that the label "advanced" is cut...and so we will see "adv...ed"

Jacques

heckflosse:
This would in my opinion be less ugly at least:

grafik

ff2000:
> Reducing the width of this combobox(es) (or moving them to a different position) probably would solve the issue:
> grafik

A solution might be to use icons for the basic/normal/expert combo box, in the style of the curve type choser (Linear/Standard/Flexible). This would free a lot of space for the tool label.

Desmis:
I repeat the biggest problem is in mask... the longest length.
Which does not mean that the rest cannot be improved

For the basic/standard/advanced problem the designer is Pierre @Pandagrapher ..to note that the "system" is : Label + complexity choice + 1:1 + X

heckflosse:
@Desmis 

> For the basic/standard/advanced problem the designer is Pierre @Pandagrapher ..to note that the "system" is : Label + complexity choice + 1:1 + X

with all respect, it does not matter who is the designer. We (all of us) have to find a solution. That's the only thing which counts. If we didn't detect this issue earlier, it's our fault, not the fault of Pierre!!!

ff2000:
> I repeat the biggest problem is in mask... the longest length.
> Which does not mean that the rest cannot be improved
> 
> For the basic/standard/advanced problem the designer is Pierre @Pandagrapher ..to note that the "system" is : Label + complexity choice + 1:1 + X

If I see correctly its 
1) the line "Mask curves" + three curve combo boxes:
   Move the "Mask curves" label in a different line
   Or place the three comboboxes in a labeled frame (don't know how it's called in GTK)
2) "local contrast (by wavelet level)":
why not just drop " by wavelet level"?

Desmis:
@heckflosse 
I agree totaly, I never said it's the fault of Pierre...

heckflosse:
@Desmis Jacques,

> I never said it's the fault of Pierre...

> For the basic/standard/advanced problem the designer is Pierre @Pandagrapher

I don't think you meant it this way, but one can easily understand it this way ;-)

All good :smile: 

Ingo

Desmis:
@heckflosse 
Ingo

My idea, as he is the one who designed this...is that he participates if he wants to. I probably misunderstood
:)

jacques

Desmis:
In fact I spent several days trying to figure out how to do!

The problem is complex, and there are at least 3 problems:
* in mask with the 3 curves....there is already a "frame", but the design of the GUI leads to 
    maskCurveEditorG(new CurveEditorGroup(options.lastlocalCurvesDir,"")),
    CCmaskshape(static_cast<FlatCurveEditor*>(maskCurveEditorG->addCurve(CT_Flat, "C(C)", nullptr, false, false))),
    LLmaskshape(static_cast<FlatCurveEditor*>(maskCurveEditorG->addCurve(CT_Flat, "L(L)", nullptr, false, false))),
    HHmaskshape(static_cast<FlatCurveEditor *>(maskCurveEditorG->addCurve(CT_Flat, "LC(h)", nullptr, false, true))),
The possibilities...put the curves on 3 lines / change the labels C(C):  / change  localy the font size... etc. / do otherwise (??)

The last solution without the label "curves;" seems to me good...

I have replace (by wavelet level) by "Ψ levels"....

* all labels with a copromise between length and comprehension

* and also the system with "complexity" + 1:1 + X

ff2000:
"Ψ Local contrast (by Ψ level)"  vs "by wavelet level"
I don't think that replacing "wavelet" with an icon helps making the functionality more understandable really works. You already had the candelabra/triton-fork symbol and added another one. Personally I wouldn't understand that it means "wavelets" in the first place, having it twice would confuse me even more ;) Is it really important that local contrast is done by wavelets?

Desmis:
@ff2000 

A good question, what is important ?

In the case of local contrast (by wavelet levels) , the double slider under necessarily refers to wavelets....We can of course suppress the "trident" (which is the common symbol for wavelet) as "f" for Fourier...
Ok, we can draw a frame "wavelet", then put a label  "local contrast" and under the levels selector......Is it better ?

Desmis:
but a "good" solution is to reduce the font size....
Personnaly I always used "8" with "Segoe UI normal" and no problem...plus I find it more aesthetic

Jacques

Thanatomanic:
> but a "good" solution is to reduce the font size....
Personnaly I always used "8" with "Segoe UI normal" and no problem...plus I find it more aesthetic

FWIW, I would not be in favor to make this change, because it affects the entire GUI, not only the size of the right panel.

Desmis:
For me it is not a problem, in all cases I find font 8 better :)

heckflosse:
Give me some time to try solving issue please. I appreciate all your suggestions!

Pandagrapher:
Maybe these 3 modifications can solve the issue:
1. Sorting LA tools according to their processing order. Like this we can remove the number in the tool adding combobox
2. Replacing tool adding combobox by a button opening a menu when clicking on it. Like this combobox size won’t be dependent from LA tool name size
3. Replacing complexity combobox by an icon (green/yellow/red colored rectangle enough?) opening a menu when clicking on it

Gonna try to work next week on these.
Pierre

ff2000:
> 3. Replacing complexity combobox by an icon (green/yellow/red colored rectangle enough?) opening a menu when clicking on it

Colors are dangerous ;) I am sure I have issues distinguishing green and yellow...
Two variants that I think might work:
Screenshot_20210314_073517
The right one is simpler but people might confuse it with just a menu to interact with the module (not sure about that).
The left one should be a gear icon (sorry for my bad inkscape skills) split up into three segments. Might look better with 6 instead 9 spikes, and split it in the spikes instead of the grooves. Or use three seperate gears...

ff2000:
I think these gears work better:
Screenshot_20210314_080409

Desmis:
@heckflosse 
I just made some (last I hope) adjustments to Labels...
I also change a little in locallabtools.cc ==>         complexity->setPreferredWidth(options.fontSize * 8.5, -1); ==> 8.5 instead of 9
With all this changes, ToolPanelWidth=429....This seems a minimum with sizefont = 9


Of course we can change the GUI, including that of the selection "basic....advanced"....This will a gain in ergonomic...But not in width right panel

@Pandagrapher 
Pierre
For the point 1) of your text, the order with the number (add tool to current spot) is given for information. And I don't think we must ordered the tools in the order of the number.
Pour être clair, je vais l'écrire en français: l'ordre dans "add tools to current spot) est donné à titre indicatif et ne correspond pas à l'usage. Ainsi les 2 premiers dans l'ordre du traitement "Log encoding" et "denoise" ne sont pas les plus utilisés. Par contre "Color and Light" est probablement le premier qui doit apparaître dans les menus et classement des outils et il est presque le dernier en traitement.
Of course we can exchanged on the order of use.

Jacques

Lawrence37:
With the current changes, the minimum width looks fine on my build. I like the idea of using icons for the tool complexity or even moving the combo-box inside the tool itself. For the icon idea, I prefer something simple like the 3 lines proposed by @ff2000. Maybe change the lines into sliders to avoid confusion?

Beep6581:
It is not a good solution changing the font size/padding of an instance of an element when an element of that type appears in other places using a standard size, e.g. if some type of combobox has a descriptive label adjacent to it, as is the case for every tool-setting combobox in RT, then every instance of that type of label should use the same font size. The theme designer relies on this. The user sets his display settings and RT font size based on this. And just because some label doesn't fit in one language doesn't mean altering font size or padding locally is the solution - other languages might not have this problem; or, other languages might have more problems. Local deviations don't work at best, and break things at worst.

I have also never been in favor of multiple levels of UI complexity (basic/standard/advanced). Complexity should be removed not compounded.

Now let's look for a good solution. A good solution keeps things consistent, and accommodates re-flowing of elements as needed. Too little space horizontally? Break it apart vertically.

Here is a simple solution:
rt_6160_2021-03-15 01-53

```diff
diff --git a/rtgui/locallabtools.cc b/rtgui/locallabtools.cc
index d7581c938..5eb01873e 100644
--- a/rtgui/locallabtools.cc
+++ b/rtgui/locallabtools.cc
@@ -132,11 +132,11 @@ LocallabTool::LocallabTool(Gtk::Box* content, Glib::ustring toolName, Glib::ustr
     complexity(Gtk::manage(new MyComboBoxText()))
 {
     // Create expander title bar
-    Gtk::Box* const titleBox = Gtk::manage(new Gtk::Box());
+    Gtk::FlowBox* const titleBox = Gtk::manage(new Gtk::FlowBox());
     Gtk::Label* const titleLabel = Gtk::manage(new Gtk::Label());
     titleLabel->set_markup(Glib::ustring("<b>") + escapeHtmlChars(UILabel) + Glib::ustring("</b>"));
     titleLabel->set_alignment(Gtk::ALIGN_START, Gtk::ALIGN_CENTER);
-    titleBox->pack_start(*titleLabel, Gtk::PACK_EXPAND_WIDGET, 0);
+    titleBox->add(*titleLabel);
 
     Gtk::EventBox* const removeEvBox = Gtk::manage(new Gtk::EventBox()); // Glue to manage mouse clicking event on remove image
     removeEvBox->set_can_focus(false);
@@ -144,7 +144,7 @@ LocallabTool::LocallabTool(Gtk::Box* content, Glib::ustring toolName, Glib::ustr
     removeEvBox->signal_button_release_event().connect(sigc::mem_fun(this, &LocallabTool::on_remove_change));
     RTImage* const removeImage = Gtk::manage(new RTImage("cancel-small.png"));
     removeEvBox->add(*removeImage);
-    titleBox->pack_end(*removeEvBox, Gtk::PACK_SHRINK, 4);
+    titleBox->add(*removeEvBox);
 
     if (needMode) {
         complexity->append(M("TP_LOCALLAB_MODE_EXPERT"));
@@ -153,16 +153,13 @@ LocallabTool::LocallabTool(Gtk::Box* content, Glib::ustring toolName, Glib::ustr
         complexity->set_active(2);
         complexity->setPreferredWidth(100, -1);
         complexityConn = complexity->signal_changed().connect(sigc::mem_fun(*this, &LocallabTool::complexityModeChanged));
-        titleBox->pack_end(*complexity, Gtk::PACK_SHRINK, 2);
+        titleBox->add(*complexity);
     }
 
-    Gtk::Separator* const separator = Gtk::manage(new Gtk::Separator(Gtk::ORIENTATION_VERTICAL));
-    titleBox->pack_end(*separator, Gtk::PACK_SHRINK, 0);
-
     if (need100Percent) {
         RTImage* const titleImage = Gtk::manage(new RTImage("one-to-one-small.png"));
         titleImage->set_tooltip_text(M("TP_GENERAL_11SCALE_TOOLTIP"));
-        titleBox->pack_end(*titleImage, Gtk::PACK_SHRINK, 0);
+        titleBox->add(*titleImage);
     }
 
     exp = Gtk::manage(new MyExpander(true, titleBox));
```

rtdata/languages/default
```diff
@@ -2920,7 +2920,7 @@ TP_LOCALLAB_MERTHR;Difference
 TP_LOCALLAB_MERTWE;Exclusion
 TP_LOCALLAB_MERTWO;Subtract
 TP_LOCALLAB_METHOD_TOOLTIP;'Enhanced + chroma denoise' significantly increases processing times.\nBut reduce artifacts.
-TP_LOCALLAB_MLABEL;Restored data Min=%1 Max=%2 (Clip - Offset)
+TP_LOCALLAB_MLABEL;Resto. data Min=%1 Max=%2(Clip/Offset)
```
Beep6581: If these values are useful at all, why not split them vertically?

rtdata/languages/default
```diff
@@ -3126,7 +3126,7 @@ TP_LOCALLAB_THRES;Threshold structure
 TP_LOCALLAB_THRESDELTAE;ΔE scope threshold 
 TP_LOCALLAB_THRESRETI;Threshold
 TP_LOCALLAB_THRESWAV;Balance threshold
-TP_LOCALLAB_TLABEL;TM Data Min=%1 Max=%2 Mean=%3 Sigma=%4 (Threshold)
+TP_LOCALLAB_TLABEL;TM Data Min=%1 Max=%2 Mea=%3 Sig=%4(thr)
```
Beep6581: "Mea" is not a standard abbreviation for anything. If trying to shorten stuff, why not use standard abbreviations ("AM" for "arithmetic mean") or symbols (Σσς). While I am against RT's long tooltips, _this_ is exactly what tooltips are for - to explain that e.g. "AM" means "arithmetic mean".

rtdata/languages/default
```diff
@@ -3032,7 +3032,7 @@ TP_LOCALLAB_SH1;Shadows Highlights
 TP_LOCALLAB_SH2;Equalizer
 TP_LOCALLAB_SHADEX;Shadows
 TP_LOCALLAB_SHADEXCOMP;Shadow compression & tonal width
-TP_LOCALLAB_SHADHIGH;Shadows/Highlights-Equalizer
+TP_LOCALLAB_SHADHIGH;Shadows/Highlights&Equalizer
```
Beep6581: Missing whitespace around "&".

rtdata/languages/default
```diff
@@ -3552,7 +3552,7 @@ TP_WAVELET_BACUR;Curve
 TP_WAVELET_BALANCE;Contrast balance d/v-h
 TP_WAVELET_BALANCE_TOOLTIP;Alters the balance between the wavelet directions: vertical-horizontal and diagonal.\nIf contrast, chroma or residual tone mapping are activated, the effect due to balance is amplified.
 TP_WAVELET_BALCHRO;Chroma balance
-TP_WAVELET_BALCHROM;Denoise equalizer  Blue-Yellow/Red-Green
+TP_WAVELET_BALCHROM;Denoise equalizer  B-Y/R-G
```
Beep6581: Double whitespace.

rtdata/languages/default
```diff
@@ -3712,10 +3712,10 @@ TP_WAVELET_MEDILEV;Edge detection
 TP_WAVELET_MEDILEV_TOOLTIP;When you enable Edge Detection, it is recommanded:\n- to disabled low contrast levels to avoid artifacts,\n- to use high values of gradient sensitivity.\n\nYou can modulate the strength with 'refine' from Denoise and Refine.
 TP_WAVELET_MERGEC;Merge chroma
 TP_WAVELET_MERGEL;Merge luma
-TP_WAVELET_MIXCONTRAST;Reference local contrast
+TP_WAVELET_MIXCONTRAST;Ref. local contrast
 TP_WAVELET_MIXDENOISE;Denoise
-TP_WAVELET_MIXMIX;Mixed 50% noise - 50% denoise
-TP_WAVELET_MIXMIX70;Mixed 30% noise - 70% denoise
+TP_WAVELET_MIXMIX;Mixed 50%noise-50%denoise
```
Beep6581: Throwing out whitespace to fit text is not a solution.

rtdata/languages/default
```diff
@@ -1965,8 +1965,8 @@ TP_BWMIX_AUTOCH;Auto
 TP_BWMIX_CC_ENABLED;Adjust complementary color
 TP_BWMIX_CC_TOOLTIP;Enable to allow automatic adjustment of complementary colors in ROYGCBPM mode.
 TP_BWMIX_CHANNEL;Luminance equalizer
-TP_BWMIX_CURVEEDITOR1;'Before' curve
-TP_BWMIX_CURVEEDITOR2;'After' curve
+TP_BWMIX_CURVEEDITOR1;'Before' curve:
```
Beep6581: Colons should generally be hard-coded, not part of the locale string.

rtgui/curveeditorgroup.cc
```diff
@@ -34,7 +34,7 @@ CurveEditorGroup::CurveEditorGroup (Glib::ustring& curveDir, Glib::ustring group
 {
 
     // We set the label to the one provided as parameter, even if it's an empty string
-    curveGroupLabel = Gtk::manage (new Gtk::Label (groupLabel + ":", Gtk::ALIGN_START));
+    curveGroupLabel = Gtk::manage (new Gtk::Label (groupLabel, Gtk::ALIGN_START));
```
Beep6581: Why? Most of RT hard-codes the ":", and that's probably a good thing.

rtgui/editorpanel.cc
```diff
@@ -637,7 +637,8 @@ EditorPanel::EditorPanel (FilePanel* filePanel)
     vboxright = new Gtk::Paned (Gtk::ORIENTATION_VERTICAL);
 
     vsubboxright = new Gtk::Box (Gtk::ORIENTATION_VERTICAL, 0);
-    vsubboxright->set_size_request (300, 250);
+    int rightsize = options.fontSize * 45;
```
Beep6581: Why this?

rtgui/options.cc
```diff
@@ -343,15 +343,15 @@ void Options::setDefaults()
...
+    toolPanelWidth = 465;
     browserToolPanelWidth = 465;
     browserToolPanelHeight = 600;
     browserToolPanelOpened = true;;
     browserDirPanelOpened = true;
     editorFilmStripOpened = true;
     historyPanelWidth = 330;
     fontFamily = "default";
-    fontSize = 10;
+    fontSize = 8;
```
Beep6581: Why, and based on what tests? If this is solely to accommodate Local Adjustment's horizontal space problem, I object.

rtdata/languages/default
ff2000: I played with the advanced retinex tools:
https://youtu.be/T7_ZAc-27dE
I see several issues here:
* Updates to that label seem to occur randomly. I had to move one point up and down several times to get a number change.
* The magnitude of the numbers changes, this moves the different numbers around in the label, making it impossible to track changes of of one particular value.
* Occasionally it gets bigger than the panel width, GTK adds scrollbars which moves the GUI the user interacts with (I couldn't check with recent changes)
* Sometimes MIN is bigger than MAX and MEAN outside [MIN,MAX] - but perhaps that's normal?

Question: Is it really important to give the user exact numbers? Couldn't this be presented simply with a horizontal bar showing two or three marks? color them e.g. min=black, max=white, mean=gray? Or might it even be possible to visualize those values in the transition map graph the user interacts with?

ff2000:
> @ff2000
> 
> A good question, what is important ?
> 
> In the case of local contrast (by wavelet levels) , the double slider under necessarily refers to wavelets....We can of course suppress the "trident" (which is the common symbol for wavelet) as "f" for Fourier...
> Ok, we can draw a frame "wavelet", then put a label "local contrast" and under the levels selector......Is it better ?

What I meant: Wavelet is the way that local contrast tool is implemented. To the user that information isn't important (IMO). He just has to know that this graph changes "local contrast".  Same with tools using Fast Fourier Transformations: It doesn't matter for the user. Just remove that extra bit of implementation information. If a LA tool has optional dependency on FFT (e.g. a checkbox "use FFT" ) the tool has to get deactivated (greyed out) instead of carrying "FFT" in the label in case the checkbox is deactivated.

IMO the labels really should be as short as possible and carry no unneeded information. Having to parse that wall of text makes it harder (and strenuous) to interact with the  tools.

ff2000:
> I prefer something simple like the 3 lines proposed by @ff2000. Maybe change the lines into sliders to avoid confusion?

Screenshot_20210315_085052

Yep, way simpler than gears and doesn't look like the standard hamburger-icon.

Desmis:
@Beep6581 and others
All the proposals I made are to reduce the size of the "right panel".
1) in "editormanel.cc" to adapt the size of the "right panel" to the size of the font. But if you have other ideas, 
2) generaly to reduce the length of the labels
3) "CurveEditorGroup", I found this solution and add a ":" to each curve, to be able to put a blank as label


As I said I'm not a GUI specialist at all...I'm trying to figure it out. If you have any solutions I'm interested. Just make the modifications

jacques

Desmis:
I made several GUI changes for the wavelet part masks
I also revised the LA-retinex labels - I put the same as Retinex (main), Wavelets levels (main), and other adjustments.

Several people are planning to work on the GUI.... so I'll let them have a go.
If necessary I'll make further adjustments to the labels (or the GUI) after

For the font size, if I chose 8 (default) , it's by taste and also to save vertical space - and not specially for LA.
On my personnal computer I choose "7" with "Segoe UI normal" (windows). 
With this choice on my 2 screens, I see 36 images in the browser and for the editor we gain space in vertical - about 4 lines...and the space for the image is bigger. But it's a choice, it's perfectly readable despite my age (73) and my eyesight problems
The right panel is 20% of screen.

filebrowser7
editor7

jacques

Desmis:
For the local contrast wavelet, there are at least 4 ways to do it...which give very different results:
* using as reference the local contrast of each decomposition, but with the same "strength" value for each level. What is taken into account is the value of the local contrast (+ and -) for each decomposition level - It is the local contrast you see in the first position in "local contrast and wavelet"
* as "CBDL" - contrast by levels - here you can change the local contrast for each decomposition level (pyramid 2)
* as Tone mapping - substantially as above, but with stronger values and transitions provided by Guidedfilter (pyramid2)
* Edge sharpness - as "contrast by levels", but you take into account the edge detection. With that you can change contrast inside a texture...(pyramid 1)

There is also - directional contrast

And each of this tool  has several "tools" (depending of the case) to modulate the action, in addition to the curves - attenuation response, offset..., gradient sensitivity, etc. And of course masks, includind "Recovery based on luminance mask"

Jacques

Desmis:
@Beep6581 
I just push a commit aab5aa4 with the change patch - "Flowbox" 
To test 
Thanks :)

jacques

ff2000:
> A good solution keeps things consistent, and accommodates re-flowing of elements as needed. Too little space horizontally? Break it apart vertically.

Look at the tool title as a title bar. Besides the label you have fixed elements - active, close, complexity, 1:1, close. They should always be at the same place. For consistency and ease of use. The reflow moves close, complexity and 1:1 all over the place.

Desmis:
@ff2000 

I think that's the price to pay for having this facility. Note that with the "7" font (windows) , there is much less variation. Nevertheless it seems to me a good solution

jacques

Desmis:
I changed curveeditorgroup.cc to allow either the use of ":" or to have an empty string. I added an optional parameter - int blank 

Of course, I removed the ":" added in the labels 

jacques

ff2000:
Here some suggestions for Tool renaming:

1) Shadows/Highlights & Tone Equalizer -> Tone Equalization
The naming is misleading as it is "OR" and not "AND"
Shad/High basically is a "dumbed down" Tone Equalizer.
I also suggest to open Shad/High by default, perhaps just for the basic complexity. (Similar as in Local Contrast & Wavelets)

2) Soft Light & Original Retinex -> Soft Light

3) Dehaze & Retinex -> Dehaze

4) Local Contrast & Wavelets -> Local Contrast
Perhaps rename implementation "Unsharp Mask" to e.g. Local Contrast or "Simple" - "Unsharp Mask" might be more accurate regarding the imploementation but the term is most known within sharpening (which is an own LA tool which AFAIK uses RL Deconvolution, correct?)

5) Dynamic Range & Exposure
This basically mixes exposure compensation + Dynamic Range Compression, the latter being the important tool. Is it important, that Exposure is in this tool? "Color and Light" already has "Lightness" and "Contrast", why not add "Exposure" here?

6) "Vibrance & Warm/Cold" -> "Vibrance & Warmth"

Concerning "Contrast By Details Level" - why is it a seperate tool? Couldn't it be added as a third implementation of Local Contrast?

2&3 : I assume very little users will know the difference between Retinex and Original Retinex. I also assume that Original Retinex is a more refined "Soft Light" and Retinex does "Dehaze" - correct?

waynesutton50:
There is a brief description of each of the tools in the Introduction to the Local Adjustments section in Rawpedia, which may be of help in answering some of your queries (e.g. Shadows/Highlights & Tone Equalizer). https://rawpedia.rawtherapee.com/Local_Adjustments#Introduction
Further information about some of the more specific aspects of the tools can also be found in the worked examples, bearing in mind that these are not exhaustive. e.g. for original retinex see https://rawpedia.rawtherapee.com/Local_Adjustments#Dodging_and_burning

ff2000:
> There is a brief description of each of the tools in the Introduction to the Local Adjustments section in Rawpedia, which may be of help in answering some of your queries (e.g. Shadows/Highlights & Tone Equalizer). https://rawpedia.rawtherapee.com/Local_Adjustments#Introduction

According to RawPedia that tool should then be called "Shadows and Highlights" because `Adjust shadows & highlights either with the shadows/highlights sliders or with a tone equalizer.`

> Further information about some of the more specific aspects of the tools can also be found in the worked examples, bearing in mind that these are not exhaustive. e.g. for original retinex see https://rawpedia.rawtherapee.com/Local_Adjustments#Dodging_and_burning

And who besides the developers know what's the difference between "Original" and "non-Original" Retinex?

To make it clear: I proposed those changes to reduce the panel width. Changing the biggest offender - Shadows/Highlights & Tone Equalizer - already would help a lot. This IMO would be the "correct" way to fix the width issues. The current solution (reflow) is a workaround and just bad UI design.

Thanatomanic:
@ff2000 @waynesutton50 Thanks for being involved in the discussion!

I must say, this PR has exploded starting from quite a simple issue. What I find mildly annoying is that no one has presented an actual analysis of the problem before solutions were being proposed - or I have missed it completely. What is the actual, factual cause of the wide content? Is it only the labels? Is it a row of buttons? And which ones? 
imo, the answer to those questions should be the focus of finding the easiest solutions.

Edit: I must admit that I have also been some of the people coming with quick suggestions... that makes me part of the problem ;-)

Another thing I would like to add is that we have no GUI guidelines in RT, nor do we have a UX expert in the team. We're just winging it and trying to stay true to how the rest of RT already works - and even there we're not consistent everywhere... Any opinion is valid in that sense. I really wish we would have somebody to fulfill the role of UX master, or that somebody takes time to write some guiding principles (either [here](https://github.com/Beep6581/RawTherapee/discussions) or [here](https://github.com/Beep6581/RawTherapee/wiki) seem like great places).

Desmis:
@Thanatomanic  @waynesutton50 @ff2000 @heckflosse @Beep6581 @Pandagrapher @rebio

I have already mentioned the problem of width, with 3 main causes:
* the masks, with in particular the 3 curves (L C LC). With the modification I made...plus others concerning the mask labels, it is this that now determines the minimum size of the window 
* LA labels -now they all fit in the above width now 
* some labels in "wavelet levels" (main) that also made the right panel wider, have been reduced

So - and this is what I said before - the problem that remains is the "title of each tool" + "1:1" + "X" + "complexity"

Basically 3 solutions (or more):
1) the current solution with the Flowbox (easy and good for me)
2) reduce the length of the titles to fit - the biggest problem is going to be "contrast by detail levels" (it was called CBDL...but we found it weird) - it may lose its meaning. To evoke to unbundle / regroup tools is extremely complex - and except in the margin I am opposed to it
3) systematically move "basic...advanced" the line below the title (I think quite easy)
4) ??

And @Thanatomanic  I agree totaly with your last paragraph (GUI guideline, UX expert...)
jacques

heckflosse:
As I already wrote in private to @Desmis , I will take a break for one week, caused by massive pain in my back. I can't concentrate on coding because of the pain which was caused by carrying too heavy stuff (please, don't do that. I learned my lession).

Hopefully I will be back in a week or so.,

Ingo

waynesutton50:
> According to RawPedia that tool should then be called "Shadows and Highlights"

this illustrates @Thanatomanic 's point about the need for GUI guidelines and at the very least a style guide for labels. I put a forward slash in the Shadows/Highlights tool label in Local Adjustments to make it consistent with the Shadows/Highlights label in the Exposure tab. The use of '/' or 'and' in Rawpedia should probably depend on whether the reference is to the module or the function but I haven't checked this. There are a number of similar issues that need to addressed as well but this is off topic.

ff2000:
> > According to RawPedia that tool should then be called "Shadows and Highlights"
> 
> this illustrates @Thanatomanic 's point about the need for GUI guidelines and at the very least a style guide for labels. I put a forward slash in the Shadows/Highlights tool label in Local Adjustments to make it consistent with the Shadows/Highlights label in the Exposure tab. The use of '/' or 'and' in Rawpedia should probably depend on whether the reference is to the module or the function but I haven't checked this. There are a number of similar issues that need to addressed as well but this is off topic.

Sorry, my bad. The point  wasn't "/ vs. and" but to drop "Tone Equalizer" from label (according to Rawpedia's description of the tool). If Exposure uses "/" then also use it in LA.
Consistency is important.

ff2000:
> As I already wrote in private to @Desmis , I will take a break for one week, caused by massive pain in my back. I can't concentrate on coding because of the pain which was caused by carrying too heavy stuff (please, don't do that. I learned my lession).
> 
> Hopefully I will be back in a week or so.,
> 
> Ingo

@heckflosse Gute Besserung. Werd wieder!

Desmis:
I just change - as mentioned above - always for testing, the Flowbox, by a vertical Box
Now in all cases "Complexity" is under label

Jacques

ff2000:
I've played with changing the ComboBox for complexity selection in the tool title.
https://github.com/ff2000/RawTherapee/tree/lasizepanel
This now uses a PopUpButton.
Issues:
* The width of the button varies and hasn't just minimal size.
* The popup menu is empty.
* Clicking the button will also pass the click to the tool title widget and collaps/expand the tool.
* The icons don't look that good. black looks bad IMO, but white/gray might not work with all themes.

Help would be appreciated (if you like the approach). I have basically no experience with GTK.

@Desmis : My Display is 1600x900. The default panel width now is way too big. Also having an additional line per tool (even if it is collapsed) leaves even less space for the controls. And that way too much emphasis is put on the mode.
I don't think it makes much sense to try get the title in multiple lines (dynamically or statically). Please just try to find shorter tool titles so that the additional controls fit in the title.

Desmis:
As a reminder, I am not a GUI specialist at all, nor am I a C++ specialist.

I work like a monkey by copying what others do, then I make trial and error.

I have taken into account the remarks of Beep6581 “Too little space horizontally? Break it apart vertically.”.

I just reset all the "right panel" and "options" to their initial values...
At least (at minimum) we see a (small) overflow as soon as we activate "mask" or wavelet level (advanced)

Ingo and Pierre will look into this and I suggest to see what solutions they propose. Of course you and others can also contribute.

jacques

Desmis:
Hello all

Where are we?  Are there other proposals?  What are we doing?

Several improvements have been made:
* change "curveditorgroup.cc" to allow curves without label;
* examine and modify all labels (including some that are not in LA) so that they fit the minimum length without loss of meaning;
* propose 3 solutions for the header of the tools: a) shorten the text of the labels (leads to a loss of meaning), b) use a "flowbox" type box, c) use a vertical box: from my point of view this last solution is the "best" (knowing that I don't know how to change the GUI characteristics of this box)

I propose to merge (if nobody objects) the current commit... :)

Jacques

heckflosse:
> Where are we? Are there other proposals? What are we doing?

I just built the branch again to test it in the evening. Atm the weather is too good to sit in front of keyboard. Going out with the dogs now :smiley:

heckflosse:
@Desmis Jacques, I tested the branch for a while and found no issues. It's clearly an improvement :+1:

Desmis:
@heckflosse 
Ingo
Thank you :)

Others opinions, if not I merge tomorrow

Jacques

Issue #6171, opened 2021-03-14
Non-local means smears image border


Lawrence37:
Steps to produce:
1. Enable local adjustments.
2. Add full image spot
3. Add Blur/Grain & Denoise tool.
4. Set mode to anything except for off.
5. Change non-local means strength.

Observe the repeated pixels along the bottom and right sides of the image. The other two edges are unaffected.
![image](https://user-images.githubusercontent.com/45837045/111083427-d19e3a00-84ca-11eb-8768-c1cb5fe712e3.png)

PR #6172, opened 2021-03-15
Local adjustments - Nlmeans - Fixed smears image borders issue 6171

Desmis:
@Lawrence37 

First, thanks for testing :)
https://github.com/Beep6581/RawTherapee/issues/6171
I think (I hope) this patch will fix this bad behavior.
But curious - apart if I if I haven't looked at ART recently....The same problem will appears...

Others things : I have been having big problems with my computer for a few days...I hope I can still be there.

Jacques

### DIFF
--- a/rtengine/iplocallab.cc
+++ b/rtengine/iplocallab.cc
@@ -9558,27 +9558,8 @@ void ImProcFunctions::DeNoise(int call, float * slidL, float * slida, float * sl
             }
 
             if(lp.nlstr > 0 && (hspot > 150 && wspot > 150)) {
-                array2D<float> *nlm = nullptr;
-                int addsiz = 10 + lp.nlrad;
-                nlm = new array2D<float>(GW + addsiz, GH + addsiz);
-#ifdef _OPENMP
-                #pragma omp parallel for if (multiThread)
-#endif
-                for (int i = 0; i < GH; ++i) {
-                    for (int j = 0; j < GW; ++j) {
-                        (*nlm)[i][j] = tmp1.L[i][j];
-                    }
-                }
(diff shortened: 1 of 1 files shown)


Lawrence37:
The results are much better now, but if you look carefully, there is still some smearing. Interesting to see ART has this problem. I wonder if the ART implementation is original or if it was ported from somewhere.

Desmis:
@Lawrence37 
This problem is due to "borders"...
I just increase the effect of my patch.

ART take this algorithm in Ipol
http://www.ipol.im/pub/art/2014/120/

I have always big problems with my computer
jacques

Lawrence37:
I experimented with several values for the borders and couldn't figure out why the problem is still there with large borders. Then I noticed it happens without non-local means, so it's probably an issue with another part of the de-noising. Also, I think I found the cause of the original smearing and fixed it like this:
```diff
diff --git a/rtengine/iplocallab.cc b/rtengine/iplocallab.cc
index 682b3b9a0..3c6776352 100644
--- a/rtengine/iplocallab.cc
+++ b/rtengine/iplocallab.cc
@@ -11016,9 +11016,9 @@ void ImProcFunctions::NLMeans(float **img, int strength, int detail_thresh, int
 #   pragma omp parallel for if (multithread)
 #endif
     for (int y = 0; y < HH; ++y) {
-        int yy = y <= border ? 0 : y >= H ? H-1 : y - border;
+        int yy = y <= border ? 0 : y - border >= H ? H-1 : y - border;
         for (int x = 0; x < WW; ++x) {
-            int xx = x <= border ? 0 : x >= W ? W-1 : x - border;
+            int xx = x <= border ? 0 : x - border >= W ? W-1 : x - border;
             float Y = img[yy][xx] / 65536.f;
             src[y][x] = Y;
         }
```
No need to create an extra buffer :)

Desmis:
@Lawrence37 
Good catch :)

Works fine...I merge soon :)

Thank you very much

jacques

PR #6173, opened 2021-03-19
LA - Move colortoning labgrid from rgb to lab - issue #6132

Desmis:
@Thanatomanic @chaimav
I made the same modification as for shadows-highlight (main) and local contrast (main).
Move the "lab" code from rgbproc to Lab 

jacques

### DIFF
--- a/rtengine/dcrop.cc
+++ b/rtengine/dcrop.cc
@@ -1171,6 +1171,10 @@ void Crop::update(int todo)
         bool cclutili = parent->cclutili;
 
         LUTu dummy;
+        bool hasColorToningLabGrid = params.colorToning.enabled && params.colorToning.method == "LabGrid";
+        if (hasColorToningLabGrid) {
+            parent->ipf.colorToningLabGrid(labnCrop, 0,labnCrop->W , 0, labnCrop->H, false);
+        }
 
         parent->ipf.shadowsHighlights(labnCrop, params.sh.enabled, params.sh.lab,params.sh.highlights ,params.sh.shadows, params.sh.radius, skip, params.sh.htonalwidth, params.sh.stonalwidth);
         
--- a/rtengine/dcrop.cc
+++ b/rtengine/dcrop.cc
@@ -1171,8 +1171,7 @@ void Crop::update(int todo)
         bool cclutili = parent->cclutili;
 
         LUTu dummy;
-        bool hasColorToningLabGrid = params.colorToning.enabled && params.colorToning.method == "LabGrid";
-        if (hasColorToningLabGrid) {
+        if (params.colorToning.enabled && params.colorToning.method == "LabGrid") {
             parent->ipf.colorToningLabGrid(labnCrop, 0,labnCrop->W , 0, labnCrop->H, false);
         }
 
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -1286,6 +1286,10 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
             nprevl->CopyFrom(oprevl);
             histCCurve.clear();
             histLCurve.clear();
+            bool hasColorToningLabGrid = params->colorToning.enabled && params->colorToning.method == "LabGrid";
+            if (hasColorToningLabGrid) {
+            ipf.colorToningLabGrid(nprevl, 0, nprevl->W, 0, nprevl->H, false);
+            }
 
             ipf.shadowsHighlights(nprevl, params->sh.enabled, params->sh.lab,params->sh.highlights ,params->sh.shadows, params->sh.radius, scale, params->sh.htonalwidth, params->sh.stonalwidth);
 
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -1286,8 +1286,7 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
             nprevl->CopyFrom(oprevl);
             histCCurve.clear();
             histLCurve.clear();
-            bool hasColorToningLabGrid = params->colorToning.enabled && params->colorToning.method == "LabGrid";
-            if (hasColorToningLabGrid) {
+            if (params->colorToning.enabled && params->colorToning.method == "LabGrid") {
             ipf.colorToningLabGrid(nprevl, 0, nprevl->W, 0, nprevl->H, false);
             }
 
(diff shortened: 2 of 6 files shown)


rtgui/colortoning.cc
```diff
@@ -346,7 +346,8 @@ ColorToning::ColorToning () : FoldableToolPanel(this, "colortoning", M("TP_COLOR
     //------------------------------------------------------------------------
     // LAB grid
     auto m = ProcEventMapper::getInstance();
-    EvColorToningLabGridValue = m->newEvent(RGBCURVE, "HISTORY_MSG_COLORTONING_LABGRID_VALUE");
+//    EvColorToningLabGridValue = m->newEvent(RGBCURVE, "HISTORY_MSG_COLORTONING_LABGRID_VALUE");
+    EvColorToningLabGridValue = m->newEvent(LUMINANCECURVE, "HISTORY_MSG_COLORTONING_LABGRID_VALUE");
```
Thanatomanic, line 350: What does this do?

Thanatomanic:
I don't quite understand the changes here, and my first impression is that shifting this tool in the pipeline breaks backward compatibility. No? In fact, have we been considering backward compatibility with all the recent pipeline changes?

Also, why is this specific to the LabGrid method and not the other methods from the color toning panel?

Desmis:
@Thanatomanic 
No it does not break compatibility, no chnages to pp3, only the placement and "event"

Why this one, because it use Lab..in RGB
It's the same problem as Local contrast, and Shadows/highlight

jacques

rtgui/colortoning.cc
Desmis: I change the event..otherwise it doesn't work

In Rgbproc ==> event RGBCURVE
in Lab ==> event LUMINANCECURVE

rtengine/improcfun.cc
```diff
@@ -3565,9 +3565,9 @@ void ImProcFunctions::rgbProc (Imagefloat* working, LabImage* lab, PipetteBuffer
         for (int i = 0; i < tH; i++) {
             Color::RGB2Lab(tmpImage->r(i), tmpImage->g(i), tmpImage->b(i), lab->L[i], lab->a[i], lab->b[i], toxyz, tW);
```
Thanatomanic, line 3566: Considering the changes below, do we still need to perform this loop?

rtengine/improcfun.cc
```diff
@@ -3184,9 +3184,9 @@ void ImProcFunctions::rgbProc (Imagefloat* working, LabImage* lab, PipetteBuffer
                         Color::RGB2Lab(&rtemp[ti * TS], &gtemp[ti * TS], &btemp[ti * TS], &(lab->L[i][jstart]), &(lab->a[i][jstart]), &(lab->b[i][jstart]), toxyz, tW - jstart);
```
Thanatomanic, line 3184: Do we still need this loop with the commented out section below?

rtengine/dcrop.cc
```diff
@@ -1171,6 +1171,10 @@ void Crop::update(int todo)
         bool cclutili = parent->cclutili;
 
         LUTu dummy;
+        bool hasColorToningLabGrid = params.colorToning.enabled && params.colorToning.method == "LabGrid";
```
Thanatomanic: Why create a new variable? This could be directly part of the if () statement below.

rtengine/improccoordinator.cc
```diff
@@ -1286,6 +1286,10 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
             nprevl->CopyFrom(oprevl);
             histCCurve.clear();
             histLCurve.clear();
+            bool hasColorToningLabGrid = params->colorToning.enabled && params->colorToning.method == "LabGrid";
```
Thanatomanic: See previous comment. No variable needed imo.

rtengine/rtthumbnail.cc
```diff
@@ -1438,6 +1438,12 @@ IImage8* Thumbnail::processImage (const procparams::ProcParams& params, eSensorT
     CurveFactory::complexsgnCurve (autili, butili, ccutili, cclutili, params.labCurve.acurve, params.labCurve.bcurve, params.labCurve.cccurve,
                                    params.labCurve.lccurve, curve1, curve2, satcurve, lhskcurve, 16);
 
+
+    bool hasColorToningLabGrid = params.colorToning.enabled && params.colorToning.method == "LabGrid";
```
Thanatomanic: Same here about the variable

rtengine/simpleprocess.cc
```diff
@@ -1347,6 +1347,12 @@ class ImageProcessor
         CurveFactory::complexsgnCurve(autili, butili, ccutili, cclutili, params.labCurve.acurve, params.labCurve.bcurve, params.labCurve.cccurve,
                                       params.labCurve.lccurve, curve1, curve2, satcurve, lhskcurve, 1);
 
+
+        bool hasColorToningLabGrid = params.colorToning.enabled && params.colorToning.method == "LabGrid";
```
Thanatomanic: Same here about the variable

rtengine/improcfun.cc
Desmis: yes because, It create Lab for after :)

rtengine/improccoordinator.cc
Desmis: Yes we can remove, I keep only for **visibility**..with "old" version.
This change is in - dcrop, improccoordinator, rtthumbnail, simpleprocess

You can modified and push this chnage if you want :)


Jacques

Desmis:
@Thanatomanic 
As I said before, 
"Yes we can remove, I keep only for visibility..with "old" version.
This change is in - dcrop, improccoordinator, rtthumbnail, simpleprocess

You can modified and push this change if you want :)

But no problem , I can do it  :)

jacques

Thanatomanic:
@Desmis For me the original bug is fixed, but when the Color Toning module is off, changing the L\*a\*b* color correction grid still triggers the progress bar. A non-active tool should not do that. Is there a fix for that?

Desmis:
@Thanatomanic 
This problem is not specific to "ColorToning labgrid". There is the same thing with "ColorToning color corrections".
It's a GUI problem (it's code implemented some years ago by Alberto), I don't think I can solve this problem

As you said, we really need a GUI specialist
Now what do we do?  I merge or not ?
This behavior has been going on for several years...

Jacques

Desmis:
If no objections I will merge tomorrow :)

Thanatomanic:
> It's a GUI problem (it's code implemented some years ago by Alberto), I don't think I can solve this problem

> This behavior has been going on for several years...

Oh interesting. Could be a faulty listener or something. That's something for another issue. No objection for a merge of this PR :+1:
@chaimav, you are also satisfied, correct?

chaimav:
The fixed the Labgrid error, so it it fine from that perspective. The DCP []Tone curve and []Look table aspect of the bug still remains. And based on the pervasiveness of the bug, I suspect it may show up elsewhere as well. You can either merge now, or keep the branch open until the bug is totally gone. (Sorry, I am not a dev so I don't know what standard practice is in a case like this)

PR #6187, opened 2021-03-26
Update camconst.json

CarVac:
Add the EOS RP (same as 6D2) and 100D/SL1/Kiss X7 (same as 700D)


Thanatomanic:
Thanks for this. Seems all good, I checked with Adobe Camera Raw that the 100D color matrix is indeed identical to the one from the 700D.

PR #6190, opened 2021-03-29
lj92: Fix decoding of 16 bit diff values

SimonSegerblomRex:
Hi!

Diff values belonging to category `SSSS=16` (always -32678) should be encoded using zero bits and always be interpreted as 32678 when decoding. This is described in the [JPEG specification](https://www.w3.org/Graphics/JPEG/itu-t81.pdf) in section _A.5_ and _H.1.2.1_.

One problem is that old DNG files (version < 1.1) might have been encoded using 16 bits for diff value -32678. This is described in _Compatibility Issue 2_ in _Appendix A_ of the [DNG specification](https://www.adobe.com/content/dam/acom/en/products/photoshop/pdfs/dng_spec_1.5.0.0.pdf). Probably you don't need to bother supporting the old behavior (since it only affects 15 and 16 bit images, which I assume are quire rare).

This is a port of the fix discussed [here](https://github.com/ilia3101/MLV-App/pull/221).

### DIFF
--- a/rtengine/lj92.c
+++ b/rtengine/lj92.c
@@ -313,6 +313,9 @@ static int decode(ljp* self) {
 }
 
 static int receive(ljp* self,int ssss) {
+    if (ssss == 16) {
+        return 1 << 15;
+    }
     int i = 0;
     int v = 0;
     while (i != ssss) {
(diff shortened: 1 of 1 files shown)


Thanatomanic:
So, thanks for this PR @SimonSegerblomRex It seems like you have good knowledge of the issue, but is there any example of a file that does not decode properly? I could also not find it in the thread you linked to.

SimonSegerblomRex:
> So, thanks for this PR @SimonSegerblomRex It seems like you have good knowledge of the issue, but is there any example of a file that does not decode properly? I could also not find it in the thread you linked to.

Good point, I'll try to create a minimal example.

SimonSegerblomRex:
Here's a minimal `*.jpg`-file: test

It's just a single 0 encoded with 16 bits.

SimonSegerblomRex:
Let me know if you need anything more to integrate this fix.

Thanatomanic:
I believe I compiled with your fix, but the test files don't even show up in the file browser...
image

Not sure why that is. The command line doesn't show errors.
So, unfortunately I cannot test whether the fix works atm

SimonSegerblomRex:
> I believe I compiled with your fix, but the test files don't even show up in the file browser...
> Not sure why that is. The command line doesn't show errors.
> So, unfortunately I cannot test whether the fix works atm

I'll try to create a minimal DNG that works in RawTherapee. (I noticed that RawTherapee only uses lj92 when "Make" begins with "Canon" or "BlackMagic"...)

SimonSegerblomRex:
@Thanatomanic, here's a file that displays in RawTherapee: [16bit-lossless-jpeg.zip](https://github.com/Beep6581/RawTherapee/files/6313360/16bit-lossless-jpeg.zip)
screenshot-rawtherapee

Thanatomanic:
Confirmed, thanks @SimonSegerblomRex for the contribution

PR #6199, opened 2021-04-03
Enable neutral colors for LA merge file

Lawrence37:
The merge file feature of the local adjustments color & light has a background mode which is useful for merging solid colors. However, setting a neutral color by double-clicking the white dot in the color grid leads to unintuitive results because part of the processing is disabled. On my machine, it forces the background color to black regardless of the luminance I select. I see no reason to disable. @Desmis, please verify my change does not break any intended behaviors.


Desmis:
@Lawrence37 

Well seen, good catch :)
Thank you very much for testing.

This patch solve the problem. I just remove 2 unused variables "a_basemerg" and "b_basemerg"

For me you can merge this PR

Thank you

Jacques

PR #6207, opened 2021-04-18
CIExy diagram displaying the primaries (interactive) and the white point

Desmis:


Thanatomanic:
Hi @Desmis, I gave your code a testrun. A few observations:
1. When you select an Abstract profile other than None, the CIExy diagram shows up, but when you return the setting to None, it stays visible
2. The item in the dropdown 'Custom settings (...)' could simply be 'Custom (...)'
3. Having both settings 'Custom settings (sliders)' and 'Custom settings (CIE xy diagram') seems unnecessary. All the GUI elements are visible in both situations, but some snap back. Why not have one situation where the sliders and diagram are updated simultaneously?
4. If you select a predefined Abstract profile, the dots on the diagram can still be moved, but they snap back. Would be better if they could not be moved at all, _or even better_ that when you move them, the setting for the primaries immediately changes to 'Custom'.
5. Since the sliders and diagram belong to the dropdown 'Destination primaries', they should preferentially be placed in the same container (as a child-element) and not in a separate container on the same level.
6. The tooltips are very weird (please try for yourself and hover over the different sliders and spinboxes to see).
6. Is there a specific reason why the white point cannot be modified? It is also not linked to the choice of illuminant.

Desmis:
For 1 and 2, it's done (easy)

For 3, 4 and 5, it's a (very) complex problem (at least for me), because the diagram is managed externally to « Icmpanel », there is no simple way to interact. I made this interaction by a complex system based on 2 "idle_register" called by improccoordinator.cc . But if someone - GUI specialist has a better solution ?

We can work on labels and tooltips.

Yes, there is a very good reason for not allowing a free « White Point » (illuminant), because contrary to the primary, it is necessary that the « White Points » are really « White Points » otherwise we end up with anything ... this is the basis of all color management. There is a unique link between the « xy » values of the white point and the illuminant, hence the preselection...Of course we can add other illuminants to the 9 chosen (which seems sufficient to me) 

But, all this work only makes sense if the team clearly considers a merge of "trcnew"...without endless procrastination

Issue #5182, opened 2019-02-19
Rescaling, option for short edge as a ruling measure...


ghost:
As RT’s batching is working well, I’d like to see an option for defining shorter edge as a ruling measure while rescaling.

Came up on me, while some on-line printing services tell you the exact pixel dimensions which they use and due to printer system I suppose, longer edge is flexible; shorter edge is constant; image orientation is ignored. 
At the moment I run command lines (Linux) and use Gimp to do the desired post processes. I like to shoot completely in RAW and while this resizing process is available in RT, I’d get also my “not so important but to be printed home photos” straight to the printer resolution.

All this to have more control of the print quality, also to save storage and such... Yet in a nutshell: 14Mp Raw -> cropped as I like --> rescaled so that shorter edge is for example 1210pix, longer edge measure is what ever, because some on-line print services sell "flexible lenghts" --> pixel per pixel process...

Thanatomanic:
You mean this? 
![image](https://user-images.githubusercontent.com/6567747/53038409-fa36df80-347c-11e9-9ba1-384cfe20bf15.png)

Unless I understand your post wrong, this is already possible, closing.

Beep6581:
@Thanatomanic though I haven't tried, I don't think it's currently possible to batch-apply resizing to images of various orientations so that the shortest side is some specified length.

Thanatomanic:
@Beep6581 You're right, I was a little hasty. The difference between RT now and the request is setting a value for the shortest side instead of the width. That's a reasonable request.

ghost:
http://www.imagemagick.org/Usage/resize/
Fill Area Flag ('^' flag)
This also represents the requested option..

for example : https://www.ifolor.fi/en/photos/photo-prints
They deliver a pile of prints, each different lenghts if they are cropped differently…
(further info describes resolution as 1210 x (y), 301dpi)

ghost:
And to be clear, for batching I ment the "Queue" processing... But the tool in "Batch Edit" list was same above.. you got the idea, just got little tangled there..

Bezierr:
Since I need this "short edge" resizing feature myself, I tried my hand on it and implemented this functionality.

![Capture](https://user-images.githubusercontent.com/38507161/115145595-e62d9f00-a041-11eb-81e5-460618abcdf4.JPG)

[...]

PR #6210, opened 2021-04-20
Addressing issue #5182

Bezierr:
Added the options "Long Edge" and "Short Edge" to the Resize functionality.

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -297,8 +297,8 @@ HISTORY_MSG_42;Exposure - Tone curve 2
 HISTORY_MSG_43;Exposure - Tone curve 2 mode
 HISTORY_MSG_44;Lum. denoising radius
 HISTORY_MSG_45;Lum. denoising edge tolerance
-HISTORY_MSG_46;Color denoising
-HISTORY_MSG_47;Blend ICC highlights with matrix
+HISTORY_MSG_46;Resize - Long Edge
+HISTORY_MSG_47;Resize - Short Edge
 HISTORY_MSG_48;DCP - Tone curve
 HISTORY_MSG_49;DCP illuminant
 HISTORY_MSG_50;Shadows/Highlights
@@ -1924,10 +1924,14 @@ TP_RESIZE_H;Height:
 TP_RESIZE_HEIGHT;Height
 TP_RESIZE_LABEL;Resize
 TP_RESIZE_LANCZOS;Lanczos
+TP_RESIZE_LE;Long Edge:
+TP_RESIZE_LONG;Long Edge
 TP_RESIZE_METHOD;Method:
 TP_RESIZE_NEAREST;Nearest
 TP_RESIZE_SCALE;Scale
 TP_RESIZE_SPECIFY;Specify:
+TP_RESIZE_SE;Short Edge:
+TP_RESIZE_SHORT;Short Edge
 TP_RESIZE_W;Width:
 TP_RESIZE_WIDTH;Width
 TP_RETINEX_CONTEDIT_HSL;HSL histogram
--- a/rtengine/jdatasrc.cc
+++ b/rtengine/jdatasrc.cc
@@ -4,6 +4,13 @@
 #include "jpeg.h"
 
 /*
+ * HA 2021-03-12
+ * Modification according to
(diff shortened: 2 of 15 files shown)


rtdata/languages/default
```diff
@@ -297,8 +297,8 @@ HISTORY_MSG_42;Exposure - Tone curve 2
 HISTORY_MSG_43;Exposure - Tone curve 2 mode
 HISTORY_MSG_44;Lum. denoising radius
 HISTORY_MSG_45;Lum. denoising edge tolerance
-HISTORY_MSG_46;Color denoising
-HISTORY_MSG_47;Blend ICC highlights with matrix
+HISTORY_MSG_46;Resize - Long Edge
```
Thanatomanic, line 300: What is going on here? You can't just change history messages...

rtengine/jdatasrc.cc
```diff
@@ -3,6 +3,13 @@
...
 #include "jpeg.h"
 
+/*
+ * HA 2021-03-12
+ * Modification according to
+ * https://github.com/Beep6581/RawTherapee/commit/be67261d0bdeb21ffe15fe968dd5bc16edd407e0
+ * because otherwise it won't compile.
+ */
+  
 /*
```
Thanatomanic, line 13: Where are the changes in this file coming from? These seem irrelevant to your PR.

rtgui/resize.cc
```diff
@@ -81,19 +83,46 @@ Resize::Resize () : FoldableToolPanel(this, "resize", M("TP_RESIZE_LABEL"), fals
     Gtk::HBox* sbox = Gtk::manage (new Gtk::HBox ());
     Gtk::HBox* wbox = Gtk::manage (new Gtk::HBox ());
     Gtk::HBox* hbox = Gtk::manage (new Gtk::HBox ());
```
Thanatomanic, line 85: This seems to be based on an older version of `resize.cc`. There have been changes https://github.com/Beep6581/RawTherapee/blob/53bb6e41bb1f4052bb85048fe72499cabcdcb1ec/rtgui/resize.cc#L91

Perhaps you need to rebase?

rtdata/languages/default
Thanatomanic: Perhaps you can in this case since they are marked obsolete later on. Still, I would like to know for sure, just in case we break some backwards compatibility... If @heckflosse has time, he might want to comment.

Thanatomanic (changes requested):
I have not tested the branch yet. There are some indications that you based this on a not-super-recent version of the `dev` branch (leading to merge conflicts). If you can look into that, that would be great. Another thing is the reuse of the history messages of which I have no clear idea if that is a good or bad idea...

rtengine/jdatasrc.cc
Bezierr: They are. But without them, it is apparently not possible to compile RT under Windows.

rtgui/resize.cc
Bezierr: It is in fact based on releases branch, not on dev branch, because the release version is what I use for actual raw processing. I'll try the rebasing as soon as I find the time.

rtdata/languages/default
Bezierr: Not familiar with the works of the message mechanism, and the way you assign/reassign message numbers, reusing obsolete messages seemed to me the safest way to avoid breaking something. I think this could be changed fairly easily if you let me know what it should be.

rtengine/jdatasrc.cc
Thanatomanic: But the commit you refer to clearly shows these changes have been part of the codebase since October last year. You probably branched off earlier than that. This is not a problem per se, just inconvenient because it introduces merge conflicts. However, adding these comments to `jdatasrc.cc` makes no sense, they are completely irrelevant to include in the final code.

Bezierr: Yes, releases branch is definitely older than that. That one, then, should also be resolved by a rebasing.

Bezierr:
Will be replaced by a new Pull Request.

rtgui/resize.cc
Beep6581: @bezierr please base new work on the `dev` branch.

PR #6211, opened 2021-04-20
dfmanager: cleanup

heckflosse:
Already some weeks ago I made a (possibly incomplete) review of the dark frame manager.

I changed the accumulator (which is used to average the dark frames) from unsigned int (relict from old times) to float.
Also some cleanups and code formating stuff


Floessie:
@heckflosse Thanks for i.a. having run a linter on the two files. Wow, there are so many bad aspects about the code, I don't know what to say. But here's an agenda for a real cleanup:

1. PIMPL `DFManager` so the crappy `dfInfo` doesn't need to be exposed.
2. Use smart pointers.
3. Don't expose non-const pointers in the interface.
4. Turn `DFManager` into a singleton w/o global variable `dfm`.
5. Use C++ stream API and `getline()` instead of reading with `fgets()` into a fixed buffer.

Best,
Flössie

Floessie:
That's all for now. 1, 3, and 4 are finished. But there's so much more that could be done... :expressionless: 

Please test.

Best,
Flössie

heckflosse:
Compiles and works fine :+1:

heckflosse:
LGTM!

Thanatomanic (approved):
I have gone through the code again and there doesn't seem to be anything strange. When running RT, the darkframes still load as expected.

Issue #4285, opened 2018-01-08
Support for JPEG-compressed DNG files


Beep6581:
Request from:
https://discuss.pixls.us/t/compressed-dng-support/6370

Raw file:
https://raw.pixls.us/getfile.php/2204/nice/Canon%20-%20EOS%205D%20Mark%20III%20-%2014bit%20(2.3471882640587).dng

The DNG file seems to come from Magic Lantern and uses JPEG compression.

```text
-TIFF:ExifByteOrder=Little-endian (Intel, II)
-TIFF-IFD0:BitsPerSample=14
-TIFF-IFD0:Compression=JPEG
-TIFF-IFD0:Make=Canon
-TIFF-IFD0:Model=Canon EOS 5D Mark III
-TIFF-IFD0:Software=MLV App
-TIFF-IFD0:DNGVersion=1.4.0.0
```

jedypod:
I would love to see support for LJ92 compressed "Cinema DNG" files in RawTherapee.

I have some more example files and some information which may help push this forward:

Compressed Cinema DNG are created by a few different cameras:
- Digital Bolex D16
- Canon DSLR Magic Lantern raw (converted to CinemaDNG/DNG)
- BlackMagic Cinema Cameras

When opening these DNG files in RawTherapee, it has problems decoding the file. See this screenshot:
![rt-lj92-dng](https://user-images.githubusercontent.com/3261432/42426597-1789c8be-82f7-11e8-870c-64394606a579.png)

There is some useful information on LJ92 DNG compression here:
https://thndl.com/how-dng-compresses-raw-data-with-lossless-jpeg92.html

That post mentions a lj92 compression library that is open sourced as part of the MLVRawViewer project:
https://bitbucket.org/baldand/mlrawviewer/src/master/liblj92/?at=master

I have also uploaded a few more example compressed cinema dng files, included below, as well as uploaded to https://raw.pixls.us/

[cinemadng.tar.gz](https://github.com/Beep6581/RawTherapee/files/2174350/cinemadng.tar.gz)

I hope this helps! Thank you for your consideration!

nvki:
I would like to +1 this request as I'm having the same issue opening .dng files from my Blackmagic Pocket Cinema Camera. In addition to jeypod's example files I can add a example file with a color checker: [bmpcc_cinemadng](https://drive.google.com/open?id=1eVqBzSsnyRiuP7tw9Tv3Rs0_bWKIPdUG)

[...]

PR #6214, opened 2021-04-22
Fix predictor modes 2-7 for lossless JPEG

### LINKED ISSUE #6036: Apple Pro RAW support
Hello, I just found samples of new raw format.
https://www.reddit.com/r/iphone/comments/jyjtt8/proraw_example_files/
Surprisingly, they are in .DNG format(though might have some extra metadata / AI stuff). 
As a comparison with normal DNG, Pro RAW appears 2 times larger per photo.
When I open using rawtherapee, the files look pinkish and with broken debayering.

What are the chances that support of those files will ever be added? 
What is your opinion on the potential of the new format and how is it better than the traditional .DNG standard?

SimonSegerblomRex:
Hi!

Predictor modes 2-7 were broken for `lossless_dng_load_raw` (which is used when "Canon" or "BlackMagic" isn't part of `Make`). I just copy-pasted one line from `dcraw.c` to fix the issue.

After this fix it should be possible to remove lj92 (the lib + `lossless_dnglj92_load_raw`) and always use `lossless_dng_load_raw` without having any special case for `Make`. (At least all the files I could find in #4285 seem to work with `lossless_dng_load_raw` after this fix.)

[predictor-example-files-make-test.zip](https://github.com/Beep6581/RawTherapee/files/6358269/predictor-example-files-make-test.zip)

Resolves #6036

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -923,7 +923,7 @@ ushort * CLASS ljpeg_row (int jrow, struct jhead *jh)
     }
     getbits(-1);
   }
-  FORC3 row[c] = (jh->row + ((jrow & 1) + 1) * (jh->wide*jh->clrs*((jrow+c) & 1)));
+  FORC3 row[c] = jh->row + jh->wide*jh->clrs*((jrow+c) & 1);
   for (col=0; col < jh->wide; col++)
     FORC(jh->clrs) {
       diff = ljpeg_diff (jh->huff[c]);
@@ -9078,7 +9078,7 @@ void CLASS adobe_coeff (const char *make, const char *model)
     RT_matrix_from_constant = ThreeValBool::T;
   }
   // -- RT --------------------------------------------------------------------
-  
+
   for (i=0; i < sizeof table / sizeof *table; i++)
     if (!strncmp (name, table[i].prefix, strlen(table[i].prefix))) {
       if (RT_blacklevel_from_constant == ThreeValBool::T && table[i].black)   black   = (ushort) table[i].black;


rtengine/dcraw.cc
```diff
@@ -9078,7 +9078,7 @@ void CLASS adobe_coeff (const char *make, const char *model)
     RT_matrix_from_constant = ThreeValBool::T;
   }
   // -- RT --------------------------------------------------------------------
-  
+
```
SimonSegerblomRex, line 9081: Sorry about this... (auto-trimming whitespace in my editor...)

rtengine/dcraw.cc
```diff
@@ -923,7 +923,7 @@ ushort * CLASS ljpeg_row (int jrow, struct jhead *jh)
     }
     getbits(-1);
   }
-  FORC3 row[c] = (jh->row + ((jrow & 1) + 1) * (jh->wide*jh->clrs*((jrow+c) & 1)));
```
SimonSegerblomRex, line 926: This was changed by @heckflosse [here](https://github.com/Beep6581/RawTherapee/commit/f794cf684b18a9c2ca9b4ac350a10fd1ed37fc90#diff-3ca4dd32af70e9e6310586f5b5d92f55fd17859a3bc5bc9eeb4924c637846907L870). It looks like it was supposed to only be an optimization, so I guess it was just a lack of images using predictor modes > 1 back then that introduced the regression.

rtengine/dcraw.cc
Floessie: No need to be sorry. This is a valid change. The one adding unnecessary whitespace made a mistake.

SimonSegerblomRex:
I'm not planning to change anything else in this pull request. Once this has been merged I'll create a new pull request for removing lj92.

Thanatomanic:
I've tried your fixes and they work. Apple Pro RAW files also load properly. Great stuff. :+1: for a merge. Let's wait a few more days to see if anybody else wants to comment or try it out.

SimonSegerblomRex:
@Thanatomanic, I think it's safe to merge this, but let me know if you need help verifying the change with a particular set of images.

Thanatomanic:
Thanks again for the PR, looking forward to the next one to further simply processing as you described!
If you do, please feel free to add your name to the `AUTHORS.txt` list

PR #6215, opened 2021-04-23
Add resizing options "long edge" and "short edge" #5182

Bezierr:
Added the options "Long Edge" and "Short Edge" to the Resize functionality.

(Rebased with dev)


Thanatomanic:
I'll give it a testrun hopefully this weekend.

rtengine/procevents.h
```diff
@@ -68,8 +68,8 @@ enum ProcEventCode {
     EvToneCurveMode2 = 42,
     obsolete_43 = 43, // obsolete
     obsolete_44 = 44, // obsolete
-    obsolete_45 = 45, // obsolete
-    obsolete_46 = 46, // obsolete
+    EvResizeLongedge = 45,
+    EvResizeShortedge = 46,
```
Lawrence37: I had to deal with this error-prone method of adding events when I was working on the perspective correction tool. There's actually a better way by using `ProcEvent ProcEventMapper::newEvent(int, const std::string &)` to create an event programmatically instead of hard-coding it here and the corresponding refresh type in refreshmap.cc.

This is the function https://github.com/Beep6581/RawTherapee/blob/8c477ffb8b813012a5a8de1ce0e47b43dd93d799/rtgui/eventmapper.h#L31 and here it is in action https://github.com/Beep6581/RawTherapee/blob/8c477ffb8b813012a5a8de1ce0e47b43dd93d799/rtgui/perspective.cc#L92

It has an added bonus of allowing you to declare descriptive history message keys!

rtengine/refreshmap.cc
```diff
@@ -72,9 +72,9 @@ int refreshmap[rtengine::NUMOFEVENTS] = {
     AUTOEXP,         // EvToneCurveMode2,
     0,                // EvLDNRadius: obsolete,
     0,                // EvLDNEdgeTolerance: obsolete,
-    0,                // EvCDNEnabled:obsolete,
-    0,                // free entry
-    RGBCURVE | M_AUTOEXP, // EvDCPToneCurve,
+    RESIZE,           // EvResizeLongedge,
+    RESIZE,           // EvResizeShortedge,
+    RGBCURVE|M_AUTOEXP, // EvDCPToneCurve,
```
Lawrence37: By programmatically creating the event, this won't be necessary.

rtgui/resize.cc
```diff
@@ -88,22 +90,49 @@ Resize::Resize () : FoldableToolPanel(this, "resize", M("TP_RESIZE_LABEL"), fals
...
 
-    Gtk::Box* sbox = Gtk::manage (new Gtk::Box ());
-    Gtk::Box* wbox = Gtk::manage (new Gtk::Box ());
-    Gtk::Box* hbox = Gtk::manage (new Gtk::Box ());
+    Gtk::HBox* sbox = Gtk::manage (new Gtk::HBox ());
+    Gtk::HBox* wbox = Gtk::manage (new Gtk::HBox ());
+    Gtk::HBox* hbox = Gtk::manage (new Gtk::HBox ());
+    Gtk::HBox* ebox = Gtk::manage (new Gtk::HBox ());
+    Gtk::HBox* lebox = Gtk::manage (new Gtk::HBox ());
+    Gtk::HBox* sebox = Gtk::manage (new Gtk::HBox ());
```
Lawrence37: Please change to `Gtk::Box`.

rtgui/resize.cc
```diff
@@ -88,22 +90,49 @@ Resize::Resize () : FoldableToolPanel(this, "resize", M("TP_RESIZE_LABEL"), fals
...
+
     w = Gtk::manage (new MySpinButton ());
+    w->set_width_chars(5);
+    setExpandAlignProperties(w, false, false, Gtk::ALIGN_END, Gtk::ALIGN_CENTER);
     h = Gtk::manage (new MySpinButton ());
-    wbox->set_spacing(3);
```
Lawrence37, line 96: Was this line deleted for a reason?

rtgui/resize.cc
```diff
@@ -431,6 +491,38 @@ void Resize::setDimensions ()
                     break;
                 }
 
+                case 4: {
+                    // Long edge mode
+                    if (refw > refh) {
+                        w->set_value (le->get_value ());
+                        const double tmp_scale = le->get_value() / static_cast<double>(refw);
+                        scale->setValue(tmp_scale);
+                        h->set_value(static_cast<double>(static_cast<int>(static_cast<double>(refh) * tmp_scale + 0.5)));
```
Lawrence37, line 502: This is some funny looking code. I wonder why the original code used casts.

rtgui/resize.cc
Thanatomanic: I think it doesn't really matter, and is actually preferable because styling is something CSS can do.

rtengine/procevents.h
Bezierr: Thanks for your feedback. I'm currently very busy with my day job but will look into this as soon as possible.

Lawrence37 (approved):

Thanatomanic:
image

Everything works fine. However, I don't like the extra clutter in the GUI. Would it be possible to hide the 'Width' and 'Height' row whenever 'Long edge' or 'Short edge' is selected and vice versa?
I also notice that when you change the value for an edge, the width and height are updated (although the boxes are greyed out), but the size of the other edge isn't (which is also greyed out). That looks inconsistent and potentially confusing. Should be easy to also update the remaining value, right?

Bezierr:
I suppose I can change it in the way you propose. But...
I personally don't like the behavior that input into one field changes the value of others. I only implemented it this way because that is how it worked with width/height. The reason that I don't like it is as follows: I may have standard values that I frequently use for width, height, long edge, and/or short edge. If these values change whenever I enter something in *another* field, I'll always have to re-enter them when I need them.

My preferred user interface actually would be: Only the input field(s) needed for the currently selected "Specify" visible; potentially and additionally showing the resulting image dimensions as an output-only information, e.g. as done in the Crop section with the PPI input. That, of course, would be a deviation from how it has been done so far (and I would have to find out first how this can be done...).

Please advise.

Thanatomanic:
> My preferred user interface actually would be: Only the input field(s) needed for the currently selected "Specify" visible; potentially and additionally showing the resulting image dimensions as an output-only information, e.g. as done in the Crop section with the PPI input. That, of course, would be a deviation from how it has been done so far (and I would have to find out first how this can be done...).

I think this is actually a good suggestion: don't reuse interface elements for different purposes (setting a value, or disabling it and informing about a changed value). I would be happy if you could try to implement your idea to see how that looks like. Maybe give it a try on a different branch, just in case the consensus is to merge _this_ branch but not (yet) your other idea. What do you think about that?

Bezierr:
Thanks. In the meantime, because it was easier to do, I changed the GUI largely to what you proposed:

- only show width/height input fields when width, height, or bounding box is selected;
- only show long edge/short edge input fields when long edge or short edge is selected;
- adjust short edge value whenever long edge value is changed, and vice versa;
- I also removed adjusting width/height (which are not visible anyway) whenever long edge/short edge value is changed.

To me, this now seems a consistent user interface. Width affects height (and vice versa), but does not affect long edge/short edge, and is not affected by it. And long edge affects short edge (and vice versa), but does not affect width/height (and is not affected by it).

Have to do some more thorough testing before uploading it; hopefully within a couple of days.

Bezierr:
When testing, I found unexpected behavior that is already there in RT 5.8 release version (i.e. not introduced by my changes). To reproduce:

- open RT
- open a picture
- resize / width / 5000
=> corresponding height is displayed correctly
- crop / do not lock aspect ratio (so aspect ratio is changed after cropping)
=> height in resize is adjusted correctly
- save partial profile with resize parameters only
- open saved .pp3 file
=> saved value for height is not the one displayed in the GUI

Is that intended behavior? If so, why? Or can I assume that this is a bug?

Bezierr:
It seems to me that the above observation has something to to with croph and cropw apparently not being re-set when a picture is loaded or a profile is changed (EvPhotoLoaded, EvProfileChanged), which apparently can lead to messed-up values when switching between cropped and non-cropped images. It also seems that, in such a situation, correctly displayed values can be forced by disabling/enabling (or the other way round) the Crop panel. It further seems that when exporting a picture the resize settings are applied correctly, even though displayed incorrectly.
Anyway, for the time being, I'll leave it at that and may again look into this later on.

Thanatomanic:
This was alternatively merged here: https://github.com/Beep6581/RawTherapee/pull/6263

PR #6219, opened 2021-04-28
RFC : Fix parametric curve

### LINKED ISSUE #5859: Error in any parametric curve
Error in any parametric curve.
Points appear instead of curves:
I think this behavior is not correct. I don't know what to call it. 
It occurs with any parametric curve: Tone, Lab, local adjustments ... etc.
Process:
**1)**
CParametrica01
.
**2)**
CParametrica02
.
**3)**
CParametrica03

-----
Version: 5.8-2374-g746be1816
Branch: dev
Commit: 746be1816
Commit date: 2020-07-19
Compiler: gcc\ 10.1.0
Processor: generic x86
System: Windows
Bit depth: 64 bits
Gtkmm: V3.24.2
Lensfun: V0.3.2.0
Build type: Release
Build flags: -mwin32\ -m64\ -mthreads\ -msse2\ -std=c++11\ -mtune=generic\ -Werror=unused-label\ -Werror=delete-incomplete\ -fno-math-errno\ -Wall\ -Wuninitialized\ -Wcast-qual\ -Wno-deprecated-declarations\ -Wno-unused-result\ -Wunused-macros\ -fopenmp\ -Werror=unknown-pragmas\ -Wno-aggressive-loop-optimizations\ -DNDEBUG\ -O3\ -ftree-vectorize
Link flags: -m64\ -mthreads\ -static-libgcc\ -mtune=generic\ -s\ -O3
OpenMP support: ON
MMAP support: ON
Build OS: Linux 4.4.0-101-generic x86_64
Build date: Sun, 19 Jul 2020 21:20:58 +0000 UTC
Build epoch: 1595193658
Build UUID: c19ead07-bb99-4cd4-a2df-8b86fb414179

jonathanBieler:
I've been looking at the parametric curve glitches, one of the main culprit are parameters at 1.0 or 0.0 (1.0 for the "smoothness" of the curve and 0.0 for the sliders). That said the clamping to 1e-14 were also causing some issues, and removing those lead to segfaults in some cases (not sure why), so I've replaced the `xlog(xexp(...` by `pow` and that solved the crashes. Is there a good reason to use those instead of pow ? Performance wise I don't really see a difference.

raw_curve

Fixes #5859

### DIFF
--- a/rtengine/curves.h
+++ b/rtengine/curves.h
@@ -464,15 +464,6 @@ protected:
             : t + (1.0 - sh) * hl * (0.5 + CurveFactory::baseu_alt(2.0 * x - 1.0) * 0.5) + sh * (1.0 - hl) * (0.5 + CurveFactory::basel_alt(2.0 * x - 1.0) * 0.5);
     }
 
-    //compute pow(x,a) for x, a in [0,1]
-    static inline double curve_pow(double x, double a){
-
-        if (x <= 0.0) return 0.0;
-        double y = xlog(x)*a;
-        if (xisinf(y) ) return 0.0;
-        return xexp(y);
-    }
-
     void fillHash();
     void fillDyByDx();
 
--- a/rtengine/curves.h
+++ b/rtengine/curves.h
@@ -142,10 +142,18 @@ protected:
         double lx = xlog(x);
         return m2 * x + (1.0 - m2) * (2.0 - xexp(k * lx)) * xexp(l * lx);
     }
+    static inline double basel_alt(double x)
+    {
+        return (2.0 - x) * x * x * x;
+    }
     // basic concave function between (0,0) and (1,1). m1 and m2 controls the slope at the start and end point
     static inline double baseu(double x, double m1, double m2)
     {
         return 1.0 - basel(1.0 - x, m1, m2);
+    }
+    static inline double baseu_alt(double x)
+    {
+        return x * (2.0 + (x - 2.0) * x * x);
     }
     // convex curve between (0,0) and (1,1) with slope m at (0,0). hr controls the highlight recovery
     static inline double cupper(double x, double m, double hr)
@@ -447,6 +455,22 @@ protected:
     static inline double pfull(double x, double prot, double sh, double hl)
     {
         return (1 - sh) * (1 - hl) * p00(x, prot) + sh * hl * p11(x, prot) + (1 - sh) * hl * p01(x, prot) + sh * (1 - hl) * p10(x, prot);
+    }
+    static inline double pfull_alt(double x, double sh, double hl)
+    {
+        double t = (1.0 - sh) * (1.0 - hl) * CurveFactory::basel_alt(x) + sh * hl * CurveFactory::baseu_alt(x);
+        return x <= 0.5
+            ? t + (1.0 - sh) * hl * CurveFactory::basel_alt(2.0 * x) * 0.5 + sh * (1.0 - hl) * CurveFactory::baseu_alt(2.0 * x) * 0.5
+            : t + (1.0 - sh) * hl * (0.5 + CurveFactory::baseu_alt(2.0 * x - 1.0) * 0.5) + sh * (1.0 - hl) * (0.5 + CurveFactory::basel_alt(2.0 * x - 1.0) * 0.5);
+    }
+
+    //compute pow(x,a) for x, a in [0,1]
+    static inline double curve_pow(double x, double a){
+
+        if (x <= 0.0) return 0.0;
+        double y = xlog(x)*a;
+        if (xisinf(y) ) return 0.0;
+        return xexp(y);
     }
 
     void fillHash();
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -104,6 +104,13 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
                 for (int i = 1; i < 4; i++) {
                     x[i] = min(max(p[i], 0.001), 0.99);
                 }
+                // the curve becomes glitchy when the highlight parameter is too high
+                // soft-clamp is to 96% (starting from 95%) and make sure we keep 2% between parameters
+                if (x[3] > 0.95){
+                    x[3] = 0.95 + 0.01*(x[3]-0.95)/0.05;
+                    x[2] = min(x[2], x[3] - 0.02);
+                    x[1] = min(x[1], x[2] - 0.02);
+                }
 
                 for (int i = 4; i < 8; i++) {
                     x[i] = (p[i] + 100.0) / 200.0;
@@ -116,8 +123,8 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
                 }
 
                 mc = -xlog(2.0) / xlog(x[2]);
-                double mbase = pfull (0.5, x[8], x[6], x[5]);
-                mfc = mbase <= 1e-14 ? 0.0 : xexp(xlog(mbase) / mc);        // value of the curve at the center point
+                double mbase = pfull_alt (0.5, x[6], x[5]);
+                mfc = xexp(xlog(mbase) / mc);        // value of the curve at the center point
                 msc = -xlog(2.0) / xlog(x[1] / x[2]);
                 mhc = -xlog(2.0) / xlog((x[3] - x[2]) / (1 - x[2]));
             }
@@ -424,7 +431,6 @@ void DiagonalCurve::catmull_rom_set()
 
 /*****************************************************************************/
 
-
 double DiagonalCurve::getVal (double t) const
 {
 
@@ -435,20 +441,20 @@ double DiagonalCurve::getVal (double t) const
             return 0.0;
         }
 
-        double tv = xexp(mc * xlog(t));
-        double base = pfull (tv, x[8], x[6], x[5]);
-        double stretched = base <= 1e-14 ? 0.0 : xexp(xlog(base) / mc);
+        double tv = xexp(max(mc * xlog(t),-236.0)); // prevents numerical issues when calling pfull, at the cost of minor artifacts
+        double base = pfull_alt (tv, x[6], x[5]);
+        double stretched = xexp(xlog(base) / mc);
 
         if (t < x[2]) {
             // add shadows effect:
-            double stv = xexp(msc * xlog(stretched / mfc));
-            double sbase = pfull (stv, x[8], x[7], 0.5);
-            return mfc * (sbase <= 1e-14 ? 0.0 : xexp(xlog(sbase) / msc));
+            double stv = xexp(max(msc * xlog(stretched / mfc),-236.0));
+            double sbase = pfull_alt (stv, x[7], 0.5);
+            return mfc * xexp(xlog(sbase) / msc);
         } else {
             // add highlights effect:
-            double htv = xexp(mhc * xlog((stretched - mfc) / (1 - mfc)));
-            double hbase = pfull (htv, x[8], 0.5, x[4]);
-            return mfc + (1 - mfc) * (hbase <= 1e-14 ? 0.0 : xexp(xlog(hbase) / mhc));
+            double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
+            double hbase = pfull_alt (htv, 0.5, x[4]);
+            return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
         }
 
         break;
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -448,9 +448,8 @@ double DiagonalCurve::getVal (double t) const
             double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
             double hbase = pfull_alt (htv, 0.5, x[4]);
             //this part of the curve isn't affected by highlight, return the base curve
-            if (hbase < 1e-14 ){ 
-                hbase = xexp(max(xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
-                return mfc + (1.0 - mfc) * hbase;
+            if (hbase < 1e-6 ){ 
+                return stretched;
             } else {
                 return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
             }
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -104,13 +104,6 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
                 for (int i = 1; i < 4; i++) {
                     x[i] = min(max(p[i], 0.001), 0.99);
                 }
-                // the curve becomes glitchy when the highlight parameter is too high
-                // soft-clamp is to 96% (starting from 95%) and make sure we keep 2% between parameters
-                if (x[3] > 0.95){
-                    x[3] = 0.95 + 0.01*(x[3]-0.95)/0.05;
-                    x[2] = min(x[2], x[3] - 0.02);
-                    x[1] = min(x[1], x[2] - 0.02);
-                }
 
                 for (int i = 4; i < 8; i++) {
                     x[i] = (p[i] + 100.0) / 200.0;
@@ -454,7 +447,13 @@ double DiagonalCurve::getVal (double t) const
             // add highlights effect:
             double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
             double hbase = pfull_alt (htv, 0.5, x[4]);
-            return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
+            //this part of the curve isn't affected by highlight, return the base curve
+            if (hbase < 1e-14 ){ 
+                hbase = xexp(max(xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
+                return mfc + (1.0 - mfc) * hbase;
+            } else {
+                return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
+            }
         }
 
         break;


rtengine/diagonalcurves.cc
```diff
@@ -106,18 +106,18 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
...
-                    x[i] = (p[i] + 100.0) / 200.0;
+                    x[i] = max((p[i] + 100.0) / 200.0, 0.001);
                 }
 
                 if (p.size() < 9) {
-                    x[8] = 1.0;
+                    x[8] = 0.99;
                 } else {
-                    x[8] = p[8] / 100.0;
+                    x[8] = min(p[8] / 100.0, 0.99);
```
Lawrence37: Clamping these values results in a very small but imperceptible change in the output. Is it possible to come up with an alternative solution, or use values closer to the original to minimize the difference?

Thanatomanic:
@jonathanBieler Thanks for looking into this issue. It's likely the segfaults are because computation can lead to attempted computation of negative or zero logarithms and resulting in NaN's.

Please do not use `pow`, but `xpow` instead. These are vector-optimized versions which usually outperform the regular functions (see `sleef.h` for details).

I find it hard to pinpoint whether the changes to the functions actually contribute to a better numerical stability. Can you comment on that?

rtengine/diagonalcurves.cc
```diff
@@ -106,18 +106,18 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
                 }
 
                 for (int i = 4; i < 8; i++) {
-                    x[i] = (p[i] + 100.0) / 200.0;
+                    x[i] = max((p[i] + 100.0) / 200.0, 0.001);
```
Thanatomanic: If I understand the whole chain of events properly, I don't think this is an acceptable change from a backward compatibility point of view: people that have previously set their sliders to -100 would get a different result when reprocessing the file.
Also: only adding a margin at the bottom seems arbitrary, because I can also get the artifacts you try to solve when setting Lights -> 100 and pushing the third locus to the far right.

rtengine/diagonalcurves.cc
Thanatomanic: Same comment for backwards compatibility.
I also cannot find where the eighth parameter is actually used in the GUI, in my quick tests it was always 1.0.

rtengine/diagonalcurves.cc
```diff
@@ -434,21 +433,22 @@ double DiagonalCurve::getVal (double t) const
...
             return 0.0;
         }
-
-        double tv = xexp(mc * xlog(t));
+    
+        double tv = pow(t, mc);
         double base = pfull (tv, x[8], x[6], x[5]);
-        double stretched = base <= 1e-14 ? 0.0 : xexp(xlog(base) / mc);
-
+        double stretched = pow(base, 1 / mc);
```
Thanatomanic: For clarity, please write `1.0`.
As far as I see, `mc` has the approximate range of 0.1 .. 69.0 (inverted between 0.0145 .. 9.966). The value of `base` can be between 0..1. So `stretched` can easily get very, very small. I'm not sure removing the guard is such a good idea...

heckflosse:
> Is there a good reason to use those instead of pow ? Performance wise I don't really see a difference.

Iirc, I used xlog(xexp(..)) beause this is calculated per thumb. Means, changing this to pow() for a single file is not relevant for performance, but for a large folder it is relevant.

rtengine/diagonalcurves.cc
jonathanBieler: The 8th parameter controls the smoothness of the curve (it becomes "blocky" if turned down), and seems to be fixed indeed. When set to 1.0 `pfull` starts to have some issue (returns 0.0 regardless of t) and thus one would have to look at this function to accommodate that value.

rtengine/diagonalcurves.cc
jonathanBieler: Basically for reasonable parameter values `base` and other variable operates in these small range of values, and clamping them produces artefacts in the curve. Maybe there's way to rework how the curve is computed to avoid that but I don't understand it enough to do it myself. That said I don't think there's really a fundamental issue with using these small values, as long as it doesn't produces infinities and NaNs.

jonathanBieler:
I've introduced a function to compute `pow` using `xexp` and `xlog` in a "safe" way (not totally sure about that, but I haven't managed to make it crash) and clamped the parameters using smaller margins. Technically the output is different but it's really tiny.

Thanatomanic:
I don't understand why you wrote a new function, when it already seems to exist? Is there an additional benefit?
Edit: the function below also makes use of a combined log and exp. _Maybe_ this is slower... 

https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/sleef.h#L625-L641

Thanatomanic:
I did some work on understanding the curve mechanics and to come up with a possible solution without having to change the current limits. Turns out, the code is quite weird and makes use of logarithms and exponents where that actually doesn't seem to be necessary most of the time (except one particular case, which I don't think ever happens).

Let's take a deep dive.

**Input parameters**
The GUI is linked to 7 variables that play a role in `diagonalcurves.cc`. First of all, the position of the anchor points of the curve: `x[1]` < `x[2]` < `x[3]`, limited between 0.001 and 0.99. I haven't been able to find how these points are guaranteed to be strictly increasing, but we can safely assume they are. From using the curve, they can also not get too close to each other, but I also don't know how that works. It think is also safe to assume their quotient is never unity.
Next, we have `x[4]` linked to the 'highlights', `x[5]` linked to the 'lights', `x[6]` linked to the 'darks' and `x[7]` linked to the 'shadows'. All these values lie between 0 and 1.
There is also `x[8]`, but I cannot find any mention in the code that this value can be modified in the GUI. Therefore, I think it is safe to assume it is always unity, as per https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/diagonalcurves.cc#L112-L116

**Auxiliary variables**
Creating and evaluating the curve takes place [here](https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/diagonalcurves.cc#L118-L122) and [here](
https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/diagonalcurves.cc#L438-L452). It's easy to do a range analysis on the variables (hint: potential numerical issues become quite obvious quickly), but it's more interesting to look at the `pfull` function which plays a central role.

**pfull**
The `pfull` function is [defined in `curves.h`](https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/curves.h#L431-L450) and it invokes a chain of related functions in various ways. This code was probably written like this to have a general and structured method to calculate a smooth function. However, if we want to look into ways to simplify the calculation and make it less error-prone, a first step is to note that `pfull`'s second argument is always `x[8]` which we assumed to be unity.
Using this assumption and substituting the four helper functions `p00`, `p11`, `p01` and `p10` directly into the definition of `pfull`, the return value looks like this pseudocode,
```
if (x <= 1/2) {
    (1-sh)*(1-hl)*clower(x,2,1) + sh*hl*cupper(x,2,1) + (1-sh)*hl*clower(2*x,2,1)/2 + sh*(1-hl)*cupper(2*x,2,1)/2;
} else {
    (1-sh)*(1-hl)*clower(x,2,1) + sh*hl*cupper(x,2,1) + (1-sh)*hl*(1+cupper(2*x-1,2,1))/2 + sh*(1-hl)*(1+clower(2*x-1,2,1))/2;
}
```
which can be rewritten using that [`clower(x,m,sr) = 1-cupper(1-x,m,sr)`](https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/curves.h#L170-L174),
```
if (x <= 1/2) {
    (1-sh)*(1-hl)*(1-cupper(1-x,2,1)) + sh*hl*cupper(x,2,1) + (1-sh)*hl*(1-cupper(1-2*x,2,1))/2 + sh*(1-hl)*cupper(2*x,2,1)/2;
} else {
    (1-sh)*(1-hl)*(1-cupper(1-x,2,1)) + sh*hl*cupper(x,2,1) + (1-sh)*hl*(1+cupper(2*x-1,2,1))/2 + sh*(1-hl)*(1-cupper(2-2*x,2,1)/2);
}
```
Of particular note here is that the second and third argument of `cupper` are _always_ 2 and 1 respectively. This can simplify the definition of the `cupper` function greatly to this,
```
    static inline double cupper(double x)
    {
        if (x >= 1.0) return 1.0;
        if (x < 0) return x * 2.0;
        return baseu(x,2.0,0);
    }
```
Where it must be noted that as far as I see, that `x` will never be greater than 1 or lower than 0. So we could immediately substitute the `cupper` function by a call to `baseu`, which is a slightly modified call to [`basel`](https://github.com/Beep6581/RawTherapee/blob/980a16abb0834b0138d7d99e6dc94e05aa8ead89/rtengine/curves.h#L133-L144). Since the second and third argument for the `basel` function are fixed, we can simplify and rewrite that,
```
    static inline double basel(double x)
    {
        return (2.0 - x) * x * x * x; // original derivation leads to (2.0 - x) * xexp(3.0 * xlog(x));
    }
```
so that after a little rewriting,
```
    static inline double baseu(double x)
    {
        return x * (2.0 - 2.0 * x * x + x * x * x); // or x * (2.0 + (x - 2.0) * x * x)
    }
```
Putting this all together, the `pfull` function can be written as follows, where all the calls to `baseu` and `basel` are simply polynomial functions,
```
if (x <= 1/2) {
    (1-sh)*(1-hl)*basel(x) + sh*hl*baseu(x) + (1-sh)*hl*basel(2*x)/2 + sh*(1-hl)*baseu(2*x)/2;
} else {
    (1-sh)*(1-hl)*basel(x) + sh*hl*baseu(x) + (1-sh)*hl*(1+baseu(2*x-1))/2 + sh*(1-hl)*(1+basel(2*x-1)/2);
}
```

Voilá, we just got rid of a lot of exponentiation and logarithms. I am currently testing whether my maths checks out, but I'm hopeful. I'll provide a patch when it's working.

Thanatomanic:
Here's a diff with `dev` to test my suggestion. This should not give any more artifacts.

```diff
diff --git a/rtengine/curves.h b/rtengine/curves.h
index 8800e54db..02503aff2 100644
--- a/rtengine/curves.h
+++ b/rtengine/curves.h
@@ -142,11 +142,19 @@ protected:
         double lx = xlog(x);
         return m2 * x + (1.0 - m2) * (2.0 - xexp(k * lx)) * xexp(l * lx);
     }
+    static inline double basel_alt(double x)
+    {
+        return (2.0 - x) * x * x * x;
+    }
     // basic concave function between (0,0) and (1,1). m1 and m2 controls the slope at the start and end point
     static inline double baseu(double x, double m1, double m2)
     {
         return 1.0 - basel(1.0 - x, m1, m2);
     }
+    static inline double baseu_alt(double x)
+    {
+        return x * (2.0 + (x - 2.0) * x * x);
+    }
     // convex curve between (0,0) and (1,1) with slope m at (0,0). hr controls the highlight recovery
     static inline double cupper(double x, double m, double hr)
     {
@@ -448,6 +456,13 @@ protected:
     {
         return (1 - sh) * (1 - hl) * p00(x, prot) + sh * hl * p11(x, prot) + (1 - sh) * hl * p01(x, prot) + sh * (1 - hl) * p10(x, prot);
     }
+    static inline double pfull_alt(double x, double sh, double hl)
+    {
+        double t = (1.0 - sh) * (1.0 - hl) * CurveFactory::basel_alt(x) + sh * hl * CurveFactory::baseu_alt(x);
+        return x <= 0.5
+            ? t + (1.0 - sh) * hl * CurveFactory::basel_alt(2.0 * x) * 0.5 + sh * (1.0 - hl) * CurveFactory::baseu_alt(2.0 * x) * 0.5
+            : t + (1.0 - sh) * hl * (0.5 + CurveFactory::baseu_alt(2.0 * x - 1.0) * 0.5) + sh * (1.0 - hl) * (0.5 + CurveFactory::basel_alt(2.0 * x - 1.0) * 0.5);
+    }
 
     void fillHash();
     void fillDyByDx();
diff --git a/rtengine/diagonalcurves.cc b/rtengine/diagonalcurves.cc
index 622683ea2..202ff62d1 100644
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -116,8 +116,10 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
                 }
 
                 mc = -xlog(2.0) / xlog(x[2]);
-                double mbase = pfull (0.5, x[8], x[6], x[5]);
-                mfc = mbase <= 1e-14 ? 0.0 : xexp(xlog(mbase) / mc);        // value of the curve at the center point
+                double mbase = pfull_alt (0.5, x[6], x[5]);
+                //double mbase = pfull (0.5, x[8], x[6], x[5]);
+                mfc = xexp(xlog(mbase) / mc);        // value of the curve at the center point
+                //mfc = mbase <= 1e-14 ? 0.0 : xexp(xlog(mbase) / mc);        // value of the curve at the center point
                 msc = -xlog(2.0) / xlog(x[1] / x[2]);
                 mhc = -xlog(2.0) / xlog((x[3] - x[2]) / (1 - x[2]));
             }
@@ -436,19 +438,25 @@ double DiagonalCurve::getVal (double t) const
         }
 
         double tv = xexp(mc * xlog(t));
-        double base = pfull (tv, x[8], x[6], x[5]);
-        double stretched = base <= 1e-14 ? 0.0 : xexp(xlog(base) / mc);
+        double base = pfull_alt (tv, x[6], x[5]);
+        //double base = pfull (tv, x[8], x[6], x[5]);
+        double stretched = xexp(xlog(base) / mc);
+        //double stretched = base <= 1e-14 ? 0.0 : xexp(xlog(base) / mc);
 
         if (t < x[2]) {
             // add shadows effect:
             double stv = xexp(msc * xlog(stretched / mfc));
-            double sbase = pfull (stv, x[8], x[7], 0.5);
-            return mfc * (sbase <= 1e-14 ? 0.0 : xexp(xlog(sbase) / msc));
+            double sbase = pfull_alt (stv, x[7], 0.5);
+            //double sbase = pfull (stv, x[8], x[7], 0.5);
+            return mfc * xexp(xlog(sbase) / msc);
+            //return mfc * (sbase <= 1e-14 ? 0.0 : xexp(xlog(sbase) / msc));
         } else {
             // add highlights effect:
             double htv = xexp(mhc * xlog((stretched - mfc) / (1 - mfc)));
-            double hbase = pfull (htv, x[8], 0.5, x[4]);
-            return mfc + (1 - mfc) * (hbase <= 1e-14 ? 0.0 : xexp(xlog(hbase) / mhc));
+            double hbase = pfull_alt (htv, 0.5, x[4]);
+            //double hbase = pfull (htv, x[8], 0.5, x[4]);
+            return mfc + (1 - mfc) * xexp(xlog(hbase) / mhc);
+            //return mfc + (1 - mfc) * (hbase <= 1e-14 ? 0.0 : xexp(xlog(hbase) / mhc));
         }
 
         break;
```

jonathanBieler:
Works well and it's certainly more elegant than my solution. That said it crashes for me if you get these settings and drag the middle point to the left (crash around 15%). That's what I was trying to fix with my curve_pow function (xpow crashes too).

crash

Thanatomanic:
@jonathanBieler I confirm the crash. Looking for a way to mitigate...

Thanatomanic:
@jonathanBieler After spending way too long on this, I propose the following compromise that I think retains the most numerical precision (and does not crash and does not change the slider limits). At extreme values of the sliders and positions of the control points it still gives minor artifacts in the curve (combination of middle control point, blacks and whites).



Could you please test?
```diff
diff --git a/rtengine/curves.h b/rtengine/curves.h
index 8800e54db..02503aff2 100644
--- a/rtengine/curves.h
+++ b/rtengine/curves.h
@@ -142,11 +142,19 @@ protected:
         double lx = xlog(x);
         return m2 * x + (1.0 - m2) * (2.0 - xexp(k * lx)) * xexp(l * lx);
     }
+    static inline double basel_alt(double x)
+    {
+        return (2.0 - x) * x * x * x;
+    }
     // basic concave function between (0,0) and (1,1). m1 and m2 controls the slope at the start and end point
     static inline double baseu(double x, double m1, double m2)
     {
         return 1.0 - basel(1.0 - x, m1, m2);
     }
+    static inline double baseu_alt(double x)
+    {
+        return x * (2.0 + (x - 2.0) * x * x);
+    }
     // convex curve between (0,0) and (1,1) with slope m at (0,0). hr controls the highlight recovery
     static inline double cupper(double x, double m, double hr)
     {
@@ -448,6 +456,13 @@ protected:
     {
         return (1 - sh) * (1 - hl) * p00(x, prot) + sh * hl * p11(x, prot) + (1 - sh) * hl * p01(x, prot) + sh * (1 - hl) * p10(x, prot);
     }
+    static inline double pfull_alt(double x, double sh, double hl)
+    {
+        double t = (1.0 - sh) * (1.0 - hl) * CurveFactory::basel_alt(x) + sh * hl * CurveFactory::baseu_alt(x);
+        return x <= 0.5
+            ? t + (1.0 - sh) * hl * CurveFactory::basel_alt(2.0 * x) * 0.5 + sh * (1.0 - hl) * CurveFactory::baseu_alt(2.0 * x) * 0.5
+            : t + (1.0 - sh) * hl * (0.5 + CurveFactory::baseu_alt(2.0 * x - 1.0) * 0.5) + sh * (1.0 - hl) * (0.5 + CurveFactory::basel_alt(2.0 * x - 1.0) * 0.5);
+    }
 
     void fillHash();
     void fillDyByDx();
diff --git a/rtengine/diagonalcurves.cc b/rtengine/diagonalcurves.cc
index 622683ea2..de947e872 100644
--- a/rtengine/diagonalcurves.cc
+++ b/rtengine/diagonalcurves.cc
@@ -116,8 +116,9 @@ DiagonalCurve::DiagonalCurve (const std::vector<double>& p, int poly_pn)
                 }
 
                 mc = -xlog(2.0) / xlog(x[2]);
-                double mbase = pfull (0.5, x[8], x[6], x[5]);
-                mfc = mbase <= 1e-14 ? 0.0 : xexp(xlog(mbase) / mc);        // value of the curve at the center point
+                double mbase = pfull_alt (0.5, x[6], x[5]); // pfull (0.5, x[8], x[6], x[5]);
+                mfc = xexp(xlog(mbase) / mc);        // value of the curve at the center point
+                //mfc = mbase <= 1e-14 ? 0.0 : xexp(xlog(mbase) / mc);        // value of the curve at the center point
                 msc = -xlog(2.0) / xlog(x[1] / x[2]);
                 mhc = -xlog(2.0) / xlog((x[3] - x[2]) / (1 - x[2]));
             }
@@ -435,20 +436,23 @@ double DiagonalCurve::getVal (double t) const
             return 0.0;
         }
 
-        double tv = xexp(mc * xlog(t));
-        double base = pfull (tv, x[8], x[6], x[5]);
-        double stretched = base <= 1e-14 ? 0.0 : xexp(xlog(base) / mc);
+        double tv = xexp(max(mc * xlog(t),-236.0)); // prevents numerical issues when calling pfull, at the cost of minor artifacts
+        double base = pfull_alt (tv, x[6], x[5]); // pfull (tv, x[8], x[6], x[5]);
+        double stretched = xexp(xlog(base) / mc);
+        //double stretched = base <= 1e-14 ? 0.0 : xexp(xlog(base) / mc);
 
         if (t < x[2]) {
             // add shadows effect:
-            double stv = xexp(msc * xlog(stretched / mfc));
-            double sbase = pfull (stv, x[8], x[7], 0.5);
-            return mfc * (sbase <= 1e-14 ? 0.0 : xexp(xlog(sbase) / msc));
+            double stv = xexp(max(msc * xlog(stretched / mfc),-236.0));
+            double sbase = pfull_alt (stv, x[7], 0.5); // pfull (stv, x[8], x[7], 0.5);
+            return mfc * xexp(xlog(sbase) / msc);
+            //return mfc * (sbase <= 1e-14 ? 0.0 : xexp(xlog(sbase) / msc));
         } else {
             // add highlights effect:
-            double htv = xexp(mhc * xlog((stretched - mfc) / (1 - mfc)));
-            double hbase = pfull (htv, x[8], 0.5, x[4]);
-            return mfc + (1 - mfc) * (hbase <= 1e-14 ? 0.0 : xexp(xlog(hbase) / mhc));
+            double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
+            double hbase = pfull_alt (htv, 0.5, x[4]); // pfull (htv, x[8], 0.5, x[4]);
+            return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
+            //return mfc + (1 - mfc) * (hbase <= 1e-14 ? 0.0 : xexp(xlog(hbase) / mhc));
         }
 
         break;
```

jonathanBieler:
It's still glitchy in some regimes but let's say it's an improvement over the current version :)

glitch

Thanatomanic:
@jonathanBieler Yes this is what remains. I don't see an easy way to remedy this, but I won't say it's impossible.

If you agree that this is a good solution for the moment, please feel free to update your branch with my patch and we can merge it into `dev`.

jonathanBieler:
I wasn't very satisfied with the remaining artefacts so I had one more take at this. What I've ended up doing is clamping only the highlight parameter to 96%. I think that's justified because that's the parameter range in which the current version starts to have glitches. If anybody was using this kind of parameter values they will have their output changed anyway if we fix the artefacts, so better to change it to something glitch free than to a different glitch. 

I made a progressive clamping so that the slider remains responsive until 100%.

Thanatomanic:
Two points about this:
* You clamp the values in the engine, not the GUI. I haven't tested it, but does that work bidirectionally?
* Clamping means we change current (long time available) behavior. @Beep6581 can you comment on backward compatibility in this case? I think it is warranted, but I'm not sure.

jonathanBieler:
@Thanatomanic actually I think I found a solution that doesn't require messing with the parameters. The glitches appears when we try to apply the highlight correction to part of the curve that are very far from the highlight tick. In these cases, the correction becomes negligible and we can just return the "base curve" instead. I've tested it and it seems to work just fine (to be confirmed though).

rtengine/diagonalcurves.cc
```diff
@@ -454,7 +447,13 @@ double DiagonalCurve::getVal (double t) const
             // add highlights effect:
             double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
             double hbase = pfull_alt (htv, 0.5, x[4]);
-            return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
+            //this part of the curve isn't affected by highlight, return the base curve
+            if (hbase < 1e-14 ){ 
+                hbase = xexp(max(xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
+                return mfc + (1.0 - mfc) * hbase;
```
Thanatomanic: If you write out the maths, this actually simplifies greatly to `return stretched`. Please check if you can.

rtengine/diagonalcurves.cc
```diff
@@ -454,7 +447,13 @@ double DiagonalCurve::getVal (double t) const
             // add highlights effect:
             double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
             double hbase = pfull_alt (htv, 0.5, x[4]);
-            return mfc + (1.0 - mfc) * xexp(xlog(hbase) / mhc);
+            //this part of the curve isn't affected by highlight, return the base curve
+            if (hbase < 1e-14 ){
```
Thanatomanic: This is a great idea and it works very well. In my tests you could even set a much higher threshold, even something like `1e-5` gives virtually negligible differences and a perfectly smooth curve. The benefit is a small speedup because you don't have to calculate `hbase`.

rtengine/diagonalcurves.cc
```diff
@@ -454,7 +447,13 @@ double DiagonalCurve::getVal (double t) const
             // add highlights effect:
             double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
             double hbase = pfull_alt (htv, 0.5, x[4]);
```
Thanatomanic, line 449: See comments below. You only need to calculate `hbase` for one branch.

rtengine/curves.h
```diff
@@ -448,6 +456,22 @@ class Curve
...
+    }
+
+    //compute pow(x,a) for x, a in [0,1]
+    static inline double curve_pow(double x, double a){
+
+        if (x <= 0.0) return 0.0;
+        double y = xlog(x)*a;
+        if (xisinf(y) ) return 0.0;
+        return xexp(y);
+    }
```
Thanatomanic: This is no longer needed, right?

jonathanBieler:
Made these changes, it seems to work fine in my tests and produce the same curves as in the previous version (minus the glitches), I went with 1e-6 just to be safe.

rtengine/diagonalcurves.cc
```diff
@@ -435,20 +434,25 @@ double DiagonalCurve::getVal (double t) const
...
+            return mfc * xexp(xlog(sbase) / msc);
         } else {
             // add highlights effect:
-            double htv = xexp(mhc * xlog((stretched - mfc) / (1 - mfc)));
-            double hbase = pfull (htv, x[8], 0.5, x[4]);
-            return mfc + (1 - mfc) * (hbase <= 1e-14 ? 0.0 : xexp(xlog(hbase) / mhc));
+            double htv = xexp(max(mhc * xlog((stretched - mfc) / (1.0 - mfc)),-236.0));
+            double hbase = pfull_alt (htv, 0.5, x[4]);
+            //this part of the curve isn't affected by highlight, return the base curve
+            if (hbase < 1e-6 ){
```
Thanatomanic, line 451: What I meant, is that you could check if `htv` is smaller than the threshold. You can then move the calculation of `hbase` to the false conditional.

Thanatomanic (changes requested):

Thanatomanic:
Hi @jonathanBieler, do you have time to fix my last comment, otherwise I'll just merge your work and apply the fix directly to `dev`?

jonathanBieler:
@Thanatomanic Sorry was busy lately, thanks for merging anyway.

Issue #6111, opened 2021-02-13
Inspector window does not update from within the editor / filmstrip


marter001:
i am confused by inspector... it should show the full view of the hovered pic, right?
pls, have a look at my pics: while editing a landscape i was hovering another landscape in the previews. What i expected, was, that one of those two files would get opened by inspector. But instead another pic, earlier edited and stored in the same folder, was shown.
And: The right side of the inspector-frame is transparent and doesnt show the whole pic.
![inspector_1](https://user-images.githubusercontent.com/78749513/107851493-b4624700-6e0a-11eb-8f8c-9bb09bf11c0d.jpg)
![inspector_2](https://user-images.githubusercontent.com/78749513/107851512-d8258d00-6e0a-11eb-8192-ee6419699a78.jpg)
This is shown, using shift-f in file browser
![inspector_3](https://user-images.githubusercontent.com/78749513/107851519-e96e9980-6e0a-11eb-91c9-12c03073e75d.jpg)

Im am using Win10 64bit, RT Version: 5.8-2751-gd369efb04 on a 4k-monitor at 3840x2160px
I was able to reproduce this in other folders too. Inspector showed pictures which had been edited earlier.

Thanatomanic:
So, the behavior in the editor mode is bugged, but I'm completely unsure what the expected behavior should be. To be honest, I had never noticed the 'Inspect' option in the context menu before and I can't seem to find an explanation in RawPedia about it.

The inspector window that opens can be panned and zoomed, but it is completely unrelated to either the selected picture in the filmstrip or the opened picture in the editor. When you go to the file browser, it does behave as expected from the inspect window: hovering over thumbnails updates the inspector.
Maybe @rfranke can take a look at what's going on? Could this introduced by your changes?

Thanatomanic:
@marter001 I have not been able to reproduce your partially transparent inspector.

marter001:
> 
> 
> @marter001 I have not been able to reproduce your partially transparent inspector.
@Thanatomanic : just scroll the mousewheel front:

![inspector_4](https://user-images.githubusercontent.com/78749513/107858148-f56e5180-6e32-11eb-8695-1aabc73c0e83.jpg)

Thanatomanic:
I'm sorry to say that scrolling and zooming still never gives me a transparent inspector window. At extreme zoom levels I can get some visible border, but otherwise the window behaves okay.

[...]

Issue #5946, opened 2020-10-11
Full-screen inspector: wrong background when embedded jpeg is too small


heckflosse:
When inspecting a file with a small embedded jpeg using the f-key, the background is wrong.
![grafik](https://user-images.githubusercontent.com/13733487/95675103-ec375480-0bb4-11eb-81d1-9b19a1a17463.png)

To reproduce, use https://raw.pixls.us/getfile.php/2764/nice/Phase%20One%20-%20P65+%20-%20IIQ%20L%20(4:3).IIQ

ff2000:
This fixes the issue for me:

```patch
diff --git a/rtgui/inspector.cc b/rtgui/inspector.cc
index 526343fca..745434dea 100644
--- a/rtgui/inspector.cc
+++ b/rtgui/inspector.cc
@@ -374,7 +374,7 @@ bool Inspector::on_draw(const ::Cairo::RefPtr< Cairo::Context> &cr)
         c = style->get_background_color (Gtk::STATE_FLAG_NORMAL);
         cr->set_source_rgb (c.get_red(), c.get_green(), c.get_blue());
         cr->set_line_width (0);
-        cr->rectangle (0, 0, availableSize.x, availableSize.y);
+        cr->rectangle (0, 0, get_width(), get_height());
         cr->fill ();
         //*/
```
But I don't understand:
the comment above states that "the new method doesn't work".
I played a little and what makes the "new method" work was to create a CSS rule with #Inspector as selector.
I have no idea why get_background_color returns a different color (black) than what render_background does (without css rulse, with css rule, both do the same).

What is the preferred way, @heckflosse @rfranke ?

Beep6581:
![screenshot_20201012_114720](https://user-images.githubusercontent.com/5844619/95733278-30445b00-0c82-11eb-92a9-f610e7dea99c.png)

The TIFF thumbnail embedded in the IIQ image has a height is 220px. The area of wrong color is 220px high, so it's size is probably derived from the thumbnail (when it should be derived from... the parent canvas size?). Why is that rectangle drawn at all? Probably to show a black background. Though I wonder whether that's really necessary: since the screen-sized canvas (gray on my screenshot) exists, typically one should be able to set its color (to black) when creating it, instead of adding a second step which draws a black rectangle.

heckflosse:
Here's an alternative solution which simply does not upsize small embedded jpegs
```diff
[...]

PR #6222, opened 2021-05-01
Inspector window fixes

Lawrence37:
Several fixes for inspector window behavior:
- Fully enabled inspector from the film strip
- Create window title from a string in the language files
- Ensure the window closes when quickly pressing and releasing the accelerator key
- Mark the window as pinned when it is opened from the context menu
- Removed the explicitly drawn background (background color can be set in the theme CSS using the `window.background#InspectorWindow` selector)
- Zooming is properly centered at the mouse pointer
- Panning speed is decoupled from the current magnification

Partially (or hopefully fully) addresses #5867, #5946, and #6111.


TechXavAL:
As I commented in #5867:

> open RT and don't select any image: upon pressing «f» the first time, a small window with its own titlebar and the image inspector is shown. When you release «f», that window closes. Press «f» again and a full screen inspector is shown

In your current `inspector-window-fixes` this is what happens: as you don't have any image selected, the inspector shows the image right below the mouse pointer in a «fit-to-screen» zoom. If the embedded jpeg (which is shown in the inspector) is smaller than the display resolution, then the jpeg gets expanded to fit the display resolution. If the embedded jpeg is larger than the display resolution, then it gets shrank to fit the screen. I'm not sure if expanding the jpeg is desired, as it gets blurred and gives the idea of an unfocused or shaken image.

And now something unexpected: if you inadvertently have the Caps Lock key active, when you press «F» the image is shown like in the previous paragraph, but after about half a second later (you still have «F» pressed so the inspector doesn't close), the image changes automatically to 100% zoom. If the embedded jpeg is small, then you automatically see the real size preview, and if the embedded jpeg is large, you automatically see just the portion of the image right below the arrowhead. Perhaps this is not desired as may lead to confusion (remember that the Caps Lock key has been pressed inadvertently).

> open RT and select an image (open it in the Editor if you want)

Even though you have an image selected (and may have worked with it in the Editor), the behavior is the same as before: the image shown in the inspector is the one below the mouse pointer. Nothing wrong with this if it's correctly documented.

I just have to add that the responsiveness of the inspector is a bit slower now, as it closes much slower than before: it takes about half a second closing the image after releasing «f». Nothing wrong with this if it's correctly documented.

Thanks for working on this

Lawrence37:
@TechXavAL

I fixed the conflicting behavior with the caps lock.

> Even though you have an image selected (and may have worked with it in the Editor), the behavior is the same as before: the image shown in the inspector is the one below the mouse pointer

I'm sure this is the intended behavior because it is the same as with the inspector tab.

I did not notice any performance difference and I'm not sure why the window would be closing slower. None of my changes should affect the closing speed.

> I'm not sure if expanding the jpeg is desired

Ingo made a similar [comment](https://github.com/Beep6581/RawTherapee/issues/5946#issuecomment-707078992). If no one has an objection, I can make the "fit to window" mode never upscale the image when fully zoomed out.

TechXavAL:
I will perform a few more checks this week just to check everything works as expected (specially with the slower performance, as it may happen only in my computer).

Thanks

Thanatomanic:
Thanks @Lawrence37 for picking up these related issues! I'll give it a testrun today.

Thanatomanic:
@Lawrence37 Things work very well in general, including the backgrounds, scaling, panning, zooming and hovering. 👍 

Some observations:
* I did manage to segfault once, while scroll-zooming very fast, but haven't been able to repeat.
* The second part of [this post](https://github.com/Beep6581/RawTherapee/issues/6111#issuecomment-778662771) has not been solved. When you have not opened the Inspector from the File browser, but go directly to the Editor and use the context menu to open the Inspector it does not show up full-screen, but in a separate square window (screenshots shown in #6111). However, if you go back to the File browser and press 'f', you get the full-screen window. If you then go back to the Editor the Inspector will be full-screen, except not entirely (some part of the desktop is still visible).
* The behavior of loading the inspector window can change. For example, if you start in the File Browser and use 'f', the window opens and closes 'cleanly': there is a delay to load the image and the window only shows after the file has loaded. However, if you use the Inspector from the Editor once and go back to the File Browser and press 'f' the window loads immediately while showing the previous image before showing the newly loaded image.
* Open the Inspector (from File Browser), hold 'f' and scroll to zoom (in or out). If you release 'f' the Inspector disappears. However, if you pan, the window stays visible. I'm not sure what the best behavior should be, but it's inconsistent now.
* For me there is no delay for closing the Inspector window, it's quite snappy.

Finally, my last point in #5867 was that when you have multi-monitor setup the Inspector always loads on the first screen. It would be very nice if the inspector always shows up on the screen where RT's main window _isn't_ located. But obviously, this is a nice-to-have, not a need-to-have.

Edit: one more point, is that in the Editor the hotkey 'f' toggles the focus-mask. The context-menu says 'F', but neither 'f' nor 'F' actually open the Inspector. Not even when the filmstrip is the active element.

Lawrence37:
I made a few more changes:
- Finally found and fixed the issue causing the window to not show up full-screen the first time. GTK can be mysterious sometimes :expressionless:.
- Launching the inspector for the first time will show the inspector full-screen regardless of how it is opened.
- Restore the window using the size/state from the previous close.
- Fit-to-window instead of 1:1 view when opening from the context menu (consistent with pressing the `f` key).
- Only show "F" in the context menu from the main file browser but not the film strip.

@Thanatomanic 

I ran into some segfaults before, but they should have been fixed before I pushed any changes. I was unable to trigger a crash anymore, even with rigorous testing.

I changed the behavior of the window size. The size is restored from when the inspector window was last closed. I hope this solves any sizing problems.

At one time, I was able to reproduce the issue of the previous image showing briefly before the new one loads, but this no longer happens (using the same commit you tested). Nothing in the code seems fishy, so I'm afraid I won't be able to fix this problem at this time.

I agree it is inconsistent that zooming does not pin the window while panning it does. In my opinion, any interaction with the window should pin it. If there are no alternative opinions, I will make this change.

Beep6581:
Just wanted to say that I'm happy to see improvements to the Inspector! I hope to be able to test this within the next 2 days.

Thanatomanic:
@Lawrence37 I really like how things are working right now! 

It seems you changed where the inspector window opens, because on my dual screen setup it now automatically loads on my secondary monitor. This is great, except for those circumstances where the main window is located on the second screen and I would like the inspector to be drawn on the primary screen. Is that something easy to fix? Otherwise, I would be :+1: for a merge.

Lawrence37:
Currently, there is no code telling the inspector to open on which screen. It is determined by the window manager. I think it would be best to save and load the inspector location just like the main window, maybe a task to be done after the 5.9 release.

ff2000:
It was raised in some occasions that the non-permanent inspector window isn't really user-friendly.
Do you keep that behaviour because you (and all others) think it's great that the inspector closes as soon as you leave the "f"-key, or is it because you want to keep the general functionality as-is?

Also if you already take care of the issues you might want to try to refactor the code. Currently the inspector panel manages its own parent and there is a lot of code that deals with the inspector being in the panel or window. From an OOP POV it would be better to split out the inspector view and use an instance of it in seperate classes for the panel and window.

Lawrence37:
@ff2000 I opened this PR mainly to fix some of the known issues with the inspector. I don't want to change too much considering how close we are to releasing 5.9. There are certainly many usability improvements possible (multiple screen support, better key bindings, changing image directly in the inspector, ratings and color labels, dynamic switching between window & tab, etc.). They can be added in a future release without impacting users much. The code refactoring can also take place then. But if someone can provide enhancements before the release, that is fine too.

Thanatomanic:
I agree with @Lawrence37, we should merge the fixes and improve later.

> Currently, there is no code telling the inspector to open on which screen. It is determined by the window manager.

Curious. Earlier the inspector always really opened on my main screen, now it consistently opens on my second one. I'll see if I can maybe bisect it.

Thanatomanic:
@Lawrence37 36cb32b31be5702574bcc32e97a82296e465a2be is the commit for which the inspector window switches screens

Don't get me wrong, it's better this way, but it changed nonetheless.

Issue #4462, opened 2018-03-23
Replace dcraw with actively maintained raw decoding engine (e.g. LibRaw)


Beep6581:
While in the past I would occasionally communicate with Dave Coffin, author of dcraw, I have not received any replies or seen any signs of progress on dcraw for over a year. It seems the project is abandoned, and we need a plan to move forward.

Options:
* RawSpeed https://github.com/darktable-org/rawspeed
* [LibRaw](https://www.libraw.org/) https://github.com/LibRaw/LibRaw
* [libopenraw](https://libopenraw.freedesktop.org/wiki/) https://github.com/hfiguiere/libopenraw

agriggio:
Good idea. I think maybe we can adopt a two-tier approach in which we use the current code for the supported cameras, and fall back to something else for the unsupported ones. We would probably need a way to still be able to use our `camconst.json` though. 

Can someone describe the main differences between rawspeed and libraw? (I know nothing about libopenraw, though from the readme it seems that it lacks support for many cameras...)

LebedevRI:
> differences between rawspeed and libraw

> rawspeed

* no releases, copylib.
* c++14 (clang-3.5+ / gcc-5+ / clang-cl; **NO** MSVC support!)
* is not a copy-paste from dcraw.
* actually readable, maintainable codebase.
* strong integration testing (see [Raw Samples Wanted](https://discuss.pixls.us/t/raw-samples-wanted/5420?u=lebedevri)), some unit tests.
* some kind of public ci setup

> libraw

* releases
* uses [old version of] rawspeed internally, unless the camera is blacklisted, then
* mostly copy-paste class-ified version of dcraw is used
* ??? fixme

There are a few more points, but i don't feel like posting them here, feel free to ping me in IRC.

I'm biased, i'm sure libraw is not all that bad :)

agriggio:
Thanks @LebedevRI, I'll follow up on IRC when time permits.

aferrero2707:
Some time ago I played with RT and RawSpeed, and came to the point where I could open and decode a RAW file with RawSpeed in RT. If you are interested, I can push my code in a dedicated branch in the next days...

heckflosse:
@aferrero2707 great :+1: But keep in mind that we will need to support both for a while. Our dcraw code has some specialities, which are not covered by rawspeed code. This has to be solved before we can cut off dcraw.

wbzqe:
hi,
[...]

PR #6225, opened 2021-05-03
Remove lj92 library

SimonSegerblomRex:
Both `dcraw` and `lj92` contain lossless JPEG decoders. After the recent fixes in #6190 and #6214 they both work fine in most cases, but I don't see any reason to keep `lj92` now. `dcraw` has been the default when `Make` doesn't contain "Canon"/"BlackMagic" and it also supports DNG version 1.0 (with non-spec lossless JPEG encoded 16 bit diff values).

I noticed the discussion in #4462 about replacing `dcraw` with `rawspeed`. Keep in mind that `rawspeed` currently (darktable-org/rawspeed#189, darktable-org/rawspeed#258) only supports predictor mode 1.

For future reference:
Take a look at [this version](https://github.com/ilia3101/MLV-App/blob/11a93832f6543402452104375d03df138cb9ceb9/src/mlv/liblj92/lj92.c) of `lj92` if you need a lossless JPEG **encoder** (that works for 16 bit diff values). ([Here's](https://github.com/thorfdbg/libjpeg) another encoder.)


SimonSegerblomRex (commented):
I don't see any big performance differences between the two implementations. Images seem to load with about the same speed (but I haven't done any real benchmarking). All images at [raw.pixls.us](https://raw.pixls.us/#repo) (searching for "blackmagic" or "mlv") get decoed properly with the `dcraw` implementation (as well as the `lj92` one).

Thanatomanic:
I see no big problems here, if nobody complains I'll merge this on Friday.

Thanatomanic:
@SimonSegerblomRex You can still add your name to the `AUTHORS.txt` file if you want

SimonSegerblomRex:
> @SimonSegerblomRex You can still add your name to the `AUTHORS.txt` file if you want

Thanks @Thanatomanic! I hope I'll be able to make some more substantial contribution to the project in the future.

PR #6227, opened 2021-05-04
Spot removal tool

Lawrence37:


rtdata/languages/Francais
```diff
@@ -663,7 +663,7 @@ HISTORY_MSG_421;Retinex - Gamma
 HISTORY_MSG_422;Retinex - Gamma
 HISTORY_MSG_423;Retinex - Pente Gamma
 HISTORY_MSG_424;Retinex - Seuille HL
-HISTORY_MSG_425;Retinex - Base Log
+HISTORY_MSG_425;Retrait de taches
```
Thanatomanic: There is a problem here. The history message should not/cannot be reused afaik. Interestingly, there is no similar change to the `default` language file.

rtengine/spot.cc
```diff
@@ -0,0 +1,513 @@
...
+
+    for (auto entry : params->spot.entries) {
+        std::shared_ptr<SpotBox> srcSpotBox(new SpotBox(entry,  SpotBox::Type::SOURCE));
+        std::shared_ptr<SpotBox> dstSpotBox(new SpotBox(entry,  SpotBox::Type::TARGET));
+        if (   !srcSpotBox->setIntersectionWith(fullImageBox)
+            || !dstSpotBox->setIntersectionWith(fullImageBox)
+            || !srcSpotBox->imageIntersects(*dstSpotBox, true))
+        {
+            continue;
+            ++i;
```
Thanatomanic: Seems fishy - LGTM warns about this, but I have no idea about the logic of this code yet.

Thanatomanic:
I made some minor fixes to the code and indicated two potential problems. Both are probably easy to fix. @Lawrence37 do you have the time to take a look?

For the rest, the tool works like a charm while testing, so I would be :+1: for a merge.

rtdata/languages/Francais
Lawrence37: This is an artifact of merging history messages from dev. Going forward, we should stop using numbered history messages.

rtengine/spot.cc
Lawrence37: It renders correctly, but if I swap the two statements, it can cause a crash. I will remove the dead code.

rtengine/improccoordinator.cc
```diff
@@ -666,12 +714,11 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
             }
         }
 
-        oprevi = orig_prev;
-
```
Lawrence37, line 670: This doesn't look right. It might have been misplaced after reordering the processes in 7d310e688c9837d31c262678df3c71a278ce11f0, but I'm not sure. Could be related to a bug I found where enabling dehaze makes spot removal ineffective.

heckflosse:
I tested. Works fine. :+1: for merge

Lawrence37:
@Hombre57 added some documentation for using spot removal. I quote it here.
https://github.com/Beep6581/RawTherapee/blob/38075be8cc5682e73bb59ddd6f784c125a401f99/rtgui/spot.h#L29-L54

I will merge once the LGTM analysis is complete.

PR #6239, opened 2021-05-16
Improves RT window position management, fixes #6233

### LINKED ISSUE #6233: Windows 10: Raw therapee window located out of screen
System: windows 10

RawTherapee versions: all recent versions for all builds, stable, official nightly, my own nightly builds

reference: https://discuss.pixls.us/t/cant-start-the-latest-dev-version-of-rawtherapee-for-win64/24746/27

Description : from time to time, it can happen that in the option file/ GUI section, the two variables WindowX and WindowY are overwritten with garbage (-32768).
The RawTherapee window is then located out of the screen and not displayed.
Users are very puzzled.
Workaround: stop RT, reset the two variables to 0? RESTART rt

How to reproduce: although I tried, I never succeeded to find some configuration permitting to trigger  the bug.

Pandagrapher:
Hello All,

I have worked on some improvements/fixes (issue #6233) in RT window position management code. Here the following updates:
- To avoid some bad window position computation in some cases, position isn't updated when RT is maximized (fixed mode detection), minimized (new mode detection) or fullscreen
- I have added some code robustness to avoid window position to be detected outside window screen (in case of...)

I have made some tests with Windows 10 and MacOS in single screen configuration. This work is still draft because I need to take into account RT mode authorizing several window instances.

Pierre


Pandagrapher:
Hello All,

I just reworked how edit window position is managed in separate edit window mode. So this pull request is ready for review before merging.

Pierre

gaaned92:
I tested the PR. One Window.

It's not a systematic test, but I can no longer reproduce the bug.

just one remark : if option file contains out of window position, when opening RT with PR, the first time, window is out of screen. When I close RT and reopen it's ok. So not a concern knowing from now it will not happen.
Thank you @Pandagrapher

Pandagrapher:
@gaaned92 Thanks for your tests :)

> just one remark : if option file contains out of window position, when opening RT with PR, the first time, window is out of screen. When I close RT and reopen it's ok. So not a concern knowing from now it will not happen.

Exactly, now, when getting RT window position, position is now saturated between 0 and monitor size. So corrupted window position saved in option file won't happen anymore. However if window position is corrupted in option file (i.e. windows position saved with negative values only) before applying this PR, RT window will be located out of screen at next start. Algorithm only tests if position (for positive values) isn't out of screen because of monitor size change between two RT sessions.

Pierre

Lawrence37:
I tested on Linux with a single monitor. It mostly works fine. There is one problem affecting the size of the tool panel in multiple editors in own window mode. Open an image in the editor. Maximize the editor window, then close RawTherapee. Re-open RawTherapee and open an image in the editor. The tool panel is much wider than it is supposed to be.

Tested by comparing the latest dev vs. the latest dev with this branch merged.

Pandagrapher:
> I tested on Linux with a single monitor. It mostly works fine. There is one problem affecting the size of the tool panel in multiple editors in own window mode. Open an image in the editor. Maximize the editor window, then close RawTherapee. Re-open RawTherapee and open an image in the editor. The tool panel is much wider than it is supposed to be.
> 
> Tested by comparing the latest dev vs. the latest dev with this branch merged.

@Lawrence37 
Thank you for your tests and returns :)

Due to different Windows Manager behavior, I cannot reproduce the bug in MacOS... However I just pushed a commit which, I think, will fix the issue. Can you test please?

Best,
Pierre

Lawrence37:
The fix works.

Thanatomanic:
I have a Windows 10 dual monitor, extended desktop setup with my primary screen in front and a secondary screen to the left. Windows can scale the GUI for high dpi screens. My primary screen is set to 100%, my secondary screen is set to 150%. When testing this PR I ran into some trouble with scaling when it shouldn't, but I have yet to find the cause. I'll report back.

Pandagrapher:
@Thanatomanic Do you have these scaling issues with the `dev` branch or it has been introduced with this PR? The problem with HiDPI management is that it's a RT custom one (and not Gtk3 native one). Scaling is determined at startup (I guess only for the main screen) and is not updated if you have two screen with different scaling, update the scaling with RT opened, move a window to another screen, etc... Once RT 5.9 is release, I have planned to work on it to use the native Gtk mechanism and improve that but that's a different issue...

Thanatomanic:
@Pandagrapher I don't have the same issues in the `dev` build. Things are correct when the screen is maximized, so I'll report what I find when the screen is not maximized. I can also confirm that the scaling itself is not the root of the issue. Most of why I report below also happens when I manually change the scaling to 100% on both screens.

* Situation in `dev`: I can put RT's main window anywhere on any screen. Close it, reopen it. The window always reappears in the same place.
* Situation in PR: if the window initially was on my right screen (100% scaled), then after reopening the window _never_ appears in the same place, but shifted to the right. It is sometimes wrongly scaled (this is the part I find hard to reproduce). If the window initially was on my left screen, then the window _always_ appears in the same place.

Before closing (1):
image

After reopening (1):
image
Even though the entire window is inside the right screen, the GUI is scaled up.

Before closing (2):
image

After reopening (2):
image

After some more attempts I actually seem to be running into the situation where _wherever_ I put the window on the right screen, it _always_ ends up somewhere half-way between both screens.

Before closing (3):
image

After reopening (3):
image

Immediately moving this screen again (4):
image

After reopening (4):
image


In any case, something weird is going on when storing and recalling the window positions.

Pandagrapher:
@Thanatomanic Thanks for your detailed report :)
I think I have found the issue origin. Can you please test this patch?
[dual_screen_patch.txt](https://github.com/Beep6581/RawTherapee/files/6682098/dual_screen_patch.txt)

If it solves the issue, I will commit the patch.

Best,
Pierre

Thanatomanic:
@Pandagrapher This solves it completely for me :+1:

(it spams width and height in console, but you're probably aware)

Pandagrapher:
@Thanatomanic Thanks for the tests: just pushed the commit (without debug prints ;)) and I think we now are ready to merge this PR :)
Just to be sure: to merge it's better to use the "squash & merge" option for a PR like this one with a lot of commits?

Pierre

Thanatomanic:
Great, I'm 👍  for a squash and merge. That doesn't pollute the main commit log as much. The PR should remain traceable.

Issue #6244, opened 2021-05-19
On-Preview Editing Widgets: Crashes and Other Issues


Lawrence37:
@AndyAstbury Let's continue the discussion from #6242 here. I managed to trigger a crash while dragging a spot and pressing the enter/return key. It also happens with the graduated filter and perspective control lines. Can you confirm if your bluetooth mice have issues with these tools as well?

**Bug Description**
1. On some hardware, releasing the mouse button while dragging a spot removal spot has no effect. The spot continues to follow the mouse. Refer to this [comment](https://github.com/Beep6581/RawTherapee/issues/6242#issuecomment-842465893) for the original issue.
2. While dragging one of the on-preview widgets, pressing the enter/return key causes a segfault here (`editSubscriber` is null) https://github.com/Beep6581/RawTherapee/blob/d11d834045ef9f6e8cbbf63f734c07be3654123e/rtgui/cropwindow.cc#L708-L714

heckflosse:
> While dragging one of the on-preview widgets, pressing the enter/return key causes a segfault here (editSubscriber is null)

This should fix the crash:

```diff
diff --git a/rtgui/cropwindow.cc b/rtgui/cropwindow.cc
index a3e00c80a..4a136266b 100644
--- a/rtgui/cropwindow.cc
+++ b/rtgui/cropwindow.cc
@@ -705,15 +705,15 @@ void CropWindow::buttonRelease (int button, int num, int bstate, int x, int y)
         state = SNormal;
         needRedraw = true;
     } else if (state == SEditDrag1 || state == SEditDrag2 || state == SEditDrag3) {
-        if        (state == SEditDrag1) {
-            needRedraw = editSubscriber->button1Released();
-        } else if (state == SEditDrag2) {
-            needRedraw = editSubscriber->button2Released();
-        } else if (state == SEditDrag3) {
-            needRedraw = editSubscriber->button3Released();
-        }
-
         if (editSubscriber) {
+            if        (state == SEditDrag1) {
+                needRedraw = editSubscriber->button1Released();
+            } else if (state == SEditDrag2) {
+                needRedraw = editSubscriber->button2Released();
+            } else if (state == SEditDrag3) {
+                needRedraw = editSubscriber->button3Released();
+            }
+
             rtengine::Crop* crop = static_cast<rtengine::Crop*>(cropHandler.getCrop());
             Coord imgPos;
             action_x = x;

```

AndyAstbury:
[...]

PR #6253, opened 2021-05-23
On-preview widget fixes

Lawrence37:
Fixes some problems with displaying on-preview editing widgets after the enter/return key is pressed while adjusting or hovering over the widgets (#6244). Still don't now how to fix the problem with dragging a spot removal spot.

### DIFF
--- a/rtgui/controllines.cc
+++ b/rtgui/controllines.cc
@@ -199,6 +199,10 @@ bool ControlLineManager::pick3(bool picked)
 
 bool ControlLineManager::drag1(int modifierKey)
 {
+    if (action != Action::DRAGGING) {
+        return false;
+    }
+
     EditDataProvider* provider = getEditProvider();
 
     if (!provider || selected_object < 1) {
@@ -261,6 +265,20 @@ bool ControlLineManager::drag1(int modifierKey)
     }
 
     return false;
+}
+
+void ControlLineManager::releaseEdit(void)
+{
+    action = Action::NONE;
+
+    if (selected_object > 0) {
+        mouseOverGeometry[selected_object]->state = Geometry::NORMAL;
+    }
+
+    edited = true;
+    callbacks->lineChanged();
+    drawing_line = false;
+    selected_object = -1;
 }
 
 bool ControlLineManager::getEdited(void) const
--- a/rtgui/controllines.cc
+++ b/rtgui/controllines.cc
@@ -16,6 +16,7 @@
  *  You should have received a copy of the GNU General Public License
  *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
  */
+#include <assert.h>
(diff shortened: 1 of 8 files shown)


rtgui/controllines.cc
```diff
@@ -260,7 +267,24 @@ bool ControlLineManager::drag1(int modifierKey)
         autoSetLineType(selected_object);
     }
 
-    return false;
+    return true;
+}
+
+void ControlLineManager::releaseEdit(void)
```
Floessie, line 320: @Lawrence37 If you're touching code, you're free to remove artifacts like this prehistoric `void` parameter.

rtgui/gradient.cc
```diff
@@ -579,6 +580,19 @@ bool Gradient::drag1(int modifierKey)
     return false;
 }
 
+void Gradient::releaseEdit()
+{
+    if (lastObject >= 0) {
```
Floessie, line 593: It's hard to tell from the diff alone, if those `object` comparisons (`>= 0`, `> 1`, `> 0`) are correct, if "invalid" has different representations. Maybe that could be better modelled with a `std::optional<>` (which is C++17, but I have a simple C++11 implementation to share).

rtgui/gradient.cc
```diff
@@ -579,6 +580,19 @@ bool Gradient::drag1(int modifierKey)
...
+        if (lastObject == 2 || lastObject == 3) {
```
Floessie: Magic numbers. Can't their meaning be expressed through an `enum` value?

rtgui/spot.cc
```diff
@@ -185,6 +186,25 @@ void Spot::resetPressed()
...
+    Geometry *loGeom = getVisibleGeometryFromMO (lastObject);
+
+    if (!loGeom) {
+        EditSubscriber::action = EditSubscriber::Action::NONE;
+        return;
+    }
+
+    loGeom->state = Geometry::NORMAL;
+    sourceIcon.state = Geometry::NORMAL;
+    EditSubscriber::action = EditSubscriber::Action::NONE;
```
Floessie: Can be moved up. It's the same in the `!loGeom` case.

rtgui/gradient.cc
Lawrence37: `lastObject` gets its value from `EditDataProvider::getObject()` and I don't feel like changing that class or working around it in this PR.

rtgui/gradient.cc
Lawrence37: Replaced with enum values.

rtgui/spot.cc
Lawrence37: Done.

Floessie:
@Lawrence37 Much better now! Thanks! :+1:

PR #6262, opened 2021-05-26
Update sleef functions

Thanatomanic:
Somewhere in the past we have adopted functions from [SLEEF](https://sleef.org), see https://github.com/shibatch/sleef. However, I noticed recently that we use a very old version, probably 2.90 dated April 2016. In the meanwhile several SLEEF functions have had improvements (probably to do with either speed or accuracy), but those were never ported to RT.
The latest release in the 2.# branch is 2.121 (dated Jan 30, 2017), and then a new major branch with breaking changes was started and continued. The latest release in the 3.# branch is 3.5.1 (dated Sep 15, 2020).

Unless somebody (hopefully @heckflosse or @Floessie) tells me it is unnecessary, I would like to continue to update our SLEEF functions to their current optimal forms. First I will bring them up to date with the 2.121 version and potentially continue making changes to adopt the 3.5.1 version.

How does that sound?

So far I have made a round of updates for `sleef.h` only which you can give a testrun. It compiles and links, so I guess I haven't already broken something. I haven't touched the SSE2 functions.

### DIFF
--- a/rtengine/sleef.h
+++ b/rtengine/sleef.h
@@ -18,15 +18,25 @@
 #include "rt_math.h"
 #include "opthelper.h"
 
-#define PI4_A .7853981554508209228515625
-#define PI4_B .794662735614792836713604629039764404296875e-8
-#define PI4_C .306161699786838294306516483068750264552437361480769e-16
-#define M_4_PI 1.273239544735162542821171882678754627704620361328125
+//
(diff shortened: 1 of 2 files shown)


Thanatomanic:
To clarify: if you look at the changed files you can see I removed some functions. These were unused in the RT source code. I kept double precision functions even if they were unused because I think they ought to be the functions you preferentially choose when developing new code.

PR #6268, opened 2021-05-31
Perspective tool: Improve control lines user experience

### LINKED ISSUE #6250: Clarify usage of perspective tool through tooltips / disabling
i just played around with the today nighly build Version: 5.8-2934-gc32623882 and tried to understand the perspective correction tool.
There are two aspects i stumbled: (method camera based)
-  drawing only one control-line isnt enough (of course), may be this could be mentioned in GUI or Tooltip. With this, the button "apply" should be activated earliest after two existing control-lines.
- the tooltip for control-lines can be found in default (TP_PERSPECTIVE_CONTROL_LINES_TOOLTIP;<b>Ctrl</b>+<b>drag</b>: Draw new line\n<b>Right-click</b>: Delete line) but in GUI only "Edit" can be read.

Lawrence37:
Makes the "apply" and "delete all" sensitive or insensitive based on the number of control lines and adds extra tool-tip text to explain why correction cannot be applied.

Closes #6250.


rtgui/controllines.cc
```diff
@@ -84,6 +85,16 @@ size_t ControlLineManager::size(void) const
     return control_lines.size();
 }
 
+size_t ControlLineManager::getHorizontalCount(void) const
```
Floessie: Please turn that into `std::size_t ControlLineManager::getHorizontalCount() const`. See [std::size_t](https://en.cppreference.com/w/cpp/types/size_t) for appropriate includes (often, they are already in place).

rtgui/controllines.cc
```diff
@@ -84,6 +85,16 @@ size_t ControlLineManager::size(void) const
...
+{
+    return horizontalCount;
+}
+
+size_t ControlLineManager::getVerticalCount(void) const
```
Floessie: Dito.

rtgui/controllines.h
```diff
@@ -52,6 +52,7 @@ class ControlLineManager: EditSubscriber
     bool draw_mode;
     bool drawing_line;
     bool edited;
+    size_t horizontalCount, verticalCount;
```
Floessie: Dito.

rtgui/perspective.cc
```diff
@@ -739,11 +741,29 @@ void PerspCorrection::setEditProvider(EditDataProvider* provider)
...
 
+void PerspCorrection::updateApplyDeleteButtons(void)
+{
+    if (batchMode) {
+        return;
+    }
+
+    bool edit_mode = lines_button_edit->get_active();
+    bool enough_lines = lines->getHorizontalCount() >= 2 || lines->getVerticalCount() >= 2;
+    const auto &tooltip = M("GENERAL_APPLY")
```
Floessie: Why `const &` on temporary? It's valid, but makes no sense here. If `set_tooltip_text()` takes the string by value, make it `auto` and `std::move()` it in the call, else make it `const auto`.

rtgui/perspective.cc
```diff
@@ -739,11 +741,29 @@ void PerspCorrection::setEditProvider(EditDataProvider* provider)
...
 void PerspCorrection::lineChanged(void)
 {
+    updateApplyDeleteButtons();
+
     if (listener) {
         listener->panelChanged(EvPerspControlLines, M("HISTORY_CHANGED"));
     }
 }
 
+void PerspCorrection::updateApplyDeleteButtons(void)
```
Floessie: Please no `void` parameter. This is so K&R-C-ish... :wink:

rtgui/perspective.h
```diff
@@ -96,6 +96,7 @@ class PerspCorrection final :
     void tweakParams(rtengine::procparams::ProcParams &pparams) override;
     void setCamBasedEventsActive (bool active = true);
     void setFocalLengthValue (const rtengine::procparams::ProcParams* pparams, const rtengine::FramesMetaData* metadata);
+    void updateApplyDeleteButtons(void);
```
Floessie: Remove the `void` parameter.

Floessie (changes requested):
Some minor cleanups requested.

Lawrence37:
@Floessie I made changes based on your comments.

Floessie (approved):
@Lawrence37 Very nice! :+1: You even caught some more `size_t`s than I found. And the additional constants are helping as well.

Thanatomanic:
@Lawrence37 :+1: for merge!

Issue #6281, opened 2021-06-18
Dynamic processing rules based on path and filename


nicolas-t:
I would like to apply a dynamic processing profile to an image based on where the image stored.

One could add a `filename` and/or `path` option to this screen (similar to camera/lens) :
![image](https://user-images.githubusercontent.com/1765930/122539221-a5dda380-d027-11eb-996f-b5a5e8d43856.png)

We could use the `re:` prefix to use a regex.

**My usecase :**
I use my Canon 5D MKIII to scan my analogue photos.

- If the raw file is in the folder `Canon 5D MKIII` (DSLR) then it's a standard photo I want a neutral profile
- If the raw file is in the folder `Leica M6` (Analogue) then i's a digitized negative I want a negative black and white profile.

If it's ok for you I can start to work on a PR, but I definitly don't have enough knowledge to code the whole feature, so you'll have to help me at some point :)

Related topic : 
https://discuss.pixls.us/t/dynamic-processing-rules-based-on-path-and-filename/25493

Thanks

PR #6284, opened 2021-06-18
Path filter in dynamic profile panel

nicolas-t:
Hello,

I started working on adding path filter to the dynamic profile panel.
I don't think I have enough knowledge to do more, so if you like this feature please feel free to finish the job.

Related issue: https://github.com/Beep6581/RawTherapee/issues/6281


Thanatomanic:
Thanks @nicolas-t for working on your own idea! Let us know when there's something to be tested or you need advice.

nicolas-t:
Hello @Thanatomanic  :)

I'm having trouble accessing the filename in the `matches` function here : 
https://github.com/Beep6581/RawTherapee/blob/1e797f11e288d0f3dd6bd4e1e4464d9f869ab430/rtengine/dynamicprofile.cc#L89

--- 

In this function I think I have `inputFile` which is the path i'm looking for: 
https://github.com/Beep6581/RawTherapee/blob/b35e0d82694c3d5a80c5decdcd2ddb096f505a07/rtgui/main-cli.cc#L744

But, should I pass it as the second argument of `loadDynamicProfile`, and then pass it again as the second argument of [`rule.matches`](https://github.com/Beep6581/RawTherapee/blob/b35e0d82694c3d5a80c5decdcd2ddb096f505a07/rtengine/profilestore.cc#L524) ?

Seems better to me to store the filename in the `FramesMetaData`/ `FramesData`, but I'm not sure how.


Also, I never did any c++, how can I dump/echo the value of a variable ? 
 
Thanks for you help.

Lawrence37:
Hi.
I don't think the file name goes in the metadata, so passing it as an argument is the better option. You can inspect variable values with `std::cout`/`printf()` or by using a debugger such as GDB on the debug build.

nicolas-t:
Hello @Lawrence37 !
I just pushed 2 commits, I'm passing the filename as the second argument.
App is working normally, and my dynamic profile is well applied depending on filename and path.

Aside from this I tried outputing the filename here : 
https://github.com/Beep6581/RawTherapee/pull/6284/files#diff-bc5e968d23075d5dae37c689a2080f5441c91eb1d3618ab661af1be30cad5c7dR740

But nothing is showing up the terminal.
I'm compiling like this : 
`~/programs/code-rawtherapee/tools/build-rawtherapee -b -d`

And running like this : 
`~/programs/rawtherapee-dev-debug/rawtherapee`
Out of curiosity: where is the output supposed to appear ? 

Anyway, the feature is working on my side, can you test and review my code ?

Thanks

Lawrence37:
The output appears in the terminal. You've placed the code in the CLI version of RT, so it will only appear if you run `rawtherapee-cli` instead of `rawtherapee`.

I briefly tested this feature and it works fine. I will add my code review comments shortly.

rtdata/languages/English (US)
```diff
@@ -62,6 +62,7 @@
 !EXIFFILTER_ISO;ISO
 !EXIFFILTER_LENS;Lens
 !EXIFFILTER_METADATAFILTER;Enable metadata filters
+!EXIFFILTER_PATH;Path
```
Lawrence37: This is usually added by running a script, but I think it's fine? Ping @Beep6581.

rtdata/languages/default
```diff
@@ -61,6 +61,7 @@ EXIFFILTER_IMAGETYPE;Image type
 EXIFFILTER_ISO;ISO
 EXIFFILTER_LENS;Lens
 EXIFFILTER_METADATAFILTER;Enable metadata filters
+EXIFFILTER_PATH;Path
 EXIFFILTER_SHUTTER;Shutter
```
Lawrence37, line 65: I'm concerned that "Path" is a little vague. Something like "File path" is more clear.

rtgui/main-cli.cc
```diff
@@ -737,11 +737,13 @@ int processLineParams ( int argc, char **argv )
         }
 
         if (useDefault) {
+            std::cout << "loading file: " << inputFile << std::endl;
```
Lawrence37: The file isn't being loaded here.

Lawrence37 (dismissed):

Lawrence37 (commented):
@nicolas-t This looks good.

Should this be included in 5.9? It's a small change but the beta version is almost here and the default language file is under review.

Thanatomanic:
@nicolas-t Thanks for the work and the patience!

PR #6302, opened 2021-07-10
Update spot name in history after rename

### LINKED ISSUE #6298: Local adjustments - renamed spot isnt shown in History
[Version]
AppVersion=5.8-2983-g3f58095b6
Version=349

when renaming a spot for local adjustments, this will be shown in the history as expected. If there is any further change on this spot, the history again will not show the new name, but (new spot) as before.

Edit: after reopen the pic in editor, the new name is shown correctly in history.

Lawrence37:
@Floessie @heckflosse
This saves the spot name in one location. Please let me know whether you prefer this approach or the patch in #6298.

### DIFF
--- a/rtgui/controlspotpanel.cc
+++ b/rtgui/controlspotpanel.cc
@@ -703,6 +703,9 @@ void ControlSpotPanel::on_button_rename()
             row[spots_.name] = newname;
             treeview_->columns_autosize();
             listener->panelChanged(EvLocallabSpotName, newname);
+            if (controlPanelListener) {
+                controlPanelListener->spotNameChanged(newname);
+            }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


Lawrence37:
@Floessie @heckflosse @Thanatomanic and others, any thoughts on this? Otherwise I will go with the patch because there is less room for bugs.

Thanatomanic:
To test this, do I need to see what happens when I create a spot, do stuff, rename the spot and then do stuff again?

Floessie:
@Lawrence37 If you've checked, that the new spot name store is accessed only from a single thread and that the destruction order is correct with no access under destruction, I'm fine with the code. :+1:

Lawrence37:
@Thanatomanic Yes

Lawrence37:
The locallab tools were not being destructed at all. They were being managed by GTK through `Gtk::manage`, but they are never attached to a parent so GTK does not know when to destroy them.

I will push a commit to fix this. The tools are owned by `Locallab` and never change, so I'm converting them to plain member variables.

PR #6305, opened 2021-07-15
Fix size of TIFFTAG_EXIFIFD as passed to TIFFSetField

### LINKED ISSUE #5141: Attempt to write value larger than 0xFFFFFFFF in 32 bit package (5.5)
I'm a packager for mageia linux. We recently attempted to backport version 5.5 to our stable branch. QA has found that 32 bit 5.5 fails to process files that worked with 5.3. 64 bit is ok. The error text is

$ rawtherapee-cli -o p.tif -t -c P7212390.ORF 
RawTherapee, version 5.5, command line.
Output is 16-bit integer.
Processing: P7212390.ORF
TIFFWriteDirectoryTagIfdIfd8Array: Attempt to write value larger than
0xFFFFFFFF in Classic TIFF file..
TIFFWriteDirectoryTagIfdIfd8Array: Attempt to write value larger than
0xFFFFFFFF in Classic TIFF file..
Error saving to: p.tif

Our bug is https://bugs.mageia.org/show_bug.cgi?id=24092

The only other thing that changed in our package aside from the rawtherapee version itself was that I attempted to use O3 as recommended if that matters. I do not have the photo they used but will try to get it if needed.

10110111:
Starting with libtiff 4.0.0, TIFFTAG_EXIFIFD has size of 64 bits, so it
must be set as a 64-bit value. Otherwise, since TIFFSetField is a
variadic function, there's no conversion, and the function reads garbage
into the higher dword.

This patch implements the fix and thus closes #5141.

### DIFF
--- a/rtengine/imageio.cc
+++ b/rtengine/imageio.cc
@@ -1399,18 +1399,19 @@ int ImageIO::saveTIFF (const Glib::ustring &fname, int bps, bool isFloat, bool u
 
             if (exif)   {
                 int exif_size = exif->calculateSize();
-                unsigned char *buffer = new unsigned char[exif_size + 8];
                 // TIFFOpen writes out the header and sets file pointer at position 8
+                const uint64_t file_offset = 8; // must be 64-bit, because TIFFTAG_EXIFIFD is
+                unsigned char *buffer = new unsigned char[exif_size + file_offset];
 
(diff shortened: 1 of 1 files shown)


Thanatomanic:
Thanks @10110111 this seems fine!

PR #6316, opened 2021-07-27
Fixes a minor issue when loading a profile with Resize in it

Bezierr:
This is a small change I propose to address the minor issue already mentioned in [https://github.com/Beep6581/RawTherapee/issues/5182#issuecomment-821986311](url) (last paragraph).

To reproduce the issue:
- Open a picture.
- Enable Crop and crop to some unusual aspect ratio.
- Enable Resize, select "Width" and set width to a reasonable value.
- Do Partial Profile Save with only selecting Resize.
- Open another picture.
- Load the partial profile you just saved.
- => Both width and height are set to the values from the loaded partial profile. Height is not recomputed to fit the dimensions of the new picture, as it would be if the same value for width were entered manually.

The proposed change (should be one line in resize.cc) seems to fix this.


Bezierr:
Sorry, still not familiar with the workings of Git. I don't want to merge 17 commits, just one...

PR #6339, opened 2021-08-18
make spot removal tool bigger

### LINKED ISSUE #6337: Make spot removal bigger
Hi,

I would like to suggest to make the spot removal tool bigger so it can be used on really large photos too, for removing larger "spots".
Now I have already done this. I created an own fork of RT with a bigger spot removal tool. It seems to be functional. 
So really I would like to create a pull request (that's the correct term, isn't it?).
What I have not yet figured out: I would also like to increase the feathering of the spot. 
So, if you want to check the big spot removal tool, you can compile it from here:
https://github.com/blitzgneisserin/RawTherapee
Really, I have only ported this from ART (I looked through the ART commits).
I thought I ask here before I create a pull request.
2021-08-16-151124_1920x1200_scrot
Hope someone likes this. I think in any case, it does not hurt to have a bigger spot removal tool, I mean it is always possible to make it smaller.

blitzgneisserin:
this change makes the spot removal tool bigger (changes the maximal radius of the spot) so it can be used for very large photos and to remove larger spots.

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -366,7 +366,7 @@ namespace procparams
 {
 
 const short SpotParams::minRadius = 5;
-const short SpotParams::maxRadius = 100;
+const short SpotParams::maxRadius = 400;
 
 
 ToneCurveParams::ToneCurveParams() :


Thanatomanic:
I guess there is very little to say against this. Let's merge.

blitzgneisserin:
thanks @Thanatomanic

Lawrence37:
Closes #6337.

PR #6345, opened 2021-08-23
New crop guide: square center, closes #6342

### LINKED ISSUE #6342: New crop guide: square center
Most of my pictures get posted on social media, and I optimise most of them to be able to fit well in a square format (instagram made me do it).

My current workflow is to:
1. lock crop ratio to 1:1
2. do the "ideal" crop for square format
3. select another crop format for the full image. Often 3:2 or 4:5

While this works, I would very much love to have a crop guide for this, so I can directly select the final crop ratio. It would also make it a lot easier to move the crop after the fact. Right now, I need to go through all steps every time I move my crop, which is somewhat time-consuming.

The ideal crop guide in my mind would be similar to the Rule of Thirds or Harmonic Means guides, except that it would only have two lines inside, defining where the center square is. Here's an approximation (suggested lines in red)

Capture

heckflosse:


Floessie:
@heckflosse Working on it...

Floessie:
@heckflosse Do you know why we compare "Harmonic means" with `strncmp()`? Was there ever something like "Harmonic means ..."?

heckflosse:
@Floessie I don't have a clue, why `strncmp` was used. :+1: for your changes. Ready to merge imho.

Floessie:
If the fixed "Harmonic means" string turns out to be a problem, we could treat the mapping in `ProcParams::load()` as regexes...

Floessie (approved):

Issue #6343, opened 2021-08-21
Empty input file message and a white jpg/tif when using CIECAM


ghost:
**Short description**

I gave this image on pixls a go: [DJI Air 2S Raw image processing](https://discuss.pixls.us/t/dji-air-2s-raw-image-processing/26520/4)

When telling RT to export this edit as JPEG it is busy for a bit, an **Empty input file** gets written in the terminal and a very small white jpg with the dimensions of the original is created. Also tried using TIF as export format; Same result.

After some experimenting I found out that if I turn off the *CIE Colour App...* module and export all is well.

~~Using the CIECAM module and the options seen in the below included sidecar are rather normal for me. Never had any issues. This is the first time, for as far as I can remember, that I edit a DNG in combination with CIECAM though.~~

_After some more testing: This is specific to this sidecar and NOT DNG specific. I can reproduce this on other RAW formats by loading this sidecar._

**Steps to reproduce**

Not sure if turning the CIECAM module on/off will do the trick but here is the sidecar that I created: [DJI_0033.with.ciecam.jpg.out.pp3.zip](https://github.com/Beep6581/RawTherapee/files/7025864/DJI_0033.with.ciecam.jpg.out.pp3.zip)

The RAW file: [DJI_0033.DNG.zip](https://github.com/Beep6581/RawTherapee/files/7025882/DJI_0033.DNG.zip)

**Expected behavior**

Being able to export after editing.

**Additional information**

* RawTherapee version : 5.8 Latest development version (also reproduced on 5.8-2995)

**Platform**

* OS                : Linux (kernel 5.9.15-1 x86_64 amd64)
* Distro            : Debian Buster (rel. 10.10)
* Graphics driver   : nvidia (460.39)

**Build info**

* Compiler      : gcc 8.3.0
* Cmake         : 3.13.4
* Glib          : 2.58.3-2
* Gtk+          : 3.24.5-1
* Exiv2         : 1.0.0.9
* Lensfun       : 0.3.2-4

**Build using:**

cmake \
    -DBUNDLE_BASE_INSTALL_DIR="$HOME/.local/${TRGT}" \
    -DBUILD_BUNDLE="ON" \
    -DCACHE_NAME_SUFFIX="5-dev" \
    -DCMAKE_BUILD_TYPE="release"  \
    -DPROC_TARGET_NUMBER="2" \
    -DOPTION_OMP="ON" \
    -DBUILD_SHARED="OFF" \
[...]

PR #6347, opened 2021-08-27
White jpg/tif when using Ciecam - Enable checkbutton preferences - ciecam artifacts - issue 6343

Desmis:
This PR solve (I hope) the problem in #6343.
* I add a checkbutton in Preferences "Avoid artifacts"
* by default CAM16 is used instead of CAM02 (CAM16 brings less artifacts)

When you enable "avoid artefacts" , the old algorithm is used (an can create white jpg...), and some processes are ignored to avoid artifacts in some cases (defringe, wavelets levels, etc.)

If nobody objects I will merge this afternoon.

Jacques


Thanatomanic:
Hi Jacques, thanks for the solution. I would like to test but can only do so tomorrow. Please delay merging for a bit.

Desmis:
@Thanatomanic 

No problem :)

jacques

ghost:
Just to be complete: I've also done some more testing and all seems to be well.

Tests included:

* Turning on/off the _Avoid artefacts_ preference. It is not needed to restart RawTherapee to make this work. If already zoomed in to 100(+)% you do need to zoom in or out one step to see the effect after changing this preference.
* This worked on the original 2 problem edits I posted in #6343.
   Both exporting and zooming now work when the _Avoid artefacts_ preference turned _off_.
* Tried to introduce this issue editing other RAWs using the known and a reduced fresh set of modules that triggered this: I could not do so.

Thanatomanic:
This solves the issue. However, I must admit that I don't understand exactly why this works. In particular, what is _automatic_ about `autocielab`? Can this variable be set through the GUI in any other way than through the new option in the preferences?

As for the new option itself: if I see something labeled "Avoid artifacts" I would think this is a good thing. However, having it enabled can actually cause (big white) problems. This seems very counter intuitive and would require some explanation at least.

Desmis:
@Thanatomanic  @jade-nl 

It solves the problem, because it's a very very old problem.
When I set up Ciecam in 2011, as time went by, bugs were discovered, which led to the deletion of upstream functions (this around 2012, 2013), these functions brought artifacts.
We gradually set up a workaround using several variables, distributed between improccoordinator, dcrop, simpleprocess and improcfun (where there is the main Ciecam code) . These 2 variables were named "autocielab" and "labpassone"...Also from memory there was a GUI part to control  which disappeared it seems to me.  I have only re-established it

We can discuss endlessly about the name of the variables, but at the time "auto" meant (I think) that automatically all the processes that brought artifacts were not taken into account.

The recent introduction of CAM16 that I added this year should normally remove the artifacts when using the incriminated functions (CA, wavelet, etc.) and therefore one can now use denoise and Ciecam at the same time (which was not the case with CAM02).  Normally the artifacts should appear (I hope) only in the case of CAM02

Of course we can change the GUI and replace Options with something else. That's what was set up 10 years ago. I don't see the point of it 

jacques

Thanatomanic:
I understand that CIECAM02 has difficulties, and I appreciate your historic efforts to counteract them to the best of your abilities 🙂 
I am a little surprised that "artifiact" means an entirely blank export image (which is very different from the preview!), but for now, I think this PR is good for a merge.

Desmis:
Artifacts does not mean white screen....when using the functions in question and Ciecam, we ended up with black spots, bad colors, etc.
"White sceen" was only discovered today if you validate all excluded functions.... which was never done before. This is a malfunction which should be rare (very very) because of CAM16 by default.
The development of ciecam02 was very laborious about ten years ago and involved the participation of many actors, who are no longer in the current team (except Ingo which is still there, fortunately)

I will merge tommrow morning 

Jacques

ghost:
Thanks for the time and effort, appreciated.

Issue #6273, opened 2021-06-05
JPEG XL (*.jxl) file format support


BR1py:
Hello,
I'm a fanboy of your program! 

For me the new JXL format ([see JXL](https://gitlab.com/wg1/jpeg-xl)) looks very promising and I like to see a full integration in rawtherapee (which means an import and export option).

Do you have any plans for the integration of the new format?

Thanatomanic:
@BR1py There are no plans currently, but help is always welcome if you're able.

See also: https://github.com/Beep6581/RawTherapee/issues/1895 (EXR), https://github.com/Beep6581/RawTherapee/issues/5931 (FITS), https://github.com/Beep6581/RawTherapee/issues/5313 (JPEG2000)

Studiedlist:
Hi! Jpeg-xl is great, I would like to see it in RawTherapee
I could do some experiments in implementing import/export features. Can you navigate me through code so I would know where to start?

Thanatomanic:
@Studiedlist The main functions for loading can be found here https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/imageio.cc

Maybe that can be a good starting point? We haven't added new file formats in years, so I cannot point you to a PR or issue that you could use as reference. Don't hesitate to ask.

Studiedlist:
Thanks, I will try to understand this part of code and implement jxl features

xiota:
@Thanatomanic  I have [JXL import working](https://github.com/xiota/RawTherapee/tree/jxl-import) when the file name is entered on the command line, but the JXL images don't show up in the file browser.  Any pointer on where to look to change that?  Thanks.

Thanatomanic:
@xiota Perhaps here? https://github.com/Beep6581/RawTherapee/blob/d053f02450f87e964aa9feb0efd80241203f43bd/rtgui/thumbnail.cc#L116

Here is series of `if`s that determine how the thumbnails are generated. 
There is also a list in your personal `options` file called `ParseExtensions` that is loaded by RT to determine which files to show. This list shows up in the Preferences.

xiota:
@Thanatomanic Thanks.  That was what I needed. Will open PR now.

beling:
Links to Dark Table experience and implementation:
https://github.com/darktable-org/darktable/issues/10866
https://github.com/darktable-org/darktable/pull/10044

0fbcb238c0:
Any progress here?

xiota:
[...]

Issue #4915, opened 2018-11-01
Why is png not enabled in the file browser by default?


Thanatomanic:
Simple question. Is there a good reason for it?

```
ParseExtensions=3fr;arw;arq;cr2;crf;crw;dcr;dng;fff;iiq;jpg;jpeg;kdc;mef;mos;mrw;nef;nrw;orf;pef;png;raf;raw;rw2;rwl;rwz;sr2;srf;srw;tif;tiff;x3f;
ParseExtensionsEnabled=1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;0;1;1;1;1;1;1;1;1;1;1;1;
```

Beep6581:
We disabled it because PNG files came in various shapes, sizes and odd aspect ratios and often lead to crashes or File Browser thumbnail screw-ups. Typical case: user opens a photo in some bad place like "Downloads", closes RT, then they download some PNG there, then RT keeps crashing on startup.

It would generally be good to enable PNG by default, especially since now RT writes metadata to PNG files (possibly even better than to TIFF files #4464).

Thanatomanic:
Okay, I see. But isn't that equally possible with jpegs of strange dimensions?

Beep6581:
Sure, but unusual JPEGs (e.g. 12-bit) are rare, and don't lead to crashes like PNGs do/did. I'm just pointing that out, I'm not defending that we should keep PNG disabled.

PR #6367, opened 2021-09-22
Add ability to import JXL images

xiota:
This allows RawTherapee to import JXL images.  Addresses [the import portion] of #6273.

Related: #7097 #6773


Thanatomanic:
@xiota Thanks for this! 

> Addresses half of #6273.

Which half? And can you elaborate a bit about what is still missing?

xiota:
#6273 requests import *and* export.  This is for import only.  Hence half.

Floessie:
@xiota Thanks for your contribution! :+1: There seem to be some (error) conditions under which `ImageIO::loadJxl()` leaks allocated memory in `buffer`. Could you double-check?

xiota:
@Floessie Thanks for letting me know.  I added a couple lines to free the buffer.

Thanatomanic:
@xiota Just out of curiosity, do you have a particular reason to force-push the last few commits instead of committing regularly with descriptions? We can squash-merge things in the end so the master commit log won't be polluted.

xiota:
@Thanatomanic Was just to keep the commit log clean.  Didn't know you would squash merge.

xiota:
This was working yesterday, but it isn't working after I recompiled today without making any changes.

xiota:
It's working again.  I was trying to make compiling with libjxl optional, but something was being cached, so it looked like it was working when it wasn't.  Someone else will have to make the changes needed to make cmake detect and load libjxl as needed.  (Right now it's required.)

rtgui/thumbnail.cc
```diff
@@ -136,50 +135,37 @@ void Thumbnail::_generateThumbnailImage ()
...
-
-        rtengine::eSensorType sensorType = rtengine::ST_NONE;
-        if ( initial_ && options.internalThumbIfUntouched) {
-            quick = true;
-            tpp = rtengine::Thumbnail::loadQuickFromRaw (fname, ri, sensorType, tw, th, 1, TRUE);
-        }
+    if ( tpp == nullptr ) {
+        quick = false;
+        tpp = rtengine::Thumbnail::loadFromRaw (fname, ri, sensorType, tw, th, 1, pparams->wb.equal, TRUE);
+    }
```
Lawrence37: Doesn't this lead to unnecessary function calls, compared to the old code?

rtdata/options/options.lin
```diff
@@ -12,8 +12,8 @@ MultiUser=true
 
 [File Browser]
 # Image filename extensions to be looked for, and their corresponding search state (0/1 -> skip/include)
-ParseExtensions=3fr;arw;arq;cr2;cr3;crf;crw;dcr;dng;fff;iiq;jpg;jpeg;kdc;mef;mos;mrw;nef;nrw;orf;ori;pef;png;raf;raw;rw2;rwl;rwz;sr2;srf;srw;tif;tiff;x3f;
-ParseExtensionsEnabled=1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;1;0;1;0;1;1;1;1;1;1;1;1;1;1;1;
+ParseExtensions=3fr;arw;arq;cr2;cr3;crf;crw;dcr;dng;fff;iiq;jpg;jpeg;kdc;mef;mos;mrw;nef;nrw;orf;ori;pef;png;raf;raw;rw2;rwl;rwz;sr2;srf;srw;tif;tiff;x3f;jxl;
```
Lawrence37: I think this should be in alphabetical order.

rtgui/thumbnail.cc
xiota: It will call functions to check if the file is a raw before trying to open non-raw file types.  But...

* It reduces nesting by removing the if-else-if-else-if-else structure.
* It allows formats to be added to imageio without having to return to this particular section to add file extensions.
* This is a raw processor, so prioritizing raw files seems reasonable.
* The raw-processing functions should exit quickly enough for non-raw files.

Lawrence37: Ok, if the raw functions exit quickly for non-raw files, it should be fine. Thanks for clarifying.

rtdata/options/options.lin
xiota: Is there a tool to keep the values in `ParseExtensionsEnabled` with whichever extensions they belong to?

Lawrence37: I don't think so. It has to be verified manually.

xiota: I put it in a csv file to look at it in LibreOffice... One of the disabled extensions is png.  Why is that?

Lawrence37: Not sure. I do remember png as one of the disabled extensions.

xiota: I moved `jxl` after `jpeg`.  I didn't check that the others are alphabetized though.

heckflosse:
> One of the disabled extensions is png. Why is that?

There was a time, where some png files led to crashes of RT when starting RT on a folder which such a file.
I think we can enable png per default now. Iirc the crashes should be fixed...

Lawrence37:
I remember now. It was issue #3794 right? There was also a question about it in #4915.

xiota:
Should I reenable PNG in this PR?  Also, any advice/help for conditional compilation?

Thanatomanic:
> I think we can enable png per default now. 

Yes, but not in this PR imo (keep it cleanly focused on JXL).

> Someone else will have to make the changes needed to make cmake detect and load libjxl as needed. (Right now it's required.)

@xiota Well, this is an issue indeed. I found out that on Windows MSYS2 does not provide the `libjxl` package yet. So anyone on Windows should first manually compile `libjxl` or we should start working with submodules or something. Not very convenient.
My suggestion is to create a build option `WITH_JXL` here https://github.com/Beep6581/RawTherapee/blob/0e122a485a2339bf2331a7c54a2f5cb2ee2f2dd3/CMakeLists.txt#L197-L225
and set it to OFF by default.
Then you should be able to conditionally check for the package, like here
https://github.com/Beep6581/RawTherapee/blob/0e122a485a2339bf2331a7c54a2f5cb2ee2f2dd3/CMakeLists.txt#L496-L498

But I'm not at all familiar with the intricacies of CMake, so perhaps this doesn't work at all...

EwoutH:
Thanks for this effort! I would love to have JPEG XL import support in RawTherapee, what's needed to move this PR forward?

Beep6581:
Builds fine in Manjaro against libjxl-0.7.0 and libjxl_threads-0.7.0, and opens a sample .jxl file fine, though without any metadata.

Thanatomanic:
> though without any metadata

Is it to be expected that once we support evix2 this automagically works for jxl too?

kmilos:
> Is it to be expected that once we support evix2 this automagically works for jxl too?

If exiv2 is built w/ BMFF support, then yes.

Lawrence37:
Converting this to a draft because the GitHub Actions workflows still need to be updated to use the JXL library. Making JXL optional may also be necessary for the moment.

Beep6581:
Moving to 6.0 due to lack of activity.

EwoutH:
@xiota Would you be able to further work on this? If so, is there anything you need to move this forward?

xiota:
Issues I think holding this back:

* [x] Conditional compilation to build libjxl support only when the library is present.
* [x] Windows library availability?
* [ ] I'd need a refresher on how rawtherapee works
* [ ] I'd need a refresher on how libjxl works
* [x] update and rebase to current rawtherapee
* [x] update for current libjxl
* [ ] I don't understand how libjxl color management works

kmilos:
> Windows library availability?

No problem w/ this one: https://packages.msys2.org/base/mingw-w64-libjxl

> I don't understand how libjxl color management works

You are not alone on that one 😉

xiota:
I force pushed to rebase to the current dev branch.  It's updated to the current libjxl API (only one function changed).  I also figured out how to make libjxl optional, so if it's not installed, compilation should continue with jpeg xl disabled.

CMakeLists.txt
```diff
@@ -552,6 +553,16 @@ if(WITH_SYSTEM_KLT)
...
 
+set(JXL_LIBS "libjxl libjxl_threads")
+if(WITH_JXL)
+    pkg_check_modules(JXL REQUIRED IMPORTED_TARGET ${JXL_LIBS})
+else()
+    pkg_check_modules(JXL IMPORTED_TARGET ${JXL_LIBS})
+endif()
+if(JXL_FOUND)
+    add_definitions(-DLIBJXL)    
+endif()
```
kmilos: Not a fan of this logic... If the user is explicitly expected to set WITH_JXL to ON, shouldn't silently do stuff behind their back when they no longer don't - they'll still get it built when they change it back to OFF just because the library is present on their system.

In fact, it might be better to default this to ON (you've added extension support), and enable the user to explicitly turn it OFF completely then.

xiota: Changed to tristate with settings: AUTO/ON/OFF

xiota:
@Lawrence37 Would you mind running this through workflows to check that the cmake files are compatible?

I updated them to add a tristate option `WITH_JXL`.  Default is AUTO, which enables JXL import when libjxl is available.  YES force enables the feature, showing error when libjxl is not available.  NO force disables the feature, even when libjxl is found.

I looked at the workflow files, and it seems for Ubuntu and MSYS, libjxl can be added to the depends lists.  However, Ubuntu has libjxl-dev 0.7.0, while MSYS has libjxl 0.10.2.  When I rebased, I updated and test built with 0.10.2.

Lawrence37:
The workflows will probably not support jxl without installing libjxl, but I'll run them anyway.

xiota:
@Lawrence37 Thanks for running the workflows earlier.  They all passed (deps not installed and build completed with jxl export disabled).

Would you mind running the workflows again?  I added libjxl to the dep lists.  Hoping they'll build with jxl export enabled.

xiota:
Looks like... [macOS](https://github.com/Beep6581/RawTherapee/actions/runs/8668458298?pr=6367) and [Windows](https://github.com/Beep6581/RawTherapee/actions/runs/8668458284?pr=6367) workflows succeeded.  I suppose people using those OSes can grab the artifacts to test?

The Linux workflows failed because they're running on Ubuntu 20.04 and 22.04, but [libjxl](https://packages.ubuntu.com/search?keywords=libjxl-dev&searchon=names&suite=all&section=all) was added in 23.04.  Should be available in 24.04 in a few weeks.

How would RT devs like to proceed?

Lawrence37:
It seems like there is a deb package for Ubuntu 20.04 (https://github.com/libjxl/libjxl/releases/tag/v0.10.2). We can use that or compile libjxl in the AppImage workflow. Note that the runner must be the oldest LTS Ubuntu available (currently 20.04) to maximize compatibility of the AppImage.

xiota:
Is figuring out the deps in the workflows important for this PR or could it be deferred?

Conditional compilation seems to be working, so the deps could be left out of the Linux workflows until someone who knows how can update them.

Lawrence37:
The AppImage workflow generates the official AppImage for releases, so it is important that libjxl is included. Otherwise, there will be no jxl support for AppImage users.

xiota:
I found the button to run workflows and got the [AppImage workflow](https://github.com/xiota/RawTherapee/actions/runs/8677760467) to complete successfully.  The resulting appimage was able to thumbnail and load a jxl file.

It uses debs from the libjxl github repo, but libhwy is from a PPA.  The highway repo doesn't publish debs, and the 22.04 deb won't install on 20.04.  The debian highway package has the same problem.

I'm unable to test the codeql workflow.  Assuming no mistakes, it should also work.  It uses libjxl from github, and libhwy from the official Ubuntu 22.04 repo.

Lawrence37:
Nice! Getting a package from a PPA is less than ideal. I see highway can be installed using `vcpkg install highway`. `vcpkg` comes preinstalled in the Ubuntu 20.04 environment. I have never used vcpkg before, but it might be worth a shot.

xiota:
Looks like vcpkg might have to be built from source.  Instead of doing that, I decided to try using [`makedeb`](https://www.github.com/makedeb/makedeb) to build, package, and install libjxl.  The resulting appimage seems to work.

kmilos:
FWIW, the libjxl project publishes deb packages for 20.04: https://github.com/libjxl/libjxl/releases/tag/v0.10.2

xiota:
@kmilos See preceding comments for why they are "less than ideal".

kmilos:
I though that was related to PPAs... But I guess that source, although "official", can be considered as one. Then again, I don't see how vcpkg would be different in essence. They are all 3rd parties...

xiota:
The libjxl release debs depend on libhwy, which is not in the 20.04 repos.  The only libhwy deb I found is from an old PPA.  The package itself is probably fine because canonical controls the build server.  But if the maintainer changes or deletes the repo, any workflows depending on it could break.

Ubuntu 22.04 does have libhwy in its repos, so the codeql workflow can use the libjxl release debs.

xiota:
When I searched earlier for how to use vcpkg, I found instructions about how to build from source or use custom actions that appear to build from source.  However, running a test workflow that calls `vcpkg install libjxl` does appear to be working (vcpkg is preinstalled).  So if preferred, I can try switching the workflow to vcpkg.

xiota:
Cannot use vcpkg...
```
-- Checking for modules 'libjxl;libjxl_threads'
--   No package 'libjxl' found
--   No package 'libjxl_threads' found
```

Lawrence37:
To clarify my previous comments:
Using the libjxl debs is fine because they are official. libjxl-dev depends on libhwy and libhwy-dev, which are not available on Ubuntu 20.04 and have no official debs. I saw on the project's GitHub that highway can be installed through vcpkg, and the Ubuntu 20.04 runner has vcpkg preinstalled. My recommendation is to try vcpkg for installing libhwy, after which the libjxl debs can be installed. I have not used vcpkg before, so I don't know if it will work well. I'm also unsure if installing highway through vcpkg will provide all the dependencies needed by libjxl-dev because I don't know if the package manager will see the highway installation. Because of the uncertainty, I'm not demanding that vcpkg be used, but I do think it is worth trying.

xiota:
vcpkg is a parallel system to apt, so whatever it installs won't satisfy apt dependencies.  It also installs somewhere that the build scripts don't normally use.  Since there's already a working solution that doesn't use a PPA, I'm not very interested in figuring out how to make vcpkg work.  Also, 20.04 is EOL next year.  22.04 has libhwy.  24.04 should have libjxl.

To clarify: I already made a few attempts to use vcpkg.  See error message [noted in previous comment](https://github.com/Beep6581/RawTherapee/pull/6367#issuecomment-2053992121).

Lawrence37:
@xiota I just reviewed the current workflow and I think it is fine because all dependencies either come from the official source or Ubuntu.

xiota:
To do?

- More user testing?
- Code review?
- Run `astyle` on added code sections.
- Anything else?

Now that workflows are creating artifacts for this, is there a way to publicize them to get users to test?

Lawrence37:
Code review and testing by developers are left. Astyle and static checkers like cppcheck are optional. You can use the official forum at https://discuss.pixls.us/c/software/rawtherapee/14 to generate publicity.

xiota:
> You can use the official forum ... to generate publicity.

[RawTherapee JPEG XL Import](https://discuss.pixls.us/t/rawtherapee-jpeg-xl-import/43189)

rtengine/imageio.cc
```diff
@@ -24,13 +24,20 @@
 #include <string>
 #include <utility>
 #include <vector>
-
```
Floessie, line 27: Please reapply this blnak line, as it seperates C++ headers from C headers.

rtengine/imageio.cc
```diff
@@ -24,13 +24,20 @@
...
 #include <fcntl.h>
 #include <glib/gstdio.h>
 #include <png.h>
 #include <tiff.h>
 #include <tiffio.h>
 
+#ifdef LIBJXL
+#include <fstream>
```
Floessie: Either add a blank line after the C++ header (vs. C headers below) or drop the header in favor of other tools (see my next comment).

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+    JxlBasicInfo info = {};
+    JxlPixelFormat format = {};
+
+    format.num_channels = 3;
+    format.data_type = JXL_TYPE_FLOAT;
+    format.endianness = JXL_NATIVE_ENDIAN;
+    format.align = 0;
+
+    // get file contents
+    std::ifstream instream(fname.c_str(), std::ios::in | std::ios::binary);
```
Floessie: I'm not sure, if this is the right way to deal with the `Glib::ustring` here. Aren't there safer GTK/Glib alternatives to read a file designated by a unicode filename? If so, you can drop the `<fstream>` include above.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+
+    if (JXL_DEC_SUCCESS !=
+        JxlDecoderSetParallelRunner(dec.get(), JxlResizableParallelRunner,
+                                    runner.get())) {
+        g_printerr("Error: JxlDecoderSetParallelRunner failed\n");
+        return IMIO_HEADERERROR;
+    }
+
+    // grand decode loop...
+    JxlDecoderSetInput(dec.get(), compressed.data(), compressed.size());
```
Floessie, line 880: This takes `const uint8_t* data`, so please make `compressed` const.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+    JxlPixelFormat format = {};
+
+    format.num_channels = 3;
+    format.data_type = JXL_TYPE_FLOAT;
+    format.endianness = JXL_NATIVE_ENDIAN;
+    format.align = 0;
+
+    // get file contents
+    std::ifstream instream(fname.c_str(), std::ios::in | std::ios::binary);
+    std::vector<uint8_t> compressed(
```
Floessie: We use `<cstdint>`, so this should be `std::uint8_t`.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+            }
+
+            JxlResizableParallelRunnerSetThreads(
+                runner.get(),
+                JxlResizableParallelRunnerSuggestThreads(info.xsize, info.ysize));
+        } else if (status == JXL_DEC_COLOR_ENCODING) {
+            // check for ICC profile
+            deleteLoadedProfileData();
+            embProfile = nullptr;
+            size_t icc_size = 0;
```
Floessie: `std::size_t`.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+                        "Warning: JxlDecoderGetColorAsICCProfile failed\n");
+                } else {
+                    embProfile = cmsOpenProfileFromMem(icc_profile.data(),
+                                                       icc_profile.size());
+                }
+            } else {
+                g_printerr("Warning: Empty ICC data.\n");
+            }
+        } else if (status == JXL_DEC_NEED_IMAGE_OUT_BUFFER) {
+            format.data_type = JXL_TYPE_FLOAT;
```
Floessie: Is this a precaution, or does `format.data_type` really change in the process?

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+int ImageIO::loadJxl(const Glib::ustring &fname)
+{
+    if (pl) {
+        pl->setProgressStr("PROGRESSBAR_LOADJXL");
+        pl->setProgress(0.0);
+    }
+
+    std::vector<uint8_t> icc_profile;
+
+    gpointer buffer = nullptr;
```
Floessie: We shouldn't use raw pointers anymore. There are two superior solutions, IMHO:

1. `std::vector<unsigned char>`
2. `std::unique_ptr<unsigned char[]>`

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+        } else if (status == JXL_DEC_NEED_IMAGE_OUT_BUFFER) {
+            format.data_type = JXL_TYPE_FLOAT;
+
+            if (JXL_DEC_SUCCESS !=
+                    JxlDecoderImageOutBufferSize(dec.get(), &format, &buffer_size)) {
+                g_printerr("Error: JxlDecoderImageOutBufferSize failed\n");
+                return IMIO_READERROR;
+            }
+
+            buffer = g_malloc(buffer_size);
```
Floessie: No raw pointers, please. See comment above.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+                g_free(buffer);
+                return IMIO_READERROR;
+            }
+        } else if (status == JXL_DEC_FULL_IMAGE ||
+                   status == JXL_DEC_FRAME) {
+            // Nothing to do. If the image is an animation, more full frames
+            // may be decoded. This example only keeps the first one.
+        } else if (status == JXL_DEC_SUCCESS) {
+            // All decoding successfully finished.
+            // It's not required to call JxlDecoderReleaseInput(dec.get())
```
Floessie: This comment seems redundant, as `dec` obviously is a smart pointer.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+            // since the decoder will be destroyed.
+            break;
+        } else if (status == JXL_DEC_NEED_MORE_INPUT) {
+            g_printerr("Error: Already provided all input\n");
```
Floessie: What about the more common "Premature end of file"?

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+            return IMIO_READERROR;
+        }
+    }  // end grand decode loop
+
+    unsigned int width = info.xsize;
+    unsigned int height = info.ysize;
+
+    allocate(width, height);
+
+    int line_length = width * 3 * 4;
```
Floessie: This should be unsigned, too.

rtengine/imageio.cc
```diff
@@ -822,6 +829,171 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+
+    for (size_t row = 0; row < height; ++row) {
```
Floessie: `std::size_t`

rtengine/imageio.h
```diff
@@ -90,6 +90,10 @@ class ImageIO : virtual public ImageDatas
     int load (const Glib::ustring &fname);
     int save (const Glib::ustring &fname) const;
 
+#ifdef LIBJXL
+    int loadJxl (const Glib::ustring &fname);
```
Floessie: Hmm, the other `load...()` extensions are uppercase. And better append the new one rather than prepend it.

rtengine/utils.cc
```diff
@@ -237,6 +237,14 @@ bool hasJpegExtension(const Glib::ustring& filename)
    return extension == "jpg" || extension == "jpeg";
 }
 
+#ifdef LIBJXL
+bool hasJxlExtension(const Glib::ustring& filename)
+{
+   const Glib::ustring extension = getFileExtension(filename);
+   return extension == "jxl";
```
Floessie: This can be a one-liner: `return getFileExtension(filename) == "jxl";`

rtgui/thumbnail.cc
```diff
@@ -205,53 +204,40 @@ void Thumbnail::_generateThumbnailImage ()
...
-        //     image out of the RAW. Mark as "quick".
-        //  2. if we don't find that then just grab the real image.
-        bool quick = false;
-
-        rtengine::eSensorType sensorType = rtengine::ST_NONE;
-        if ( initial_ && options.internalThumbIfUntouched) {
-            quick = true;
-            tpp = rtengine::Thumbnail::loadQuickFromRaw (fname, sensorType, tw, th, 1, TRUE);
-        }
+    if ( tpp == nullptr ) {
```
Floessie: `if (!tpp) {`

Floessie (changes requested):
Quick review of the diff with some minor change requests, most prominently the change from raw pointer `buffer` to something more sane. :wink:

rtengine/imageio.cc
xiota: glib has a function to read files, but it's pointer heavy.
```
gboolean
g_file_get_contents (
  const gchar* filename,
  gchar** contents,
  gsize* length,
  GError** error
)
```

rtgui/thumbnail.cc
xiota: Okay... but... I didn't write that line...
```
$ git blame -L241,+1 0fdac3797aebbba4a751184e60278e5460bc56b7 -- rtgui/thumbnail.cc
0731975ff0 (Flössie 2016-10-12 17:48:40 +0200 241)         if ( tpp == nullptr ) {
```

rtengine/imageio.cc
xiota: I don't remember why I did that.  I can change to `assert()`.

rtengine/imageio.cc
xiota: Could you point me to usage examples?  Is it safe to feed vector into a function that takes buffer pointer and size arguments?

rtengine/imageio.cc
xiota: The decode loop is from libjxl examples.  Those are the status messages they used. In `jxl/decode.h`, I see nothing corresponding to "premature end of file".  The closest looks like `JXL_DEC_NEED_MORE_INPUT`, but that only occurs when an embedded preview is requested.  Since this isn't asking for a preview, that status would/should never occur.  I guess the result would be a partially decoded image or `JXL_DEC_ERROR`.

rtengine/imageio.cc
xiota: The decode loop, including a variation of that comment, is from libjxl examples.

`JXL_DEC_SUCCESS` and "All decoding successfully finished." seems more obvious.

What about shortening it to: "Decoder will be released automatically."

rtengine/imageio.cc
Floessie: How about `Gio::File` that is used in various places? But maybe @Lawrence37 has an idea on how problematic the `filename.c_str()` actually is...

rtengine/imageio.cc
Floessie: Yes, it is. And if I'm not mistaken, you already do that with `icc_profile`. :wink:

rtengine/imageio.cc
Floessie: That's fine. :+1:

rtengine/imageio.cc
Floessie: Okay. If I read "Error: Already provided all input" somewhere this would leave me at a loss. Glad to hear that this should never occur. :sweat_smile:

rtgui/thumbnail.cc
Floessie: I know, I know. But while you're at it...

And actually, I didn't write that line either, just fired clang-tidy at it. :smiley:

rtengine/imageio.cc
xiota: Assuming I'm reading the documentation correctly.  I'll try to test with some truncated files to see what happens.

rtgui/thumbnail.cc
xiota: In this file, there are five instances of `tpp == nullptr` and two of `!tpp`.  Should I change them all to `!tpp`?

(I'm going to change all of them.)

rtengine/imageio.cc
xiota: Maybe, `JXL_DEC_NEED_MORE_INPUT` == "premature end of file"?

(I think I didn't understand the intent of your initial comment.)

rtengine/imageio.cc
xiota: I found this, but I'm pretty sure there's a pointer to remember to free: `bool Gio::File::load_contents(char *&contents, gsize &length);`

rtengine/imageio.cc
xiota: My left brain doesn't know what my right brain is doing.

rtengine/imageio.cc
xiota: This seems to work... don't know if it's safe passing everything through std::string.
```C++
std::string const contents = Glib::file_get_contents(fname);
std::vector<std::uint8_t> const compressed(contents.begin(), contents.end());
```

xiota:
@Floessie Thank you for reviewing.  Let me know if I missed something or there's more.

rtengine/imageio.cc
Floessie: Hmm, `Glib::file_get_contents(fname)` takes a `std::string` not a `Glib::ustring` and also thows. Additionally, this doubles the amount of memory used. So no, this is not a solution, I'm afraid.

rtengine/imageio.cc
```diff
@@ -967,26 +964,24 @@ int ImageIO::loadJxl(const Glib::ustring &fname)
...
 
     allocate(width, height);
 
-    int line_length = width * 3 * 4;
+    std::size_t line_length = width * 3 * 4;
 
-    for (size_t row = 0; row < height; ++row) {
-        setScanline (row, ((const unsigned char*)buffer) + (row * line_length), 32);
+    for (std::size_t row = 0; row < height; ++row) {
+        setScanline(row, ((const unsigned char*)buffer.data()) + (row * line_length), 32);
```
Floessie: I don't think this C-style cast is needed anymore. That's why I suggested `buffer` should be `std::vector<unsigned char>`...

Floessie:
@xiota It's quite okay from my point of view now. Let @Lawrence37 comment on the `Glib::ustring` problem and functionality.

Thanks for your contribution! :+1:

rtengine/imageio.cc
xiota: I will try removing the cast.  If no problems, will bundle with other changes.

rtengine/imageio.cc
xiota: Would be nice to have a function to read a file into a vector that could be used with all decoders that work with memory buffers.  Would such a function go in rtengine/utils.cc?  Maybe I should look...

I think I misunderstood your initial comment (again).  I didn't notice any issues reading files with Chinese and Japanese filenames using either fstream/c_str or file_get_contents.  I don't recall the reason for using c_str().  I vaguely remember having problems with casting or auto conversion to std::string not working.  Might be a glitch that's been resolved by now.

Lawrence37: `Glib::ustring::c_str()` returns a UTF-8 encoded Unicode string, but `std::ifstream`'s expected encoding is platform-dependent (I think?). Maybe [`Glib::filename_from_utf8`](https://developer-old.gnome.org/glibmm/2.48/group__CharsetConv.html#ga2bd94064ad97b43324a7854b62f0835b) can be used to ensure the correct encoding for the file name? I'm not sure if the encoding is meant for the platform, Gio, or both. For a full Gio method, you can try getting a file with [`Gio::File::create_for_path`](https://developer-old.gnome.org/glibmm/2.48/classGio_1_1File.html#a1b1a4553b226689d50b9525488223e25), opening an input stream with [`Gio::File::read`](https://developer-old.gnome.org/glibmm/2.48/classGio_1_1File.html#a0395b3b0b8cec5dc1fd9953e52e5fbc1), and finally read the file content into your buffer with [`Gio::InputStream::read_all`](https://developer-old.gnome.org/glibmm/2.48/classGio_1_1InputStream.html#a6da0fc57d1a5df7107370c67e84aa496).

xiota: I added a function to `utils.cc` because it's potentially useful elsewhere.  It uses `Glib::filename_from_utf8` to convert ustring to std::string.  It returns an empty vector if there's any error.

```c++
std::vector<std::uint8_t> getFileData(const Glib::ustring &filename);
```

xiota:
Checking in, since it's been nearly a month... anything else needed on my end for this?

Lawrence37:
No, you are just waiting for functionality testing. It's unfortunate that no one has commented about their experience on the Pixls topic. I will test if no one else does, but I do have a backlog of other pull requests to review.

xiota:
@Lawrence37 Thanks for letting me know.  No hurry...  Just wanted to know the status.

rtengine/imageio.cc
```diff
@@ -822,6 +827,172 @@ int ImageIO::loadTIFF (const Glib::ustring &fname)
...
+
+    format.num_channels = 3;
+    format.data_type = JXL_TYPE_FLOAT;
+    format.endianness = JXL_NATIVE_ENDIAN;
+    format.align = 0;
+
+    std::vector<std::uint8_t> const compressed = getFileData(fname);
+
+    if (compressed.empty()) {
+        g_printerr("Error: loadJXL failed to get data from file\n");
```
Lawrence37: Elsewhere we are using `std::cerr` or `fprintf`

Lawrence37:
I got this error compiling with libjxl 0.8.2.
```
/RawTherapee/rtengine/imageio.cc: In member function ‘int rtengine::ImageIO::loadJXL(const Glib::ustring&)’:
/RawTherapee/rtengine/imageio.cc:831:19: error: cannot convert ‘JxlColorProfileTarget’ to ‘const JxlPixelFormat*’
  831 | #define _PROFILE_ JXL_COLOR_PROFILE_TARGET_ORIGINAL
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                   |
      |                   JxlColorProfileTarget
/RawTherapee/rtengine/imageio.cc:904:60: note: in expansion of macro ‘_PROFILE_’
  904 |                     JxlDecoderGetICCProfileSize(dec.get(), _PROFILE_, &icc_size)
      |                                                            ^~~~~~~~~
In file included from /usr/include/jxl/decode_cxx.h:20,
                 from /RawTherapee/rtengine/imageio.cc:29:
/usr/include/jxl/decode.h:779:50: note:   initializing argument 2 of ‘JxlDecoderStatus JxlDecoderGetICCProfileSize(const JxlDecoder*, const JxlPixelFormat*, JxlColorProfileTarget, size_t*)’
  779 |     const JxlDecoder* dec, const JxlPixelFormat* unused_format,
      |                            ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
/RawTherapee/rtengine/imageio.cc:831:19: error: cannot convert ‘JxlColorProfileTarget’ to ‘const JxlPixelFormat*’
  831 | #define _PROFILE_ JXL_COLOR_PROFILE_TARGET_ORIGINAL
      |                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                   |
      |                   JxlColorProfileTarget
/RawTherapee/rtengine/imageio.cc:920:40: note: in expansion of macro ‘_PROFILE_’
  920 |                             dec.get(), _PROFILE_,
      |                                        ^~~~~~~~~
/usr/include/jxl/decode.h:799:50: note:   initializing argument 2 of ‘JxlDecoderStatus JxlDecoderGetColorAsICCProfile(const JxlDecoder*, const JxlPixelFormat*, JxlColorProfileTarget, uint8_t*, size_t)’
  799 |     const JxlDecoder* dec, const JxlPixelFormat* unused_format,
      |                            ~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~
```

kmilos:
There was a change in API at some point, 0.9 I think...

xiota:
I test built only with 0.7 and 0.10 because those are the versions readily available in repos.  Looks like the cutoff version I ~~got from examples~~ [chose] isn't accurate.
```
#if JPEGXL_NUMERIC_VERSION < JPEGXL_COMPUTE_NUMERIC_VERSION(0, 8, 0)
```

I'll look for the commit that made the change

kmilos:
You can also compare to the darktable #ifdefs, I think we got it right there: https://github.com/darktable-org/darktable/blob/master/src/imageio/imageio_jpegxl.c

xiota:
Maybe I misinterpreted how to choose the cutoff before.  The changes were made in v0.9-snapshot-409-gb08a7049.  darktable uses 0,9,0.  Will change it.

Lawrence37:
They make a note of API changes in their change log: https://github.com/libjxl/libjxl/blob/main/CHANGELOG.md
This can be helpful for determining if there are other places we can improve compatibility. It can also be used for setting a minimum required version if necessary.

Lawrence37 (commented):
The functionality seem fine. It properly loads the embedded color profiles. EXIF data reads without issue (images without a container load without bugs). HDR images retain data. Animated and transparent images load (with necessary limitations).

2 things that could be improved but do not block this pull request:
- Transparent parts of images are not completely black. There is some weird smearing of colors from opaque areas. See the Dice image from https://jpegxl.info/test-page/.
- JXL status/library version could be added to AboutThisBuild.txt.

xiota:
Switched from `g_printerr` to `std::cerr`.  Added libjxl version to `AboutThisBuild.txt`.

I suspect the color smearing may be how the file is actually encoded.  To get rid of it, might have to decode and process the alpha channel.  darktable shows the same effect.  Does RT have any existing code that deals with transparency that I could look at?

> It properly loads the embedded color profiles.

~~I think this is done by code elsewhere.~~

Lawrence37:
I don't know if the code is in RawTherapee itself, but transparency is handled for pngs and tiffs.

Lawrence37:
Do you know what it shows in AboutThisBuild.txt if jxl support is disabled or if libjxl is not installed?

One more possible improvement. Users who are upgrading will not see jxl images because their settings carry over. It would be good to automatically add the jxl extension to the list of file types to load if the version in the options file shows 5.10 or older.

xiota:
> Do you know what it shows in AboutThisBuild.txt if jxl support is disabled or if libjxl is not installed?

When not installed, it shows something like: `libjxl: V` (the version number is missing).  Should be the same for the other libraries.

When disabled, it shows the version that would have been installed.

I'll change it to something nicer like `n/a` for both cases.

> Users who are upgrading will not see jxl images because their settings carry over.

How is this handled when new raw formats are added?

Lawrence37:
> How is this handled when new raw formats are added?

I don't think this was ever handled. The last addition was .ori, which is disabled by default anyway. The other extension in the git history is .cr3. Nothing special was done for that even though I think something should have been.

xiota:
Maybe enabling new file extensions could go in a separate PR that takes care of all of them together?

Handling the transparency case could also go in a separate PR someday, like after jxl export is added.

Lawrence37:
Yes, it's fine to do that in a separate PR.

CMakeLists.txt
```diff
@@ -552,6 +555,30 @@ if(WITH_SYSTEM_KLT)
...
+    else()
+        message(STATUS "  JXL support disabled by WITH_JXL = ${WITH_JXL}")
+        set(JXL_VERSION "Disabled")
+        set(JXL_INCLUDE_DIRS "")
+        set(JXL_LIBRARIES "")
+        set(JXLTHREADS_INCLUDE_DIRS "")
+        set(JXLTHREADS_LIBRARIES "")
+    endif()
+else()
+    set(JXL_VERSION "Not installed")
```
Lawrence37: It is not installed at compile time, but libjxl could be installed later or RawTherapee could be distributed to someone who does have libjxl installed. "Disabled" suffices for the user.

xiota: I'd rather keep the strings different.  AboutThisBuild reflects the state of the system at build.  Having different strings is useful to distinguish which branch was taken.  The steps to enable JXL would be slightly different for each.

In a few years, when libjxl is standard across all distros, the AUTO setting can be removed.

Lawrence37: There is a potential problem with a simple "Not installed" message. Let's say a package distributor shares a build of RawTherapee without jxl support due to libjxl not being installed on the machine used for compiling. A user who wants to open jxl images will find that they don't show up in the file browser. Naturally, they may ask about it (on Pixls, here, or elsewhere) and share their operating system, hardware specifications, where they got RawTherapee from, and the build information  in the splash dialog. They, or someone else, will notice that it says libjxl is not installed. Next they install libjxl and restart their computer, but it still says libjxl is not installed!

This is confusing for people, so it should either be specific that it was not installed during the build process and therefore there is no jxl support, or it should not reference the installation status.

Apologies for the long post, but I hope I'm clear about what my concern is.

xiota: I'll change it to "Disabled (Auto)" for the unavailable / not installed case.  That should prevent the confusion you describe as well as differentiate for later troubleshooting.

xiota:
```

# installed, enabled or auto
libjxl: 0.10.2

# installed, disabled (WITH_JXL=FALSE)
libjxl: 0.10.2 - Disabled

# not installed, auto
libjxl: Disabled, Auto

# not installed, disabled (WITH_JXL=FALSE)
libjxl: Disabled
```

Lawrence37 (approved):

Lawrence37:
Looks good to me. You are welcomed to add yourself to AUTHORS.txt.

xiota:
I'll leave it alone for now... can always add in a future...

Does github have a way to track todos? jxl export, enable new file types in config, process jxl alpha channel

kmilos (approved):
LGTM

Lawrence37:
> Does github have a way to track todos?

You can just create issue(s) for them. If you want just one issue, GitHub will pick up any checklist in the original post and show a progress widget for it. See #6563 for an example.

Lawrence37:
@Floessie There are a few minor changes since your last review. Do you want some time to go over the recent updates, or are you ok for merging?

Lawrence37:
Merged. Thank you @xiota for working on this!

rtgui/thumbnail.cc
Lawrence37: @xiota FYI this may be causing some jpg files to open incorrectly because they are treated as raw files. See https://discuss.pixls.us/t/rt-problem-nikon-z8-w-z50mm-1-8s-lens/43933

xiota: Thanks.  Will look at more.

xiota: I have a potential fix.  Do you want it in a separate PR?  Would it be okay to add to #7096, along with a couple other misc changes, like a workflow update that needs to be in dev to make testing libjxl 0.7.x easier.

Issue #6376, opened 2021-10-25
Add a average filter for pixel shift to process the moving  parts


huahuasousou:
In the current version5.8, we can use median for moving parts to remove some objects. but some photography don't want to remove the moving object, and they want to have a Motion blur. So maybe we can add an average filter for moving parts. and it can achieve the slow shutter effect that simulates long exposure.

heckflosse:
@huahuasousou 

I created psaverage branch with the requested feature

heckflosse:
@huahuasousou Any interest to test this?

huahuasousou:
> @huahuasousou Any interest to test this?

Hi, I can test this feature

huahuasousou:
> @huahuasousou Any interest to test this?

Hi, I just finished the test, result it's good. you can check it in attachement.
![_DSC8766](https://user-images.githubusercontent.com/35621281/141019085-53e1ea9f-83e4-4bc9-9135-e07f23733c48.jpg)

![_DSC8766_PSMS](https://user-images.githubusercontent.com/35621281/141019070-2e55d07a-65bc-422b-9fca-1450d895f88a.jpg)

heckflosse:
@huahuasousou Thanks for testing 👍

PR #6380, opened 2021-11-10
Add an average filter for pixel shift to process the moving parts, cl…

heckflosse:
…oses #6376


Floessie (approved):
LGTM. Very straight forward. The manual radio buttons could be changed into real ones. It's a pitty, we can't have a `pixelShiftMethod` in the PP3 for compatibility reasons.

heckflosse:
@Floessie Thanks for the review 👍

PR #6382, opened 2021-11-18
Support Dnggainmap for bayer files

heckflosse:


heckflosse:
Build works fine here. No idea why checks fail ... :-(

heckflosse:
@Floessie Do you have an idea why checks fail?

Lawrence37:
Because `rawimagesource.h` now includes `rawimage.h`, `dcp.cc` includes `myfile.h` (`dcp.cc` < `rawimagesource.h` < `rawimage.h` < `dcraw.h` < `myfile.h`). `dcp.cc` has `using namespace rtengine;`, so it finds the `rtengine::fopen`s in `myfile.h` instead of looking in `stdio.h`. Something like this will fix the compilation error, but I don't know if this is the right approach.
```diff
diff --git a/rtengine/dcp.cc b/rtengine/dcp.cc
index b65bb5f72..f8b5456c5 100644
--- a/rtengine/dcp.cc
+++ b/rtengine/dcp.cc
@@ -741,7 +741,7 @@ DCPProfile::DCPProfile(const Glib::ustring& filename) :
         1.00000f
     };

-    FILE* const file = g_fopen(filename.c_str(), "rb");
+    FILE* const file = ::g_fopen(filename.c_str(), "rb");

     if (file == nullptr) {
         printf ("Unable to load DCP profile '%s' !", filename.c_str());
```

Lawrence37:
I noticed a faint grid pattern after applying the gainmap. Using the flat-field image here https://filebin.net/5avurwfzusukehbl and starting with the neutral profile, I enabled flat-field correction from metadata, set the white balance, reduced the exposure, and applied +100 contrast.
image
Enhanced further:
image

Edit: A similar thing happens in ART. Is this just how the gainmap works?

heckflosse:
@Lawrence37 Thanks for testing. Looks like a bug. I'll investigate.

Lawrence37:
I'm guessing the bilinear interpolation is not smooth enough.

heckflosse:
I double checked the bilinear interpolation code. Code is ok.

Entropy512:
That's probably just a limitation in the design of GainMap - bilinear interpolation with a limited number of datapoints may simply not be able to model the actual optical behavior of the lens.

heckflosse:
Ok to merge?

Thanatomanic:
@heckflosse Do you have a set of files to generally test this functionality of the gainmap? I know there is a random assortment in a thread on Pixls... but that's not an easy test suite.

Otherwise, I haven't really tested the code. Therefore not sure about a merge.

heckflosse:
@Thanatomanic 

> Do you have a set of files to generally test this functionality of the gainmap?

Yes, I have such a set. But it's huge and my upload speed is quite slow. It's about 1 GB. I can create a 7z from that an upload it tomorrow. Ok ?

heckflosse:
@Thanatomanic Looks like the compressed size is a bit lower. I will start upload soon. 

Edit: I will send you the link to the filebin as soon as the upload is completed via pixls.us...

Edit: Compressed size is 608 MB. Uploading now. Not all files contain valid gainmaps. You should set verbose=true in options file to check

Lawrence37:
The gainmap can be applied in addition to a flat-field file. Should it be one or the other, or is there a use case for using both simultaneously?

Entropy512:
I can't think of any usecases for both simultaneously.  Of course, Murphy's Law says that if it's decided to force these to be mutually exclusive, someone will find a usecase...

Thanatomanic:
Flat-field can still be used for dust correction, no? That is independent of the gainmap. There's your usecase, maybe ;-)

Entropy512:
Hmm...  Maybe, although you'd have to apply the gainmap to both the flatfield image and the target image or you'd be double-correcting for the optical effects that are corrected for in the gainmap.  I think that would basically cancel out the gainmap operation, leaving you using a flatfield for both vignetting correction and dust correction.  Unless the gainmap was per-illuminant (some comments imply that if the phone does things correctly, it will be) and you want to apply a flatfield collected under one illuminant to an image collected under a different one for dust collection.  That might work?  Or some way to have the flatfield tool determine what was "residual" correction in a flat after gainmap is applied?

Lawrence37:
I completely forgot about dust removal. I use flat-field images to remove dust spots and vignetting together, but using the flat-field image just for dust removal is valid.

Thanatomanic:
Hi @heckflosse I (finally) gave this a testrun. I have gone through your test set and found no immediate errors or crashes. However, two comments:
1. The placement of the checkbox is now above the flat-field controls, though I don't think these controls affect the application of the gainmap. I would suggest to put the checkbox "From Metadata" on the top and then put the rest of the controls in a groupbox (see profiled lens correction for example). Maybe other distinctions are possible, but something must be done I think.
2. The application of the gainmap can push nearly-clipped pixels past the point of no return. For example, the file `htc_u12_plus_04.dng` that you send me has an almost blown-out sky. But some careful application of highlight recovery and compression can get this back to something acceptable and blue. However, after application of the gainmap, this data is irretrievable lost. I didn't check the code, but I assume some hard clipping is applied.
image
Is this expected behavior? Or is there a way to apply the gainmap, but prevent the hard clipping?

heckflosse:
> I didn't check the code, but I assume some hard clipping is applied.

No, there is only hard clipping for the darks (to prevent them to go negative when subtracting black point). At the highlights there is no hard clipping.

```
    for (std::size_t y = 0; y < static_cast<size_t>(H); ++y) {
        const float rowBlack[2] = {black[FC(y,0)], black[FC(y,1)]};
        const float ys = y * row_scale;
        float xs = 0.f;
        for (std::size_t x = 0; x < static_cast<std::size_t>(W); ++x, xs += col_scale) {
            const float f = getBilinearValue(mvals[y & 1][x & 1], xs, ys);
            const float b = rowBlack[x & 1];
            rawData[y][x] = rtengine::max((rawData[y][x] - b) * f + b, 0.f);
        }
    }

```

Also, the htc u12 files have a wrong white level. dcraw code reads wl = 1023, while in my tests it is 959. That makes me think the camera subtracts perhaps a black level of 64 before writing the files.

Anyway, clipped areas in an image can not be correctly flatfield corrected, because flatfield correction means multiplications and on clipped areas the multiplication also clips (shit in => shit out).

If no area in the image is clipped, it should be possible to apply a 'Clip Control' as we already have for the Flat field corection using flat field shots.

Thanatomanic:
@Lawrence37 @Beep6581 @Entropy512  This PR is working and should be ready for a merge in 5.10. As Ingo explained, the issue with highlights is a side-effect of multiplication and something that a little HR recovery can easily mitigate.

I will propose a change for the GUI to change the position of the checkbox.
Other than that I would like your thoughts on making this setting Enabled in our bundled profiles. It makes sense to apply this automatic correction to provide a (better, less vignetted) starting point for editing.

Thanatomanic:
image

New layout

Lawrence37:
The separator bar is a nice addition. Regarding having the gainmap enabled, some of our profiles have automatic profiled lens correction with vignette correction on. There would be double correction. Since the Lensfun correction is the more common case, I think it would suffice to keep the profiles as-is.

Thanatomanic:
Thanks @heckflosse! (I hope you will return one day...)

Beep6581:
> @Lawrence37 @Beep6581 @Entropy512 This PR is working and should be ready for a merge in 5.10. As Ingo explained, the issue with highlights is a side-effect of multiplication and something that a little HR recovery can easily mitigate.
> 
> I will propose a change for the GUI to change the position of the checkbox. Other than that I would like your thoughts on making this setting Enabled in our bundled profiles. It makes sense to apply this automatic correction to provide a (better, less vignetted) starting point for editing.

Is this new functionality explained anywhere?

PR #6383, opened 2021-11-22
Favorites preferences

### LINKED ISSUE #5256: GUI for adding tools to Favorites tab
The Favorites tab functionality was finished in #5016 and will be available in RawTherapee 5.6. Tools must currently be added manually: http://rawpedia.rawtherapee.com/Favorites_Tab

This issue is for creating a GUI to simplify adding tools to the Favorites tab.

Lawrence37:
A way to edit the favorites panel in Preferences.

image

This is a work in progress. I'm opening this pull request to get feedback in case I'm going in the wrong direction.


ghost:
Just merged and build what you have thus far. Except for the following warning in the terminal it configures and builds without a problem:

```
[100%] Building CXX object rtgui/CMakeFiles/rth.dir/zoompanel.cc.o
/home/jade/.local/git/RawTherapee/rtgui/toolpanelcoord.cc:36:47: warning: declaration 'struct ToolPanelCoordinator::ToolTree' does not declare anything
 using ToolTree = struct ToolPanelCoordinator::ToolTree;
```
Actually using it also works as expected.

Not sure if I like the variable width panels, but I can understand why it would be implemented this way. Then again, how long ("wide") can the name of a tab/module/tool be before it gets too long in the first place?

---

The following might not be part of the scope of this PR but I do want to mention it just in case:

I'm not at all a fan of the fact that the tools are removed from their original tab when setting up favourites. From a users perspective I cannot find a good reason why this is the case, to be honest.

Anyway: Nice to see that this is implemented in an easy-to-use form!

Thanatomanic:
Great to see you took this up @Lawrence37 ! I'll try to test it out soon-ish.

I agree with @jade-nl on his last point: I don't see a good UX reason why panels should not be copied to favorites, instead of moved there. I can imagine this can be tricky to implement though...

heckflosse:
This fixes the warning

```diff
diff --git a/rtgui/toolpanelcoord.cc b/rtgui/toolpanelcoord.cc
index 586d4755f..23b297d20 100644
--- a/rtgui/toolpanelcoord.cc
+++ b/rtgui/toolpanelcoord.cc
@@ -33,7 +33,7 @@
 using namespace rtengine::procparams;

 using Tool = ToolPanelCoordinator::Tool;
-using ToolTree = struct ToolPanelCoordinator::ToolTree;
+using ToolTree = ToolPanelCoordinator::ToolTree;

 const std::vector<ToolTree> EXPOSURE_PANEL_TOOLS = {
     {

```

Lawrence37:
@jade-nl @Thanatomanic
Thanks for the feedback.

The variable width tool lists is something I don't like as well. Eventually, this will be changed to fixed-width. I need to figure out how the width behaves with GTK scaling (HiDPI).

I will be touching some of the code that adds the tools to the favorites panel. It may not be easy to make copies of the tools. GTK might not allow widgets to be placed in multiple locations at once. More research needed...

arturoisilvia:
I've been using favorites for a long time and as it is now it's fine for me: move modules (not copy them to favorites). But it is a personal preference.
And also the order of the modules is important.
These are my favorites: With these from top to bottom I finish an edition quickly.
Favoritos_EN

ghost:
@Lawrence37 

This might be due to the WIP/Draft state this is in but: When you open preferences it opens with the focus on the *Favourites* tab and not the *General* tab.

Floessie:
@Lawrence37 I can reproduce the macOS build failure with a libc++-Toolchain on Linux. This is due to `enum class` not being hashable (a known [defect in C++11](http://www.open-std.org/jtc1/sc22/wg21/docs/lwg-defects.html#2148) fixed in C++14). I'll see what I can do...

Floessie:
My first idea was this:

```patch
diff --git a/rtgui/guiutils.h b/rtgui/guiutils.h
index 126cf672b..83d473abc 100644
--- a/rtgui/guiutils.h
+++ b/rtgui/guiutils.h
@@ -20,6 +20,7 @@
 
 #include <functional>
 #include <map>
+#include <type_traits>
 
 #include <gtkmm.h>
 
@@ -30,6 +31,21 @@
 #include "../rtengine/coord.h"
 #include "../rtengine/noncopyable.h"
 
+namespace std
+{
+
+template<typename T, typename enable_if<is_enum<T>::value && !is_convertible<T, int>::value, int>::type = 0>
+struct hash {
+    size_t operator ()(T val) const noexcept
+    {
+        using type = typename underlying_type<T>::type;
+
+        return hash<type>{}(static_cast<type>(val));
+    }
+};
+
+}
+
 namespace rtengine
 {
 
```

But that leads to ambiguity. I think best would be an own `std::unordered_map<>` typedef (using `using`, of course) with a custom hasher like [here on SO](https://stackoverflow.com/a/24847480).

Floessie:
@Lawrence37 Should be fixed now. `toollocationpref.cc` is nicely PIMPL'ed but you could also make use of the anonymous namespace for helpers.

Lawrence37:
@jade-nl The preferences window opens on the General tab now.

Lawrence37:
Restarting is no longer required after editing the favorites in preferences. The dynamic re-location of tools should also make it easy to keep favorite tools in their original panels. I will add an option to select between the current behavior or duplicate the tools.

I nested some widgets in more GTK boxes (to keep the tools separate from the decoration in the tabs, and to separate the sub-tools from the rest of the parent tool's widgets). This broke some of the RawTherapee theme. I fixed a border issue, but there is still a margin/padding issue I need to look at. Maybe @Thanatomanic can help.

Thanatomanic:
@Lawrence37 Great work as always! I'll take a look at the CSS.

ghost:
Nice updates!

Glad to see that you found a way to fix the variable window width "issue" and I'm looking forward to the *duplicate tool* solution.

Lawrence37:
Favorite tool duplication is done. Apart from the CSS issue(s), I consider this complete. Now is the time for testing and last-minute feature requests. I'll write up some documentation and maybe dig into the CSS problem again.

ghost:
I'm really happy with the way this turned out and I'm sure the CSS issue(s) will be dealt with!

Haven't found any mishaps and I can't think of any add-ons/changes that I would like to see.

Thanks for the time spend working on and implementing this!

heckflosse:
We should put a hint in the docs (rawpedia), that on Windows the duplication of tools will increase the number of required GDI handles. For SETM this should not be a real issue, but for METM it may reduce the number of editors a user can open.

heckflosse:
@Lawrence37 I have a suggestion. In my preferences dialog it looked like this

grafik

Then I added some tools

grafik

Then I wanted to change the order of the tools (e,g, put Raw Black Points at the end). To do that I have to use the scroll bar or widen the preference dialog

grafik

Would it be possible to change this?

grafik

Imho this would reduce the questions from users we have to answer because they don't find the move up/down or remove buttons.

Anyway. Thanks a lot for this work. Really appreciated 👍

Lawrence37:
> the duplication of tools will increase the number of required GDI handles

The tools aren't actually duplicated, just moved from tab to tab. Am I correct to think that the number of GDI handles is not affected?

> Would it be possible to change this?

Yes, I will move the buttons to the middle. Your screenshots alerted me to the fact that I used DPI scaling while setting the column widths. I will need to readjust them.

heckflosse:
> The tools aren't actually duplicated, just moved from tab to tab.

Oh, cool 👍

heckflosse:
@Lawrence37 : LGTM 👍

Floessie:
Builds fine with GCC 11.2 and Clang 13.0 on Debian Unstable PPC. :+1:

Lawrence37:
I think I fixed the theme issues.

Lawrence37:
Here's some documentation for editing the Favorites tab in preferences. These two pages will need to be updated.
- http://rawpedia.rawtherapee.com/Preferences
- http://rawpedia.rawtherapee.com/Favorites_Tab

## Favorites
This tab allows you to choose which tools appear in the Favorites tab of the editor toolbox.

### Available Tools
Checkboxes indicate which tools and sub-tools are in the Favorites tab. Only tools that can be marked as a favorite will have a checkbox. Toggling a checkbox adds or removes the corresponding tool to/from the Favorites tab.

### Favorites Panel
Favorite tools are listed here in the same order as they will appear in the Favorites tab. Selecting a tool from the list and clicking the up or down button located to the left of the list moves the tool up or down the list. It is also possible to rearrange tools by dragging them. Tools can be removed from favorites by selecting them and clicking the remove button. Use the Ctrl key to select multiple tools or the Shift key to perform a range selection.

### Options
- **Keep favorite tools in original locations**: If set, favorite tools will appear in both the favorites tab and their original tabs.  Enabling this option may result in a slight delay when switching tabs.

---

This PR is targeted for 6.0, so I will wait until after 5.9 has been tagged before merging (unless there is agreement that this should go in the upcoming release).

Lawrence37:
This PR is good to go and seems to work well with the current `dev` branch. It will cause conflicts with the global tone equalizer (#6641) due to differences in how tools are added to the tool panel, but I will resolve them.

ghost:
I just logged to mention this and see that GitHub picked up on the unresolved conflict (you don't see this message when not logged in). Not sure if there's an auto-update thingy send to you, hence this reply.

This conflict showed up after this PR was merged on  Jan 31st: #6673 

> Auto-merging rtgui/labcurve.cc
> CONFLICT (content): Merge conflict in rtgui/labcurve.cc
> Auto-merging rtgui/icmpanel.h

BTW: Is there a reason why this one hasn't been merged yet? It is tagged for 5.10 and there don't seem to be any issues with it (have been using it ever since you created it roughly a year ago).

Floessie:
Builds fine with GCC 12.1 and Clang 15.0 on Debian Unstable PPC. :+1:

ghost:
@Floessie : Interesting. 

I'm curious how merging in PPC differs from what GitHub and I do:

Screenshot_2023-02-02_12-47-12

I always assumed that merging files is the same allover.

Floessie:
@jade-nl I was building the current `favorites-gui` branch, not a merged one. Sure, `dev` should be merged to `favorites-gui` again and its conflicts resolved...

Lawrence37:
@jade-nl There were no comments or reactions since my last commits and I didn't want to take unilateral action by merging to dev. Now that there has been plenty of time to raise concerns, I'm comfortable with merging this within the next week.

Lawrence37:
> A graphical user interface for easily marking tools as "favorited" is being worked on, but due to time constraints related to the release date, this GUI is not ready for 5.6 and will feature in a future release.

A bold prediction on RawPedia :slightly_smiling_face:

Floessie:
Thank you, @Lawrence37! :+1:

Issue #6386, opened 2021-11-30
Faster dual-demosaic/capture sharpening blend mask


heckflosse:
Dual demosaic and Capture sharpening build a mask to blend between demosaic algorithms and before/after sharpening.

The mask is built using a sigmoid function which uses the expensive e^x function.
I made a try to use a less expensive sigmoid function, which uses 1/sqrt(x), for which at least in SSE code I could use the fast reciprocal of sqrt. Eyeballing the results they are equal. Only when looking at the differences between both methods by using an extreme brightness curve I can (hardly) see very small differences.

Here's a plot of both functions (keep in mind that the result will be blurred with a gaussian sigma of 2.0 afterwards anyway)

![grafik](https://user-images.githubusercontent.com/13733487/144044727-b9817c75-35ec-430d-a12a-ce9d949fd170.png)

Speedup for a 45.6 MP file on my machine is from 178 ms to 118 ms (timings include the gaussian blur)

PR follows...

Edit: I forgot to mention that red is the old expensive 1 / ( 1 + e^-x) and blue is the new faster x/sqrt(1 + x* x)

PR #6387, opened 2021-11-30
faster sigmoid function to create the contrast mask used in dual demo…

heckflosse:
…saic and capture sharpening, #6386

### DIFF
--- a/rtengine/rt_algo.cc
+++ b/rtengine/rt_algo.cc
@@ -35,13 +35,23 @@
 
 namespace {
 
-template<typename T>
-T calcBlendFactor(T val, T threshold) {
+float calcBlendFactor(float val, float threshold) {
     // sigmoid function
     // result is in ]0;1] range
     // inflexion point is at (x, y) (threshold, 0.5)
-    return 1.f / (1.f + xexpf(16.f - (16.f / threshold) * val));
+    const float x = -16.f + (16.f / threshold) * val;
+    return 0.5f * (1.f + x / std::sqrt(1.f + rtengine::SQR(x)));
 }
+
+#ifdef __SSE2__
+vfloat calcBlendFactor(vfloat val, vfloat threshold) {
+    // sigmoid function
+    // result is in ]0;1] range
+    // inflexion point is at (x, y) (threshold, 0.5)
+    const vfloat x = -16.f + (16.f / threshold) * val;
+    return 0.5f * (1.f + x * _mm_rsqrt_ps(1.f + rtengine::SQR(x)));
+}
+#endif
 
 float tileAverage(const float * const *data, size_t tileY, size_t tileX, size_t tilesize) {
 


Desmis:
@heckflosse 
Ingo
I just tested on 2 images from Z6-2 with  - "Amaze + VNG4" and contrast threshold about 25


Gain in processing time - about 5 to 10%
No visible difference in TIFF
Very small differences in layers difference 16 bits

Good improvment :)

Jacques

Floessie:
@heckflosse Somehow the GitHub interface for reviews has changed or I'm too dumb to find the right button... Nonetheless, I'd like to approve the change code-wise. I test-built it on PPC with Clang 13: not worse than before. :wink:

Floessie (approved):
Found the button at last. :facepalm:

Thanatomanic:
Hi @heckflosse, to alleviate the mismatch between the original function and the faster one, you might want to consider to change the factor 16.0 to something more around 9.58. Over the entire range of threshold values, this is what my simple find-fit function gives me as a pretty decent fitting parameter. As you already predicted, the behavior near the edge cases is different.

image

heckflosse:
> to change the factor 16.0 to something more around 9.58

The factor 16.0 was (is) just an optimization thing, see here, where 0.0625 = 1 / 16

https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/rt_algo.cc#L105

heckflosse:
@Thanatomanic I have to think a bit more about your comment...

heckflosse:
@Thanatomanic 

I don't understand what you mean, because you refer to the threshold.

The current formula is:

`x = -16.f + (16.f / threshold) * val`

If `threshold == val` that leads to x = 0, which leads to a value of 0.5 in both sigmoid functions.

Thanatomanic:
@heckflosse I tried to find a different value for the `16.f` in order to make the two curves match each other more closely (least-squares minimization). If the 16 has a specific origin, no change is necessary of course, I cannot judge that.

My 'more optimal' suggestion would be to use
`x = -9.58f + (9.58f / threshold) * val`
Which still gives `x=0` when `threshold = val`.

Thanatomanic:
An example of how the curve matches more closely (threshold = 0.8)
image

threshold = 0.4
image

Or if you want to play yourself: https://www.desmos.com/calculator/gdrzan4ubk

heckflosse:
@Thanatomanic 

Ok, now I know what you mean. Factor 9.58 matches better in the mid range, but it converges worse against 0 or 1

heckflosse:
@Thanatomanic Honestly I can not say, which of the sigmoid functions is better suited for the purpose.
`Original` was already an arbitrary choice in the sense that it is one of a lot of possible sigmoid functions.

Thanatomanic:
Sure, it may not matter at all in the end. Feel free to keep it as is, I just had some fun with maths :-)

heckflosse:
> I just had some fun with maths

same here :-)

jenshannoschwalm:
I cross checked this in dt before doing the pr. I would think a "as close as possible to 1 / 0 in the flat regions" is desired for minimal blending cross-over. So i would opt for taking 16.

PR #6406, opened 2022-01-11
Mac: set fftw3f_omp linker flags

Benitoite:
Conditionally sets linker flags for macOS for fftw3f_omp when OPTION_OMP==ON

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -487,12 +487,11 @@ endif()
 
 pkg_check_modules(LCMS REQUIRED lcms2>=2.6)
 pkg_check_modules(EXPAT REQUIRED expat>=2.1)
+pkg_check_modules(FFTW3F REQUIRED fftw3f)
 
 #Set the appropriate FFTW flags on macOS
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -487,7 +487,14 @@ endif()
 
 pkg_check_modules(LCMS REQUIRED lcms2>=2.6)
 pkg_check_modules(EXPAT REQUIRED expat>=2.1)
-pkg_check_modules(FFTW3F REQUIRED fftw3f)
+
+#Set the appropriate FFTW flags on macOS
+if(APPLE AND OPTION_OMP)
+        set(EXTRA_LIB "-L${LOCAL_PREFIX}/lib -lfftw3f_omp -lfftw3f -lm")
+else()
```
Thanatomanic: You don't need to check for the existence of `fftw3f` in general on macOS?

Benitoite: Thanks for taking a look... revised to check existence.

PR #6413, opened 2022-01-23
Automated Windows builds

Lawrence37:
This uses GitHub Actions to perform release and debug builds. Builds are automatically triggered when a new tag is created, the `dev` branch is updated, or a pull request targeting `dev` is updated. A build can also be manually triggered. Release builds for tags and `dev` also have an installer. Debug builds come with GDB. Signed-in GitHub users can download any builds from the Actions tab. The latest tag and `dev` builds will be publicly available at [nightly-github-actions](https://github.com/Beep6581/RawTherapee/releases/tag/nightly-github-actions).

These automated builds can potentially replace the official ones on Travis CI. I don't anticipate any issues that would require significant re-writing of the build workflow.

I updated the Inno Setup file to reflect their current [list of officially maintained languages](https://jrsoftware.org/files/istrans/).

The list of DLLs in the bundle is slightly different from the list on RawPedia due to missing library errors. We should update the RawPedia page accordingly.


Thanatomanic:
Excellent @Lawrence37! 

> Release builds for tags and dev also have an installer.

Does this effectively mean we can move to something of a rolling release? And seeing as this is unfamiliar territory for me, are there foreseeable complications when using the installer to install a nightly build over an existing installation? E.g. potential for orphaned files in the installation directory when things are replaced/deprecated?

> We should update the RawPedia page accordingly.

Do you have editing rights? If not, could you please indicate which should be removed?

Lawrence37:
The official builds also create installers for the release and dev versions. I'm not sure if installing over older versions is a good idea or not. I believe it's recommended to uninstall the old version first.

I don't have write access to RawPedia. I changed libffi-6 to libffi-7 and added libbrotli* among others (some related to file formats I think). I would do a diff between the two lists.

Lawrence37:
These are the dependencies.

New:
```
libbrotlicommon.dll
libbrotlidec.dll
libdeflate.dll
libjbig-0.dll
libLerc.dll
libwebp-7.dll
```

Removed:
```
libcroco-0.6-3.dll
```

Updated:
```
libffi-6.dll -> libffi-7.dll
```

Thanatomanic:
RawPedia has been updated. If there are no more comments here, feel free to merge this important step back on the rails @Lawrence37!

Beep6581:
Fantastic!

PR #6414, opened 2022-01-24
A Black, Coloured, Grey and a White border preset

Desmis:
This pull request has been posted on behalf of @Jade_NL

A Black, Coloured, Grey and a White border preset.

This PR, an original idea of @arturoisilvia, consists of four Black,  Coloured, Grey and White border presets to be included in the existing Bundled Profiles (/Misc).
    
Jacques (@jade-nl) will explain how this one can be used in a separate reply.
Thanks to @arturoisilvia

Jacques


Thanatomanic:
Applying a border by using Local Adjustments seems a novel thing to do, though to me complete overkill and inappropriate usage of the module. So I'm really curious for more explanation.

It will also always cut away part of your image, no?

ghost:
Just to set the record straight: This PR **was not**  posted on my behalf! I did just about everything to keep my name out of this.

I'm somewhat busy at the moment and will post some info on how this works and how to use it later.

@Thanatomanic 
> It will also always cut away part of your image, no?

Yes, it does. I think that is also true for other RAW editors that have a dedicated facility to do this.

Thanatomanic:
> Yes, it does. I think that is also true for other RAW editors that have a dedicated facility to do this.

darktable adds it outside https://darktable-org.github.io/dtdocs/en/module-reference/processing-modules/framing/

ghost:
_In general:_

I agree with @Thanatomanic that this is a novel way to use Local Adjustments, a bit of a poor man's way of creating borders in RawTherapee.

Not sure if this is really overkill or inappropriate use. There is at present no functionality in RawTherapee that can create borders and people, often being lazy, might not want to use a third party program to do this. I personally use a flexible script written around convert (an ImageMagick tool) to create inner and outer borders.

For me this is just a way to see what you can do with the tools at hand and learn a lot in the process about those tools. They can often do more then expected at first glance.

This way of creating borders came up in a three-way conversation and @Desmis decided to push it as a PR.

_About the profiles:_

They are Local Adjustments based and use 2 spots: the first spot being *Color & Light* using a Full Image method and the second an excluding spot using a Rectangle.

The black, grey and white profiles are basically the same. The only difference being the position of the horizontal line in the *Color & Light -> Curves -> RGB Tone Curve* part

The colour profile works differently: It also uses the *Color & Light* tool, but only uses the tools that are used during, what most would consider normal operation. The sliders are set to their extremes, though. The colour itself can be set/changed by using the *Color correction grid*

Both of the above methods fully change the image to the chosen colour.

The second spot makes sure that this colour is excluded except for the border part.

Because *Shape method* is set to *Symmetrical (mouse + sliders)* in the *Specific cases* section, new sliders became accessible. These, *Right* for the verticals and *Bottom* for the horizontals, can be used to set the wanted border width.

As mentioned earlier, this does remove the outer part of the image.

Desmis:
@Thanatomanic 

The use that is made of LA here, is not a marginal use, but totally corresponds to the conception I made of it

If there were masks and layers in RT, everyone would find their use normal to make these “borders”.
Each Spot can be considered (I explained it in part in the French documentation) as a layer and their superposition is similar to Photoshop’s blending modes. In fact each spot corresponds, but seen in another way, to a mask/layer couple.

On the other hand, if the collective obviously doesn't like this PR, I will quickly close it.

Jacques

Thanatomanic:
After giving the profiles a testrun, I have a few additional remarks:
1. The coloured profile has a different default 'border size' than the others. Could be easily fixed I guess.
2. The borders are not uniform in size (top/bottom vs. left/right). Due to the nature of the LA spots, the sizes will behave differently for images with different aspect ratio's and orientations.
3. Because crop comes after LA in the pipeline, the border completely disregards cropping. You will have to set the spot size and position manually after you choose your crop. This is very inconvenient if you just want to add a 100px border on all sides.

In general I still think that LA is not the right tool for the job and these profiles are not worth including in RT. It would be much more worthwhile to have a dedicated tool for this that works at the very end of the pipeline. But let's hear some other opinions. 

> If there were masks and layers in RT, everyone would find their use normal to make these “borders”.

I would never consider to use a mask or layer to add a border to my image. GIMP, Photoshop, even IrfanView, all have tools to add a border or increase canvas size.

ghost:
@Thanatomanic 

_Point 1:_  Easily fixed, but you are correct. They should be the same when first loaded.
_point 2:_  Aspect ratios and also landscape vs portrait are not in any way taken into consideration. These are starting points at bests and need changing according to the image.
_point 3:_  It is a poor man's solution at best........

> In general I still think that LA is not the right tool for the job and these profiles are not worth including in RT. It would be much more worthwhile to have a dedicated tool for this that works at the very end of the pipeline. 

This, especially the second part, and the other points you've made are rather fair.

Desmis:
Certainly the LA solution is imperfect, even if point 1 can be easily improved. It allows everyone to customize if he wants, width, height, color of the borders,  and then use with partial paste, and solves a majority of cases.

However it seems that it is better to have no solution than an imperfect solution.

So, unless other opinions come in I'll close this PR tomorrow.

Jacques

PR #6418, opened 2022-01-31
Allow really tiny spots for removal

sobukus:
I may be a very picky person, but I want to repair really tiny specks of dust on my negative scans.
I tried setting the SpotParams::minRadius to 1 and this seems to work fine. I can zoom in in the
UI and still manipulate things with point and click. I guess 1 is a natural lower boundary for the
radius that should be explored;-)

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -365,7 +365,7 @@ namespace rtengine
 namespace procparams
 {
 
-const short SpotParams::minRadius = 5;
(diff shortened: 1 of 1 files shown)


sobukus:
My thoughts on thin scratches and tiny spots could be of relevance here: https://github.com/Beep6581/RawTherapee/issues/6419  While I do maintain that the artificial limit of radius of 5 needs justification I don't see yet, a different algorithm for these very small features might be more appropriate.

Issue #6417, opened 2022-01-30
Film Simulation ComboBox does not cope well with long HaldCLUT names


Bezierr:
This is a minor GUI issue that I encountered while fighting with the Film Simulation tool.

If the name of at least one of the HaldCLUTs found is too long to fit into the right panel, all names are truncated on the left side, which somewhat messes up the GUI and may leave the names illegible. See "Before" attachment.
![Film Simulation Before](https://user-images.githubusercontent.com/38507161/151713404-83fe1f71-6f38-4de4-aac5-f2edff5b511e.JPG)
It seems that a small change can fix that. See "After" attachment.
![Film Simulation After](https://user-images.githubusercontent.com/38507161/151713423-c598965c-2ac1-4ad9-b275-4aa7526caa35.JPG)
I'm going to propose a pull request that seems to fix this issue one of these days. Of course, it will be up to you to decide whether you prefer the old or the new look.

Observed on
RT 5.8 release and newer
Windows 10 20H2

PR #6423, opened 2022-02-04
Film Simulation ComboBox for long HaldCLUT names

Bezierr:
I propose a fix for issue #6417.

As indicated in the comment in the code, I found this solution on stackoverflow. I don't know if this is the best possible solution, but it seems to work.

Feel free to not use this fix if there is a reason why the current behavior is like it is.

### DIFF
--- a/rtgui/filmsimulation.cc
+++ b/rtgui/filmsimulation.cc
@@ -64,9 +64,6 @@ FilmSimulation::FilmSimulation()
     :   FoldableToolPanel( this, "filmsimulation", M("TP_FILMSIMULATION_LABEL"), false, true )
 {
     m_clutComboBox = Gtk::manage( new ClutComboBox(options.clutsDir) );
-	// The following line, inspired by https://stackoverflow.com/questions/43414163/gtk-mm-limit-width-of-combobox,
-	// makes the Film Simulation ComboBox look better if there are HaldCLUT names that are too long to fit into it
-	((Gtk::CellRendererText*)m_clutComboBox->get_first_cell())->property_ellipsize() = Pango::ELLIPSIZE_END;
 
     int foundClutsCount = m_clutComboBox->foundClutsCount();
 
@@ -222,7 +219,13 @@ ClutComboBox::ClutComboBox(const Glib::ustring &path):
     set_model(m_model());
 
     if (cm->count > 0) {
-        pack_start(m_columns().label, false);
+		// Pack a CellRendererText in order to display long Clut file names properly
+		Gtk::CellRendererText *renderer;
+
+		renderer = new Gtk::CellRendererText;
+		renderer->property_ellipsize() = Pango::ELLIPSIZE_END;
+		pack_start(*renderer, false); 
+		add_attribute(*renderer, "text", 0);
     }
 
     if (!options.multiDisplayMode) {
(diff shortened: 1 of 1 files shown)


rtgui/filmsimulation.cc
```diff
@@ -64,6 +64,10 @@ FilmSimulation::FilmSimulation()
     :   FoldableToolPanel( this, "filmsimulation", M("TP_FILMSIMULATION_LABEL"), false, true )
 {
     m_clutComboBox = Gtk::manage( new ClutComboBox(options.clutsDir) );
+	// The following line, inspired by https://stackoverflow.com/questions/43414163/gtk-mm-limit-width-of-combobox,
+	// makes the Film Simulation ComboBox look better if there are HaldCLUT names that are too long to fit into it
+	((Gtk::CellRendererText*)m_clutComboBox->get_first_cell())->property_ellipsize() = Pango::ELLIPSIZE_END;
```
Lawrence37: Rather than cast the result of `get_first_cell()` into a `CellRendererText*`, it would be better, if possible, to explicitly create a `CellRendererText`, set its ellipsize property, and add it to the combo box.

Currently, we are packing the model column.
https://github.com/Beep6581/RawTherapee/blob/bab5516871ebe0eca63e9f97689468f6de0a818c/rtgui/filmsimulation.cc#L221
You can replace the column with a new `CellRendererText` and set its text property, with [add_attribute](https://developer-old.gnome.org/gtkmm/3.16/classGtk_1_1CellLayout.html#ae5f84edaa30e5c586ca9266961bb4f96) for example.

Lawrence37:
Thanks @Bezierr for this PR. I left a suggestion for making your fix less prone to problems that may come from casting pointers.

Bezierr:
Hi,
Thanks for your comments.

Sorry didn't have much time recently so have only now managed to look at it. Have followed your suggestions (I think). Is there a way to update this pull request, or do I just close it and start a new one?

Bezierr:
Ok, just noticed that the pull request is updated automatically :-)

rtgui/filmsimulation.cc
```diff
@@ -222,7 +219,13 @@ ClutComboBox::ClutComboBox(const Glib::ustring &path):
     set_model(m_model());
 
     if (cm->count > 0) {
-        pack_start(m_columns().label, false);
+		// Pack a CellRendererText in order to display long Clut file names properly
+		Gtk::CellRendererText *renderer;
+
+		renderer = new Gtk::CellRendererText;
```
Lawrence37: Wrap this in a [`Gtk::manage()`](https://developer-old.gnome.org/gtkmm/3.16/namespaceGtk.html#a753d0ddca83bd0d0decdfdcd1e93076a).

Lawrence37 (commented):
Once my comment gets addressed, this PR has my approval. Since this is a small change, I think it can go into 5.9.

Bezierr:
Ok. I understand the intention, but given that this seems to be only instantiated once at the start of RT, I don't really see why it is necessary. Better style? "Just in case"? Or am I missing something?

Anyway, it can't harm, either.

Lawrence37:
It's instantiated multiple times in multiple editors mode. Not a big issue, but the principle is important. If one widget is not managed here, why manage other widgets elsewhere? Pretty soon it adds up and we get horrible memory leaks.

Lawrence37 (approved):

Floessie:
@Bezierr Just a quick note: Could you compact this

```c++
Gtk::CellRendererText *renderer;

renderer = Gtk::manage(new Gtk::CellRendererText);
```

to this

```c++
Gtk::CellRendererText* const renderer = Gtk::manage(new Gtk::CellRendererText);
```

?

Best,
Flössie

Bezierr:
@Lawrence37: Thanks; I can understand that.
@Floessie: Done.

Issue #5670, opened 2020-02-27
Regression in lens identification


Barugon:
Version: 5.8 AppImage for Linux.
OS: Fedora 31, x86_64.

The lens EXIF shows as "Unknown".

![lens_exif](https://user-images.githubusercontent.com/16503728/75406319-07bf4a80-58c5-11ea-97ca-85c5ef8b976d.png)

This seems to cause Profiled Lens Selection to not be able to automatically find the correct lens (even thought it's in the list).

![lens_select](https://user-images.githubusercontent.com/16503728/75406700-ff1b4400-58c5-11ea-93b5-3f92af715fd5.png)

You can get a raw file [here](https://www.dpreview.com/reviews/image-comparison/download-image?s3Key=c44354db78824e85b37b34253e5d593b.arw) that will show this issue.

Beep6581:
```text
$ exiftool -G5 -s -args ~/downloads/a7M3Studio0008.ARW | grep -E "Lens(Info|Model|Type|Spec|ID)"
-TIFF-IFD0-ExifIFD:LensInfo=85mm f/1.8
-TIFF-IFD0-ExifIFD:LensModel=FE 85mm F1.8
-TIFF-IFD0-ExifIFD-MakerNotes:LensType=E-Mount, T-Mount, Other Lens or no lens
-TIFF-IFD0-ExifIFD-MakerNotes:LensSpec=FE 85mm F1.8
-TIFF-IFD0-ExifIFD-MakerNotes-Tag940c:LensType3=Sony FE 85mm F1.8 or Viltrox PFU RBMH 85mm F1.8
-TIFF-IFD0-ExifIFD-MakerNotes-Tag9050b:LensType2=Sony FE 85mm F1.8 or Viltrox PFU RBMH 85mm F1.8
-TIFF-IFD0-ExifIFD-MakerNotes-Tag9050b:LensSpecFeatures=FE
-Composite:LensID=Sony FE 85mm F1.8
```

The lens metadata is in RT:
```
$ grep -rni viltrox rtexif/sonyminoltaattribs.cc
1059:        choices.insert (p_t (32796, "Sony FE 85mm f/1.8 or Viltrox PFU RBMH 85mm f/1.8"));
```

aferrero2707:
@Beep6581 Is this a problem that needs to be investigated? Could this be due to the packaging process for the AppImage?

Barugon:
I switched to Fedora's package and it's still a problem.

[...]

PR #6437, opened 2022-03-17
Fix so E-Mount lens names are retrieved

Bezierr:
This is related to issue #5670 though I don't know whether it can be considered a definite fix or not.

Problem: Sony E-Mount lenses are not retrieved correctly. Instead, something like "Unknown (28mm f/2.0)" is returned (see attached).

The reason for this apparently is that for E-Mount lenses Sony uses a "lenstype2" mechanism which is not implemented in RT. Instead, RT tries to retrieve a lens with lens ID 65535 (which is usually -- always? -- assigned to E-Mount lenses), finds more than one, can't decide which one is correct, and returns this "Unknown ..." string as lens name. A fallback mechanism to get the lens model from the exif tag is implemented but checks for a string _equal _to__ "Unknown" instead of _beginning _with__ "Unknown", and fails.

The change I propose here fixes that.

----------------------------------
I personally don't like this Makernotes-based mechanism at all. It is apparently undocumented and may change, and apparently has in the past, with any new camera. The way it is implemented in RT, with hard-coded lens lists, also means that new lenses can only be added with new releases of RT, which haven't been that frequent recently.

So, in my locally compiled version of RT, I bypass this mechanism and immediately read the lens model tag, just as it is done for Fujifilm lenses. I know that for my lenses this works; however, I can't be sure that this is _always_ the case. Which is why I propose a different solution above that seems to be more in line with the original developer's intention.

Capture
.

### DIFF
--- a/rtengine/imagedata.cc
+++ b/rtengine/imagedata.cc
@@ -557,7 +557,7 @@ FrameData::FrameData(rtexif::TagDirectory* frameRootDir_, rtexif::TagDirectory* 
                 } else if (!make.compare (0, 4, "SONY") || !make.compare (0, 6, "KONICA")) {
                     if (mnote->getTag ("LensID")) {
                         lens = validateUft8(mnote->getTag("LensID")->valueToString());
-                        if (lens == "Unknown") {
+                        if (!lens.compare (0, 7, "Unknown")) {
                             lens_from_make_and_model();
                         }
                     }


Entropy512:
Heh, I was just about to start looking into this myself - https://discuss.pixls.us/t/a-bit-confused-about-auto-matched-camera-profile-amcp-option/29836/36?u=entropy512

Longterm solution is a move to exiv2, but short term (before 5.9), that's probably up to @Thanatomanic

Thanatomanic:
This is fine for now as a fix, thanks @Bezierr

PR #6442, opened 2022-03-25
Handle linear DNG as other raw images

### LINKED ISSUE #6441: Highlight reconstruction with color propagation/Retinex does nothing for Linear DNGs
**Short description**
Highlight reconstruction with color propagation or Retinex does nothing for Linear DNGs

**Steps to reproduce**

1. Open a linear dng (PhotometricInterpretation:  Linear Raw)
2. Enable `Highlight reconstruction` with `color propagation` or `Retinex`.
3. Nothing happens.

**Expected behavior**
 `Highlight reconstruction` with `color propagation` and `Retinex` works like other raw images.

sgotti:
RawTherapee already handle linear DNGs as a RawImageSource but the red,
green, blue RawImageSource arrays aren't populated.
So operations that rely on them like Highlight recovery with color
Propagation and Retinex have no effect.

This patch populates the above arrays by calling nodemoasic for images
that have 3 color and not a CFA. Then these channels will be used by
RawImageSource::getImage like done with CFA or monochrome images

Fixes #6441

Question: Is there a test suite to check that this PR doesn't causes some regressions?

### DIFF
--- a/rtengine/demosaic_algos.cc
+++ b/rtengine/demosaic_algos.cc
@@ -878,7 +878,7 @@ void RawImageSource::nodemosaic(bool bw)
         for (int j = 0; j < W; j++) {
             if (bw) {
                 red[i][j] = green[i][j] = blue[i][j] = rawData[i][j];
-            } else if(ri->getSensorType() != ST_FUJI_XTRANS) {
+            } else if(ri->getSensorType() == ST_BAYER) {
                 switch( FC(i, j)) {
                 case 0:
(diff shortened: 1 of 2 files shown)


Thanatomanic:
For reference: according to [this](http://www.barrypearson.co.uk/articles/dng/linear.htm) it is not required that a linear DNG has three channels, but your code assumes that.

I am happy to merge this though, since we do not support non-RGB camera's anyway.

PR #6446, opened 2022-03-27
Logo fixes

Thanatomanic:
Today I found out that the RawTherapee logo has some improperly matched geometry.
image

The segments are also not symmetric and properly aligned:
image

I fixed this by recreating the logo from the ground up. It should have fully symmetric geometry now. 
All the changes have been propagated to wherever the logo is used.

Additionally:
1. I move the `rt-logo.svg` to the main `images` folder, instead of the `svg` subfolder, because it is a template and not converted to a png.
2. Two additional files `rt-logo-text-black.svg` and `rt-logo-text-white.svg` are added to the repo for whenever the logo + text is needed (e.g. on the website banner).
3. Finally, there was an `xcf` file containing all sized of the logo icon which sole purpose seemed to be to generate the `rawtherapee.ico` file. I have removed it, as creating the file is trivial from the logo versions in `images\png` (GIMP > Open as Layers).

_Note_: curiously, the font that is being used in the logo for RT is not open source. In fact it is a commercial font, ITC Eras Std. I think a subset of this font may be freely available, but this doesn't seem to be the case for the Ultra-Bold variant. I am really surprised this was chosen a few years ago when the redesign was made...


Lawrence37:
@Thanatomanic `rt-logo.svg` seems to be missing and the new splash is slightly bigger than the old one.

Thanatomanic:
@Lawrence37 Thanks for noticing! I just added the missing files.

Do you think the slightly bigger splash screen causes issues?

Lawrence37:
I don't know. The About dialog will be slightly bigger, I think. The font size is 60pt in both the new and old splash, but the output has them at different sizes. How do we know which one is correct?

Thanatomanic:
As long as nobody complains... 😉

Desmis:
@Thanatomanic  @Lawrence37 

I tested on windows 8, all seems to work fine :)

jacques

PR #6456, opened 2022-04-12
Remove GDB from Windows debug build

### LINKED ISSUE #6455: Automated debug builds fail
@Lawrence37 Perhaps you have seen it too, but the automated debug builds fail on the following:

image

Seems a bit weird to me. Any clues?

Lawrence37:
*waiting for build to complete*

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -89,26 +89,6 @@ jobs:
           echo "Copying Lensfun database to the build directory."
           cp -R "/C/msys2/msys64/mingw64/var/lib/lensfun-updates/version_1" 'build/${{matrix.build_type}}/share/lensfun'
 
-      - name: Restore GDB from cache
-        id: gdb-cache
-        if: ${{matrix.build_type == 'debug'}}
-        uses: actions/cache@v2
-        with:
-          key: gdb-windows-64
-          path: gdb-windows-64.exe
-
-      - name: Download GDB
-        if: ${{matrix.build_type == 'debug' && steps.gdb-cache.outputs.cache-hit != 'true'}}
-        run: |
-          echo "Downloading GDB."
-          curl --location 'http://www.equation.com/ftpdir/gdb/64/gdb.exe' > 'gdb-windows-64.exe'
-          echo "Verifying download."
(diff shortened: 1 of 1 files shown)


Beep6581:
Hi @Lawrence37 

What was the reason for removing `gdb.exe`, and what is the alternative?

Lawrence37:
@Beep6581
The build process used to download GDB and verify its checksum. The site publishing GDB regularly updates the binaries which causes the verification to fail periodically. As discussed in the linked issue, we decided to remove GDB and have people use their own copies. If bundling the debugger is preferable, then an alternative download site can be used (one where we can control which version to download). I recall RawPedia has a link to such as site.

Beep6581:
http://rawpedia.rawtherapee.com/How_to_write_useful_bug_reports#Step_by_step

> Our debug Windows builds usually include gdb.exe, so you don't need to do anything to get it. If they don't,

@Lawrence37 we should remove that part.

The next part explains where to get GDB:

> install [TDM-GCC](http://tdm-gcc.tdragon.net/) making sure to install the GDB package as well.

Is that still the simplest way to get GDB in Windows?

Keep in mind that the instructions are meant to guide the user in the simplest way possible in doing something which is otherwise quite technical. Choices are confusing.

Lawrence37:
Going through the installation steps for TDM-GCC and copying gdb.exe isn't quite as simple as it could be. One can download gdb.exe directly from this page: http://www.equation.com/servlet/equation.cmd?fa=gdb. It's the same GDB bundled with the automated builds before this PR (and also the one from the days of Travis CI).

Beep6581:
@Lawrence37 who is Equation Solution, what is their relationship with GCC, why would we and RT's users trust them?

Lawrence37:
@Beep6581 I actually don't know. I got the [link](https://github.com/aferrero2707/rt-win64/blob/c0d5e4e4540d2a2b37349faec60efc6503b1da24/ci/package-msys2.sh#L109) from @aferrero2707 's Travis builds.

Beep6581:
I updated the documentation.
http://rawpedia.rawtherapee.com/How_to_write_useful_bug_reports#Step_by_step

Issue #6457, opened 2022-04-13
Removal of obsolete strings


Entropy512:
There's been some discussion of this over on pixls, but we should probably track it as its own issue.

I'm in the process of writing a tool to identify obsolete strings for removal, and have made some progress already on it.

Current version is here, I'll PR it as a WIP once I've made some more progress, right now I'm only targeting the history events table and there are already a decent number of findings:
```
#!/usr/bin/env python3

import clang.cindex
import subprocess

index = clang.cindex.Index.create()
procevents = index.parse('rtengine/procevents.h',args=['-x', 'c++'])

for chld in procevents.cursor.get_children():
    if(chld.displayname == 'rtengine'):
        for c in chld.get_children():
            if(c.displayname == 'ProcEventCode'):
                for pec in c.get_children():
                    #print(pec.kind, pec.displayname, pec.enum_value)
                    grp1 = subprocess.Popen(('grep', '-ro', '--exclude=procevents.h', pec.displayname), stdout=subprocess.PIPE)
                    wcr1 = subprocess.check_output(('wc', '-l'), stdin=grp1.stdout)
                    grp1.wait()
                    grp2 = subprocess.Popen(('grep', '-ro', '--exclude=procevents.h', '--exclude=refreshmap.cc', pec.displayname), stdout=subprocess.PIPE)
                    wcr2 = subprocess.check_output(('wc', '-l'), stdin=grp2.stdout)
                    grp2.wait()
                    print(pec.enum_value, pec.displayname,int(wcr1), int(wcr2))
```
[results.txt](https://github.com/Beep6581/RawTherapee/files/8482163/results.txt)

Grepping this for items with zero hits gives us
[unreferenced_in_source.txt](https://github.com/Beep6581/RawTherapee/files/8482169/unreferenced_in_source.txt)

So over 20 events that appear to be orphaned,  many of them orphaned for a few releases at this point

Next up is looking through `default` for unreferenced strings beyond the history events

Entropy512:
[results.txt](https://github.com/Beep6581/RawTherapee/files/8484971/results.txt)
Now includes a traversal of `default`

Grepped version at:
[orphans.txt](https://github.com/Beep6581/RawTherapee/files/8484976/orphans.txt)

Thanatomanic:
[...]

PR #6458, opened 2022-04-13
WIP:  Script to find orphaned strings

### LINKED ISSUE #6500: Local adjustments CIECAM-Model Mask is declared as „ZCAM“
May be i am totally misunderstanding this function….

Using the CIECAM module in local adjustments, you can activate and  use the mask as in each tool of LA.
But:
any change here is shown as a ZCAM-function in the history. 
The C-curve shows up as ZCAM brightness, 
The L-Curve  shows up as ZCAM contrast J
The LC(h)-curve shows up as  ZCAM contrast Q
The contrast-curve shows up as ZCAM chroma

Is ZCAM used as a mask here? If so, are the designations really correct 
If not, where to find the meaned ZCAM functions?

Version: 5.8-3096-gf86a45528
Branch: dev
Commit: f86a45528
Commit date: 2022-06-10
Compiler: cc 12.1.0
Processor: generic x86
System: Windows
Bit depth: 64 bits
Gtkmm: V3.24.6
Lensfun: V0.3.3.0
Build type: release
Build flags:  -std=c++11 -fno-tree-loop-vectorize -mtune=generic -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:  -mtune=generic
OpenMP support: ON
MMAP support: ON
Build OS: MINGW64_NT-10.0-20348 3.3.5-341.x86_64 x86_64
Build date: Fri, 10 Jun 2022 16:57:11 +0000 UTC
Build epoch: 1654880231
Build UUID: 533d5020-e1f8-46d6-9873-366a2b41db0a

Entropy512:
WIP script to support issue #6457

### DIFF
--- a/tools/findorphans.py
+++ b/tools/findorphans.py
@@ -1,0 +1,46 @@
+#!/usr/bin/env python3
+
+import clang.cindex
+import subprocess
+import sys
+
+index = clang.cindex.Index.create()
+procevents = index.parse('rtengine/procevents.h',args=['-x', 'c++'])
+
+if(1):
+    for chld in procevents.cursor.get_children():
+        if(chld.displayname == 'rtengine'):
+            for c in chld.get_children():
+                if(c.displayname == 'ProcEventCode'):
+                    for pec in c.get_children():
+                        #print(pec.kind, pec.displayname, pec.enum_value)
+                        #print(pec.displayname, file=sys.stderr)
+                        grp1 = subprocess.Popen(('grep', '-ro', '--exclude=procevents.h', '--exclude-dir=.git', pec.displayname), stdout=subprocess.PIPE)
+                        wcr1 = subprocess.check_output(('wc', '-l'), stdin=grp1.stdout)
+                        grp1.wait()
+                        grp2 = subprocess.Popen(('grep', '-ro', '--exclude=procevents.h', '--exclude=refreshmap.cc', '--exclude-dir=.git', pec.displayname), stdout=subprocess.PIPE)
+                        wcr2 = subprocess.check_output(('wc', '-l'), stdin=grp2.stdout)
+                        grp2.wait()
+                        print(pec.enum_value, pec.displayname,int(wcr1), int(wcr2))
+
+with open('rtdata/languages/default', 'r') as deflang:
+    for line in deflang:
+        if(line[0] == '#'):
+            continue
+        if(line[0:2] == '//'):
+            continue
+        if(line[0:2] == '/*'):
+            #our language files support comment blocks?????????????????????????????
+            #or is this commented block bogus?
+            continue
+        if(line[0:2] == '*/'):
+            continue
+        else:
+            stringid = line.split(';')[0]
+            if(stringid.startswith('HISTORY_MSG')):
+               continue
+            #print(stringid, file=sys.stderr)
+            grp1 = subprocess.Popen(('grep', '-ro', '--exclude-dir=languages', '--exclude-dir=.git', stringid), stdout=subprocess.PIPE)
+            wcr1 = subprocess.check_output(('wc', '-l'), stdin=grp1.stdout)
+            grp1.wait()
+            print(stringid, int(wcr1))


Entropy512:
I could use some input on some oddities I discovered, there seem to be multiple comment types, and I question the validity of some of them:

The `default` file starts with lines that are commented with #
There are a few lines with // comments - are these actually treated as comments, or just being ignored since nothing references the "commented" ID?
There is also a block of a few items that are commented with C style /* and */ - are these working at all like intended, or again, getting ignored?

I'm guessing the /* must be working, because there are duplicate entries between the commented HISTORY_MSG items and the ones that are not

Thanatomanic:
Only `#` and `!` style comments seem supported: https://github.com/Beep6581/RawTherapee/blob/b1e7860a234109761fb9e693e0c603cf0c58827f/rtgui/multilangmgr.cc#L187

The `//` and `/*` lines should get ignored like you say, because they don't form a key-value pair.

Entropy512:
Will have to do a deep-dive on that, // comments would create a mismatch, but I'll need to read through multilangmgr to figure out why this didn't break:

```
/*
HISTORY_MSG_1114;Local - ZCAM lightness
HISTORY_MSG_1115;Local - ZCAM brightness
HISTORY_MSG_1116;Local - ZCAM contrast J
HISTORY_MSG_1117;Local - ZCAM contrast Q
HISTORY_MSG_1118;Local - ZCAM contrast threshold
HISTORY_MSG_1119;Local - ZCAM colorfullness
HISTORY_MSG_1120;Local - ZCAM saturation
HISTORY_MSG_1121;Local - ZCAM chroma
*/
HISTORY_MSG_1113;Local - HDR PQ
HISTORY_MSG_1114;Local - Cie mask enable
HISTORY_MSG_1115;Local - Cie mask curve C
HISTORY_MSG_1116;Local - Cie mask curve L
HISTORY_MSG_1117;Local - Cie mask curve H
```
I am guessing that if a key occurs twice, the second value somehow overwrites the first match, but I'll have to dig through the code to confirm that.  Either way it looks like anything with // or /* really needs to be nuked

Entropy512:
https://github.com/Beep6581/RawTherapee/blob/b1e7860a234109761fb9e693e0c603cf0c58827f/rtgui/multilangmgr.cc#L201
If I'm reading that and https://www.cplusplus.com/reference/map/map/emplace_hint/ is correct (insertion only happens if the key is not already present), I think we're probably going to see some wrong behaviors.  Time to read the local adjustments manual and try to figure out how to trigger "CIE mask enable" and determine if it actually shows up correctly, or if I see ZCAM stuff.

Thanatomanic:
image

This should not show ZCAM lightness at all, but Cie mask enable. So there you have the first bug due to bad comments.

Entropy512:
That's what I suspected might happen, you confirmed it before I could get to that point in the UI.  :)

It's a lot easier to fix a bug when you find it by looking at the code and saying "that looks funny, how is it working?" and you confirm expected broken behavior.

I am curious as to how that particular section affects the existing script's behavior.  I think having an explicit sanity checker for duplicate keys would be beneficial to add to the script.  I've seen some of the translators concerned about duplicate strings - I wonder how often that is a problem, I'm going to poke at looking to see if any particular string in default is present for multiple keys.

Desmis:
All ZCAM code is disabled.

But, if users request it, it is possible to reactivate (GUI, rtengine, datas...), but, I warn in advance, the results are from my point of view (and other testers) bad. It seems that the algorithm provided by the researchers is not good (of course to be confirmed)
Which doesn't mean that I didn't make mistakes in assigning events


In this case, it is obvious that the programmer will have to review the "events".

jacques

Entropy512:
@Desmis - the issue as far as mistakes is that when someone (you?) tried to remove them, they used C-style block comments (/* and */) which are not actually valid for this purpose.

Since ZCAM is disabled/removed, I'm going to be removing that block of strings completely instead of moving it to proper commenting in a PR later this week.  If it ever comes back, it can be re-added with appropriate history event table numbering.

Desmis:
@Entropy512 

No problem for me :)

I think that in the current state of ZCAM (researchers), it is not "tomorrow" that Zcam will be incorporated into RT.

Thanks

Thanatomanic:
I just tried to run this on my Windows machine, but I run into trouble.

Running from the Mingw64 MSYS2 prompt does not seem to work at all - it does not seem to find the module clang, though I successfully completed `pip install libclang` as I think is required.

Doing so from the Windows cmd prompt gives the following traceback:
```
c:\msys64\home\Roel\RawTherapee>python tools/findorphans.py
Traceback (most recent call last):
  File "c:\msys64\home\Roel\RawTherapee\tools\findorphans.py", line 18, in <module>
    grp1 = subprocess.Popen(('grep', '-ro', '--exclude=procevents.h', '--exclude-dir=.git', pec.displayname), stdout=subprocess.PIPE)
  File "C:\Program Files\Python39\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Program Files\Python39\lib\subprocess.py", line 1420, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified
```

I assume the issue is with "subprocess.PIPE", but I am not sure at all. Not a Python wizard...

Edit: in any case, not super urgent...

Entropy512:
looks like it can't find grep?  Path issue combined with python's subprocess possibly?

Thanatomanic:
I see now, the script depends on a few Unix specific tools (grep, wc). I don't mind an OS-specific script, it just makes it harder for me to test 😄 

Edit: after making https://sourceforge.net/projects/grep-for-windows/ and http://gnuwin32.sourceforge.net/packages/coreutils.htm available on my system (add to PATH), I am able to run the script successfully from the Windows PowerShell.

Entropy512:
Since it's just a developer support tool, feel free to skip merging for 5.9 unless you want to rerun it for further sanity checks while I'm away.

I'll try to start taking a crack at gettext when I get back from vacation which would effectively obsolete all of the current translation-related tools anyway.

Thanatomanic:
Enjoy your trip @Entropy512! I think I'll just merge. I fully agree that we should move on to a more common and user-friendly system after 5.9.

PR #6460, opened 2022-04-20
Maintenance 5.8.x

pmjdebruijn:
I noticed the current RT release isn't particularly recent, so I tried cherry-picking most camera support enhancements, a few build fixes and a few bugfixes into a seperate branch.

This needs review, and still some testing. But maybe you would consider this to do a point release soonish...


Thanatomanic:
Hi Pascal @pmjdebruijn. Not to sound stupid here, or to devalue your effort to cherry pick this, but: how exactly is this different than preparing a full release in terms of the actual steps needed to get the release out? Our current hangup is not the lack of testing or functionality that needs polishing, it is getting things ready to really build, publish and communicate about the damn thing. So, I don't see much value in figuring out the same stuff for a minor version. But perhaps you have some other ideas?

Entropy512:
@Thanatomanic I think the only thing that I can see as a possible benefit here is that it does not touch rtdata/languages in any way, so the current main thing remaining on a TODO list for 5.9 does not apply here.

pmjdebruijn:
So I forgot a few things (see above)

@Thanatomanic I wasn't aware of what issues remain. And I'm still not entirely to be honest...

In a general sense, people may think a project has died when there haven't been any releases in a long time (minor or major), even if it really hasn't. And a new release (even if minor) may be a good place/time to sound a call for help (if needed)...

Entropy512:
https://discuss.pixls.us/t/the-problem-with-no-releases/29781 may help you as far as background on the current situation.

Edit:  The TL;DR is that while over the years many developers have been given increased roles and responsibilities, one person coordinated actual releases.  That person has basically disappeared without a word, only to reappear suddenly without explanation for why they were missing for months, only to disappear again once everyone thought they were finally back.

Others are trying to pick up the slack - for example I'm trying to make educated guesses as to exactly what is needed for "revising default" per https://github.com/Beep6581/RawTherapee/issues/5632#issuecomment-1006086423 since that was the last communication on the subject.

While it may not have been intentional, one benefit I personally see in your effort is that it is not hitting the languages directory, and thus in theory it can probably go forward without any involvement from the translators team, since translations appears to be one of the blockers for a 5.9 release.  But that's only my personal opinion, the decision is not mine.  For the moment I'm a fairly minor contributor, although I do intend on increasing that now that the pandemic seems to be heading in a good direction...  (Famous last words...  BA.2 is surging here now...)

pmjdebruijn:
I've avoided gui/lang changes semi-intentionally yes :)

Entropy512:
It's not my decision, but in some ways this might make a good way for a new release manager to practice/dry-run a release process without having to worry about that translations situation.

As far as camera support that reminds me I need to stop procrastinating on doing a StdA A7M4 colorchecker shot...

pmjdebruijn:
With regard to 18066755c7da76994bf5ff6a412d79e5571e433b were the 'or' bits removed in the past?

pmjdebruijn:
> With regard to [1806675](https://github.com/Beep6581/RawTherapee/commit/18066755c7da76994bf5ff6a412d79e5571e433b) were the 'or' bits removed in the past?

It seems those were included as it did, so that should be fine then. You'll probably want to pick that into the dev branch...

pmjdebruijn:
As far as I can tell, this branch should be "ready" now.

pmjdebruijn:
So I found another two :)

For arguments sake, say you'd consider this branch, what would need to happen to actually do a release?

pull my branch into the main repo (probably not into 'releases' as some cherries aren't exact 1:1 picks)
update RELEASE_NOTES.txt
tag 5.8.1
run tools/generateReleaseInfo
run tools/generateSourceTarball
(optionally) build AppImage
(optionally) build Windows version
(optionally) build OS X version
Update website
Announce on various places

Did I miss anything?

Entropy512:
Well, it's not my call, but I personally would have a concern regarding the amount of regression testing done (or not done) when cherry-picking/backporting this many commits back to an old baseline.

I believe the CI system should handle most builds now (possibly even macOS?), although the last time a release was done, it was not in the state that it is now, so I think that this might be the first time the CI system might be used for a release.

@Thanatomanic your thoughts?  Even if a 5.8.1 maint isn't done since an attempt is being made to get 5.9 out, since M is missing in action, thinking through what would be required for a release checklist is beneficial.

Semi-offtopic:  Sorry for not moving forward on the translations cleanup work.  Freak snowstorm that brought down trees everywhere + some minor health issues has had me a bit distracted the past week.

pmjdebruijn:
@Entropy512  In principal you're right of course (WRT regression). Practically speaking the risk probably isn't that big, given that most cherries (except a few) picked clean, and only a few patches were rather large. And the risk of regression could easily be managed by releasing a release candidate first, and after that waiting a week or two before final release.

pmjdebruijn:
And something to consider, not having had a release a long time, it's worse for raw conversion software than most other software. Say if GIMP were not to have a new release, the old version would for some time remain as functional as it has always been. However for raw converters, if someone buys a new camera, suddenly the raw converter stops being useful entirely.
Which is why I see value in a point release, as that may be achievable in manageble timeframe.

The another point of practical concern: should I try to solicit some testing of my branch from other folks in the meanwhile?

Thanatomanic:
I am happy to see some new thoughts in the discussion about the release and even a seemingly solid approach to actually get a working release, thanks @pmjdebruijn.
However, after some more pondering I am still not convinced a point release is worth it over just pumping out the full release. The point release 'only' patches some bugs and improves camera support. This is definitely valuable to those that have been stuck on 5.8 and haven't found the path to the (very stable) development builds. No clue how big this group of 5.8 users is, but if they can get better camera support _and_ plenty of new features, and it doesn't cost us much more work to get that out, why not just do that?
Also, those using the development builds (essentially everyone on Pixls) won't start using 5.8.1 because they have gotten used to the new tools already. For them, 5.8.1 is useless and perhaps even ridiculous because "why can they do that, but not 5.9?!?!" 🤷 

Anyway, were essentially in a feature-freeze state and no critical bugs have been found in the last year (?) (that weren't fixed) even thought the dev builds have been in use by a rather sizeable group users. As it stands, I would be happy to release it as-is, even without merging the currently open PRs. If we get the ball rolling ourselves, it should be more straightforward to put out a 5.9.1 release some time after that with some more PRs merged and inevitable bug fixes.

* As for release notes: some time ago I had also already started compiling them here https://github.com/Beep6581/RawTherapee/discussions/6331
* I have push rights [to the website](https://gitlab.com/patdavid/rawtherapee-web), but I'm not entirely sure yet how to properly work with deployable websites (I only ever just wrote plain HTML/CSS/JS by hand). Also, to upload files I think we still need FTP access, which I don't have. Perhaps @heckflosse?
* Builds for Windows and Linux are working perfectly through GitHub Actions thanks to the work by @Lawrence37. Same goes for macOS thanks to @Benitoite. The latter does require some authorization by Apple before it can become available in the app store afaik.
* If the only additional actions are to run the two scripts (tools/generateReleaseInfo and tools/generateSourceTarball), things should be easy...

Thanatomanic:
Also, @pmjdebruijn do you have experience on how a new software release is picked up by the different Linux package repos? Is there anything we need to provide for that? I have zero experience and wouldn't know where to obtain this knowledge...

Lawrence37:
@Thanatomanic made some good points. `dev` is basically 5.9 without some UI string revisions and translations. It's hard to justify 5.8.1 when `dev` is just as stable and has more features. If `dev` is a better candidate for an intermediate release, why even bother when 5.9 is just a step away.

mattiaverga:
> Also, @pmjdebruijn do you have experience on how a new software release is picked up by the different Linux package repos? Is there anything we need to provide for that? I have zero experience and wouldn't know where to obtain this knowledge...

Speaking as the one who package RT in Fedora Linux, we're stuck with 5.8 with a few backported patches that were requested by some users.
I would more than happy to package a '5.9 alpha' version, but didn't do that because I was unsure if things in `dev` branch were all to be included in the next 5.9 final. What about creating a `5.9` branch, so that packagers can start to provide an alpha version from that branch, while development can go on on `dev`, backporting just the translations to the 5.9 branch?

Entropy512:
I am GUESSING that something along those lines was the intended strategy, and that the plan was something like:
"review `default`"
create release branch
kick off the translators

The problem is that the remaining developers are kind of recreating the process from bits and pieces here.  I'm focusing right now on strings/default translations cleanup under the assumption it was the main remaining blocker.  For the most part, per https://github.com/Beep6581/RawTherapee/issues/5632#issuecomment-1038383953 - dev is already in feature freeze, so just bugfixes and low-risk camera support (camconst and DCP profiles).

Pr #6382 and #5904 are, from a rough look at the PRs, the most potentially invasive PRs that are tagged for 5.9 - not sure what Ingo and Roel's thoughts are on those.

pmjdebruijn:
> Also, @pmjdebruijn do you have experience on how a new software release is picked up by the different Linux package repos? Is there anything we need to provide for that? I have zero experience and wouldn't know where to obtain this knowledge...

Well, I do have some thoughts and things you should take into consideration:

We can categorize Linux distro in two main categories, traditional release, and rolling release.

And we can categorize the distribution path as in-band (official distro repos) and out-of-band (third party repos).

# For traditional distros (and in-band)

Lots of distros have a release cycle of about 6 months, therefore it would be a good idea to release new RT versions at least as often, so each new distro release also has a new RT release. Keep in mind that a release does not need to be a new major release, if a major release isn't ready, a point release just containing some bugfixes and some new camera support would be quite fine.

Given that this 6 month cycle isn't synchronised between distros, it's probably even a good idea to try to roughly target quarterly releases. Again point release would be fine. And I'd like to emphasize that adding camera support is probably the most important feature you could ever add.

Keep in mind once a new distro version is released, typically package versions are locked in, except for rare patches which are cherrypicked most commonly. Some distro's like Ubuntu are slowly becoming a little more flexible in moving new software versions into already released distro versions, this still seems to be the exception rather than the rule. 

With regard to to Debian, they have a backports repo these days, which makes it fairly easy for users to opt-in to newer software versions on an already released Debian version.

However, to increase changes for new RT version to be made available in the backports repo, you should try to avoid bleeding edge library dependencies. A good rule of thumb would be the make sure new RT releases build fine on Debian Stable and Ubuntu LTS. Test build environments can be fairly easily be created using debootstrap/chroot/apt-get build-dep rawtherapee.

# For traditional distros (and out-of-band)

Something like SuSE's Open Build Service might be worth considering at some point in the future, though I have no experience with that.

# For rolling distros

These should usually require the least consideration, given that new library versions are available fairly quickly, and there is usually no release timing to take into account.

Once a 5.9 release happens, it might also be worth it to contact package maintainers of the bigger distros like Fedora, Debian (and thus indirectly Ubuntu and other derivatives), Arch, FreeBSD (I know it's not Linux, but for the purposes of this discussion, that doesn't matter all that much). Given there hasn't been a release in a while, not all of them might be paying that close attention to RT. Though @mattiaverga sortof preemptively weakened this point :) But that's a good thing :)

Entropy512:
FYI, I think 6 months was the original average/target release cadence, that just hasn't happened due to a combination of factors, most notably extended absences  that had no advance warning from the person who usually took the release lead role.  Looking at history, up to 5.8, it was an average of roughly 6 months.

Thanatomanic:
Despite the work @pmjdebruijn put into this, I think we can safely shelf this, because we are increasingly close to releasing 5.9 in full.

PR #6466, opened 2022-04-25
WIP: Remove orphaned strings and history events

Entropy512:
Actual changes as part of #6457 
So far I've just done history events 1-99, starting work on the rest.
No history event from "Profile Loaded" looks fishy to me, is this real?  Is that referenced in some unusual way in the code?

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -265,7 +265,6 @@ HISTORY_CUSTOMCURVE;Custom curve
 HISTORY_FROMCLIPBOARD;From clipboard
 HISTORY_LABEL;History
 HISTORY_MSG_1;Photo loaded
-HISTORY_MSG_2;PP3 loaded
 HISTORY_MSG_3;PP3 changed
 HISTORY_MSG_4;History browsing
 HISTORY_MSG_5;Exposure - Lightness
@@ -279,9 +278,6 @@ HISTORY_MSG_12;Exposure - Auto levels
 HISTORY_MSG_13;Exposure - Clip
 HISTORY_MSG_14;L*a*b* - Lightness
 HISTORY_MSG_15;L*a*b* - Contrast
-HISTORY_MSG_16;-
-HISTORY_MSG_17;-
-HISTORY_MSG_18;-
 HISTORY_MSG_19;L*a*b* - L* curve
 HISTORY_MSG_20;Sharpening
 HISTORY_MSG_21;USM - Radius
@@ -307,10 +303,6 @@ HISTORY_MSG_40;WB - Tint
 HISTORY_MSG_41;Exposure - Tone curve 1 mode
 HISTORY_MSG_42;Exposure - Tone curve 2
 HISTORY_MSG_43;Exposure - Tone curve 2 mode
-HISTORY_MSG_44;Lum. denoising radius
-HISTORY_MSG_45;Lum. denoising edge tolerance
-HISTORY_MSG_46;Color denoising
-HISTORY_MSG_47;Blend ICC highlights with matrix
 HISTORY_MSG_48;DCP - Tone curve
 HISTORY_MSG_49;DCP illuminant
 HISTORY_MSG_50;Shadows/Highlights
@@ -318,7 +310,6 @@ HISTORY_MSG_51;S/H - Highlights
 HISTORY_MSG_52;S/H - Shadows
 HISTORY_MSG_53;S/H - Highlights tonal width
 HISTORY_MSG_54;S/H - Shadows tonal width
-HISTORY_MSG_55;S/H - Local contrast
 HISTORY_MSG_56;S/H - Radius
 HISTORY_MSG_57;Coarse rotation
 HISTORY_MSG_58;Horizontal flipping
@@ -330,7 +321,6 @@ HISTORY_MSG_63;Snapshot selected
 HISTORY_MSG_64;Crop
 HISTORY_MSG_65;CA correction
 HISTORY_MSG_66;Exposure - Highlight reconstruction
-HISTORY_MSG_67;Exposure - HLR amount
 HISTORY_MSG_68;Exposure - HLR method
 HISTORY_MSG_69;Working color space
 HISTORY_MSG_70;Output color space
@@ -341,12 +331,10 @@ HISTORY_MSG_74;Resize - Scale
 HISTORY_MSG_75;Resize - Method
 HISTORY_MSG_76;Exif metadata
 HISTORY_MSG_77;IPTC metadata
-HISTORY_MSG_78;-
 HISTORY_MSG_79;Resize - Width
 HISTORY_MSG_80;Resize - Height
 HISTORY_MSG_81;Resize
 HISTORY_MSG_82;Profile changed
-HISTORY_MSG_83;S/H - Sharp mask
 HISTORY_MSG_84;Perspective correction
 HISTORY_MSG_85;Lens Correction - LCP file
 HISTORY_MSG_86;RGB Curves - Luminosity mode
@@ -393,12 +381,6 @@ HISTORY_MSG_127;Flat-Field - Auto-selection
 HISTORY_MSG_128;Flat-Field - Blur radius
 HISTORY_MSG_129;Flat-Field - Blur type
 HISTORY_MSG_130;Auto distortion correction
-HISTORY_MSG_131;NR - Luma
-HISTORY_MSG_132;NR - Chroma
-HISTORY_MSG_133;Output gamma
-HISTORY_MSG_134;Free gamma
-HISTORY_MSG_135;Free gamma
-HISTORY_MSG_136;Free gamma slope
 HISTORY_MSG_137;Black level - Green 1
 HISTORY_MSG_138;Black level - Red
 HISTORY_MSG_139;Black level - Blue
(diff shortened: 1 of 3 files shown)


rtengine/procevents.h
```diff
@@ -89,7 +89,7 @@ enum ProcEventCode {
     EvCrop = 63,
     EvCACorr = 64,
     EvHREnabled = 65,
-    obsolete_66 = 66, //obsolete
+    obsolete_66 = 66, // obsolete
```
Lawrence37, line 96: The corresponding refresh map entry is still there. Can you remove it?

rtengine/procevents.h
```diff
@@ -24,7 +24,7 @@ namespace rtengine
 // Aligned so the first entry starts on line 30
 enum ProcEventCode {
     EvPhotoLoaded = 0,
```
Lawrence37, line 30: While you're revising this file, could you add a few blank lines to bump this one down to line 30 like the comment says?

There's also line 203 in refreshmap.cc which should be removed to preserve the alignment.

Lawrence37:
> No history event from "Profile Loaded" looks fishy to me, is this real?

Sounds like an event for when the image is first loaded and the default profile is applied. I don't recall the history message being used, but I will dig into the code when I have time.

rtengine/procevents.h
Entropy512: Good catch, it's a bit harder to find refreshmap entries that are historically orphaned.

Will move EVHRAmount to category 0 and mark as obsolete tonight

rtengine/procevents.h
Entropy512: Hmm, before I do that, we should figure out *why* it starts on line 30.  Perhaps someone attempted to do an auto-parser script as a companion to generateUnusedKeys, similar to my findorphans?

No such script currently exists (and I see a decent amount of evidence it never did...), it may be better to simply delete the comment and have a script that can handle the start line moving around and/or commented lines.

As to line 203, that needs some investigation.  It looks like EvCATEnabled was moved from ALLNORAW to LUMINANCECURVE - looks like @Desmis did that in 5abd92209 - so that appears to be recent and intentional, @Desmis  is there a reason the old value was left in there but commented, as opposed to just changing it?

heckflosse: > we should figure out why it starts on line 30

I did this long time ago to make it easier to (manually) check for consistency between procevents.h and refreshmap.cc
For example get an entry in procevents.h like `EvClip = 12,`. If the corresponding entry in refreshmap.cc is not in line 30+12 = 42, something is wrong.

Entropy512: Longterm we might want to figure out a way to automate that a bit.  I'll work on fixing the alignment for now as libclang's comments support is...  Hard to understand to say the least so automating that task is probably a post-5.9 thing.

Thanks for the information!

Desmis:
Hello all
I come to add my grain of salt in the debate, not that I contest in any way, the simplification, the wish of clarification, etc.
But we have to be very careful, otherwise RT will not work properly anymore.

There are 2 ways to proceed with the "events
* either the one mentioned with the correspondence between "procevents.h" and "refreshmap.cc" and you have to be very careful 
* or to use a single function incorporated in the GUI, for example as we find in "wavelet.cc" (there are many other cases...), where we declare in a single "message", "events", the level of this "event" which will condition the execution order and the message which appears in the GUI.
* example
 EvWavenaclari = m->newEvent(DIRPYREQUALIZER, "HISTORY_MSG_WAVCLARI");
 EvWavushamet = m->newEvent(DIRPYREQUALIZER, "HISTORY_MSG_WAVUSHAMET");
 EvWavballum = m->newEvent(DIRPYREQUALIZER, "HISTORY_MSG_WAVBALLUM");


Of course in both cases it is necessary to check that the system works, especially from "improcoordinator.cc".

And as a reminder, I am not a good programmer at all, but I had a lot of difficulties to understand and make the system work, especially when you want to change the process order :)

jacques

rtengine/procevents.h
Entropy512: So I just looked at this - the comment is on line 28, but the first entry already is on line 30, no action needed there.

rtengine/procevents.h
Entropy512: Done in the latest commit.

rtengine/procevents.h
Lawrence37: The one in refreshmap.cc starts on 30 but this one is on line 26.

Entropy512: Aha, missed that.

Fixed.  I amended the last commit and forcepushed since this one was so small.

Lawrence37:
I could not find any references to the "Profile loaded" event or history message, even on a very old version of RawTherapee. There's only one place where the history message key gets generated. It uses the event code, but `EvProfileLoaded` is never referenced directly or indirectly via a cast. I think it's safe to remove.

Entropy512:
Hmm, where was it referenced numerically?  I'd like to take a look just to be sure, might turn out to be a possible failing of my analysis scripts.

Lawrence37:
It is not referenced in any way. Sorry if I was unclear.

Entropy512:
Ah, so when you say "the key gets generated" I'm assuming you mean by something that is traversing the events table?  e.g. generated from the table but never used?

Lawrence37:
> There's only one place where the history message key gets generated.

I should reword this. I mean there is one piece of code responsible for generating messages keys.

It does so by taking in an event enum. Since `EvProfileLoaded` is never used directly or from a cast, HISTORY_MSG_2 never gets generated.

Entropy512:
I'll try to do a few more 100-message chunks tonight.  Hopefully will finish all orphans by the end of the weekend.

Entropy512:
HISTOGRAM_TOOLTIP_RAW is listed as an orphan, this kinda looks strange to me that there isn't a tooltip for this...

Most orphaned strings that aren't history events are locallab stuff - not surprising given how much work was done on that since 5.8, and some of the "orphaned" strings have wording indicating that they were experimental so it's no surprise they were dropped.

Entropy512:
Also, as you go further into the events table, instead of moving things to obsolete_NN, they are instead commented out in procevents.h

I'm thinking that maybe we want to rework all of these commits to use commenting-out, so there's better history of what it used to be?  Thoughts?

Lawrence37:
I don't think it matters if we comment them or our change them to obsolete. It's not like anyone needs to know that they were. If they do, it's simple enough to do a `git blame`.

Entropy512:
In that case I'll continue on with the current method unless @heckflosse , @Floessie , or @Thanatomanic disagree.
Planning on finishing the orphaned strings tonight, hopefully tomorrow I'll take a crack at running generateUnusedKeys as a sanity check and generateTranslationDiffs as a final cleanup phase.

Thanatomanic:
I agree with @Lawrence37: if it becomes necessary, things are easy to add back in again. In general, there is sooo much commented code in RT that is simply not useful anymore, nor will it likely become relevant in the future.

marter001:
just to make sure: this proceeding will not find missing entries, right?
Would it be better to open new issues for this?

Entropy512:
Do you mean something that is in the code and missing from `default`, or something that is in `default` but not other language files?

generateTranslationDiffs should handle the second case

marter001:
first one: choosing profile "none" in colormanagement, in history is shown (none) but there is no entry for this in default. The same for (camera)

Entropy512:
Hmm...  That might be a second phase after this round of cleanup.  It's probably best to open a separate issue for those.

marter001:
ok, i opened an issue here: #6472 

Since i am already testing the history entries for the localisation, and didn't find a lot missings, it could be, that a second phase for explicite finding missing entries will not be necessary.

Lawrence37 (approved):
Looks good to merge.

Entropy512:
Technically there's sttill more to do, although I'm OK doing a second round PR for that.

Sorry I haven't been around as much as I thought to finish this, a bit of general burnout lately.

Entropy512:
Apologies for taking so long to finish this.

I squashed the history events commits, since the last bit of orphan removal hasn't been reviewed yet I haven't squashed it.  I intend to squash prior to someone merging.

Also rebased on latest dev

Entropy512:
Looks like the CI system is borked for MacOS - since I didn't touch anything related to what bombed.

rtdata/languages/default
```diff
@@ -3604,11 +3538,9 @@ TP_RAW_3PASSBEST;3-pass (Markesteijn)
 TP_RAW_4PASS;3-pass+fast
 TP_RAW_AHD;AHD
 TP_RAW_AMAZE;AMaZE
-TP_RAW_AMAZEBILINEAR;AMaZE+Bilinear
```
Lawrence37: `TP_RAW_AMAZEBILINEAR;AMaZE+Bilinear`
`TP_RAW_DCBBILINEAR;DCB+Bilinear`
`TP_RAW_RCDBILINEAR;RCD+Bilinear`
These three dual-demosaic keys are in use.

Entropy512: Re-added in next commit.

Lawrence37:
`HISTORY_MSG_112` can be removed and `EvPdShrMaskToggled` is obsolete.

Beep6581:
Hello, I would like to review `default` and it seems like this PR needs to be merged before that. @Lawrence37 could you do that if ready?

Entropy512:
I'll leave it up to you and @Lawrence37 whether you want to handle the small remaining issues he found in a separate PR or not.

If you want me to handle them I'll tyr to take care of that tonight, sorry I've been a horrible procrastinator (AGAIN).

Lawrence37:
@Beep6581 @Entropy512 One obsolete string is not a big problem. Feel free to merge.

Lawrence37:
@Entropy512 Ok to squash and merge?

Entropy512:
Yeah.  I feel like there's more to be done, but that could be in another PR.

Entropy512:
Squashed and rebased, cleaned up the commit message a bit

Lawrence37:
@Beep6581 Merged.

PR #6473, opened 2022-05-07
Camconst support for multiple crops

Lawrence37:
Allows different raw crops and mask areas to be applied based on the image size. Ported from ART.


Thanatomanic:
I had no idea Alberto @agriggio had implemented something, super cool! Thanks for porting it @Lawrence37. Will test soon...

Thanatomanic:
I finally got around to test this and it works very well! Propose to merge soon 👍 

There are a few more camera's for which we can probably include different crops (according to samples on raw.pixls.us):
* Canon EOS R
* Canon EOS RP
* Canon EOS 250D
* Canon PowerShot G5 X Mark II
* Canon EOS M6 Mark II
* Many Panasonic models
* Sony ILCE-7S
* Sony DSC-RX100M6
* Sony ILCE-7RM5

Lawrence37:
Good to hear that it works well. In that case, we can merge and add support for the other crops when time (and motivation) permits.

Issue #6475, opened 2022-05-08
Another crashtest


marter001:
**Short description**
I know, it is not fair: no edit will contain more than 10 local spots, but since ist possible, i tried to have all LA tools in one list. 
Then i suddenly got a crash, which is clearly reproducable on most raw files:

**Steps to reproduce**
Open a raw file (i used cr2 or arw) in editor
Apply attached sidecar
Goto Local adjustments
Activate the spot (Dynamik Belichtung) (top down in the pic)
expand „mask and modifications“
choose the drop down „show mask“
RT will crash at this point

**how the crash will not happen:**
Open a raw file (i used cr2 or arw) in editor
Apply attached sidecar
Goto Local adjustments
Uncheck „enable spot“ (Kontrast u. wavelets)
Uncheck „enable spot“ (Neuer Spot)
Activate the spot (Dynamik Belichtung) (top down in the pic)
expand „mask and modifications“
choose the drop down „show mask“
the mask will be shown

**Additional information**
I was not able to reproduce this by creating a similar situation in a new pp3 file.

Version: 5.8-3091-gfeada4cd0
Branch: dev
Commit: feada4cd0
Commit date: 2022-04-28
Compiler: cc 11.3.0
Processor: generic x86
System: Windows
Bit depth: 64 bits
Gtkmm: V3.24.5
Lensfun: V0.3.3.0
Build type: release
Build flags:  -std=c++11 -fno-tree-loop-vectorize -mtune=generic -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:  -mtune=generic
OpenMP support: ON
MMAP support: ON
Build OS: MINGW64_NT-10.0-20348 3.3.4-341.x86_64 x86_64
Build date: Thu, 28 Apr 2022 06:02:37 +0000 UTC
Build epoch: 1651125757
Build UUID: 85be1867-6a6b-4912-92d7-389a34703721

[crashtest.pp3.txt](https://github.com/Beep6581/RawTherapee/files/8646554/crashtest.pp3.txt)

Lawrence37:
Crash confirmed. Here's the backtrace.
```
[...]

PR #6478, opened 2022-05-13
LA - Take into account Ingo patch Ipwavelet issue #6475- improve behavior mask when for a same Spot several mask

Desmis:
See issue #6475 Another crash test


Thanatomanic:
Thanks Jacques!

PR #6481, opened 2022-05-15
Enable installation on arm64 processors on windows

jrockwar:
When trying to install RawTherapee on Windows ARM64 devices, the following error appears when opening the setup application:

image

However, Windows 11 has x64 emulation enabled by default on ARM devices, meaning that RawTherapee actually works under emulation (albeit with a performance penalty). Prior to this, I was using RawTherapee on my Surface Pro X from the .zip file as the architecture check only prevents you from running the set-up, but the program runs fine.

This PR adds arm64 to the architecture check, so that it doesn't error out and it can be installed on arm64 devices. (I forked onto a separate repo so that I could build artifacts without cluttering the main repository, in case something did not work).

I have tested this on my Surface Pro X - packages are published, the install executes successfully and the program can be run with no issues. This is of course still under emulation, this is a very small PR that doesn't enable arm-native compilation.

### DIFF
--- a/UpdateInfo.cmake
+++ b/UpdateInfo.cmake
@@ -106,9 +106,9 @@ if(WIN32)
     elseif(BIT_DEPTH EQUAL 8)
         set(BUILD_BIT_DEPTH 64)
         # Restricting the 64 bits builds to 64 bits systems only
-        set(ARCHITECTURE_ALLOWED "x64 ia64")
+        set(ARCHITECTURE_ALLOWED "x64 ia64 arm64")
         # installing in 64 bits mode for all 64 bits processors, even for itanium architecture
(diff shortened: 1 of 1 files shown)


Thanatomanic:
Well, if this is the only change necessary to have RT working (suboptimally) on Windows 11 with ARM, I'm happy to merge the change. We can even claim to support Win 11 with the upcoming release.

jrockwar:
@Thanatomanic this is definitely suboptimal, but my Surface Pro X has been my main device for Raw development for a few months now. It is totally useable under emulation.

Do you need any logs or other proof? I am quite a data-driven person so I'm happy to "produce the evidence" if you point me in the right direction. Incidentally, I am also happy to test anything W11 or ARM-related on my device.

PR #6483, opened 2022-05-18
Better ARQ Support

Calandracas606:
Added Support for ARQ files from ILCE-1, both 4 and 16 shot
<strike>Added Support for ARQ files from ILCE-7M4</strike>

"Fixed" possible bug where tiff_ifd[].samples was getting overwritten. This probably hasn't been noticed since it requires the TIFF tags to be parsed in a specific order, and affects files with 4 samples per pixel (pixel shift)

Some pixel shift data about Sony bodies other than ILCE-7RM3 is missing, added some TODOs to show where to add the data. 

Sony bodies without the proper pixel shift data will default to ILCE-7RM3. Pentax still defaults to Pentax.

This is my first time contributing to an open source project, so please let me know if I've made any mistakes during this process.

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -6590,9 +6590,10 @@ int CLASS parse_tiff_ifd (int base)
 	break;
       case 258:				/* BitsPerSample */
       case 61443:
-	tiff_ifd[ifd].samples = len & 7;
+        if(!tiff_ifd[ifd].samples || tag != 258)
+            tiff_ifd[ifd].samples = len & 7;
 	if ((tiff_ifd[ifd].bps = getint(type)) > 32)
-        tiff_ifd[ifd].bps = 8;
+            tiff_ifd[ifd].bps = 8;
 	if (tiff_bps < tiff_ifd[ifd].bps)
 	    tiff_bps = tiff_ifd[ifd].bps;
 	break;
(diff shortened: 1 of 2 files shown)


Entropy512:
Is ILCE-7M4 a typo?  Did you mean ILCE-7R4?  I was fairly certain pixel shift was exclusive to the R family and the 1.

Calandracas606:
> Is ILCE-7M4 a typo? Did you mean ILCE-7R4? I was fairly certain pixel shift was exclusive to the R family and the 1.

My mistake, for whatever reason I thought the ILCE-7M4 had pixel shift, but it appears that I was incorrect

Calandracas606:
Removed the comments and check for ILCE-7M4

Calandracas606:
This branch should be ready for dev review when someone has the chance

rtengine/dcraw.cc
```diff
@@ -6439,7 +6439,8 @@ int CLASS parse_tiff_ifd (int base)
 	break;
       case 258:				/* BitsPerSample */
       case 61443:
-	tiff_ifd[ifd].samples = len & 7;
+	if (tiff_ifd[ifd].samples != 4)
+	    tiff_ifd[ifd].samples = len & 7; // possible bug? should this be .bps instead of .samples ?
```
Lawrence37: FYI LibRaw avoids resetting the samples if the tag is 258 and it is already set.
https://github.com/LibRaw/LibRaw/blob/6fffd414bfda63dfef2276ae07f7ca36660b8888/src/metadata/tiff.cpp#L438-L439

Lawrence37: This will change the samples if it's anything other than 4, whereas LibRaw only does so if samples hasn't been set. I wonder if this will be problematic.

Calandracas606:
Made the suggested changes based on the LibRaw code.

Honestly haven't done any work with ARQ in a long time, but i've tested and it works with both 4 and 16 shot pixel shifts from my ILCE-1

Lawrence37:
Thank you @Calandracas606, this looks good. GitHub makes it very hard to compare relevant changes when commits are rebased, but fortunately this PR only has a few lines of code.

Issue #4650, opened 2018-06-30
White balance Auto - new algorithm Itcwb


Desmis:
I just commit in branch "autowblocal" a big improvment to "Itcwb" - Auto iterate temperature correlation"
To test :)

Here the summary:

This algorithm try to find temperature correlation between 20 to 40 color between 98 spectral color and about 20 to 40 color found in the image, I just found the idea in the web "correlate with chroma" instead of RGB grey point,but I don't use any algo found on the web.

I have test many many algo to find the first one that work :)
Probably (sure) there are improvment to do...

I have create a table temperature with temp and white point with 100 values between 2000K and 12000K we can obviously  change these values, more...with different steps
I have create or recuparate and transformed 98 spectral colors from Colorchecker24, others color and my 468 colors target, or from web flowers, etc. with a step of 5nm, I think it is largey enough.

1) for the cuurent raw file we create a table for each temp of RGB multipliers
2) then, I choose the "camera temp" to initialize calculation (why not ?)
3) for this temp, I calculated XYZ values for the 98 spectrals datas
4) then I create for the image an "histogram", but for xyY (Cie 1931 color space)
5) for each pixel (in fact to accelerate only 1/10 for and 1/10 for y), I determine for each couple xy, the number of occurences
6) I sort this result in ascending order
7) in option we can sort in another maner to take into account chroma : chromax = x - white point x, chromay = y - white point y
8) then I compare this result, with spectral datas found above in 3) with deltaE (limited to chroma)
9) at this point we have xyY values that match Camera temp, and spectral datas associated
10) then I recalculate RGB values from xyY histogram
11) after, I vary temp, between 2000K to 12000K
12) RGB values are recalcualted from 10) with RGB multipliers, and then xyY are calcualted for each temp
13) spectral datas choose are recalculated with temp betwen 2000K to 12000K with matrix spectral calculation, that leeds to xyY values
14) I calculated for each couple xy, Student correlation (without Snedecor test)
15) the good result, is the best correlation
16) we have found the best temperature where color image and color refences are correlate
17) after we pass this value to improccoordinator

Some variables or function are not used, keep in case of.
I have test with cat02 but result are not stable enough ! why ??, therefore cat02 neutralized.
This operation is done (actually) 100 times and compare Student coefficient, and keep the absolute  minimum, We can probably optimize....
But actually the goal is to find the good algorithm!

I think, this algo is very good in most cases :) ...to verify of course.
[...]

PR #6484, opened 2022-05-19
Fix missing coefficient for Daylight spectral data

KLZ-0:
Filled in a missing coefficient value for `s1`:

wavelength | s1
--- | ---
770 | -11.20
775 | -11.80
**780** | **-10.40**
785 | -10.50
790 | -10.60

and `s2`:

wavelength | s2
--- | ---
590 | 2.10
595 | 2.65
**600** | **3.20**
605 | 3.65
610 | 4.10

The old initializers only contained 96 out of the expected 97 values, `s0` seems to be fine.

The correct values were taken from [this book](https://onlinelibrary.wiley.com/doi/book/10.1002/9781119975595), specifically [appendix 5](https://onlinelibrary.wiley.com/doi/pdf/10.1002/9781119975595.app5).


Lawrence37:
Nice catch. The values from the book seems to match the ones in the code except for S1 wavelengths 365, 370, and 375. If I understand correctly, the original data is in 10nm increments and the ones in between are linearly interpolated. This implies the 370nm value was incorrect somewhere at one point.

|Wavelength|Book|RT     |
|-------------|------|------|
|365             |40.20|40.70|
|370             |42.40|43.40|
|375             |40.45|40.95|

http://www.brucelindbloom.com/index.html?Eqn_DIlluminant.html agrees with the book. Perhaps we should update these three values too?

KLZ-0:
You are right, those values seem definitely wrong. I suppose they were originally transcribed from a book or pdf. I will add a commit for that.

Lawrence37:
These spectral values are used by the temperature correlation white balance (#4650). With this fix, the output is sometimes slightly different usually when the light is significantly different from daylight. I cannot say if there is an improvement. For some images, it is better. For others it is worse.

Issue #6411, opened 2022-01-21
Loading partial profile using Local Adjustments spot(s) = Double/Unwanted result.


ghost:
**Short description**
Loading a partial, LA related, profile results in double/merged spots instead of the current spot(s) being replaced by those in the partial profile.

**Steps to reproduce**
1. Load any RAW,
2. Activate Local Adjustments,
3. Add 2 spots,
4. Save these 2 spots using *ctrl-click* on *Save current profile*,
5. Reset RAW. Result should be a clean LA module with no spots,
6. Create 2 new spots,
7. Load the partial profile saved in step 4 (make sure the *Preserve* mode is selected),
8. You now have 2x2 spots instead of the 2x1 spots that are in the partial profile.

**Expected behavior**
The behaviour for all the other modules is as follows: _Partial loaded profiles fully replace the module/tool they target._

I tested this with many modules, but especially  with the 2 other modules that have "spots" or "instances", namely *Wavelet Levels* and *Color Toning* I do get the expected behaviour (what is already there is fully replaced).

**Additional information**
A visualization of the above 8 steps:

https://user-images.githubusercontent.com/56407344/150508480-5fa795c8-808a-4e22-b818-4479f299d033.mp4

- The amount of spots is irrelevant (so not necessarily 2+2 as shown above).
- If the partial file has less spots then currently used, the extra spot(s) are removed and the overlapping is merged as described above.
- It does not matter if there are more then 1 tool attached to spot(s).

I initially thought that @Desmis was responsible for this, but he told me that this part of the code was created by @Pandagrapher.

There is another, related(??), issue that deals with LA and partial profiles: #6150

Lawrence37:
This comment is relevant: https://github.com/Beep6581/RawTherapee/issues/6136#issuecomment-787483194. To summarize, this is currently the intended behavior but in the future, applying partial profiles should either add or replace spots. I think there should be at least an option to add spots since it would be difficult to achieve the same result otherwise.

ghost:
@Lawrence37 : Thanks for pointing this out.

I have to agree to there being the ability to choose, but: I really think that the way it is implemented now is the wrong way around. Everywhere in RT what is there is fully replaced by the loaded profile and this should be the default for the Local Adjustments too.

Fixing this seems to be a lot of work and, if I understand correctly, will not be started on until the new stable 5.9 rolls out.

So, I'll shut up for now and patiently wait (I'm going to leave this one open for the time being).

[...]

Issue #6136, opened 2021-02-25
LA: New activated modules are not restored when loading a previous pp3 profile.


arturoisilvia:
When restoring a pp3 profile, later activated modules in Local Adjustments are not restored:
I have the first 4 modules, of which I have activated the first 2:
![1_Los2primerosModulosActivados](https://user-images.githubusercontent.com/55624244/109184397-406f5980-778f-11eb-9b56-bc77d5f352d7.JPG)

I save this profile with name: TheFirst2ModulesActivated.pp3
Next, I activate the third module and take 2 more:
![2_Con5Modulos](https://user-images.githubusercontent.com/55624244/109184430-49f8c180-778f-11eb-890a-6d1e5e24a4d7.JPG)

Now I load the previous profile: TheFirst2ModulesActivated.pp3
And the result is that the last 2 modules are still activated.  They should not be or be off.
![3_Resultado](https://user-images.githubusercontent.com/55624244/109184463-511fcf80-778f-11eb-9a9f-210a719ee109.JPG)

Greetings, Arturo.
Version: 5.8-2776-gf72747771 win64.zip

Thanatomanic:
@arturoisilvia I assume you have 'partial filling' enabled? I can replicate your behavior. 
![image](https://user-images.githubusercontent.com/6567747/109187956-b2956d80-7792-11eb-8f86-ddcbc69fd7f7.png)

There is something far worse though: if you enable 'full mode' _nothing_ is applied. Try to create a few spots, save the pp3, apply Neutral profile and then load the pp3. LA is activated, but no spots are created!
![image](https://user-images.githubusercontent.com/6567747/109188014-c214b680-7792-11eb-99ed-3464c8e738f6.png)

@Desmis @Pandagrapher Can either of you take a look? This seem quite problematic...

Pandagrapher:
@Thanatomanic @arturoisilvia 
For me the case reported by arturoisilvia is not a bug. When loading a profile with "partial filling" enabled, a merge is performed between your profile and the actual parameters (like all other tools). In the given example, all module activation/added state is kept except for the third module which is deactivated to match the loaded profile.

But the fact, that loaded profile does not erase completely when "full mode" is selected is more annoying... I am investigating...

arturoisilvia:
@Thanatomanic 
Yes, the profile processing mode is partial.

Pandagrapher:
@Thanatomanic 
This patch solves the issue when loading a profile when "full mode" is activated:

[...]

Issue #6150, opened 2021-02-28
Local adjustments lost when applying partial profile


epadepa:
**Short description**
Local adjustments are reset to null  when applying partial profile even if the partial profile doesn't touch local adjustments.

**Steps to reproduce**
1. Save partial profile containing only say a resize or a colour toning
2. Open new image or reset to neutral profile
3. Add one or more local edit spot with one or more tools associated to them.
4. Ensure Processing profile fill mode is set to partial
5. Load profile from file selecting profile saved in step 1 above
6. Observe how previously created Local adjustment spots have been removed

**Expected behavior**
In step 6 above I expect the previously created spots to remain in the saved profile doesn't touch the Local Adjustments module

**Additional information**
- Version of RawTherapee: 5.8-2896-g50d040fff
- Version of your operating system: Debian Sid

```
Version: 5.8-2896-g50d040fff
Branch: dev
Commit: 50d040fff
Commit date: 2021-02-28
Compiler: cc 10.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.2
Lensfun: V0.3.2.0
Build type: release
Build flags: -std=c++11 -Wno-deprecated-declarations -Wno-unused-result -O3 -msse2 -std=c++11 -march=native -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -flto -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:  -march=native -flto
OpenMP support: ON
MMAP support: ON
Build OS: Linux 5.10.0-3-amd64 x86_64
Build date: Sun, 28 Feb 2021 20:22:17 +0000 UTC
Build epoch: 1614543737
Build UUID:
```

Thanatomanic:
~~This was reported literally a few days ago. https://github.com/Beep6581/RawTherapee/issues/6136~~

Thanatomanic:
Thanks to a friendly message on Pixls, this is actually a bug. @Desmis Can you take a look why all Local adjustment spots are removed when a simple partial profile is applied even when the partial copy mode is active?

You can test with this pp3 if you like:
[testprofile-RGBCurves.pp3.txt](https://github.com/Beep6581/RawTherapee/files/6693832/testprofile-RGBCurves.pp3.txt)

[...]

PR #6525, opened 2022-07-14
Local adjustments - Fix apply/paste actions with LA

Pandagrapher:
The Pull Request fixes the following issues:
- Fixes LA lost when applying partial profile (#6150)
- Fixes LA behavior when using apply/paste actions: spots are now totally replaced and not anymore merged (#6136, #6411)
- Fixes a crash in specific situations when saving/copying a partial number of spots
- Partial paste dialog: Fixes Locallab button remaining inconsistent if all the spots are deselected

@jade-nl @Lawrence37 @epadepa

### DIFF
--- a/rtgui/paramsedited.cc
+++ b/rtgui/paramsedited.cc
@@ -3382,14 +3382,15 @@ void ParamsEdited::combine(rtengine::procparams::ProcParams& toEdit, const rteng
 
     if (locallab.enabled) {
         toEdit.locallab.enabled   = mods.locallab.enabled;
+
+        // In that case, locallab is impacted by the combine:
+        // Resizing locallab spots vector according to pedited
+        toEdit.locallab.spots.resize(locallab.spots.size());
     }
 
     if (locallab.selspot) {
         toEdit.locallab.selspot   = mods.locallab.selspot;
     }
-
-    // Resizing locallab spots vector according to pedited
-    toEdit.locallab.spots.resize(locallab.spots.size());
 
     // Updating each locallab spot according to pedited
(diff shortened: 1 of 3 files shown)


ghost:
Ah, good to notice that this one is being picked up!

I just merged, build and tested your PR: This does not, however, fix #6411.

If I repeat the exact steps shown in 6411 I end up with the same unwanted result: Double entries, old and new, were I expect the current ones to be overwritten by the newly imported ones.

Pandagrapher:
@jade-nl Thanks for your testing. However I just tested again and I cannot reproduce the issue #6411 with this PR. Are you sure you have the `f3d7f6a` commit? (because the `6a03d57` one was only fixing the issue during copy/paste actions)

ghost:
@Pandagrapher : Apologies! Problem is found on my side of things.

Turns out that I made a typo which was not picked up by my semi-automated merge/build script (script is now fixed....).

(Correctly) Merged, build and tested your changes: Works nicely and this fixes #6411

Thanks for the time and effort.

PS: Although the other fixed issues aren't mine I'll test those as well. Will be later today or maybe during the upcoming weekend.

ghost:
Tested and this PR also fixes #6136 on my side of things.

@arturoisilvia mentions the following as an expected result: _They should not be or be off._

The actual result is that the situation is restored to what is in the loaded pp3 file (amount of tools, status etc.). All else present is removed (the *not be* part).

---

Tested #6150 as well: Also fixed on my side: Local Adjustment spots are _not removed_ when using fill mode set to *preserve (partial)* and they _are removed_ when mode is set to *fill*.

---

All test were done on Linux (Debian).

Pandagrapher:
@jade-nl Thanks for all your testings

If no one objects, I gonna merge the PR tomorrow

PR #6536, opened 2022-07-29
Relative pathnames to .dcp and .lcp files in .pp3 for interchangeability between Windows and Linux

Bezierr:
This is related to my forum question [https://discuss.pixls.us/t/using-rt-both-on-windows-and-on-linux/31678](https://discuss.pixls.us/t/using-rt-both-on-windows-and-on-linux/31678). Since I happened to have some time, I implemented a solution for this.

**Background:** .pp3 files that were created under Windows cannot be used under Linux if they contain (absolute) file names, in particular to camera profile .dcp files, or lens profile .lcp files.

-------------------------------------------------------------------------
### Implemented solution part 1:

In Preferences / Image Processing, in the Directories section, added two directories "Camera profiles directory" and "Lens profiles directory".

Processing:

When reading from .pp3:
	IF (InputProfile is a relative path
		AND CameraProfilesDirectory is defined in Preferences
		AND CameraProfilesDirectory is an absolute path)
	THEN
		create the absolute path to the input profile by prepending CameraProfilesDirectory to InputProfile
	ELSE
		unchanged behavior compared to today
		
When writing to .pp3:
	IF (CameraProfilesDirectory is defined in Preferences
		AND CameraProfilesDirectory is a prefix of the absolute path to the input profile)
	THEN
		write a relative path to InputProfile that exists of what comes after CameraProfilesDirectory
	ELSE
		unchanged behavior compared to today
		
Same for LensProfilesDirectory.

Consequently, backward compatibility is assured: If you don't want this "relative path" behavior, simply don't put anything into the corresponding settings in Preferences.

-------------------------------------------------------------------------
### Implemented solution part 2:

Since RT Windows version uses "\\" as separator (which is not understood under Linux) but also understands "/", replace all "\\" by "/" when writing InputProfile or LensProfile to .pp3 file.

Again, this is only done if the new directories are not empty.

-------------------------------------------------------------------------
### Implemented solution part 3:

FilmSimulation already has such a "base directory" mechanism; however, under Windows, does not accept "/" as separator. So when reading HaldClut path from .pp3 under Windows, replace all "/" with "\\", and under Now-Windows, replace all "\\" with "/".

This, once again, means that if you use RT either under Windows or under Linux (but not on both on the same .pp3 files), the behavior will be unchanged to what it is today.

-------------------------------------------------------------------------
### Words of caution:

I tested this with Windows and Linux, but have no access to Mac. I understand that Mac file paths look like Linux file paths, so it *should* work for Mac, too. But it is not tested.

There may be other places in .pp3 files where (absolute) file paths are stored which prevent interchangeability between Windows and Linux. If you like my approach, please let me know if there are more, and I may look into it whenever I find the time.

Finally: Feel free to reject this Pull Request if you think it is not practical for general use. It would still continue to work for me ;-)


Lawrence37:
Is it safe to simply convert forward and backslashes? Backslash is valid in Linux filenames. Even though nobody really uses it, it's still a possibility.
If I understand the current behavior, the profile paths can be relative to the pp3 if they are saved next to each other. Setting the base directory in preferences will mean it is no longer possible to bundle them together and move them without breaking the pp3.

Benitoite:
I would avoid `\` on any POSIX because it is a reserved escape character, and as you can see is even escaped itself in c++, and `/` on windows because it is was at one time reserved for switch/command flag.   

I didn't notice any functional change with this PR on macOS.

Bezierr:
> Backslash is valid in Linux filenames. Even though nobody really uses it, it's still a possibility.

Using a filename with backslash in RT on Linux did not give me any result, i.e. the file was apparently not found. That's why I came up with the conversion.

> If I understand the current behavior, the profile paths can be relative to the pp3 if they are saved next to each other. Setting the base directory in preferences will mean it is no longer possible to bundle them together and move them without breaking the pp3.

Yes, this is my understanding, too. I you want to use this functionality (I don't), simply don't specify the base directories, and it should work as before. Unless there is a bug in my implementation, of course.

Bezierr:
Additional remark:

I‘m sure you noticed that for the camera profile / lens profile paths, I do the (back)slash conversion when *writing* to the .pp3 file, whereas I do it when *reading* the HaldCLUT path.

The rationale behind this: These camera profile / lens profile paths are new, so why not create them in an OS independent way? HaldCLUT profile paths, on the other hand, exist already, so you should not have to change them if you want to read a .pp3 file created under Windows, under Linux.

Thinking about this again, however, for consistency reasons, maybe it is better to always do the conversion when *reading* the .pp3 file? What is your opinion?

Lawrence37:
It could be useful to merge the existing and new functionality. If the profile is next to the pp3, use a path relative to the pp3. Otherwise, check if it's in the configured directory. Just a suggestion, if you think it's a valid use case.

In my opinion, it doesn't really matter if all the conversions are done while reading or if there is a mix of reading and writing.

Entropy512:
> Is it safe to simply convert forward and backslashes? Backslash is valid in Linux filenames. Even though nobody really uses it, it's still a possibility.

Probably not.  Might be useful to look at how wslpath works - Microsoft took quite a lot of care in designing that to maximize on-the-fly interoperability between a Linux instance within WSL and Windows

I'm trying to figure out where the source for the actual MS implementation is - https://github.com/laurent22/wslpath is NOT it.

Bezierr:
> It could be useful to merge the existing and new functionality. 

Looking into this, it seems to me that the "existing" functionality actually... does not exist.
It is implemented (I use it myself in this PR), but functions `save` and, inside it, `relativePathIfInside` in `procparams.cc` apparently never are called with `fnameAbsolute=false`. It is either explicitly `true` or defaults to `true`.

At least that's my impression looking at the code, and indeed I haven't managed to get a relative path for Camera Profile.

Am I wrong?

Bezierr:
Coming back to this, I made a couple of changes.
- Always do conversion of "/" or "\\" when reading the profile.
- Referring to the comment by @Lawrence37 , I changed the processing order to, as suggested, "if the profile is next to the pp3, use a path relative to the pp3. Otherwise, check if it's in the configured directory." I repeat that it is my impression that the first "if" always fails because fnameAbsolute is always true. I haven't changed that since there may be a reason for it.
- I found that FlatField and DarkFrame do have base directories in Preferences which, however, apparently are not used to create relative filenames. Added this functionality.
- I found that dynamicprofile.cfg is not portable between Windows and Linux, either. Added  conversion of "/" or "\\" when reading it.
- I found that when dynamicprofile.cfg is a symlink (which, in my case, it is on Linux), it is overwritten instead of followed. Changed this. (Works only on Linux since on Windows, symlinks apparently are not recognized.)

Tested extensively on Linux and Windows, not on Mac. Seems to work well. It's fun to seamlessly switch between RT on Linux and RT on Windows :-)

Again, if you don't consider this pull request suitable for the "official" RawTherapee, feel free to reject.

rtengine/dynamicprofile.cc
```diff
@@ -218,6 +219,22 @@ bool DynamicProfileRules::loadRules()
 
         try {
             rule.profilepath = kf.get_string (group, "profilepath");
+			#if defined (WIN32)
+			// if this is Windows, replace any "/" in the path with "\\"
+			int pos = rule.profilepath.find("/");
+			while (pos != -1) {
```
Lawrence37: Replace `-1` with `Glib::ustring::npos`

Lawrence37 (commented):
Nice updates @Bezierr. I left one small review comment.

Bezierr:
Thanks. Sure. Done.

Lawrence37:
I do believe this pull request is ready for merging.

Lawrence37:
I don't know if there is a formal threshold for `AUTHORS.txt`, but @Bezierr you have made contributions in the past and I don't see your name there. I think it is a good idea to add yourself to the list of developers with this pull request.

Lawrence37:
Merged. Thanks @Bezierr!

PR #6545, opened 2022-08-11
Local Adjustments : Optimize FFTW (when used) - when full image is selected

Desmis:
This PR aims to improve the use of FFTW when the user selects "Full image" in "Local adjustments - settings".

Until now the dimensions taken into account when using a FFTW function (e.g. Blur...) were slightly smaller than the real size of the image. As a consequence, some pixels (near the edges) were not processed.

This PR adds a few tens of pixels so that the whole image is processed (small impact on the consumption of resources)

This PR has no impact on the rest of the code: rtengine (except iplocallab.cc), rtgui is unchanged, rtdata is unchanged.

I tested to check the function on the images of my old D200, those of a D800, and those of a Z9 

Jacques

### DIFF
--- a/rtengine/iplocallab.cc
+++ b/rtengine/iplocallab.cc
@@ -617,6 +617,7 @@ struct local_params {
     float laplacexp;
     float balanexp;
     float linear;
+    int fullim;
     int expmet;
     int softmet;
     int blurmet;
@@ -895,7 +896,15 @@ static void calcLocalParams(int sp, int oW, int oH, const LocallabParams& locall
     lp.laplacexp = locallab.spots.at(sp).laplacexp;
     lp.balanexp = locallab.spots.at(sp).balanexp;
     lp.linear = locallab.spots.at(sp).linear;
-
+    if (locallab.spots.at(sp).spotMethod == "norm") {
+        lp.fullim = 0;
+    } else if(locallab.spots.at(sp).spotMethod == "exc"){
+        lp.fullim = 1;
+    } else if (locallab.spots.at(sp).spotMethod == "full"){
+        lp.fullim = 2;
+    }
+   // printf("Lpfullim=%i\n", lp.fullim);
+    
     lp.fftColorMask = locallab.spots.at(sp).fftColorMask;
     lp.prevdE = prevDeltaE;
     lp.showmaskcolmet = llColorMask;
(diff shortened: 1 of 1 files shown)


Desmis:
@Lawrence37 
Hello
Can you explain why all builds fail, and for the same reason, which has nothing to do with my modification of iploccalab.cc.
Every time it seems to be due to "include lensfun".

Thank you

Jacques

Lawrence37:
One of the websites used by lensfun to update its database is down. We should wait for the website to come back up then rerun the builds.

Lawrence37:
@Desmis I just merged the PR that works around the Lensfun issue. You can pull from dev if you want the Windows and AppImage builds to succeed. The MacOS one will still fail due to a different issue.

Desmis:
@Lawrence37 

I just try to re-build Windows builds, and also Appimage

When **"Include lensfun"** the system says 
Error: Process completed with exit code 1.

Am I making a mistake?

**Oh I just see I must I must pull from dev...
But excuse my ignorance, but how do you make a  "Git pull" when I am in the branch "la_optfftw"
Can you give me the git command: git merge etc.**
Of course before i have
Git checkout dev
Git pull
Git checkout la_optfftw

Another question, can I merge this PR without problems? It's only a little modification to iplocallab.cc, that I think it is advisable to do before 5.9

Thank you

Jacques

Lawrence37:
@Desmis
After `git checkout la_optfftw ` you can use `git merge dev`

The code looks good, but I have not tested this branch. Maybe someone else can, otherwise I will try to give it a quick test run later.

Lawrence37:
I gave this a brief test. Although I wasn't able to reproduce the issue in `dev`, I didn't notice any problems with this branch. It is OK to merge this PR for me.

Desmis:
@Lawrence37 

Hello
Thank you for the service you are providing. My health will be improved by less stress. Thank you again.

You are a person I appreciate very much, your remarks are always relevant, and you are really a nice guy.

As far as the usefulness of the modification is concerned, it is only useful for Blur. 

For other uses in Original retinex, Blur Mask (Color & Light), or PDE functions (Dynamic Range & exposure) this has no or negligible impact.

Some comparisons of the image size, of the concerned area without modification, and with modification

Nikon D200
Image size : 3892x2608
Whithout optimization : 3888x2600
With optimization : 3969x2688


Nikon D800
Image size : 7370x4916
Whithout optimization : 7350x4914
With optimization : 7500 x 5040

If you use Blur with very strong "radius", and you make layers / differences you will see small gaps

For information the FFTW function is optimized if the numbers H and W are multiples 
2^n * 3^p *5 ^v * 7^k * 11^y * 13^x
And the optimization is even greater when using multiples of 2, 3 and 5, and exceptional if it is only a multiple of 2^n for example :16384... 4096...
So the larger size, can be in some cases faster.

Thank you for testing. If no one objects, I'll do the merge tomorrow

Jacques

Floessie:
> You are a person I appreciate very much, your remarks are always relevant, and you are really a nice guy.

I can only agree with that. :+1:

PR #6547, opened 2022-08-14
Castellano (Spanish) language file

### LINKED ISSUE #6530: New Spanish GUI translation
This is the brand new «Spanish» translation of the GUI for version 5.9

Note that the language file must be called _Castellano_ and not _Spanish_ (this is due to the large number of official languages here in Spain, in addition to the existence of other variants of Spanish in different countries around the world).

[Castellano.zip](https://github.com/Beep6581/RawTherapee/files/9169936/Castellano.zip)

Lawrence37:
Provided by Francisco Lorés and Javier Bartol, with `generateTranslationDiffs` applied. Closes #6530.

@TechXavAL I would like to add automatic language detection. Currently, RawTherapee associates the es_ES locale code with the `Espanol` language file (Latin American Spanish). I did some research and believe es_ES is more closely related to `Castellano` while es_419 corresponds to `Espanol`. Do you know if this is the case, at least for Castellano?


TechXavAL:
That's correct.

I would say `es_419` is too generic for all the latin american variations of Spanish, but it may cover pretty well all the languages in Latin America, if nobody has any objection.

On the other hand, `es_ES` is the correct code for Spanish in Spain (or more precisely, Castellano in Spain).

Lawrence37:
The code for choosing the language file from the locale is a bit weird, which is one reason to consider replacing it with something like gettext as @Entropy512 suggested. Nevertheless, automatic detection of Castellano and Latin American Spanish should work now. `es_419` or any variation of it matching `es.419*` will use Espanol. `es_ES`, its variations, and all other `es*` locales will use Castellano (because the country code `ES` matches the language code `es` so RawTherapee interprets this as the main locale).

TechXavAL:
Sorry for the late reply. My computer dead a few days ago. Back to work now...

Your solution is Ok to me and I will be happy with it as it is, but I think it's not how it should work, if I understand the locale_country concept: if I'm right, `es` is related to the broad language (the locale) used in a document or file. And the second part after the underscore (`_XX`) is related to the country that uses that language.

Well, _devil is in the details_ and here lies the problem: all countries (Spain and all of latin american countries) talk in a generic _Spanish_ (`es`), which is differentiated by the second part (`_XX`).

In Spain we talk in `es_ES` (castellano), in Mexico they talk in `es_MX` (which is slightly different to castellano), in Colombia they talk in `es_CO`, and so on.

I had no previous knowledge of `es_419` even though it exists and it is a valid common locale for all latin american countries. However, I haven't found any `es_419*` locale anywhere.

So maybe a good compromise would be setting `es_ES` for `Castellano`, and Latin-american for anything else.

Thanatomanic:
Wouldn't it make more sense to stick to the descriptions used by the ICU Locale system? So "Spanish (Spain)" for `es_ES` instead of "Castellano", and "Spanish (Latin America)" for `es_419` instead of non-specific "Spanish"?

Edit: we already do this for English and it is probably what people are more familiar with anyway from other occurrences in their OS or software.

TechXavAL:
I don't want to be argumentative, as this is not something you have to spent many time with.

The main fact is that in Spain there are 5 official languages, they are all spoken in Spain, and they are all _Spanish_. Technically _Castellano_ is the «top» official language, so `es_ES` should technically make reference to _Castellano_.

However, I guess nobody outside Spain would give a damn about that, so `es_ES` could perfectly be linked to `Spanish`.

On the other hand, as I said, what would RT do when it founds a localization for `es_MX`, `es_CO`, `es_AR`, ...?

Moreover, [according to ICU](https://unicode-org.github.io/icu/userguide/locale/#language-code) it takes it as if Spanish was a uniform language all over the world, which is not, and even RT distinguises between Brittish English and American English... In the case of Spanish, there are more than two options...

If I had to decide, I would bind `es_ES` to `Spanish` and anything else to `Spanish (Latin America)`. Note I said _«anything else»_ to take into account all the different Latin American locales which wish to be more specific than `es_419`.

Lawrence37:
The naming of the language files doesn't follow any particular convention. However, I remember seeing a comment in the code that describes what the expectation is. The first word is the language. It can be followed by another word in parentheses that describes the variation of that language. If using the parentheses, there should be a main language file. This "requirement" isn't even followed by the current file names, so I don't know what guidelines we should follow. If we use the existing file names as an example, then I think the Spanish files should be `Spanish (Castellano)` (from what I'm understanding from @TechXavAL, "Spain" is too broad) and `Spanish (Latin American)`/`Spanish (Latin America)`.

For the locale code associations, I'm fine with changing the default Spanish language to the Latin America one.

TechXavAL:
@Lawrence37 , @Thanatomanic , if you think this is the right solution, then I'm Ok with that. My main concern is that the new translation will be available to the end user. Don't spend more time on this, please.

Thanatomanic:
This just shows we need to have a clear policy on how to handle stuff. Let's talk about that after 5.9 drops.

For now, @Lawrence37 , could you please commit the proposed changes to the names of the languages? ~~I think what Lawrence proposes works.~~

Edit: pinged Xavier, but it's Lawrences' PR.

Thanatomanic:
Thanks @Lawrence37, this seems good to merge now! (and I think this completes all of the open localization issues)

Issue #6048, opened 2021-01-04
Panel background changes color erratically when scrolling


ghost:
RawTherapee_dev_5.8-2701-gb8d5a3d_20210102.AppImage running on Ubuntu 18.04

In the Advanced, Local Adjustments, Transform and RAW tabs, some of the panel colors change from medium grey #565656 to light grey #636363 on mouseover, further reducing the contrast and readability. Scrolling the affected area out of sight and back again resets all or part of the panel to the medium grey color. The easiest way to observe it is by scrolling the Post-Resize sharpening panel in the Transform tab in and out of sight. The behavior does not occur in any of the first three tabs (Exposure, Detail, Color) where all the panels are consistently #565656.
The behavior has been tested on two different machines both running updated Ubuntu 18.04

Thanatomanic:
Can you provide screenshots of this behavior? There is no hover CSS style defined for the panel backgrounds, so I'm not sure what is going on.

ghost:
The problem is that activating the screenshot tool and moving the mouse towards the panel immediately makes it change state. I did manage to take this with my phone but it's not very clear.
![Screenshot from 2021-01-04 12-44-36](https://user-images.githubusercontent.com/65723152/103545581-0a55dd80-4ea2-11eb-973c-977f0a02a6ee.png)

heckflosse:
> The problem is that activating the screenshot tool and moving the mouse towards the panel immediately makes it change state.

Use the PrintScreen Key instead?

ghost:
Thanks, that worked :-) The edge sharpness is #636363 and Denoise & Refine is #565656. If I mouse over it, it becomes #636363
![Screenshot from 2021-01-04 16-34-02](https://user-images.githubusercontent.com/65723152/103552091-a801da80-4eab-11eb-99d0-052ecee93aa6.png)

Thanatomanic:
Hmmmm @pseudopanax I just built RT on Linux Mint in VM and encounter no such problems. @TechXavAL @afr-e @soupygit is any of you using Linux and able to reproduce this?

TechXavAL:
This is what I see (Linux Mint 20, not VM, RT commit cc7f1ecda):

![colors-labels](https://user-images.githubusercontent.com/52250392/103568275-90375000-4ec5-11eb-9729-841b364e0ff8.png)

[...]

PR #6555, opened 2022-08-20
Fix for white outlines and background in AppImage

### LINKED ISSUE #6526: AppImage: White outlines and layers in UI
**Short description**
UI elements in the **AppImage** have a white outline/layer to it

**Steps to reproduce**
1. Open the AppImage
2. Check the UI (eg. Editor)

**Observed behavior**
White outlines layers as shown in the image
editor slider
editor save image

**Expected behavior**
Same UI theme as in 5.9 with no white outline

**Additional information**
Issue was observed in KDE neon, but was also observed with a live Ubuntu 22.04

Already noted by Lawrence37 in the 5.9 thread.
> > I prefer builds from an automated source
> 
> The automated builds for Windows and Linux AppImages will also build releases. If you plan on using them, verify the quality first. I have encountered issues such as weird crashes on the Windows version and **white outlines with the default theme on the AppImage**.

I did not find a bug report for this.

My guess that this is a bug and not a style change is based on that the AppImage didn't have that issue in the past. It has changed with some update between April and June around when the new system for automated updates got in place.
The same dev version on Windows (from ZIP) does not show this behaviour

Only the RT and RT legacy profiles are effected. None of the TooWaGrey profiles show this behaviour.

Lawrence37:
I managed to track down the difference between the AppImage and the other builds. It sets the `GTK_THEME` environment variable according to the GTK themes it can find with `gsettings`. On my computer, `gesettings` does not find all the themes, which causes the AppImage to default to `Adwaita:light`. I patched the GTK plugin for linuxdeploy to remove the `GTK_THEME` environment variable. It will now use the system theme, which I believe is the same behavior as the other builds including native ones. Uses can override the theme with the environment variable.
To avoid theme issues, the RawTherapee CSS themes should control all the colors instead leaving some to the default theme. We should consider fixing the RawTherapee theme. The TooWa* themes are unaffected.
Closes #6526.

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -101,7 +101,9 @@ jobs:
           echo "Downloading linuxdeploy."
           curl --location 'https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage' > linuxdeploy-x86_64.AppImage
           echo "Downloading GTK plugin for linuxdeploy."
-          curl --location 'https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh' > linuxdeploy-plugin-gtk.sh
+          curl --location 'https://raw.githubusercontent.com/linuxdeploy/linuxdeploy-plugin-gtk/master/linuxdeploy-plugin-gtk.sh' \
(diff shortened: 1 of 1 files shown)


Thanatomanic:
This makes perfect sense, thanks for digging @Lawrence37! This might also fix https://github.com/Beep6581/RawTherapee/issues/6048

PR #6558, opened 2022-08-24
Increment AppImage tools cache key

### LINKED ISSUE #6526: AppImage: White outlines and layers in UI

Lawrence37:
Closes #6526

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -89,7 +89,7 @@ jobs:
         id: appimage-tools-cache
         uses: actions/cache@v2
         with:
-          key: appimage-tools
+          key: appimage-tools-1
           path: |
             ./build/linuxdeploy-x86_64.AppImage
             ./build/linuxdeploy-plugin-gtk.sh


Lawrence37:
@Thanatomanic I believe the reason why the white lines still exist is because the build was caching the part I had changed.

Thanatomanic:
Let's see then :-)

Lawrence37:
I can confirm it solves it for me.
image
(Yes, I tried the previous build and saw the white lines.)

Issue #5658, opened 2020-02-19
Canon EOS 100D dual ISO DNG regression 


madarexxx:
Hi :) 
I am using Rawtherapee for a while and it is the great example of free software -- developing, bringing new ideas and possibilities to users. 

Today I have met with the great regression --- Rawtherapee refused to open Dual ISO DNG file, created by Canon SL1 and converted to DNG by:

```
cr2hdr: a post processing tool for Dual ISO images
Last update: 185da25 on 2018-02-04 22:52:42 UTC by Dan:
```
I am running:
- Linux Mint 19.3, amd64
- Intel Core i7 8750H
- 16 GB of DDR4 RAM

I have tested both 5.8 AppImage release from official site and 1:5.6-git20200218-1794~ubuntu18.04.1
 version from PPA https://launchpad.net/~dhor/+ppa-packages  and have got completely same behavior. 

Following screenshots are demonstrating RT reaction on DNG file with a slightly modified Auto-Matched Curve Low ISO and Neutral profiles. The last one is Darktable, which opened DNG correctly.

![Screenshot at 2020-02-19 23-16-18](https://user-images.githubusercontent.com/36810599/74862745-7c96ff80-536e-11ea-9c96-d30ae80b7117.png)
![Screenshot at 2020-02-19 23-16-40](https://user-images.githubusercontent.com/36810599/74862740-7bfe6900-536e-11ea-900b-71d5cee65b1e.png)
![Screenshot at 2020-02-19 23-16-53](https://user-images.githubusercontent.com/36810599/74862735-799c0f00-536e-11ea-97b8-de483c046271.png)

Here is the DNG file I'm trying to process
https://www.dropbox.com/s/gs5s0i4wug24maa/IMG_5431.DNG?dl=0

Thanks in advance :)

Beep6581:
@madarexxx which version of RawTherapee opened it correctly?

Beep6581:
The file you uploaded is not the same as the file in your screenshots, but the problem looks the same.

Beep6581:
Confirmed that it looks correct in 5.7.

Thanatomanic:
Bisected it to this seemingly unrelated commit https://github.com/Beep6581/RawTherapee/commit/81f9ab73152c8896fd59c37e3adb1b8e3d563563

heckflosse:
It also fails in ART, where I stole this from ;-)

heckflosse:
@Thanatomanic 
[...]

PR #6570, opened 2022-08-27
Canon 1000D White Level Regression Fix

Lawrence37:


Thanatomanic:
Looks good and doesn't seem to give issues for other Canon files I tested. It also still fixes the original bug in #5658. Feel free to merge!

PR #6571, opened 2022-08-28
Fix for Nikon D5100 black image regression

### LINKED ISSUE #5654: Regression: NIKON D5100 14-bit uncompressed raw files open black - Corrupt data near ...
Nikon NEF file made with Nikon D5100 open as if they were completely black. In the file browser they are displayed correctly, but after opening a file in the editor by enter, the image becomes black. After that, they display as black rectangles also in the file browser. This happens only after the update to version 5.8. Version 5.7 and before worked correctly with the same raw files. As far as I can determine, this happens to all of my NEF files (several of which were edited with previous versions of Rawtherapee). I attached screenshots and uploaded a sample NEF file with an autogenerated pp3 file to: [here](https://drive.google.com/open?id=1ycFVH3q0iOQBffita41E1UWUnuOLHZXf)
After opening the file for editing, the following error message appears in the terminal: 
```
/home/duri/Képek/2019_tel/rtbug/new/newest/DSC_9677.NEF: Corrupt data near 0x20879a0
/home/duri/Képek/2019_tel/rtbug/new/newest/DSC_9677.NEF: Corrupt data near 0x20879a0
/home/duri/Képek/2019_tel/rtbug/new/newest/DSC_9677.NEF: Corrupt data near 0x20879a0 
```

I'm using linux (Manjaro, kernel 5.4.18-1 but I could reproduce it with openSUSE 15.0 (linux 4.12.14)), and tried to use the official packages, as well as compiling Rawtherapee 5.8 from source. 

RT-browser-screenshot
RT-editor

Lawrence37:
This removes the last row containing corrupt data instead of removing 2 for an even image height. It preserves more pixels which I believe is in the spirit of RawTherapee.

Closes #5654.

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -9763,6 +9763,8 @@ void CLASS identify()
      if(!dng_version) {top_margin = 18; height -= top_margin; }
   if (height == 3014 && width == 4096)	/* Ricoh GX200 */
 			width  = 4014;
+  if (height == 3280 && width == 4992 && !strncmp(model, "D5100", 5))
+    { --height; } // Last row contains corrupt data. See issue #5654.
   if (dng_version) {
     if (filters == UINT_MAX) filters = 0;
     if (filters) is_raw *= tiff_samples;


Thanatomanic:
Thanks for the fix, works!

Issue #6384, opened 2021-11-22
Return of the loop vectorization compiler issue using gcc 11.2.0 on Windows MSYS2


Thanatomanic:
I updated my MSYS2 installation today and lo and behold, the [`-fno-tree-loop-vectorize` compiler bug is back](https://github.com/Beep6581/RawTherapee/issues/5749). This time with a green color cast instead of magenta.

**To reproduce**
Fully update MSYS2, compile RT `dev` branch, open any Bayer file, apply Neutral, toggle 'Auto-correction' for RAW Chromatic Aberration Correction. Voilá, green pixels.

![image](https://user-images.githubusercontent.com/6567747/142936465-855ae06e-2e83-48e1-8c5d-d792d9f1f340.png)
[AboutThisBuild.txt](https://github.com/Beep6581/RawTherapee/files/7584446/AboutThisBuild.txt)

**Temporary solution**
Building with `-DCMAKE_CXX_FLAGS='-fno-tree-loop-vectorize'` removes the issue.

Can anyone confirm first before we may need to head back to the people upstream?
@heckflosse

heckflosse:
I can not confirm the green cast, because my msys2 gcc 11.2 RT builds crash as soon as I open a raw :-(

Thanatomanic:
I cannot reproduce any longer... not sure what was going on. Closing for now.

heckflosse:
@Thanatomanic 
> I cannot reproduce any longer... not sure what was going on. Closing for now.

But I can now reproduce it when built with gcc 11.2.0 ...
![grafik](https://user-images.githubusercontent.com/13733487/144852304-8becad46-1ec6-43f1-aa0f-a18c93c099f6.png)

Reopening...

heckflosse:
@Thanatomanic I already work on narrowing it...

Edit: It indeed looks like the old gcc issue, but this time with doubles instead of floats. Maybe they fixed it only for single precision floats... or they reinvented the bug because their regression test did only cover single precision float, who knows.

Anyway, I'm on it...

heckflosse:
@Thanatomanic Applying this silly patch solves the issue (not that we want to apply it) and clearly shows, that there's an issue in gcc 11.2

```diff
diff --git a/rtengine/gauss.cc b/rtengine/gauss.cc
index 99201a860..a27fc5a9c 100644
--- a/rtengine/gauss.cc
+++ b/rtengine/gauss.cc
@@ -19,7 +19,7 @@
[...]

PR #6583, opened 2022-09-18
Set -ffp-contract=off compiler flag for GCC >= 11 (#6384)

Lawrence37:
Replace the previous workaround of setting -fno-tree-loop-vectorize for a GCC optimization bug.


Thanatomanic:
I think this makes sense! 👍

blitzgneisserin:
I think the issue is solved now but it wouldn't hurt if someone else tested this fix tool

Lawrence37:
The GCC version check is working for the automated builds and I can confirm this change still fixes the green image issue.

Thanks @blitzgneisserin for testing the performance.

Merging now.

PR #6585, opened 2022-09-24
Color Appearance and lighting : Remove cat02preset and all dependencies issue 5664

Desmis:
Hello 

As per @Beep6581  request in issue 5664, I have removed anything that could refer to "presetcat02" in Color Appearance & Lighting".

I remind you that **I am against this deletion**, nevertheless I do it so as not to delay anything in the release of 5.9.

I return to my quote in this issue:

_[@Beep6581](https://github.com/Beep6581)
This is your point of view, not mine. It turns out that we don't have the same conception (to put it mildly) of complex subjects...
So tomorrow when I have time, I delete...
Another detail, as you have been away for a long time, you are probably unaware of my state of health...I have been (very) ill for more than 2 years now. Currently I am on radiotherapy combined with chemotherapy...and the future is more than unlikely. I could have a lack of dynamism, even burnout...especially in front of inaction, but I try by all means to keep my head above water. Fortunately and finally it seems to be moving forward.
Jacques_

I don't want to dwell on my state of health which began to deteriorate around 2017, I'm not looking to make myself complain, but simply to explain certain "whys"...

For me the most important is the difference in the way of seeing (and therefore of doing) the evolutions of Rawtherapee. 
For a majority of people (and there is no value judgment), compliance with the rules, procedures (which emerge from an application process) is essential. 

For others (minority), the steps of adaptation and creation are essential associated with a necessary pragmatism. Since the beginning of my professional life, I have essentially been in this type of approach which has always caused problems with my professional entourage (misunderstandings, conflicts, sidelining, etc.). I t's essentially the same thing with RT and I believe that rather than seeking to oppose, we must seek complementarity.

Also, I'm not a computer scientist, my code is bad and by evolutionary design, necessarily "never ready". I saw myself being asked several times to split into 2 or 3 pieces of code which for me was an inseparable whole - which from my point of view is a mistake...

Fortunately there is a team of which a majority of members supported me and without which nothing would have been possible.
It is enough to see the excellent text produced by @patdavid-test Pat David to present the main evolutions of rawtherapee for 5.9. Thanks to him.

https://rawtherapee.com/2022/07/rawtherapee-5.9-wip-and-project-updates/

if no one objects, but everyone can do it, I wil merge tomorrow.

Note: My computer is very old (like me) and very sick (like me).  It heats up a lot and often crashes for no apparent reason. I ordered a new model, I hope to be delivered before the next hospitalization;
jacques


Thanatomanic:
I see this is prompted by @Beep6581's remark about leaving commented code in the software. I generally agree with him, because, over time, people tend to forget why things are commented and forget that it once was part of an unfinished plan. However, I do not think it is necessary to remove it, certainly not _now_, and _most definitely_ not because out of a grudge.

There are many more places in RawTherapee with 'bad' code, simply because many people with different styles contributed over the years and because it is a complex system and nothing is ever perfect. I would argue that there are more urgent issues with the software than to enforce a certain code quality - although improving code quality is certainly something we need to do in the future. So, I would not merge this PR.

Lawrence37:
Deleted code can be recovered with a git command, so it's not a big issue if we need the code in the future. My concern with this PR is that we are making non-trivial code changes so close to the release (although I did review the code and it looks good). It is sufficient to remove the entries from `default` for now.

Desmis:
@Thanatomanic  @Lawrence37 @Beep6581 

I will create a PR with only remove the 3 labels from default
HISTORY_MSG_CAT02PRESET
TP_COLORAPP_PRESETCAT02
TP_COLORAPP_PRESETCAT02_TIP

Jacques

Entropy512:
I agree about the concerns about non-trivial changes shortly before release.

If I am understanding this correctly, it is merely for a preset?  If it was decided to remove it after 5.9, would it break any existing edits?  e.g. would it only affect the UI?

If removing it later would not break anything, maybe we can put it as a proposed cleanup when there's less risk of breakage?  But if formally releasing it would mean that maintenance was required for the rest of time, that's a different story.

Beep6581:
I agree to not merge this until after 5.9 is released.

Desmis:
Hello
In fact, removing it before or after doesn't change anything to the GUI or functionality.
The advantage of keeping it as it is is to maintain true compatibility with CIECAM in symmetrical mode. To "simplify" the use, I disabled this part of the code....But it is very likely that one day someone will (re)request this feature.

Note: people who fully master CIECAM are very rare (I modestly think I am one of them). To my knowledge Rawtherapee is the only (??) software that offers a complete solution whether in "main" or in "LA".

In french we say "la critique est aisée mais l’art est difficile"  (Philippe Néricault 18th century)

I agree also to not merge this until after 5.9 is released

Jacques

Desmis:
Some (messy) thoughts on Ciecam, Colored Appearance Models (CAM), Chromatic Adaptations (CAT) and semantics

First, I do not claim to have the truth. This is complex, subject to everyone's interpretation and the origin is old (several decades, when French researchers worked on the difference in a cartesian conception of colors, contrasts... and taking into account account of human senses: sight, brain,... hence the term CIE – Commission Internationale de l'éclairage).

Originally (about 80 years ago) was born the concept of XYZ and its variants, notably xyY.
The great surprise of the time was that it was necessary to enter negative values (we called this in 1931) "stimulus", but moreover it depends on a series of other parameters such as the observer (with "arbitrary" choices 2° and 10°), and also mathematics in the sense of matrix calculation which does not make things simple either to read or to explain.

In short, I pass the steps and about 45 years ago "Lab" was born, which is part of a CAM (colored appearance model), but which is not one. Its original purpose was to measure color deviations (deltaE : in this case it is relevant). This is much less true for the rest, in particular the color drifts when increasing the saturation (this is compensated in RT by the Munsell LUTs). 

But Lab (or its LCH variants) has become a model notably promoted by Adobe in Photoshop and used by many software programs (including Rawtherapee, Darktable). Lab is not "bad", quite the contrary, but you have to know its limits... The main one (which is not linked to Lab but to the history of cathodic monitors) is that at the time we didn't know how to make monitors with more than 120cd/m2... Here we enter the chicken and the egg debate, but the fact remains that the piloting of the monitors remained as it was at the time coded on 8 bits... From there to say it's Lab's fault, it's a very bad cognitive inference.

En 1997 was born CIECAM97 which was the first model taking into account human perception (simultaneous contrast, surround, etc.).

In 2002 was born CIECAM02, full of bugs, but which has the merit of being explained with "formulas" more or less recognized by a majority of people.

By simplifying from XYZ we act on a non-symmetrical model taking into account the shooting conditions, and those where we will look at the result (screen, TV, printer...) with the environmental conditions (is it dark or light...). 

In summary we start to calculate parameters related to the shot, where often it is only the person who took the photo who knows... We introduce several concepts:
CAT scene chromatic adaptation, absolute luminance, average luminance and surround

Then after many intermediate calculations, to take into account the physiological part of man, we use a conversion matrix, which can be called CAM.
Of course after there is the reverse approach (with iCAM) and all the parameters of Viewing conditions (CAT, absolute luminance, average luminance and surround..).

In 2016, a big simplification was made making CIECAM16 with its CAM and iCAM matrices as well as many intermediate calculations removed or simplified...
Moreover, in order to be able to use a satisfactory chromatic adaptation, the "symmetrical" model was born.... Clearly nothing that can simplify a Cartesian vision... since nothing is Cartesian.

In the meantime, the notion of HDR has appeared (a rather vague notion which has brought a multitude of more or less recognized concepts)
Some have quickly had the reflex to say Lab is to be put in the trash, and CIECAM is one of the antiques. Certainly not everything is perfect but in 95% of cases, Lab and CIECAM16 will satisfy everyone with a very good rendering.

Several concepts have been born recently including Jzazbz and ZCAM (derived from Jzazbz). The latter ZCAM is totally unusable, full of errors...
As for Jzazbz (under its variant JzCzHz) is not a CAM because it does not take into account the parameters mentioned above (like surround, etc.)

The development of JzCzHz was enamelled with (very) many difficulties and with the help and the many tests of Wayne and Jacques, we "agreed" on a semantics (in order not to scare, and to simplify the vocabulary) . In addition, it had to "fit" into the space available at the GUI level. Hence the current names (Color appearance), even if some concepts are foreign to it.

Moreover I added my grain of salt in front of the dysfunctions of Jz, by introducing a rather personal interpretation, but which I think is not false to better take into account HDR and SDR as well for CAM16 as Jz.

So of course everyone can criticize it's very easy, but as I said in another post “La critique est aisée, l'art est difficile”.

Of course we can develop the vocabulary, and also the methods of calculation.... If I had to summarize:
* CIECAM in main, is complex. But is essential in many cases
* Cam16 in ”LA” seems intuitive to me and brings good results
* JzCzHz is complex...but works

The main problem to solve is to modify the output (monitor) of RT so that it supports HDR mode... For my part, I don't know how to do... because we use LCMS and 8-bit encoding. ...But maybe someone knows....Knowing that HDR monitors are (as of today) out of budget for 95% of users.

Jacques

Entropy512:
HDR monitor output is going to be a problem for at least a few more years.  The methods of doing it are highly platform-dependent at the moment and nonexistent on Linux.

It is why, for example, you basically need an Ultrastudio or Decklink hardware unit to do HDR video output from DaVinci Resolve - OS color management has been such a mess for so many years that Blackmagic just gave up and said "you need dedicated hardware we designed to bypass everything".

Desmis:
After reflexion I propose to restore this PR as it is, that is to say delete the unused code.
If it becomes necessary one day, I may rewrite the necessary functionality in another way.

I remove from "draft" and put the tag 5.10

Jacques

ValZapod:
>is why, for example, you basically need an Ultrastudio or Decklink hardware unit to do HDR video output from DaVinci Resolve

Just need to connect mpv to it somehow. You can already conenct mathematically lossless x264 encoder to it through Voukoder. Still WCG works already, no support for Adobe RGB is hillarious though, but still you can 3DLUT it.

And they just added Dolby Vision profile 5 support tjat uses IPTPQc2 color space, only for MacOS for now. So it is fast evolving.

Thanatomanic:
@ValZapod Please stop commenting if you do not contribute to resolving the actual issue. This is going off-topic again on something extremely specific.

Desmis:
@Thanatomanic 
Roel, I totally approve of your intervention. :)


Just a parenthesis: the HDR part is only a (small, but important and complex) part of the Color Appearance Models (CAM).

Jacques

ValZapod:
I am trying to be helpful. Dunno why you do not like it, and I understand you do not know much about video transfer functions which are different from photo. Still, I am not going to shut up if I see some wrong, horribly wrong info, like 2.4 gamma stuff.

As for HDR, all displays that are above 100 nits are HDR, and that is all of them. All.

Thanatomanic:
I did not like the comment because this issue is not about HDR and who supports what, but about how to go forward with having a modern color adaptation model in RawTherapee. Your comment does not add anything to that discussion, because it was only specifically aimed at a remark by @Entropy512.

PR #6587, opened 2022-09-24
Fix Windows build libpcre DLL error

Lawrence37:
Updates the PCRE DLL file name that need to be bundled with the executable.


Beep6581:
Seems good to merge.

PR #6593, opened 2022-10-01
Cleanup of obsolete language keys in translations

Thanatomanic:
When trying my hand at updating the Dutch translations, I noticed that the translation files typically still contain very old or obsolete language keys. I have therefore inspected `procevents.h`, looked at the `obsolete_` keys and removed those from all translations. All these keys were already absent in `default`.

For reference, the obsolete and therefore removed keys:
```
HISTORY_MSG_2
HISTORY_MSG_16
HISTORY_MSG_17
HISTORY_MSG_18
HISTORY_MSG_44
HISTORY_MSG_45
HISTORY_MSG_46
HISTORY_MSG_47
HISTORY_MSG_55
HISTORY_MSG_67
HISTORY_MSG_78
HISTORY_MSG_83
HISTORY_MSG_112
HISTORY_MSG_125
HISTORY_MSG_131
HISTORY_MSG_132
HISTORY_MSG_133
HISTORY_MSG_134
HISTORY_MSG_135
HISTORY_MSG_136
HISTORY_MSG_209
HISTORY_MSG_250
HISTORY_MSG_274
HISTORY_MSG_275
HISTORY_MSG_300
HISTORY_MSG_409
HISTORY_MSG_473
HISTORY_MSG_604
HISTORY_MSG_848
```


Thanatomanic:
A point of discussion is what to do with the history language keys now marked `--unused--`. These keys are already either commented out in `procevents.h` or marked obsolete.

As far as I understand the code, these could be removed without loss of functionality. It alsodoesn't really make sense to keep them in... Does it? Some thoughts @Beep6581 @Entropy512

Entropy512:
They aren't commented out in procevents.h - there are still placeholder entires (with comments indicating they are a placeholder for an obsolete/removed entry) in both procevents.h and refreshmap.cc - because deleting the entries would require renumbering all of the history events.

Technically you could remove the keys from the history events in default, but that would cause the history events part of default to not be as many lines as the tables in procevents.h and refreshmap.cc.  See the comments by @Lawrence37 and @heckflosse regarding the spacing at the top of https://github.com/Beep6581/RawTherapee/pull/6466/commits/3bfcc66e465b8438fc2b76a8ed7f1c83d6f7107e#diff-349153db3b16da71194b5d6e16784b85423b5002b682546f806fef2e2d78a8ff

(bah, their comments don't seem to show if you follow the link?)

Thanatomanic:
I see, what a ridiculously hard system to maintain 😄 
I hope we can rework this in the future...

Beep6581:
Looks fine to merge.

> these could be removed without loss of functionality. It alsodoesn't really make sense to keep them in...

Seems that way to me. The history entries in `default` are already non-contiguous, so if that's not a problem then I see no reason to keep the `--unused--` keys.

Issue #6567, opened 2022-08-26
Linux automated build needs upgrade (to work with GitHub Actions)


Thanatomanic:
In a similar vein as #6566, we also need to update our Linux Automated build to something more current to ensure nightly will be built.

See https://github.com/actions/runner-images/issues/6002 for more information.

Lawrence37:
The upgrade can and should wait until after the release. In the long term, we should containerize the build with Docker, just like with the old automated builds.

Some context:

It is recommended to package AppImages in the oldest LTS operating system version possible. The reason is because AppImages are not completely self-contained and rely on host system libraries that should already exist in every Linux installation. From https://docs.appimage.org/reference/best-practices.html:
> Some core libraries, such as glibc, tend to break compatibility with older base systems quite frequently, which means that binaries will run on newer, but not on older base systems than the one the binaries were compiled on.

Therefore, it's best to use the oldest operating system that we plan to support.

Ubuntu is a popular Linux distribution with long-term-support (LTS) versions released every 2 years. They reach end-of-life after 5 years. GitHub Actions supports the latest two LTS versions of any operating system they offer. We have reached a point when GitHub is removing the Ubuntu 18.04 release from Actions, but that operating system will not reach its end-of-life until April of next year. If we move the AppImage builds to Ubuntu 20.04, we risk breaking the AppImage for those who are still on Ubuntu 18.04 (and possibly other OSs). Fortunately for us, the 18.04 runners will still be available (with brownout periods) until the end of November.

We can use the current automated builds for the 5.9 release, then upgrade the runner to Ubuntu 20.04. The next step is to containerize the build so we have more control over the build environment.

PR #6605, opened 2022-10-25
CI: Update GitHub actions environments and used actions

EwoutH:
A bit of maintenance on the CI:
 - Updates the used actions to their latest versions.
 - Update the images to the `-latest` tags.

This currently means Ubuntu will use 20.04 LTS (and soon 22.04 LTS) and macOS will use macOS 12. This is useful since that means tooling is upgraded automatically, as are check that RawTherapee works well with recent OSes.

Also adds a `workflow_dispatch` trigger to the main workflow (the others already have it).


Thanatomanic:
This has already been discussed elsewhere, but we won't merge this until after the 5.9 release. 

Also, updating CI to comply with recent OSes must not mean the old ones suddenly become incompatible. I think it is easier to assume a modern OS has some form of backwards compatibility with their predecessors, than assuming everyone will always use the latest updates.

EwoutH:
Thanks for the quick reply! That's perfectly fine, and not wanting to change things before a release is good practice.

The macOS upgrade can't wait too long however, GitHub Actions has already deprecated the environment and is now doing rolling brownouts. As you can see, [some builds](https://github.com/Beep6581/RawTherapee/actions/workflows/main.yml?query=branch%3Arelease-5.9) already failed because of that.

> This is a scheduled macOS-10.15 brownout. The macOS-10.15 environment is deprecated and will be removed on December 1st, 2022. For more details, see https://github.com/actions/virtual-environments/issues/5583

Furthermore, almost all build tools have backwards compatibility options. In the macOS build `-mmacosx-version-min=10.15` is used for example. Using the latest build tools is best practice, they contain the latest optimizations, bug fixes and most importantly security mitigations.

Lawrence37:
The AppImage build environment should use the oldest active LTS version of Ubuntu to maintain compatibility with older systems. See my comment [here](https://github.com/Beep6581/RawTherapee/issues/6567#issuecomment-1229066766).

Note that while older operating systems like Ubuntu 18.04 or 20.04 may stay with lower build tool versions, they update many with back-ported bug fixes and security improvements.

Benitoite:
- PR #6613 macos autobuild now uses the macos11 *Big Sur* image. This is the oldest macos version currently supported by the CI. 
- The previous 10.15 *Catalina* is now deprecated, but would have been preferred due to backwards compatibility.  (10.15 itself will run on 2012 macs.)
- Big Sur starts on 2013 macs.  Monterey starts on 2015 macs.
#### The `-latest` tag should not be utilized on the macos autobuild for these reasons.

Beep6581:
@EwoutH thank you for the PR. However, I too am under the impression that using `latest` could result in an unnecessary lack of compatibility with older distros. If I'm wrong please correct me. If the other changes are compatible with the non-latest distros then we could merge this if the `latest` chunks are dropped from the PR.

EwoutH:
First of all, congratulations on the [5.9](https://github.com/Beep6581/RawTherapee/releases/tag/5.9) release!

I rebased this PR, but parts of it where already updated in the mean time, so it has become a bit smaller.

The main question is if you want to maintain a push or pull model for CI environments; push meaning using the `latest` versions, and getting a new default image when it's mature, and pull meaning defining the version explicitly (i.e. `22.04`)

As for updating the build images, it will result in newer built tools being used (by default). This means practically:

- For Linux, this will move from the Ubuntu 18.04 image (GCC 7) to 20.04 (GCC 9) and [soon](https://github.com/actions/runner-images/issues/6399) to 22.04 (GCC 11).
- For macOS, this will move from macOS 11 (Xcode 13) to [soon](https://github.com/actions/runner-images/issues/6384) macOS 12 (Xcode 14).
- For Windows, it will stay the same, since Windows is already on the Server 2022 image.

Updating GCC and Windows images is generally recommended, since their toolchains have very long periods of backwards compatibility, while also supporting newer features and platforms (for example Windows Arm64).

For macOS the situation is a bit complicated right now, since they're transitioning from architecture, and dropping backwards compatibility faster then usual. So on the one hand you want to use current versions to support the newest platforms, and on the other hand you want a version that supports old platforms to a certain extend.

Xcode 13 supports macOS 10.9 to 12 as deployment targets, Xcode 14 macOS 10.13 to 13.

macOS 10.13 (High Sierra) supports the following devices:

> - [iMac](https://en.wikipedia.org/wiki/IMac_(Intel-based)#3rd_generation:_Unibody) (Late 2009 or later)
> - [MacBook](https://en.wikipedia.org/wiki/MacBook#Discontinued) (Late 2009 or later)
> - [MacBook Air](https://en.wikipedia.org/wiki/MacBook_Air) (Late 2010 or later)
> - [MacBook Pro](https://en.wikipedia.org/wiki/MacBook_Pro) (Mid 2010 or later)
> - [Mac Mini](https://en.wikipedia.org/wiki/Mac_Mini) (Mid 2010 or later)
> - [Mac Pro](https://en.wikipedia.org/wiki/Mac_Pro) (Mid 2010 or later)

So the question becomes if for new RawTherapee versions, it's worth it dropping support for devices older than listed above, in order to support current and newer devices and macOS versions better.

Benitoite:
The more your drop the faster the default code is- minimum SDK for macOS 11 should be included in the build tools. Building for the test unit only it totally doesn’t matter on macOS.  On Nov 30, 2022, at 1:12 PM, Ewout ter Hoeven ***@***.***> wrote:﻿
First of all, congratulations on the 5.9 release!
I rebased this PR, but parts of it where already updated in the mean time, so it has become a bit smaller.
The main question is if you want to maintain a push or pull model for CI environments; push meaning using the latest versions, and getting a new default image when it's mature, and pull meaning defining the version explicitly (i.e. 22.04)
As for updating the build images, it will result in newer built tools being used (by default). This means practically:

For Linux, this will move from the Ubuntu 18.04 image (GCC 7) to 20.04 (GCC 9) and soon to 22.04 (GCC 11).
For macOS, this will move from macOS 11 (Xcode 13) to soon macOS 12 (Xcode 14).
For Windows, it will stay the same, since Windows is already on the Server 2022 image.

Updating GCC and Windows images is generally recommended, since their toolchains have very long periods of backwards compatibility, while also supporting newer features and platforms (for example Windows Arm64).
For macOS the situation is a bit complicated right now, since they're transitioning from architecture, and dropping backwards compatibility faster then usual. So on the one hand you want to use current versions to support the newest platforms, and on the other hand you want a version that supports old platforms to a certain extend.
Xcode 13 supports macOS 10.9 to 12 as deployment targets, Xcode 14 macOS 10.13 to 13.
macOS 10.13 (High Sierra) supports the following devices:


iMac (Late 2009 or later)
MacBook (Late 2009 or later)
MacBook Air (Late 2010 or later)
MacBook Pro (Mid 2010 or later)
Mac Mini (Mid 2010 or later)
Mac Pro (Mid 2010 or later)


So the question becomes if for new RawTherapee versions, it's worth it dropping support for devices older than listed above, in order to support current and newer devices and macOS versions better.

Lawrence37:
> The main question is if you want to maintain a push or pull model for CI environments

For Ubuntu and macOS, we definitely want to pin the OS version. I'm ambivalent about the others.

Beep6581:
This PR should go into 5.10-beta1, else closed. @EwoutH @Benitoite @Lawrence37 what are your thoughts?

Benitoite:
I'm fairly ambivalent on this for the following reason:

Due to the new macOS security arrangements (certified code signature, notarization, Apple Dev identification) that have progressively rolled-out since RT v5.8 and 5.9 in MacOS which is now available as high as 14-beta, the CI artifacts are not really useful beyond a verification that the base executable is *buildable* on a Mac.  The CI does not provide any information regarding the actual intended default *usability* of the bundle generated.

macOS-11 is the minimum system version on x86_64 that works with the app presently, and that is why it is currently used. On arm64 we can use macos-latest and get support for most of the Apple Silicon. However the CI runs on x86-64 and there is no presently supported cross-compilation in any of our scripting between x86_64 and arm64 macOS.

With being said, if this gets merged and something breaks in the macOS CI it should be very trivial to fix before release.

My personal preference is to leave macos-11 pinned on the x86_64 CI runner for as long as possible.  

(edit: my guess is that macos-11 will be depricated and later dropped when macOS 14 *Sonoma* is released, as Apple support for macOS only lasts for 3 major versions.)

Lawrence37:
I'm ok with upgrading the actions, but the Linux runner must be pegged to the oldest Ubuntu available.

PR #6608, opened 2022-10-31
Add raw_crop and masked_areas for Canon EOS R7 and R10

alexforencich:
Add raw_crop and masked_areas settings to camconst.json for Canon EOS R7 and R10.  raw_crop settings determined from R7 and R10 raw files. masked_areas uses the same relative offsets as the R5.


Thanatomanic:
Thanks!

PR #6611, opened 2022-11-18
Try to fix bug when using CBDL main and Local Sharpening

Desmis:
This PR does not solve the bug itself, but avoids the bug.

For this by default "CDBL main" is set to "After black-and-white".

But if the user chooses "Before black-and-white" and chooses in "Local adjustments" - "sharpening", then the process is changed to "after".

This PR can be done before 5.9 or after...

jacques

### DIFF
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -731,14 +731,14 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
         // Remove transformation if unneeded
         bool needstransform = ipf.needsTransform(fw, fh, imgsrc->getRotateDegree(), imgsrc->getMetaData());
 
-		int sharplablocal = 0;//detect if Locallab sharpening is active
+		bool sharplablocal = false;//detect if Locallab sharpening is active
         for (int sp = 0; sp < (int)params->locallab.spots.size(); sp++) {
 			if(params->locallab.spots.at(sp).expsharp) {
-				sharplablocal = 1;
+				sharplablocal = true;
 			}
 		}
 				
-		if (sharplablocal == 1) {
+		if (sharplablocal == true) {
 			params->dirpyrequalizer.cbdlMethod = "aft";
 		}
 
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -731,16 +731,6 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
         // Remove transformation if unneeded
         bool needstransform = ipf.needsTransform(fw, fh, imgsrc->getRotateDegree(), imgsrc->getMetaData());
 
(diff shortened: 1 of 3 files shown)


ghost:
Merged and tested on a JPG and a RAW:

This does fix the artifact issue mentioned on [pixls](https://discuss.pixls.us/t/ugly-artifacts-on-laptop-with-alder-lake-p-graphics/33620/) on my Linux box.

rtengine/improccoordinator.cc
```diff
@@ -731,6 +731,17 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
         // Remove transformation if unneeded
         bool needstransform = ipf.needsTransform(fw, fh, imgsrc->getRotateDegree(), imgsrc->getMetaData());
 
+		bool sharplablocal = false;//detect if Locallab sharpening is active
+        for (int sp = 0; sp < (int)params->locallab.spots.size(); sp++) {
+			if(params->locallab.spots.at(sp).expsharp) {
+				sharplablocal = true;
+			}
```
Beep6581: Maybe a `break;` inside the `if` to not check remaining spots unnecessarily?
Or ditch `sharplablocal` entirely and
```
        for (int sp = 0; sp < (int)params->locallab.spots.size(); sp++) {
            if(params->locallab.spots.at(sp).expsharp) {
                params->dirpyrequalizer.cbdlMethod = "aft";
                break;
            }
        }
```

Beep6581:
The problem here is that this doesn't fix the bug, as you noted, and it leads to a GUI which doesn't tell the truth:

screenshot_20221119_223515

screenshot_20221119_223534

After using Local Adjustments, I set CbDL to "Before" (RGB) as the GUI shows, but there are no artifacts when I zoom 1:1 so it must be secretly using "After" (Lab).

As "This is probably a very old bug" [[1](https://discuss.pixls.us/t/ugly-artifacts-on-laptop-with-alder-lake-p-graphics/33620/5?u=morgan_hardwood)], maybe it's better to leave it now and deliver a better fix for post-5.9.

Desmis:
Hello

I agree
I am examining how to solve the bug.. Maybe I have a track

Jacques

Desmis:
Hello
I push a change...I think it solves the bug...
I change "damping" (deconvolution sharpening) in Local adjustments. 
By default "damping = 0", I change to 1, if CBDL is used with "before black-and-white"....
The impact of "damping" on the image when going from 0 to a low value seems negligible (usable in "advanced" mode)

Jacques

Beep6581:
Hey @Desmis , that seems to work well, I can't reproduce the artifacts. Fine to merge to dev, and I'll cherry-pick that into 5.9.

Desmis:
hello
@Beep6581 , thank you for your evaluation.
If no others remarks, I will merge, late in the afternoon

Jacques

PR #6615, opened 2022-11-25
Lensfun DB directory fallback

Lawrence37:
If the Lensfun database fails to load using the path set in the `options` file, this will try to load the database bundled with the application (the path set with the CMake option `LENSFUNDBDIR`). No fallback occurs if the bundled location is not set or is the same as the path in `options`.

This PR should fix the [issue some 5.9-rc1 testers found](https://discuss.pixls.us/t/rawtherapee-5-9-rc1-ready-for-testing/33631/66) in which a [blank database directory in `options`](https://discuss.pixls.us/t/rawtherapee-5-9-rc1-ready-for-testing/33631/76) causes the profiled lens corrections to have no lenses available.

### DIFF
--- a/rtengine/init.cc
+++ b/rtengine/init.cc
@@ -58,10 +58,20 @@ int init (const Settings* s, const Glib::ustring& baseDir, const Glib::ustring& 
 #pragma omp section
 #endif
 {
+    bool ok;
+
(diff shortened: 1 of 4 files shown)


Thanatomanic:
At first glance this seems perfectly reasonable, thanks Lawrence! I have to time to actually test though...

Beep6581:
I won't have time to test, but sounds good.

PR #6616, opened 2022-11-27
Improves sharp in LA to avoid long processing if user chooses fullimage

Desmis:
I think this improvment can wait 6.0

Jacques


Beep6581:
@Desmis would you like to merge?

Desmis:
@Beep6581 
Yes, I wish. This is (hopefully) the last process which takes a huge amount of time if fullimage is selected.

Thank you

Jacques

Beep6581:
Then go ahead :)

PR #5889, opened 2020-08-18
Adopt Exiv2 for metadata handling

Beep6581:
Solves #3801 and many metadata-related issues.

PR #6631, opened 2022-12-04
Exiv2 updates

Lawrence37:


Lawrence37:
I addressed some of the comments I left in #5889 and cherry-picked metadata related commits from ART (excluding ExifTool).

Thanatomanic:
Excellent work @Lawrence37! This is a pretty heavy beast that requires careful testing. But I assume you would only want to do that after this PR is merged into `metadata-exiv2`, right?

Lawrence37:
Yes. The main branch has an incomplete implementation. It's just copying all the data to the output image. The commits here finish the metadata handling and add some additional features. I opened this PR to allow comments on which commits we want in the initial Exiv2 integration.

Lawrence37:
Any opinions on whether we should merge just the bare minimum for metadata handling or include some or all additional commits?

Lawrence37:
I intend to merge this to the main Exiv2 branch soon.

Issue #5769, opened 2020-05-20
Poor man dehaze (and remove uniform lens flare)


heckflosse:
I made some tries with a very simple method to remove haze (and uniform lens flare) from raw files. Well, at least partly. It's a good starting point to add the already existing dehaze tool in RT.

The method is super simple. It just calculates the mininum value from each raw channel and subtracts that value from the corresponding channel.

The idea behind this approach is:

If there is a region in the scene, which does not reflect enough light to the camera sensor, to be recorded by the sensor, then, if there is a sensor value > 0, this value must be caused by haze or flare and not by the scene.

Having this situation and assuming the haze or flare is somehow evenly distributed, we can just subtract it, and surprisingly, it works very well for long distance landscape shots, where there is a lot of haze between camera and subject, and maybe also a lot of flare (internal lens reflections) by using a long focal length which are more prone to internal reflections.

Of course we can not use this method for every image.
1) It does not work on high-iso files where the minimum value is always around zero because of noise
2) it does not work on underexposed files for the same reason

I just works for the conditions I described above.

Anybody interested in a branch? In first step the branch would just spit out the raw black levels for the `dehaze/deflare` in console.

Ingo

Floessie:
I already had the pleasure to play with Ingo's patch and it really makes a difference for the type of shots he mentioned. When it comes to me, yes, I'm interested! :smile:

Thanatomanic:
@heckflosse Will you integrate this with the current dehaze tool as a variant (like unsharp mask and deconv for sharpening)? Or will it become a separate module?

In any case, a branch would be nice to test.

heckflosse:
@Thanatomanic 
> Will you integrate this with the current dehaze tool as a variant (like unsharp mask and deconv for sharpening)? Or will it become a separate module?

Don't know at the moment where to put it. Maybe it will just become an automatic checkbox in raw black levels tool (because it does exactly that).

heckflosse:
Branch follows soon. Have to code three variants (Bayer, Panasonic Bayer where raw value 0 means a bad pixel, and xtrans...)

heckflosse:
I just created the `poor_man_dehaze` branch.

To test, you just have to look at the console output and fill in the values in ui by hand.
Example:

Console output:
```
red : 57
green : 155
blue : 112
```

ui:
![grafik](https://user-images.githubusercontent.com/13733487/82562364-5f441e00-9b75-11ea-8815-9ffe8bf2c7cb.png)

[...]

PR #6635, opened 2022-12-06
Add "Inpaint opposed" to Highlight reconstruction and improved Itcwb

Desmis:
For some time several people have (again) looked into the reconstruction of the highlights. It must be said that the problem is difficult, to reconstruct pixels that have disappeared (or are corrupted), there are even cases where it is impossible.

A team made up of members of Gmic and Darktable have developed "Inpaint opposed". Alberto Grigio @agriggio  ported it on ART, and I present it to you in Rawtherapee. Thanks to them.

**Inpaint opposed (IO)**
This method restores clipped pixels by looking at selected pixels that are next to clipped regions and uses the average of all of these pixels to estimate the proper colour of the highlights. This works well for most images, but images that have clipped areas of significantly different colours may not work as well.

Inpaint opposed (IO) is clearly a good algorithm, which in terms of results can be compared to "Color Propagation (CP)". It has the advantage of consuming less resources, around 40% less than CP.

On several images tried, sometimes, taking into account subjectivity, CP is the best, sometimes IO…

I think that by default "Inpaint opposed" can replace "Blend". This is what I propose, by putting this selection in "Procparams.cc", "Tonecurve.cc" and all the files of Profile pp3. Of course we can change and put back "Blend".

To facilitate the comparison I activated BENCHFUN both for "Color Propagation" and for "Inpaint opposed" (hilite_recon.cc)

Jacques


Thanatomanic:
Merci Jacques, interesting addition. I will try to take a look soon!

waynesutton50:
I tried out half a dozen problem images using both Inpaint Opposed and Color Propagation. The results varied depending on the image and the degree of clipping and while I think I generally prefer Color propagation there are instances where Inpaint Opposed seemed to do a better job, especially when Color Propagation fails as in the example below: 

**Color propagation**
image

Inpaint Opposed

image

I also found that in some cases you can use a lower level of highlight compression using Inpaint Opposed compared to Color propagation. 

The following example were based on the image used in this thread: https://discuss.pixls.us/uploads/short-url/lNREaRUPRBaUXDNC6Sa06qc4nRz.cr2
**Color propagation**
image

Inpaint Opposed
image

waynesutton50:
Trying it out on a few images it seems to be useful when there is very severe clipping and Color Propagation can no longer cope. e.g. 
Color Propagation
image
Inpaint Opposed

image

Or with the image in this thread:  https://discuss.pixls.us/uploads/short-url/lNREaRUPRBaUXDNC6Sa06qc4nRz.cr2 

Color Propagation
image
Inpaint Opposed
image

ghost:
Just merged this one and played with it for about an hour:

This is a nice addition to the methods already available and Inpaint Opposed seems to be able to "fix" things where Colour Propagation (and the other methods) fails.

It, IO, does produce decolourisation at times or banding, but that is to be expected. Each highlight issue needs its own method and there's no one fixes all solution.

I need to test this one a lot better/longer before I can say anything definitive about it. First impressions are positive, though.

About defaults: I use *Colour Propagation* as default with *Clip out-of-gamut colours* turned off. This gives me the best results at the moment, much better then the out-of-the-box  defaults. Haven't played with this new one long enough to decide if I would make it my preferred method.

BTW: I initially wasn't able to merge this one due to a merging conflict. Turns out that this one and #5769 (the not yet merged *poor_man_dehaze*) don't like each-other all that much:

```
Auto-merging rtengine/rawimagesource.h
CONFLICT (content): Merge conflict in rtengine/rawimagesource.h
Auto-merging rtengine/rawimagesource.cc
Automatic merge failed; fix conflicts and then commit the result.
```

Not sure if that is your problem, though. Just mentioning it in case.

Desmis:
@waynesutton50 @jade-nl 

Thank you for all these tests.
But in fact there is no miracle recipe.

I don't have "merge" problems, but since there is an added "linalgebra.h" file, this supposes to redo Cmake...Maybe that's what's happening to you.

Jacques

rtdata/profiles/Pop/Pop 2 Lab.pp3
```diff
@@ -1,4 +1,5 @@
 [Exposure]
+[Exposure]
```
Lawrence37:
```suggestion
```

Desmis: Well seen. I change and push the change

Thank you
Jacques

Thanatomanic:
In my tests I have found only a single case where IO works badly: https://raw.pixls.us/getfile.php/755/nice/Nikon%20-%20D800%20-%2014bit%20uncompressed%20(3:2).NEF

You get very ugly purple/blue banding that none of the other HR methods have.

image

Desmis:
@Thanatomanic 

Thank you for testing :)

Unfortunately, there is no miracle recipe. Reconstructing missing data is a significant challenge.
Nevertheless, the 2 algorithms "Color Propagation" and "Inpaint opposed" together manage to solve the majority of the problems.

Inpaint Opposed consuming less resources, I put it by default in the profiles... But we can change this choice...

jacques

Lawrence37:
In my experience, inpaint opposed is somewhere between blend and color propagation in terms of quality and robustness. Blend is very reliable but does not reconstruct much detail. Color propagation can produce great results, but is not very reliable. Inpaint opposed offers a good balance.

Lawrence37:
There must be some caching issue related to white balance. If you turn off white balance, turn on highlight reconstruction, then turn white balance back on, you get this:
image
Toggle the highlight reconstruction off and on and the result changes:
image

Desmis:
@Lawrence37 

Thank you for testing :)
You have found a malfunction, certainly rare to use, but it is a bug.
Maybe this modification (adding "ctemp.getTemp() >= 0"  when Inpaint Opposed is used) allows to circumvent the problem which appears only with IO.
I will push the change quickly

Thanks

Lawrence37:
The bug appears with any white balance change.

Start with Neutral, drop exposure to -1.0, set white balance temperature to 4000, enable inpaint opposed, and reset white balance to camera:
image

Toggle highlight reconstruction off and on:
image

Desmis:
@Lawrence37 

Unlike other methods "Inpaint opposed" needs a white balance to work...
It must be somewhere in the code an "event" problem... not easy to find..

Thank you again

Jacques

Thanatomanic:
There is still something not right in combination with the WB. The latest commits did not solve the issue. ART has the same problem btw.

To reproduce:
1. Load a RAW, apply Neutral
2. Enable HR Inpaint Opposed
3. Change Temperature or Tint by a large amount (lower values seem to have a bigger effect)
4. Disable and enable HR
5. Observe a clear difference in colors before and after

If the IO algorithm is WB dependent, maybe it should always be performed with fixed multipliers? Or perhaps it should always be re-triggered once the WB is changed?

Desmis:
@Thanatomanic 
As I said above, "inpaint-opposed' takes white balance into account, which brought the rawimagesource.cc rewrite. A bad WB brings bad results.
Unlike Color Propagation.
I modified the events to try to take into account the WB changes, but that doesn't make much sense...
You can review what i did to "events" in improccoordinator.cc, refresmap, procevents.h and make changes if i made a mistake.
But **good WB is essential**
Happy new year
Now, the family has arrived, I will be in difficulty to continue.

Jacques

Lawrence37:
The latest commits seem to have solved the WB issue. I suppose the refresh map changes can be reverted?

Lawrence37:
The latest commits seem to have solved the WB issue. I suppose the refresh map changes can be reverted?

Desmis:
@Lawrence37 
Thank you for  testing :)

I reverted the refresh map...it's better, faster response

But I found a malfunction, with the current code "getimage" is called several times with "Inpaint opposed", hence resource consumption.
I modified the "getimage" function (there are many files where it is present), adding a variable 'opposed' (default 0, 1 with Inpaint Opposed ) to only call 'Inpaint Opposed' once.

jacques

jenshannoschwalm:
> In my tests I have found only a single case where IO works badly: https://raw.pixls.us/getfile.php/755/nice/Nikon%20-%20D800%20-%2014bit%20uncompressed%20(3:2).NEF

Unfortunately downloading this gives a small tiff not the raw :-( Would like to test this image ...

Thanatomanic:
@jenshannoschwalm This is a problem of raw.pixls, or your browser, or a combo I'm not sure. In fact it is the NEF, just rename the extension.

jenshannoschwalm:
Thanks.

BTW I will push a PR with faster masking operations (as we know the dilating size we can avoid the generic Algo and borders) to dt the next days, also reducing memory footprint.

jenshannoschwalm:
> You get very ugly purple/blue banding that none of the other HR methods have.

@Thanatomanic thanks for the image, indeed one of those problematic pictures. The reason is pretty simple if you understand the algo, btw the dt implementation has the same problem with "inpaint opposed" - segmentation would be the way to go.

> This method restores clipped pixels by looking at selected pixels that are next to clipped regions and uses the average of all of these pixels to estimate the proper colour of the highlights. This works well for most images, but images that have clipped areas of significantly different colours may not work as well.

I think the above is either a misconception or at leat somewhat misleading. I would describe it as:

1. For every blown photosite location find the means of the other channels in a 3x3 box.
2. replace blown data with above value **plus** some global, magic chroma correction.

The chroma correction is calculated from locations "around" blown out areas in the whole image. This simplification is good for most lightning situations. If you want to avoid the artefacts you need some "local" correction --> use segmentation algo.

Desmis:
@Thanatomanic @Lawrence37 @waynesutton50 

@jenshannoschwalm 
Thank you for this insightful review.

The RT version has the same characteristics as the ART version, and this is normal, since the code is almost identical.

Even though I'm a (very) bad computer person, I saw that the "chroma" part was with differences from that of DT.
I'll look, but:
* from Monday I resume my treatments for the disease (stopped during the holiday season). Today i'm leaving Paris to go home
* I have trouble understanding the DT code…

Nevertheless, this simplified version of "Inpaint Opposed" is better than "Blend" in a majority of cases, and often a little worse than "Color Propagation" (origin Emil Martinec).

Jacques

Desmis:
As I said above I don't feel capable of adapting the "chrominance" part of DT.

Nevertheless, on the evoked image (Nikon - D800 - 14 bit uncompressed(3_2).NEF) **it is enough to modify in "White balance" - "Tint" about 1.0 and all is good**. We can if necessary with other tools then (Local adjustment, Abstract profile, channel Mixer,...) retouch the global colorimetry if necessary (it is not useful in this case). Just try **White-balance - automatic -** **Temperature correlation.**

So in a very large majority of cases "Inpaint opposed" works very well...and there is always "Color Propagation" which also works very well.

I propose to merge this PR, any improvements can be made if necessary later

Jacques

Desmis:
Note that the White Balance automatic method, "Temperature correlation" only works correctly with 2 majority illuminants : Daylight and Blackbody
Temp : between 2000K and 12000K
Tint (green) : between 0.4 to 4.0

In other cases : fluorescent, LED,... the results are unpredictable

jacques

Lawrence37:
@Desmis Are the two latest commits meant to be in this pull request? They look unrelated and are similar to some changes in #6643.

jenshannoschwalm:
@Desmis There is now https://github.com/darktable-org/darktable/pull/13254 implementing a better solution for the "chromacity stuff" that should be easier to understand & port :-)

Desmis:
@Lawrence37 
Yes, those 2 commits are part of the PR. Indeed "Inpaint opposed" takes into account - and this is an essential element - the whole of the white balance which is not limited to the temperature. There are other criteria at least as important as "green / tint" and the CRI (Color Rendering Index which is good for Daylight and Blackbody, but can be very bad for example F illuminants).

Itcwb (temperature correlation) that I developed (assisted by Ingo to optimize the code) in 2018, should solve the problem of the right "WB", by acting on Temperature and Tint to find the best compromise. At the time I had imagined 9 parameters, present in 'option' to optimize the algorithm.

Several (new) tests and the adaptation to "Inpaint Opposed", lead me to change some default values (those of the commit). The algorithm will thus work better, if the "Camera" settings are not good, and/or if there are few colors to sample.

As it seems to me:
* very complex at the code level (at least for me) to integrate these settings in the main menu (9 parameters which the user will not understand anything in general)
* burden the GUI of White-balance to modify (if any 1 parameter)

This is why I leave them in "option" (it will be necessary to update the documentation) and I modified the name of a variable to be sure that it is taken into account by default.

As the PR #6643 modify these same values,  I only harmonized the code, adding (at least) 1 parameter to the GUI interface, which would not update the user 'options' (even by merge) file, hence the same modification, but with GUI.

I'll re-examine Itcwb's code because it doesn't always behave well when there are few colors to compare with Student's law.
For #6643 ,  I maintain a) either we leave it as it is (Observer settings, Itcwb settings... that the user can modify manually in 'options') b) or we put it in GUI "options'....this which seems more relevant to me, and even though it says 'options' should not modify the image (virtually everything in 'Preferences/ Color Management' alters the image.

@jenshannoschwalm 
I saw that the DT code for this part was simpler, nevertheless still too complex, for my old brain :)
Thank you  for your attention to this PR and your advice.

Jacques

Desmis:
Hello, 
From my point of view, Inpaint opposed works very well in the majority of cases, and thanks again to the designers. This algorithm has the particularity of being (very) sensitive to white balance settings.

Admittedly the DT code, which my old brain struggles to understand, contains a "chroma" part which is missing from the ART/RT version - if anyone can do that, of course I'm all for it. But running "WB auto temperature correlation" upstream solves (in my opinion) the problem in the difficult case, in particular the Nikon D800 image in test.

I did a complete review of the code - Itcwb -  I designed in 2018 (optimized by Ingo afterwards), so that the algorithm fails as little as possible in difficult cases. As a reminder, this algorithm is innovative, and I have not copied code anywhere...

I made 3 major changes
1) increase the sampling for reading the image in xyY – change from 190 to 235
2) give the possibility in this sampling to eliminate or not the purple colors.
3) at the end of the treatment, treat differently the balance between the values of the correlations (Student) and the value of the green.

Other adjustments have been made by default to optimize the algorithm in 95% of cases
* use the "Extra" value which only partially takes into account the basic settings (temperature, tint)
* increase accuracy by reducing the scale of analysis
* sort the data to be compared for the comparison with the 200 spectral data, by order of chromaticity, rather than by number.
* A total of 9 settings (available in Settings, and can be in Preferences if the other 'attached' PR is accepted) allow you to fine-tune the algorithm and the adaptation to specific cases (example of Observer 2° and 10° ). For me it would be a mistake to put these settings in the main menu even if it changes (a little) the image.

For comparison on the proposed Nikon D800 image:
* Inpaint opposed: 600ms
* Color Spread: 800ms
* Itcwb (depending on settings): 60 or 30 ms.

I did a brief update from Rawpedia.

If no one objects, I propose to merge this PR in a few days

jacques

garagecoder:
@Desmis 
Ignoring how it's coded, the chroma correction is a really simple idea; minimise the discontinuity (i.e. gradient) at the edges of the inpainted regions. Easiest way is to get the average difference at all inpainted edges (globally), then just add that value to the inpaints (again, globally). Here's a basic description which roughly fits the DT stuff:

- Calculate the recovered areas as usual from r = power_mean(p,q) (I’ll call this refavg)
- Get a per-channel mask of pixels surrounding the clips (using a small dilate)
- Get the per-channel _average_ difference (difference = original - refavg) of _all_ pixels which lie in the surround mask.
- Simply add the difference constant to each channel of refavg before drawing it in (with the usual constraint of > clip)

I think there's probably some stuff about a dark cutoff too, but I didn't check the most recent code. @jenshannoschwalm may correct some of that.

jenshannoschwalm:
> * Get the per-channel _average_ difference (difference = original - refavg) of _all_ pixels which lie in the surround mask.


> I think there's probably some stuff about a dark cutoff too, but I didn't check the most recent code. @jenshannoschwalm may correct some of that.

While doing the "Get the per-channel _average_ difference" you should take into account data that less than the channel's clip value **and** larger than a "dark" threshold (to avoid taking data at harsh transitions), the threshold is not so critical, around 0.5 of clip.

Desmis:
@garagecoder @jenshannoschwalm  Thank you for this advice, which is invaluable to me as well as the exchanges by PM.

-----
I think that the current solution allows to take advantage of the excellent "Inpaint Opposed IO" especially if it can be combined with "Itcwb" (WB auto temperature correlation: for Daylight and Blackbody illuminants).

I took advantage of this PR to make a complete review of "Itcwb" which should see the cases where the algorithm fails become very rare.

Of course there is always room for improvement, and making "IO" more perfect may be a goal for the future. Of course if a developer wants to take on the task he will be welcome.

In the meantime, I propose to merge this PR quickly, by the end of the week, if nobody objects.

Jacques

ghost:
Thanks Jacques, and other collaborators, for implementing and now merging this one.

One question/remark: I still see these timing messages in the terminal:

> HLRecovery_inpaint took 278 ms
> highlight_recovery_opposed took 155 ms
> highlight_recovery_opposed took 150 ms
> highlight_recovery_opposed took 153 ms

Should those still be shown?

The first one (*HLRecovery_inpaint*) belongs to *Colour Propegation*, the second to the newly added *Inpaint Opposed*. The other 3 reconstruction options do not show any timings in the terminal.

Desmis:
@jade-nl 

I voluntarily left this marking, but it can easily be deleted...

jacques

Desmis:
I deleted now :)

PR #6641, opened 2022-12-12
Global tone equalizer

Lawrence37:
Global version of the tone equalizer in local adjustments (ported from ART) with recent enhancements including pivot, colormap, and fix for darkening at extreme luminances.


Entropy512:
As a side note - a lot of the local adjustments modules have global use cases, and while it is possible to use a "global" spot to use them in this way, it's pretty ugly UI-wise.

There are a lot of capabilities in LA that I simply ignored for a long time because of this.

I'm not sure what the long-term solution to this is - somehow making any of the LA capabilities with global applicability available from the general UI???

Desmis:
@Lawrence37 
Ok , Agreed in principle, this appears to be an "old" idea, from someone 4 or more years ago.

But, in this case, there are many other examples with important differences, like Original retinex....or wavelet (richer in LA than main),or denoise,  or Ciecam (Jz...), etc. What do we do ?.
For me it's mostly a communication and/or GUI issue..
I generally approve of any innovation, but don't we make things more complex for the user

---------
Nevertheless, being a poor computer scientist can someone tell me how to load the code for this version on the fork of lawrence37

git checkout ..... (what?)

Thank you

reminder: I am very tired, due to ongoing treatments and lack of sleep


Jacques

Desmis:
I found a way to compile, maybe there is easier?

exemple
git clone http://github.com/Lawrence37/RawTherapee /e/code/repo-law
jacques

Thanatomanic:
@Desmis The 'official' way to do it, is:
```
git fetch origin pull/6641/head:some-name-for-your-local-branch
git checkout some-name-for-your-local-branch
```

Desmis:
@Thanatomanic 

Thank you :)

Beep6581:
Easy to use, runs fast, great results! Looking forward to merge.

Thanatomanic:
@Lawrence37 All seems to work well and you've seem to have taken care of the rest too (partial paste dialog, favorite panel, etc.). Great! 

For me it's not too clear what the underlying working principle is, nor what pivot and regularization do. But since things come from darktable originally, I guess that's something their documentation can provide.

👍 for merge

Lawrence37:
The basic idea is the same as darktable. It is parametric exposure ("gain" is more accurate) adjustment.

The implementation details are somewhat different. It works like this. First, it calculates the XYZ Y luminance for the pixel and uses it for the tone/norm. Based on that and the tone curve represented by the 5 equalizer adjusters, it determines the EV gain to apply.
The equalizer adjusters are similar to the parametric tone curve, each affecting a range of tones. Specifically, the 3 middle adjusters (shadows, mid-tones, highlights) are centered at -6, -4, and -2 EV (where white is 0 EV) with a Gaussian window of standard deviation σ=2 EV. The outer 2 adjusters are the combinations of every other window, i.e from -8 to -∞ EV (blacks) and 0 to +∞ EV (whites). Although the UI shows a range of -100 to +100 for each of the adjusters, they represent EVs of gain. Blacks and shadows have a range of -2 to +3 EV (i.e. -100 is -2 EV and +100 is +3 EV), mid-tones have -2.5 to +2.5 EV, and highlights and whites have -3 to +2 EV. For those who want a graphic visualization of the tone curve, here's one (ignore everything after the pivot slider): https://www.desmos.com/calculator/vjdyrxsyzq
Regularization is analogous to the tone equalizer masking in darktable. It preserves local contrast by applying a guided filter to the Y luminance. This works because small details are blurred away from the tone/norm image, so the tone adjustments leave (most of) the local contrast alone. The specifics of the guided filter parameters are complex and weird. I want to improve this part in the future, but I'm leaving it as-is to be consistent with the implementation in local adjustments.

Alberto introduced "pivot" to ART. It shifts the EVs affected by each equalizer adjuster, allowing finer adjustments to either the shadows or highlights. Alberto also added a colormap for visualizing which parts of the image is affected by each equalizer adjuster. It can also be used to understand the function of the pivot.

Desmis:
Just a note that doesn't take anything away from @Lawrence37  excellent work. The essential difference with the current version of LA is "pivot".
For the rest, apart from the particularities of LA (deltaE, choice of Rt-spot, etc.) the algorithm is the same.

Ok for merge

Jacques

ghost:
@Lawrence37 

Just to let you know: I've been trying out this new (global) tone equalizer for a few days now and have not come across anything untoward.

Merge worthy as far as I'm concerned.

ghost:
I have a question about the *show tonal map* functionality:

* Is it supposed to be static?

If so: what am I actually looking at (the state just after the demosaic step or another, later step in the pipeline)?

* Would it be feasible to make this dynamic?

Not sure how hard/easy this would be, it also does not change when adjusting, for example Exposure or LAB settings. Implementing this would thus not be limited to just the slider changes done in this specific module.

Implementing/merging this PR as it is right now has, for me, a higher priority then adding a dynamic tonal map, so if this is possible it might need to be a separate Feature Request.

Lawrence37:
The colormap shows which zone(s) affect which parts of the image. It is calculated from the tone equalizer's input image. If it doesn't respond to tonal changes from earlier parts of the pipeline, then there is a bug. It would also show up as a misbehavior of the tone equalizer because the tone equalizer is responsible for creating the colormap.

Desmis:
Just for information, and regardless of what it can be done for "Colormap" / "Tone Equalizer" (main). You have this feature (or similar) in "Local Adjustments" / "Tone Equalizer".

In "standard" or "advanced" mode => "Mask and modifications" / "Show modified areas wihtout mask"

ghost:
@Lawrence37 

> If it doesn't respond to tonal changes from earlier parts of the pipeline, then there is a bug.

I think we ran into a bug:

https://user-images.githubusercontent.com/56407344/225536940-d78f1b76-2cbd-4f46-9e07-68d489a7f387.mp4


As you can see the map does not react to changes made by the Exposure module or the sliders in the Tone Equalizer module. Only exception being the *pivot* slider.

I'm not at all sure what you mean by this:

> It would also show up as a misbehavior of the tone equalizer because  [...]

As you can see in the video, the sliders seem to be working as expected when not using the map.


Just to make sure we are talking about the same thing here: I expect the map to reflect/change in relation to:

1. Changes made in the module(s) that come before the Tone Equalizer in the pipeline (Exposure in the video),
2. Changes made with all the sliders in the Tone Equalizer.

Lawrence37:
It's a bit difficult to explain, but I will try. When the tonal map is enabled, the tone equalizer

Lawrence37:
It's a bit difficult to explain, but I will try. When the tonal map is enabled, the tone equalizer has a different role. Instead of affecting the tones, it creates the tonal map, and that becomes the output of the tone equalizer. If the tonal map does not respond to any adjustments, it's logical to guess that the tone adjustment won't either.

Your screencast shows the exposure having an effect on the brightness of the tonal map. This indicates exposure is applied *after* the tone equalizer.

The tonal map helps you see which zones you need to adjust to affect a certain area in the image. It is expected behavior that the zone sliders do not affect the tonal map. The blacks slider, for example, will always affect the darkest parts of the image as seen before applying the tone equalizer. Having the zone shift in response to the blacks slider would be confusing (not to mention, it would be logically impossible to compute like a chicken or egg problem). Pivot *does* affect the tonal map because its purpose is to shift the zones. Likewise, regularization affects how pixels are mapped to a zone according to neighboring pixels. Therefore, it also has an effect on the tonal map.

I hope this explaination is clear.

Lawrence37:
That reminds me, the location in the pipeline is not finalized. I'm open to suggestions on its placement. I feel strongly that it should come before the tone curves (and probably highlight compression too) due to its ability to recover highlights. Tone curves can clip and they are useful for final tone adjustments. Think scene-referred vs. display-referred. Other than that, I'm not sure where to put it.

Desmis:
The location in the pipeline of this process (or another) is still subject to debate - the good example with its advantages and disadvantages and the position of the 3 "denoise" modules (Noise reduction, LA denoise, Wavelet levels).

In this case, if you want to have a noticeable effect compared to other similar features (tone curves, Local Adjustments), you have to place it early in the process. Can be just after Highlight reconstruction (Inpaint opposed or Color propagation), or just before.

The visualization of the affected areas (again whatever the process) is something complex. Indeed we must be able (?) to show the situation "before" to decide what to do, and "after" to show the incidences. This, but it is not the same thing here, was difficult to develop in "LA" by differentiating for example the "Preview Delta E" and "Show modified areas without mask"

Jacques

ghost:
@Lawrence37 

> I hope this explanation is clear.

Yes, it is. It also, partially, addresses another uncertainty I had, namely the place this module has in the pipeline (no info yet on RawPedia). I presumed, based on looking at the Exposure tab, that it was placed after exposure and before Lab, which turns out to be a wrong presumption.

> That reminds me, the location in the pipeline is not finalized.

Good of you to mention this: My first reaction was to ask for it to be placed after exposure, but you make a really good technical point why it should not. Scene-referred seems to be the more obvious and correct choice.

How did Alberto tackle this (exposure does change the tonal map in ART)? Or did he opt for the display-referred approach?

Anyway: The way I use this module at the moment (primarily without the tonal map) works very well and as I'm not technical enough to add something useful about the pros-cons of its pipeline position I'll let the experts deal with that one.

Thanks for the clarification!

Desmis:
I think, if you have to change the place in the process of "Global tone equalizer" is to put it in "getimage" (Raw process) just before the two robust "highlight reconstruction" tools (Color propagation & Inpaint-opposed) to improve (or not ?) the result.

"Pivot" acts to position itself at the zero point of the DR (Dynamic Range). The Colormap helps to position itself. This zero point corresponds to :
I correct my previous error, by this principle formula (I went too fast)

        white_Ev = log(maxLum / yb) / log2;
        black_Ev = white_Ev - DR;

Not adjusting it, changes the response (but does not make it bad) of the extreme levels (Black,..., White)

Jacques

Lawrence37:
@jade-nl Alberto has it after exposure. He made significant pipeline changes in ART which makes it easier to insert the tone equalizer between exposure and the tone curves. In RawTherapee, most if not all the RGB processes are computed together using image tiles. It's difficult to make a tiled version of the tone equalizer. I also saw some code suggesting the highlight compression is tied to exposure which could make it impossible to insert the tone equalizer between exposure and highlight compression. The current location is as close to exposure as possible without coming after the tone curves or being inside the tile process, so maybe it is already in the best place.

@Desmis I think placing it before highlight reconstruction would confuse the reconstruction algorithms because they won't know which pixels are clipped.

Desmis:
@Lawrence37 @jade-nl 
In fact, the place where this module is suitable. 
The features are fine.

For my part, we can "merge" as is. 

We can always improve and put a visualization of the effect of actions linked to sliders, as in LA, but this does not specifically affect this module.

Thank you :)

Jacques

ghost:
@Lawrence37 / @Desmis 

OK, I've read the above explanations (thanks!) and been  playing some more with the tonal map. Now that I know what to expect and why, I don't see any reason not to leave it as-is.

Entropy512:
> In RawTherapee, most if not all the RGB processes are computed together using image tiles. It's difficult to make a tiled version of the tone equalizer.

I'm wondering if it might be a good idea to fix that stuff so that tiling is only done on a per-operation basis, if at all, because there are other potential operations that would ideally be done just before or just after the tone curves, or even "wrapped around" the tone curves.

(Something that I've been considering is something similar to the "exposure fusion" function in darktable's basecurve wrapped around RT's tone curves, or something else that I've been tossing around in my head and might proof-of-concept with some Python later - doing either would require NOT having a pipeline that does "tile, do lots of stuff, untile" with the tone curves being part of the "do lots of stuff" phase.)

Lawrence37:
The tiling may have performance benefits, but it could be possible to split it up into smaller chunks and insert full image operations between without significant penalty. Only benchmarks will tell us for sure.

Lawrence37:
I want to move the tone equalizer after exposure because it's easier to adjust the exposure to the ballpark range before applying finer tonal adjustments. Since highlight compression is tied to exposure and shadow compression is logically linked to them, the tone equalizer would come after all three. It is ok in my opinion because they are all tonal adjustments so there's no good reason to use both simultaneously.

The pipeline will change from
... > Tone equalizer > Channel mixer > Exposure / Highlight compression > Shadow compression > DCP tone curve and look table > Clip gamut > Brightness and contrast > Tone curves 1 and 2 > RGB curves > ...
to 
... > Channel mixer > Exposure / Highlight compression > Shadow compression > Tone equalizer > DCP tone curve and look table > Clip gamut > Brightness and contrast > Tone curves 1 and 2 > RGB curves > ...

Note: The pipeline in RawPedia looks wrong because it places tone curve before highlight and shadow compression:
Channel mixer > Tone curve > Highlights > Shadows > RGB curves

Lawrence37:
There's definitely a performance penalty for splitting up the tiled operations. The overhead can be as bad as a 50% increase in processing time for the tiled part, but of course this worst-case percentage scenario is when the "useful" processing time is at its minimum and therefore was fast to begin with. On my computer, processing a 13MP image went from ~150ms to ~220ms. Adding a few adjustments takes it up to ~370ms, which, if the overhead is also 70ms, corresponds to a ~25% performance penalty.

For the tone equalizer, I made two code paths. The first logically splits the tiled part into two and inserts the tone equalizer between. The second executes the whole tiled part as one unit. If the tone equalizer is enabled, the first path is taken. Otherwise the second one is used to minimize the performance impact. Benchmarks show there is negligible change in performance when the tone equalizer is disabled, i.e. the second code path is taken.

I'm now satisfied with both the pipeline placement and performance.

Lawrence37:
I will merge this soon.

Issue #6644, opened 2022-12-14
Input/output color profile support has a clunky UI


Entropy512:
This is connected to/impacts https://github.com/Beep6581/RawTherapee/issues/5575 among others

The UI/UX for input/output color profile handling is really clunky.  Rather than specify gamut and transfer function for output, the user must go to the ICC profile creator, generate a file, save it, then manually move it to a location where it will be used for output (as the default directory for output profiles on Linux is not normally user-writable).

Also, transfer function should probably be overridable in the batch process dialog where output file format is selected, as transfer function is often tied to output file format (e.g. JPEG is often gamma-encoded as is 8-bit TIFF, but 16+ bit TIFF is usually linear)

Similarly, for input, the default location saved to by the ICC profile generator is difficult to reach in the file dialog on Linux as it is in a hidden directory under $HOME ($HOME/.config )

Long-term (6.0?), a user should be able to select common gamuts and transfer functions from the UI without generating a custom ICC profile and then pointing another component of RT to the resulting file.

Short-term, usrICCDir in https://github.com/Beep6581/RawTherapee/blob/dev/rtengine/iccstore.cc#L450 should perhaps be renamed sysICCDir (as it is where system-installed profiles are located), and iccstore should additionally look in options.rtdir/profiles/output like the existing code that looks at profiles/output for output profiles.  I'm going to take a crack at this later in the week.

(Is there a reason that input profiles are pulled from the system store but output profiles are pulled from an RT-specific user location? 
 Also is there a reason that the ICC profile generator defaults to saving files in a location not used for input OR output profiles in iccstore?)

Entropy512:
Some aspects of this may be a dupe of some of @Beep6581 's comments related to UI in https://github.com/Beep6581/RawTherapee/issues/5518

Thanatomanic:
I agree that it would be very convenient to skip the ICC profile step and have the option to generate them on the fly.

We should probably also reconsider the TRC we embed in our profiles by default. Currently, these are always the same, but Rec2020 or ProPhoto or sRGB standards expect different TRCs. Users probably don't know we don't respect that.

Entropy512:
Yeah, Rec2020's default TRC is the same as Rec709, which I don't think we support at all.

The majority of content isn't even that - it's Rec. 2020 primaries combined with HLG or PQ TRC.

[...]

PR #6645, opened 2022-12-15
iccstore: Allow loading profiles from user-writable configuration directory

Entropy512:
In addition to bundled profiles and the system ICC profile store, load profiles from a user-writable/user-specific directory

On Linux, this is $HOME/.config/RawTherapee/iccprofiles/output - corresponding to "input" being already supported

Partial fix for part of #6644

### DIFF
--- a/rtengine/iccstore.cc
+++ b/rtengine/iccstore.cc
@@ -454,6 +454,8 @@ public:
         if (loadAll) {
             loadProfiles(profilesDir, &fileProfiles, &fileProfileContents, nullptr, false);
             loadProfiles(userICCDir, &fileProfiles, &fileProfileContents, nullptr, false);
+            Glib::ustring user_output_icc_dir = Glib::build_filename(options.rtdir, "iccprofiles", "output");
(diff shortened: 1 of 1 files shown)


Beep6581:
I found references of standard user locations being
`~/.local/share/color/` and
`~/.local/share/icc/`
but can't tell just how standard they are - they don't exist on my Manjaro system with a few color-related apps installed.

Entropy512:
I guess the question is whether there should be an RT-specific profiles directory, or a general one.

Honestly, I think having an RT-specific directory is even more important for output profiles than input since it's via a dropdown right now...  I'm not sure what the history is on the current default setting for RT pointing to a system ICC store.  (Possibly related to proofing profiles maybe coming ideally from the system store???)  Otherwise having a common/standard location for multiple apps will likely lead to that dropdown becoming really unwieldy really fast.

Lawrence37:
This looks fine to me.

Lawrence37:
Merged! This is a small but good step toward a better color management UX.

Issue #5575, opened 2019-12-21
Reference TIFFs are not tagged with an ICC profile that indicates linear data


Entropy512:
This is semi related to issue #5518 - reference TIFFs saved as part of the profiling process are not tagged with an ICC profile that signals that the file contains linear data as opposed to an sRGB TRC.

Now it's understandable why the first thought would be not to tag the images with a profile, since we don't know what the color primaries are until after profiling is complete.  However, anything that tries to display or process the images (such as defishing with Hugin - see https://discuss.pixls.us/t/color-profiling-with-a-fisheye-lens/14880 ) will assume sRGB primaries anyway.  We also DO know that the data should be linear.

If the images are tagged with an ICC profile that indicates sRGB primaries and a gamma of 1.0, then the images are much easier to view in RT's file browser and also are much easier to work with in Hugin for the purposes of defishing if the camera in question is one that has fixed fisheye lenses.  Similarly, opening the image in GIMP or RT to identify the location of the fiducials so they can be passed to argyll-scanin's -F parameter is much less painful if the image is tagged with a linear profile.

TBD:  If a reference TIFF is tagged with an ICC profile as opposed to untagged, does it interfere with argyll-scanin or dcamprof?

Beep6581:
RawTherapee is far more flexible today than it was back when the "Save reference image for profiling" button was implemented. Is that button even still necessary? To be checked whether the same can be accomplished using the traditional saving method.

Entropy512:
Good idea.

I'll post some examples tomorrow, but I THINK that starting from your idea I found a combination of settings that results in the same output data, but with the TIF being properly tagged as being linear.

If you don't want to try waiting:
Start with neutral profile.  Disable WB module, set input profile to No Profile, working profile to sRGB, and output profile to an ICC profile with sRGB colorspace but gamma=1.0 (I had to manually create such an ICC profile).

Probably could be done simply with a bundled preset.  Need to do a little more experimentation.

Thanatomanic:
> Start with neutral profile. Disable WB module, set input profile to No Profile, working profile to sRGB, and output profile to an ICC profile with sRGB colorspace but gamma=1.0 (I had to manually create such an ICC profile).

This does not give identical output to the Reference Image button though (note: you have to set sRGB linear TRC for the working profile as well). I tried looking through the code, but couldn't come up with a good reason why it doesn't match. It does look very similar, but not identical, which would defy the point of a 'reference' image of course.

[...]

PR #6646, opened 2022-12-15
WIP: Bundled profile for reference image

Entropy512:
Solves #5575 

TBD:  Remove the reference image code now that it is superseded?

Also:  Currently everything is set to sRGB - we can basically choose any colorspace as long as output = working and the input camera profile is disabled.  Any preferences?


Beep6581:
> TBD: Remove the reference image code now that it is superseded?

Yes, less duplication means less bloat, less code maintenance, less documentation maintenance.

> Currently everything is set to sRGB - we can basically choose any colorspace as long as output = working and the input camera profile is disabled. Any preferences?

sRGB seems to mimic the workings of the now-obsolete button, so I'd stick to that.

I proposed a change to the PP3 to strip all irrelevant keys. This makes it clear what the profile is mean to do when you look at it in a text editor, and ensures the profile won't break in RT when a tool is added or modified.

I'm not sure whether you want to include the `HLRecovery` section which activates highlight reconstruction - it might screw with the colors? Did "Save Reference Image" do that?

Entropy512:
HLRecovery was the default for what I started with - in theory it probably should be disabled, but in practice any image that has been exposed in such a way as for it to matter is one that is fundamentally not suitable to being used as a reference image.  I'm fine with nuking it since it's basically a NOP for a proper target shot anyway.

Also fine with stripping the unused/irrelevant keys - I'll review/test that later this week.  I'm entering "holiday vacation packing" phase soon.  :)

As to sRGB - in general, as long as output = working it should be functionally identical, since "no profile" on input and output=working leads to an effective NOP with respect to gamut transforms.

sRGB linear as an output profile is probably more useful to people for bundling as there are other use cases for it until a better long-term solution for #6644 can be determined and implemented

ProPhoto as working matches the current default working profile of RT (e.g. it isn't a change from the defaults while sRGB is)

Probably usefulness of the bundled profile wins, unless you've got thoughts as far as a more "common" use case as far as gamut for output profile.

ValZapod:
>sRGB linear

BTW, this combination of words makes little sense, sRGB is R'G'B', i.e. unlinear. The moment we talk about linear RGB (without prime ') it is just BT.709 primaries RGB. It is sRGB input (on the image) profile that makes it decode to RGB the correct (sRGB transfer) way.

Entropy512:
@Beep6581 thinking about this more - I'm not sure what Hugin will do if the input gamut is other than sRGB.  If the attached ICC profile indicates linear encoding it does respect that, but who knows what for the gamut.

I do know that if the input is flagged as linear with sRGB gamut, Hugin's output will be linear data that has not had any gamut transformations applied (since we all know it isn't actually sRGB, we just want to ensure that all gamut transforms are no-ops).  My use case for the original issue was making it less painful to use Hugin to defish the reference images from a fixed-lens fisheye camera, and it also makes pulling fiduciary coordinates with GIMP a lot easier.

Also, do you know if there's an "ignore user" function in github?  There's a block feature but that's probably overkill for the moment.

Beep6581:
> HLRecovery was the default for what I started with - in theory it probably should be disabled, but in practice any image that has been exposed in such a way as for it to matter is one that is fundamentally not suitable to being used as a reference image. I'm fine with nuking it since it's basically a NOP for a proper target shot anyway.

@Entropy512 I agree that not performing HLrec is probably for the best considering this processing profile's use case. Thumbs up for merging if you either remove the `[HLRecovery]` section from the PP3, or set `Enabled=false`, whichever is more appropriate.

> Also, do you know if there's an "ignore user" function in github? There's a block feature but that's probably overkill for the moment.

If you block the user, from your perspective it will be like an ignore.

Beep6581:
@Entropy512 please provide info on how to use this profile before merging to dev. Either directly to RawPedia, or here and then I will add it to RawPedia.

Entropy512:
Apologies for not finishing up/cleaning things on this including some of the other outstanding comments.  I'll try to revisit this over the weekend.

I'm finally going to start to have more free time again soon to resume the 2137931270721390 different things I've been procrastinating on lately.

I don't think I have rawpedia access - I have been planning on requesting access, but decided to put it off based on the discussion over on pixls regarding a likely migration to a static site with hugo + pull requests.  (that is a separate issue/discussion but I'm willing to help with that if needed, I'm planning on pushing for a similar migration at my dayjob.)

Entropy512:
I've noticed you're back and active again @Beep6581 - any news on the replacement for rawpedia discussed in https://discuss.pixls.us/t/an-update-on-rawpedia/34168 ?  It would be preferable to add the documentation in question to whatever replaces Rawpedia, instead of to the old system.

Beep6581:
@Entropy512 nothing in the form of news yet, but certainly much motivation (inversely proportional to available time).

Would be good to have this in 5.10, and it's the kind of thing that does not need to be in beta1.

You could provide just a stub paragraph/list of steps here in plaintext describing how to use it, and I'll add it to the docs.

PR #6647, opened 2022-12-17
Speed up compilation of rtengine/procparams.cc

### LINKED ISSUE #6548: Speed up compilation for release builds with GCC 12.1
After upgrading GCC to 12.1, the compile time for release builds increased. `rtengine/procparams.cc` is responsible for a significant percentage of the build time. The difference is quite dramatic and can be seen in the automated Windows builds.

Here's an example using two consecutive `dev` builds.
GCC 11.3: https://github.com/Beep6581/RawTherapee/runs/6540668974
- Release: 12m 54s
- Debug: 11m 39s

GCC 12.1: https://github.com/Beep6581/RawTherapee/runs/6834837161
- Release: 21m 40s
- Debug: 11m 19s

We could probably speed up the build by disabling some optimizations for `rtengine/procparams.cc` or splitting it into multiple files. Thoughts?

Lawrence37:


rtengine/CMakeLists.txt
```diff
@@ -176,6 +176,16 @@ if(LENSFUN_HAS_LOAD_DIRECTORY)
     set_source_files_properties(rtlensfun.cc PROPERTIES COMPILE_DEFINITIONS RT_LENSFUN_HAS_LOAD_DIRECTORY)
 endif()
 
+if(CMAKE_CXX_COMPILER_ID STREQUAL "GNU" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL "12.0")
```
Thanatomanic, line 179: Thanks for the patch. Perhaps it's worthwhile to add a comment referring to the issue? Or else explain why this code is here.

Lawrence37: Added a comment

Issue #6656, opened 2023-01-01
Activating Ciecam introduces color cast in RT 5.9


stefan-i:
Hi
I noticed an unwelcome change in the Ciecam module in 5.9 that was not present in 5.8.

In 5.8 and earlier, just activating Ciecam with all sliders to default values did not change the image (provided "mean luminance Yb%" was set to 18). Now in 5.9, when Ciecam is activated a noticeable color shift occurs. 

As an example, take the image of the color checker here:
https://www.file-upload.net/download-15068738/Sonya7IVDaylightDSC00122.DNG.html

When setting the white balance on the second left grey patch, the RGB values are identical (as they should be):
<img width="387" alt="Clipboard01-NoCiecam" src="https://user-images.githubusercontent.com/13986045/210179271-9bc92e2b-7a92-4ac7-ba13-c4efae07b508.png">

However, just activating the Ciecam module introduces a color cast:
<img width="387" alt="Clipboard02-Ciecam" src="https://user-images.githubusercontent.com/13986045/210179294-dc5a6cf2-01a4-4b8f-9d3e-c4e05a0d4ced.png">

So with Ciecam on, it is no longer possible to set a  neutral white balance by e.g. clicking on a grey patch, as the module introduces a color shift.

The color shift is quite noticeable in real world images (looks like magenta shift). Just look at the asphalt background of the provided Raw file to get an idea.  It does not seem to depend much on the white balance setting of the image: When you set white balance of the image in RT to random values the strength of the shift by activating Ciecam seems to stay roughly the same, so it does not seem to be an effect of the normal color adaptation behavior of Ciecam but an unwanted, fixed shift.

I think that at default values there shouldn't be these kind of color shifts for supposedly neutral patches, so could you please take a look at the cause of this problem? Thank you!

P.S.: I played around a bit in the Ciecam settings, and the shift can mostly (but not entirely) be avoided by changing the following settings from the default values:
<img width="287" alt="Clipboard03-SettingsNeutral2" src="https://user-images.githubusercontent.com/13986045/210179830-027e858e-4936-4d9b-b735-644f509b6e7b.png">
However, this is just a workaround for now and may not work as intended for all images, so it would be preferable to properly fix this issue.

RT versions tested: 

Version: 5.9
Branch: 5.9
Commit: 9b8583988
Commit date: 2022-11-27
Compiler: cc 12.2.0
Processor: generic x86
System: Windows
Bit depth: 64 bits
[...]

PR #6658, opened 2023-01-04
CIECAM - Changes such as symmetric works and temperature output makes sense - issue #6656

Desmis:
See text in issue #6656 

With the modifications made, in "automatic symmectric" mode there may remain small differences. These are related to the precision of the algorithms (normal and inverse) and the change from "double" to "float" a few years ago (2015?) to save time and memory.

Admittedly, it's not "perfect", but very close to the objectives.
Reminder, it is more than normal that in "Classic" mode there are differences in colorimetry and contrast... this is the purpose of a CAM

jacques


Desmis:
After the test by @stefan-i.... in issue. I propose to merge this PR, if nobody objects tomorrow :)

Jacques

PR #6661, opened 2023-01-09
External editor fixes for Windows

Lawrence37:
See https://discuss.pixls.us/t/5-9-can-not-launch-external-editor/34606. For some reason, certain applications do not launch properly when using the improved external editor feature.

This introduces a fallback to the previous application launcher and fixes the executable file picker filter so that `.exe`s are shown by default.

### DIFF
--- a/rtdata/themes/TooWaBlue-GTK3-20_.css
+++ b/rtdata/themes/TooWaBlue-GTK3-20_.css
@@ -1278,6 +1278,11 @@ menuitem:hover > * {
     color: @text-hl-color;
 }
 
+menu menuitem > radio + * image:not(.dummy),
+#MyExpander menu menuitem > radio + * image:not(.dummy) {
+    margin-left: 1pt;
+}
+
 menu image:not(.dummy),
 #MyExpander menu image:not(.dummy) {
     min-height: 2em;
--- a/rtdata/themes/size - Legacy.css
+++ b/rtdata/themes/size - Legacy.css
@@ -383,6 +383,11 @@ menu arrow {
     margin: 0 -0.25em 0 0;
 }
 
+menu menuitem > radio + * image:not(.dummy),
+#MyExpander menu menuitem > radio + * image:not(.dummy) {
+    margin-left: 1pt;
+}
+
 menu image:not(.dummy),
 #MyExpander menu image:not(.dummy) {
     min-height: 2em;
--- a/rtdata/themes/size.css
+++ b/rtdata/themes/size.css
@@ -351,6 +351,11 @@ menu arrow {
     margin: 0 -0.25em 0 0;
 }
 
+menu menuitem > radio + * image:not(.dummy),
+#MyExpander menu menuitem > radio + * image:not(.dummy) {
+    margin-left: 1pt;
+}
+
 menu image:not(.dummy),
 #MyExpander menu image:not(.dummy) {
     min-height: 2em;
--- a/rtgui/editorpanel.cc
+++ b/rtgui/editorpanel.cc
@@ -905,7 +905,6 @@ EditorPanel::EditorPanel (FilePanel* filePanel)
     send_to_external = Gtk::manage(new PopUpButton("", false));
     send_to_external->set_tooltip_text(M("MAIN_BUTTON_SENDTOEDITOR_TOOLTIP"));
     setExpandAlignProperties(send_to_external->buttonGroup, false, false, Gtk::ALIGN_CENTER, Gtk::ALIGN_FILL);
-    send_to_external->addEntry("palette-brush.png", M("GENERAL_OTHER"));
     updateExternalEditorWidget(
         options.externalEditorIndex >= 0 ? options.externalEditorIndex : options.externalEditors.size(),
         options.externalEditors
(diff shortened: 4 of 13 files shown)


Stuntflyer:
Thanks for working on this. Will test and report back when the work around is available.

Desmis:
@Lawrence37 

Since the recent changes I have some difficulties with "Photoshop" when I launch the external editor (Windows 11). My version is very old CS5, but is enough for me

Previously, Photoshop launched, then the current file in the preview was displayed directly in Photoshop.

Now, with or without the change made by the PR, Photoshop launches without problems, is opened by default in the chosen folder (for me C:\msys64\tmp).
The TIF file is present in C:\msys64\tmp, but it does not open automatically... I have to do: file - open - find the file.
It's not a big deal :)


Jacques

Lawrence37:
Maybe Photoshop has some issue with the file paths sent by RawTherapee. What happens if you add ` %f` or ` %u` to the command? For example, `"C:\My Program\myprogram.exe" %f`

Desmis:
@Lawrence37 
if i change in 'options' (Rawtherapee5-dev/options) - with an editor:
Commands="C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)\\Photoshop.exe";
with
Commands=C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)\\Photoshop.exe;

All seems to work fine :)

Jacques

Desmis:
@Lawrence37 

I don't know what I did, but now I'm unable to set up Photoshop in Preferences....no more than Gimp
I no longer have access to the external editor
 My "options" file is broken...

when I am in Preferences
* if I check "change application",  none fit
* if I check "Change executable" , the system finds the directory "Adobe Photoshop CS5 (64 Bit)", but after ???

here options file, what is wrong ?
[External Editor]
EditorKind=2
GimpDir=GimpDir=C:\\msys64\\home\\jdesm
PhotoshopDir=C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)
CustomEditor=
OutputDir=2
CustomOutputDir=C:\\msys64\\tmp
Float32=false
BypassOutputProfile=false
Names=Photoshop;
Commands=“C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)\\Photoshop.exe”;
IconsSerialized=;
EditorIndex=0

On my computer Photoshop is in
C:\Program Files\Adobe\Adobe Photoshop CS5 (64 Bit)
and executable file is in this path : Photoshop.exe

*So I use : Save current image... no problem, but it is better with "Edit in current editor"


Thank you
image

jacques

Desmis:
@Lawrence37 

And now with this change in 'options'
[External Editor]
EditorKind=2
GimpDir=GimpDir=C:\\msys64\\home\\jdesm
PhotoshopDir=C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)
CustomEditor=
OutputDir=2
CustomOutputDir=C:\\msys64\\tmp
Float32=false
BypassOutputProfile=false
Names=Photoshop;
Commands=C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)\\Photoshop.exe
IconsSerialized=;
EditorIndex=0

With branch external-editor-fixes
Photoshop runs with no problem...
Curious :)

But if I change the branch to "dev" or something else... does not work...
Jacques

Stuntflyer:
@Lawrence37 

With 5.9-60 working well, I'm still trying get dev 5.9-57 working with the external editor.

Is it possible that the "editor" showing one backslash and the config file showing two backslashes could create a problem?

It seems that Jacques was able to get a single backslash in the config file yet I can get no less than two.

Lawrence37:
The backslashs are not the issue. I think the options snippet that Jacques shared actually has two backslashs but they were not escaped before posting the comment. This is two: \\, but you see only one.

The reason why it is working for Jacques is because there are no quotation marks. Without them, any spaces in the path break the application launcher and forces RawTherapee to use the fallback. You can see this by trying without the quotes in dev. It won't work because dev doesn't have the fallback mechanism.

If you want to find a solution that works in dev, try the things I mentioned in this Pixls comment: https://discuss.pixls.us/t/5-9-can-not-launch-external-editor/34606/58

Stuntflyer:
Okay, I will certainly try those things you commented about. If for some reason they don't work for me, what options will I have going forward, if you decide not to add the "work around" to the dev releases?

I tried all of your suggestions and still not working. Always getting the external error message.

The only thing I was able to do was to launch PS along with a tif or nef file from a cmd prompt.

Lawrence37:
I added radios to hopefully prevent people from being confused why the editor doesn't launch after clicking an editor.
image

@Stuntflyer @Desmis I do plan on merging this into `dev` even if we can't find a "proper" solution. With this fix, everything should work automatically if you were receiving the error message. If there was no message, the hack-y but usable solution is to remove the quotes from the command.

Desmis:
@Lawrence37 
I pull your change.

Some remarks:
When in Preferences , Photoshop is described as: (without "")  
**C:\Program Files\Adobe\Adobe Photoshop CS5 (64 Bit)\Photoshop.exe**
**Photoshop launch, then the file is open but in console I have this message**

"Output profile_: "RTv4_sRGB"
Using "RTv4_sRGB" output profile
loading dynamic profiles...
saving dynamic profiles...
Saved 0 dynamic profile rules
saving dynamic profiles...
Saved 0 dynamic profile rules
**Error launching external editor.
Error code #8: Failed to execute child process (No such file or directory)
Unable to launch external editor with Gio. Trying custom launcher."**


When in Preferences , Photoshop is described as: (with "")  
**"C:\Program Files\Adobe\Adobe Photoshop CS5 (64 Bit)\Photoshop.exe"**
**Photoshop launch, then the file does not open**, I must in photoshop "open file"
**No error message in console**

On the other hand **my screen (preview) is different from** what I see in your case
image

When I edit "options"
[External Editor]
EditorKind=2
GimpDir=GimpDir=C:\\msys64\\home\\jdesm
PhotoshopDir=C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)
CustomEditor=
OutputDir=2
CustomOutputDir=C:\\msys64\\tmp
Float32=false
BypassOutputProfile=false
Names=Photoshop;
Commands="C:\\Program Files\\Adobe\\Adobe Photoshop CS5 (64 Bit)\\Photoshop.exe";
IconsSerialized=;
EditorIndex=0

is there an error ?

If I change : EditorIndex from 0 to -1, Photoshop does not launch

But ok to merge, because a minima it works :)

jacques

Lawrence37:
@Desmis That is expected behavior. In the first case, the lack of quotes makes the path broken up at each space character. It causes the editor to fail to launch and the workaround is used. In the second case, the path is fine so the editor opens. For some reason that I don't understand, your copy of Photoshop doesn't work well with the new application launcher.

Since the workaround is functional and is better than no solution, I am merging this pull request. I still hope we can find a proper solution in the future.

Desmis:
@Lawrence37 

OK for me, thank you

Jacques

Stuntflyer:
@Lawrence37 

Ok for me too

Thank you

PR #6665, opened 2023-01-15
Abstract profile - Gamut improvment

Desmis:
This PR improves "Abstract profile".

Let us first recall the 3 used hands.

* TRC: adjust gamma and slope to: a) modify the image rendering, which by default is g=2.4 s=12.92; b) raise the shadows and modify the lighter tones while preserving the colorimetry and the gamut.
*  Illuminant (white point): a) adjust the gamut of the Working profile to adapt it to the shooting conditions taking into account the colorimetry. If an image is taken at 4000K, in sRGB (which is D65), the gamut may not be suitable. The joint action of the Illuminant and CIECAM (symmetrical mode) allows you to ensure good chromatic adaptation; b) used in conjunction with the modification of the Primaries below.
 * Primaries: this is where the notion of Abstract profiles comes into its own. Several possibilities are available: a) Use this module as a Channel Mixer in order to create special effects similar to Color Toning, with or without CIECAM; b) More simply restore colors (saturation) to an image, in most cases maintaining the balance of colors; c) Use it to modify or refine the result of the Input profile to improve the colorimetry (calibration).

But, in some cases, for example when the **difference between the "working profile" and "Destination primaries" is important, or the change of illuminant important, the conversion can bring artefacts** (which are no longer special effects…), despite the profile conversion performed by LCMS. 

So I added a **checkbox -Gamut Contro**l which is disabled by default. 
How does it work ? 
1) I use **Gamutmap, designed by Emil Martinec** in 2010/2011. This gamut protection works in XYZ and ensures absolute colorimetry.
2) I uploaded a procedure that calculates the XYZ-RGB matrix based on the primaries (which are free) and the illuminant (free too)**: Color::primaries_to_xyz**.

Of course my code can be improved...as you know I'm a poor programmer, but it works.

jacques

### DIFF
--- a/rtengine/color.cc
+++ b/rtengine/color.cc
@@ -24,6 +24,7 @@
 #include "sleef.h"
 #include "opthelper.h"
 #include "iccstore.h"
+#include <iostream>
 
 using namespace std;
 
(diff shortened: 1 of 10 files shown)


Desmis:
If nobody objects, I will merge this PR sunday afternoon :)

Jacques

rtengine/color.cc
```diff
@@ -1910,6 +1911,147 @@ void Color::Lch2Luv(float c, float h, float &u, float &v)
     v = c * sincosval.y;
 }
 
+void Color::primaries_to_xyz (double p[6], double Wx, double Wz, double *pxyz)
+{
+	//calculate Xr, Xg, Xb, Yr, Yb, Tg, Zr,Zg Zb
+	double Wy = 1.0;
+    double Xr = p[0] / p[1];
```
Thanatomanic, line 1918: I am correct to assume that the values of `p` can be user-specified? If so, are values of 0 acceptable? If so, we need guards here to prevent DIV/0.

Thanatomanic:
@Desmis I have no objections against merging, but there are quite some weird indentations in your commits. From a style perspective it would be nice to fix these.

rtengine/color.cc
Desmis: Well seen, good catch.

Thank you.
I just push a commit to solve this case.

Jacques

Desmis:
@Thanatomanic 

I don't see where there are quite some weird indentations ? Can you show them to me.

Thank you

Jacques

rtengine/color.cc
```diff
@@ -1910,6 +1911,147 @@ void Color::Lch2Luv(float c, float h, float &u, float &v)
...
+    double Zb = (1.0 - p[4] - p[5]) / p[5];
+
+    using Triple = std::array<double, 3>;
+
+    using Matrix = std::array<Triple, 3>;
+
+    Matrix input_prim;
+    Matrix inv_input_prim = {};
+
+	input_prim[0][0] = Xr;
```
Thanatomanic: Example for indentation

rtengine/color.cc
```diff
@@ -1910,6 +1911,147 @@ void Color::Lch2Luv(float c, float h, float &u, float &v)
...
+    Matrix Ma_oneBradford = {};
+    Ma_oneBradford[0][0] = 0.9869929;
+    Ma_oneBradford[0][1] = 0.4323053;
+    Ma_oneBradford[0][2] = -0.0085287;
+    Ma_oneBradford[1][0] = -0.1470543;
+    Ma_oneBradford[1][1] = 0.5183603;
+    Ma_oneBradford[1][2] = 0.0400428;
+    Ma_oneBradford[2][0] = 0.1599627;
+    Ma_oneBradford[2][1] = 0.0492912;
+	Ma_oneBradford[2][2] = 0.9684867;
```
Thanatomanic: Another indentation

rtengine/color.cc
```diff
@@ -1910,6 +1911,147 @@ void Color::Lch2Luv(float c, float h, float &u, float &v)
...
+    Matrix cone_ma_one = {};
+    for (int i = 0; i < 3; ++i) {
+        for (int j = 0; j < 3; ++j) {
+            cone_ma_one[i][j] = 0;
+
+            for (int k = 0; k < 3; ++k) {
+                cone_ma_one[i][j] += cone_dest_sourc [i][k] * Ma_oneBradford[k][j];
+            }
+        }
+	}
```
Thanatomanic: Another indentation

rtengine/color.h
```diff
@@ -1847,6 +1847,13 @@ static inline void Lab2XYZ(vfloat L, vfloat a, vfloat b, vfloat &x, vfloat &y, v
     */
     static void gamutmap(float &X, float Y, float &Z, const double p[3][3]);
 
+	/**
```
Thanatomanic, line 1850: This whole function is indented differently

rtengine/iplab2rgb.cc
```diff
@@ -520,7 +523,25 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
         greyy = rtengine::LIM(greyy, 0.5f, 1.f);
     }
     //fixed crash when there is no space or too small..just a line...Possible if bx, by aligned with Gx,Gy Rx,Ry
-    float ac = (greyy - redyy) / (grexx - redxx);
+	//fix crash if user select 0 for redyy, bluyy, greyy
```
Thanatomanic: Indentation is one block more than expected from surrounding code

rtengine/iplab2rgb.cc
```diff
@@ -520,7 +523,25 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
...
+		bluyy = epsil;
+	}
+	if(greyy == 0.f) {
+		greyy = epsil;
+	}
+
+	//fix crash if	grexx - redxx = 0
+	float grered = 1.f;
+	grered = grexx - redxx;
+	if(grered == 0.f) {
```
Thanatomanic: Space after `if`

rtengine/iplab2rgb.cc
```diff
@@ -621,8 +641,10 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
             D60 = 6005  // for ACES AP0 and AP1
 
         };
-        double tempv4 = 5003.;
-        float p[6]; //primaries
+		double tempv4 = 5003.;
```
Thanatomanic: Indentation

rtengine/iplab2rgb.cc
```diff
@@ -802,55 +855,91 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
...
         switch (ColorManagementParams::Illuminant(illum)) {
-            case ColorManagementParams::Illuminant::DEFAULT:
-            case ColorManagementParams::Illuminant::D55:
+            case ColorManagementParams::Illuminant::DEFAULT:{
+				break;
+			}
+            case ColorManagementParams::Illuminant::D55:{
+                Wx = 0.956565934;
+                Wz = 0.920253249;
+				break;
```
Thanatomanic: Especially these two blocks are strangely indented

rtengine/iplab2rgb.cc
```diff
@@ -802,55 +855,91 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
...
+                Wx = 0.964295676;
+                Wz = 0.825104603;
                 break;
             }
 
             case ColorManagementParams::Illuminant::D60: {
-                xyD = {0.32168, 0.33767, 1.0};
+                Wx = 0.952646075;
+                Wz = 1.008825184;
+				xyD = {0.32168, 0.33767, 1.0};
```
Thanatomanic: Indent

rtengine/iplab2rgb.cc
```diff
@@ -802,55 +855,91 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
...
 
             case ColorManagementParams::Illuminant::D65: {
+                Wx = 0.95045471;
+                Wz = 1.08905029;
                 xyD = {0.312700492, 0.329000939, 1.0};
                 break;
             }
 
             case ColorManagementParams::Illuminant::D120: {
+				Wx = 0.979182;
```
Thanatomanic: Indent

rtengine/iplab2rgb.cc
```diff
@@ -907,10 +998,15 @@ void ImProcFunctions::workingtrc(const Imagefloat* src, Imagefloat* dst, int cw,
                     const float r = src->r(i, j);
                     const float g = src->g(i, j);
                     const float b = src->b(i, j);
-
-                    *(p++) = toxyz[0][0] * r + toxyz[0][1] * g + toxyz[0][2] * b;
-                    *(p++) = toxyz[1][0] * r + toxyz[1][1] * g + toxyz[1][2] * b;
-                    *(p++) = toxyz[2][0] * r + toxyz[2][1] * g + toxyz[2][2] * b;
+					float X = toxyz[0][0] * r + toxyz[0][1] * g + toxyz[0][2] * b;
```
Thanatomanic: Too much indentation

Desmis:
@Thanatomanic 

Thank you for your help.

I have to admit that I don't always understand the reason for the indentations (I had paid attention to this point when writing this code)
In particular the code of iplab2rgb underwent a complete review during the creation of Abstract Profiles.... But I fully agree that this should be done.

I think it's easier if you make the "right" changes directly. Otherwise I'll do what I understand.  :)

Jacques

Desmis:
I formatted the code with the provided formatter for RT (Astyle / astylert.bat) which is recognized for C++ code. I used the recommended settings for RT: "rawtherapee.astylerc".

On the other hand, the code of ""workingtrc" has been reviewed for better understanding (as a C++ language) by the RT team. I only made minor changes to the code.

Jacques

Lawrence37:
I noticed that black pixels suddenly become green when Gamut control is enabled.

Desmis:
@Lawrence37 
Thank you for testing :)

Yes, it is quite possible, depending on the images, the "working profile" and the primaries chosen.
The algorithm that I chose which is simple (Emil Martinecc in 2011) is not perfect, in particular because it does not use intent and works in absolute colorimetry (we cut everything that exceeds).

This is why at the same time (2011) I imagined another type of control which uses relative colorimetry, but which manages HL very badly.

The only thing that works well, and is much debated, is Lab's lack of linearity (which resulted in it being rejected in e.g. DT). In 2011, 2012 I implemented this LUT (200)  based fix, which works well and makes Lab "linear"

In this case, after testing, I chose Emil's (Gamutmap), leaving the user the choice to use it or not. I remind that in many cases it is not strictly useful, because there is the output profile, but it fixes something else....

Abstract Profile, is located at the end of the process and does not interfere with the upstream work, only Ciecam is behind, to be able to make chromatic adaptation corrections if necessary.

But I think that this justified remark does not call into question this addition?

Thank you again

Jacques

Lawrence37:
Hi Jacques, I have no objections to the gamut control. I just think there is a bug in the computation somewhere that causes completely black pixels to become bright green.

Desmis:
@Lawrence37 
Could you join, raw and pp3
Thank you

Jacques

Lawrence37:
Pretty much any image will work as long as you can create black pixels. For this pp3, I started with Neutral, increased Exposure > Black, and enabled Abstract Profile with Gamut control.
[amsterdam.pef.pp3.txt](https://github.com/Beep6581/RawTherapee/files/10476290/amsterdam.pef.pp3.txt)

Desmis:
@Lawrence37 
Thanks for this patch. Effectively the black image becomes green.

I made changes to (the excellent work of Emil Martinec).

Indeed in theory at least the values X, Y, Z and the sum X + 15 * Y + 3 * Z, cannot be negative.

I also avoided a division by 0, as well as some minor changes.

This seems to fix the problem.

Thank you

Jacques

Lawrence37:
@Desmis Looks good now

Desmis:
@Lawrence37 

OK, thank you very much :)

If nobody objects I will merge this PR this afternoon.

Jacques

PR #6670, opened 2023-01-25
Add missing include for GCC13 compatibility

mattiaverga:
This #include is needed for correct build on Fedora 38 with GCC 13, see https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes

Signed-off-by: Mattia Verga <mattia.verga@tiscali.it>

### DIFF
--- a/rtengine/dcraw.h
+++ b/rtengine/dcraw.h
@@ -19,6 +19,7 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


Floessie:
@mattiaverga Thank you, Mattia! I should have caught that.

PR #6673, opened 2023-01-27
Avoid Color shift - Gamut and Munsell Review in RT - branch Munsellgamut

Desmis:
Excuse my bad english.

Some comments on this PR:
As you probably know, there have been many debates in RT and elsewhere about Lab / RGB and Cam (color appearance models).
I'm not going to go over my point of view again (which some may criticize), but in summary for Lab:

1) if we do nothing, saturation changes (chroma) bring about Hue drift especially in reds/oranges and blues/purples

2) Lab is an open space and nothing prevents - if you push the settings of L, a, b to the extremes - to produce imaginary colors... With Prophoto as "Working space", you really have to go for it. But from my point of view this is not a major problem because the output profiles (for printer or screen) bring the whole thing within acceptable limits.

3) Lab, if you want to produce and visualize HDR images, is not very appropriate, because if the low lights are well treated, it is less the case of the high lights, which are penalized by the gamma of 3.0 of Lab. Note that Lab  (float or double) does not restrict the DR "Dynamic Range". Nevertheless in "LA", Jz has been set up, and the processes which strongly use Lab, can have a weakened gamma from 3.0 to 1.0... But there remains the problem of the output profile towards an HDR monitor which is not currently solved in RT (I did not give up to find an original solution).

4) in principle, CAMs : Ciecam (main) or LA " Color Appearance (Cam 16 & JzczHz)" don't need Munsell correction...since they are supposed to be " good " in colorimetry by adding the perception of the individual. Nevertheless, in the end, we can overflow the gamut if we have made excessive settings upstream ... 

For 1) , RT is equipped since 2011, 2012 of a control made with the help of LUT Munsell which gives the hues of Lab without drifts.
For 2) if you wish, there is in particular in "Lab adjustments" and in Settings of "Local adjustments" a gamut control which was supposed to solve this problem.

**I just did a complete review of 1) and 2), because there were many bugs....** :

For "Lab adjustments" and Settings "Local adjustments" I replaced "Avoid Colorshift" with a combobox that offers 4 choices:
* None
* Lab - which makes a gamut control in relative colorimetry based on Lab, and then a Munsell correction. The main problem with this gamut correction is with images with HL reconstruction.
* XYZ - which makes a gamut control in absolute colorimetry based on XYZ, and then a Munsell correction. The main defect, also concerns the reconstruction of HL. The difference between Lab and XYZ processing (in fact "uv"), is that "uv" (XYZ) is theoretically less efficient, because it takes less account of luminance than Lab.
* Munsell: only performs the Munsell correction, without touching the gamut control. Only Hue corrections.

**By default, and I think it is desirable, "Munsell" is activated.**

I also modified a bit the control present in Ciecam…

And you also have as a reminder "Abstract profile" which with its gamut control allows (if you are not looking for special effects), a limitation of the gamut of type XYZ (uv). For example, when you want to change the data from Prophoto to Rec2020...where the color balance is theoretically respected.

Jacques


Desmis:
Some additional information

I added a tooltip in "Lab adjustments", the same as in "Local adjustments"

I added a line to the possible choices, in addition to XYZ Absolute, there is also XYZ Relative (or something close to it).

There's nothing "groundbreaking" or big innovation in all of this code, other than:
* a new feature review after several years (2012)
* the "Local adjustments" part was bugged (poor taking into account of the basic data to apply Munsell).
* a GUI adapted to more choices - it's the same (within context) in Local Adjustments and Lab Adjustments.
* The 195 Munsell LUTs are a quick (?) and easy way to process Lab : "Uniform Perceptual Lab". Deviations from "perfection" are minimal and imperceptible. All colors are examined according to the Munsell color base, deviations may reach (blue / purple) of course before correction: 0.42 radians)

So, if nobody objects, I will merge this PR soon :)


Jacques

Lawrence37:
@Desmis I noticed the previous "Avoid color shift" setting does not get translated. If I understand correctly, disabled corresponds to "None" and enabled corresponds to "Lab". However, if I apply an old pp3, it sets the value to "Munsell only".

rtdata/languages/default
```diff
@@ -2904,6 +2906,13 @@ TP_LOCALLAB_GAMM;Gamma
 TP_LOCALLAB_GAMMASKCOL;Gamma
 TP_LOCALLAB_GAMMASK_TOOLTIP;Adjusting Gamma and Slope can provide a soft and artifact-free transformation of the mask by progressively modifying 'L' to avoid any discontinuities.
 TP_LOCALLAB_GAMSH;Gamma
+TP_LOCALLAB_GAMUT_TOOLTIP;Default : Munsell.\nMunsell: fixes Lab mode hue drifts due to non-linearity, when chromaticity is changed.\nLab: applies a gamut control, in relative colorimetric, Munsell is then applied.\nXYZ Absolute, applies gamut control, in absolute colorimetric, Munsell is then applied.\nXYZ Relative, applies gamut control, in relative colorimetric, Munsell is then applied.
+TP_LOCALLAB_GAMUTTYPE;Avoir Color Shift
```
Lawrence37: We can reuse TP_LABCURVE_AVOIDCOLORSHIFT and TP_LOCALLAB_AVOID

rtgui/controlspotpanel.cc
```diff
@@ -402,12 +404,29 @@ ControlSpotPanel::ControlSpotPanel():
...
+
+ Gtk::Frame* const avFrame = Gtk::manage(new Gtk::Frame());
     ToolParamBlock* const avbox = Gtk::manage(new ToolParamBlock());
     avFrame->set_label_align(0.025, 0.5);
-    avFrame->set_label_widget(*avoid_);
+   // avFrame->set_label_widget(*avoid_);
+    avbox->pack_start(*ctboxgamut);
     avbox->pack_start(*avoidrad_);
-    avbox->pack_start(*avoidmun_);
+    //avbox->pack_start(*avoidmun_);
```
Lawrence37: Better to remove `avoid_` and `avoidmun_` completely to avoid creating unnecessary widgets and causing memory leaks. Even with `Gtk::manage`, GTK will never delete the widgets because they are not attached to anything.
https://github.com/Beep6581/RawTherapee/blob/6f2ca3f2581e6f6dcecb00ffb6bb6a730988c972/rtgui/controlspotpanel.cc#L87-L88

rtgui/controlspotpanel.h
```diff
@@ -57,6 +57,7 @@ class ControlSpotPanel:
         int sensiexclu;
         int structexclu;
         int shapeMethod; // 0 = Independent (mouse), 1 = Symmetrical (mouse), 2 = Independent (mouse + sliders), 3 = Symmetrical (mouse + sliders)
+        int avoidgamutMethod; // 0 = Independent (mouse), 1 = Symmetrical (mouse), 2 = Independent (mouse + sliders), 3 = Symmetrical (mouse + sliders)
```
Lawrence37:
```suggestion
        int avoidgamutMethod;
```

rtgui/controlspotpanel.h
```diff
@@ -293,6 +295,7 @@ class ControlSpotPanel:
         Gtk::TreeModelColumn<int> sensiexclu;
         Gtk::TreeModelColumn<int> structexclu;
         Gtk::TreeModelColumn<int> shapeMethod; // 0 = Independent (mouse), 1 = Symmetrical (mouse), 2 = Independent (mouse + sliders), 3 = Symmetrical (mouse + sliders)
+        Gtk::TreeModelColumn<int> avoidgamutMethod; // 0 = Independent (mouse), 1 = Symmetrical (mouse), 2 = Independent (mouse + sliders), 3 = Symmetrical (mouse + sliders)
```
Lawrence37:
```suggestion
        Gtk::TreeModelColumn<int> avoidgamutMethod;
```

rtgui/labcurve.cc
```diff
@@ -63,7 +67,28 @@ LCurve::LCurve () : FoldableToolPanel(this, "labcurves", M("TP_LABCURVE_LABEL"),
 
     avoidcolorshift = Gtk::manage (new Gtk::CheckButton (M("TP_LABCURVE_AVOIDCOLORSHIFT")));
     avoidcolorshift->set_tooltip_text (M("TP_LABCURVE_AVOIDCOLORSHIFT_TOOLTIP"));
-    pack_start (*avoidcolorshift, Gtk::PACK_SHRINK, 4);
+//    pack_start (*avoidcolorshift, Gtk::PACK_SHRINK, 4);
```
Lawrence37: Completely remove the `avoidcolorshift` widget too

rtengine/procevents.h
```diff
@@ -1177,6 +1177,7 @@ enum ProcEventCode {
     Evlocallabsigjz = 1147,
     Evlocallabsigq = 1148,
     Evlocallablogcie = 1149,
+	EvLocallabavoidgamutMethod = 1150,
```
Lawrence37: It's better to create the procevent like you did here:
https://github.com/Beep6581/RawTherapee/blob/6f2ca3f2581e6f6dcecb00ffb6bb6a730988c972/rtgui/labcurve.cc#L39

That way, you can have a more descriptive history name like `HISTORY_MSG_AVOID_GAMUT_METHOD` instead of `HISTORY_MSG_1151`.

Lawrence37 (commented):
There are some indentation inconsistencies in this PR as well. I think your text editor is using tab characters instead of spaces.

Desmis:
@Lawrence37 
Thank you for all these relevant comments and the time you spent on it.
The recent change of computer (and instalation of software) and my illness (and my age) must partly explain certain anomalies.

I will try to fix it.

But the basic problem is the colorimetric behavior, both in 'Lab adjustments' and 'local adjustments' there were anomalies bugs, due to bad choices (often GUI) on my part. It is remedied, but it is necessary tidy up the code.

thanks again

Jacques

rtengine/procevents.h
Desmis: Done

rtdata/languages/default
Desmis: Done

rtgui/controlspotpanel.h
Desmis: done

rtgui/controlspotpanel.h
Desmis: done

rtgui/controlspotpanel.cc
Desmis: done... I hope :)

rtdata/languages/default
Desmis: done...But my english ???

rtgui/labcurve.cc
Desmis: Done , I hope :)

Desmis:
@Lawrence37 

I think, I hope I made all suggested changes, hope I didn't make mistakes in "procevents" and "refreshmap"

The wording of the tooltip is certainly a bit long, but it explains (maybe badly) what it really is. You can change if need.

Thank you for all this suggestion (good) :)
I change setup of my editor!!


I remind you, the essential purpose of this PR is to solve problems that have been present for many years:
* make good use of the Munsell correction "Uniform Perceptual Lab", without complexity in GUI : default choice
* making gamut corrections (lab or XYZ) only possible options, especially with the problem of "highlight recovery"
* make gamutmap work - which it didn't before, even if it wasn't enabled
* and also "none"...

Jacques

Lawrence37:
@Desmis Thanks for addressing my review comments. They look good. One more suggestion I have for migrating older pp3s:
```diff
diff --git a/rtengine/procparams.cc b/rtengine/procparams.cc
index 7196b9ca6..9dce2ff23 100644
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -7910,7 +7910,25 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
             assignFromKeyfile(keyFile, "Luminance Curve", "hhCurve", pedited, labCurve.hhcurve, pedited->labCurve.hhcurve);
             assignFromKeyfile(keyFile, "Luminance Curve", "LcCurve", pedited, labCurve.lccurve, pedited->labCurve.lccurve);
             assignFromKeyfile(keyFile, "Luminance Curve", "ClCurve", pedited, labCurve.clcurve, pedited->labCurve.clcurve);
-            assignFromKeyfile(keyFile, "Luminance Curve", "Gamutmunse", pedited, labCurve.gamutmunselmethod, pedited->labCurve.gamutmunselmethod);
+            if (keyFile.has_key("Luminance Curve", "Gamutmunse")) {
+                assignFromKeyfile(keyFile, "Luminance Curve", "Gamutmunse", pedited, labCurve.gamutmunselmethod, pedited->labCurve.gamutmunselmethod);
+            } else {
+                if (ppVersion < 303) {
+                    if (keyFile.has_key("Luminance Curve", "AvoidColorClipping")) {
+                        labCurve.gamutmunselmethod =
+                            keyFile.get_boolean("Luminance Curve", "AvoidColorClipping") ? "LAB" : "NONE";
+                        if (pedited) {
+                            pedited->labCurve.gamutmunselmethod = true;
+                        }
+                    }
+                } else if (keyFile.has_key("Luminance Curve", "AvoidColorShift")) {
+                    labCurve.gamutmunselmethod =
+                        keyFile.get_boolean("Luminance Curve", "AvoidColorShift") ? "LAB" : "NONE";
+                    if (pedited) {
+                        pedited->labCurve.gamutmunselmethod = true;
+                    }
+                }
+            }
 
         }
 
@@ -8421,7 +8439,16 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
                 assignFromKeyfile(keyFile, "Locallab", "StructExclu_" + index_str, pedited, spot.structexclu, spotEdited.structexclu);
                 assignFromKeyfile(keyFile, "Locallab", "Struc_" + index_str, pedited, spot.struc, spotEdited.struc);
                 assignFromKeyfile(keyFile, "Locallab", "ShapeMethod_" + index_str, pedited, spot.shapeMethod, spotEdited.shapeMethod);
-                assignFromKeyfile(keyFile, "Locallab", "AvoidgamutMethod_" + index_str, pedited, spot.avoidgamutMethod, spotEdited.avoidgamutMethod);
+                if (keyFile.has_key("Locallab", "AvoidgamutMethod_" + index_str)) {
+                    assignFromKeyfile(keyFile, "Locallab", "AvoidgamutMethod_" + index_str, pedited, spot.avoidgamutMethod, spotEdited.avoidgamutMethod);
+                } else if (keyFile.has_key("Locallab", "Avoid_" + index_str)) {
+                    const bool avoid = keyFile.get_boolean("Locallab", "Avoid_" + index_str);
+                    const bool munsell = keyFile.has_key("Locallab", "Avoidmun_" + index_str) && keyFile.get_boolean("Locallab", "Avoidmun_" + index_str);
+                    spot.avoidgamutMethod = avoid ? (munsell ? "MUNS" : "LAB") : "NONE";
+                    if (pedited) {
+                        spotEdited.avoidgamutMethod = true;
+                    }
+                }
                 assignFromKeyfile(keyFile, "Locallab", "Loc_" + index_str, pedited, spot.loc, spotEdited.loc);
                 assignFromKeyfile(keyFile, "Locallab", "CenterX_" + index_str, pedited, spot.centerX, spotEdited.centerX);
                 assignFromKeyfile(keyFile, "Locallab", "CenterY_" + index_str, pedited, spot.centerY, spotEdited.centerY);
```

For L\*a\*b\* Adjustments, this will set the method to "Lab" if "Avoid color shift" was enabled or "None" if disabled. It's the same for Local Adjustments, except if "Munsell correction only" is also enabled, the method becomes "Munsell". I hope this is correct.

Desmis:
@Lawrence37 

I had the same question that you are solving. Should past user settings be taken into account or not?
If we answer "yes", it is because we think in particular if "disabled" that the user has understood the implications of these choices, particularly for "Local Adjustmenst". It is also for convenience that I did not does.

However, this choice is as relevant as the other, the tooltip (always present in Lab adjustments) to adjust the choices. So I'm going to make the change.

Thanks again. I appreciate :)

jacques

Desmis:
If no other remarks, I will merge this afternoon :)

Jacques

Issue #6653, opened 2022-12-29
LA Color Appearance Cam16 – improvments Sigmoid Q, Log encoding Q, mask:  branch lacam16


Desmis:
**Purpose of this change – lacam16 branch**
Make the LA module "Color Appearance", for the moment for its part Cam16, more efficient (the last important modifications date from summer and autumn 2021):
* especially for difficult images (strong dynamics, deep shadows, etc.) by optimizing the processes already in place, Log Encoding Q, and Sigmoid Q;
* by allowing the masks of this module, an extension to all the attributes possible to date.
* later extend, if necessary, some of these improvements to JzCzHz and the Log Encoding module

**Two major difficulties:**
**1. Cam16 and the 6 dimensions**
A point of programming is necessary to understand certain difficulties and constraints of working with Cam16 (it is identical with Cam02). During the development of Ciecam, around 2011, 2012, it quickly became apparent that unlike working in RGB, XYZ, or L*a*b* mode which requires declarations of 3-dimensional variables, Ciecam requires 6 dimensions: J luminance, Q brightness, C chromaticity, S saturation, M colorfullness, h hue. This implies a considerable memory occupation and significant processing times. The choice was made to use these variables one by one and sequentially in a global loop, which limits the memory occupation, but has a major drawback: the optimization of a procedure called in the middle of the loop is almost impossible. As a result, optimizations of functions such as contrast, sigmoid, etc. are limited and global.

**2. Peculiarities of Q**
* By design Q (brightness) unlike J (lightness), L* (L*a*b*), and luminance calculations in RGB, HSV,... mode is calculated from absolute luminance and does not has no fixed limits (0..1, 0..65535). It is an 'open' variable.
* Q (when J is in the range [0, 1]) can reach values of 5 or 10, depending on scene conditions. This reality is a very strong constraint on the calculations of the references of the algorithms (average, thresholds, etc.) and makes it difficult to automate/optimize the algorithms.
* but Q results in a majority of cases, in a more natural rendering, closer to human perception.

**Constraints that I imposed on myself**
* avoid compatibility breaks with version 5.9, especially at the GUI level;
* reduce tweaks needed to make Sigmoid Q, and Log encoding Q work properly;
* propose a possibility to combine Sigmoid Q and Log Encoding Q;
* and of course try to get good results with the default settings, which is no small feat.

**General principle**
[...]

PR #6678, opened 2023-02-05
Improvments Local adjustments - Color Appearance Cam 16 and Log encoding

Desmis:
Hello, and excuse my bad english.

This PR aims at both, simplification and performance improvement of " Local adjustments ", " Color appearance - Cam16 ".
Branch "lacam16"

These improvements affect the GUI - which is still to be improved (because it is complex due to the mix of Cam16, Jz codes and the possibility of Zcam).

But obviously the main improvements concern the algorithms. As a reminder on the vocabulary and some principles (use of log), they are close to " Filmic ", " Sigmoid ", " New colorimetry " in Darktable. On the substance, the principles (use of Lab, Cam, Brightness, etc.), they are different. 
I also enriched the "masks" part.

For more information see issue #6653.

To summarize you can now:
* have settings that in "basic" mode should be suitable in a majority of cases (which does not mean "all cases")
* the possibility in Cam16 - which, let's remember, is advanced CAM (one can certainly say... "there is better"!), of :  Sigmoid, Log encoding, Sigmoid + Log encoding, TRC (close to "abstract profile"), ...

I don't give by principle, any name to make a review, whether it is "code" or "algorithms", free to each one to express himself.

One can also tell me that :
* "LA" is complex (in basic mode ??) it's a bit true, but requires unlearning and relearning.
* using the "Full-image" mode is not intuitive...(if someone can improve the GUI!). Just select a choice...in a "Settings combobox", and offers in relation to "main" the possibilities of LA (deltaE, excluding spot, etc.).
* not many people are interested - see https://discuss.pixls.us/t/local-adjustments-color-appearance-cam16-improvments-sigmoid-q-log-encoding-q-mask/34377 0 responses on Pixls.us (and 1 positive review) out of 197 views, and 0 interventions from #6653. Why not?

I have also update rawpedia (in french)
https://rawpedia.rawtherapee.com/Local_Adjustments/fr#Utilisation_de_Cam16_et_des_fonctionnalit%C3%A9s_HDR
jacques


Lawrence37:
I think there are several reasons why this has received less interest compared to the equivalents in darktable
- There are fewer RawTherapee users than darktable users
- darktable has underwent a major editing workflow change ("display-referred" to "scene-referred") in which filmic and sigmoid are crucial components. RawTherapee has always had a linear RGB section of the pipeline along with "Clip out-of-gamut colors" which make features like filmic and sigmoid merely alternatives to existing tools in RawTherapee.
- The authors of filmic and sigmoid have been very active in discussing their creations with darktable users
- There are many examples and tutorials of filmic and sigmoid for people to see
- filmic and sigmoid are simpler and therefore require less learning than their counterparts in RawTherapee

If you want to increase engagement, I suggest sharing simple examples of how the Color Appearance tools work better than alternatives on discuss.pixls.us. I know there is extensive documentation on RawPedia, but I think they will get much more exposure if they are posted directly in Pixls. Comparisons of results would be very helpful.

jonathanBieler:
I'd give it a try but I can't compile on my windows machine (msys2 issues), there's no artifacts built for this, right ?

Desmis:
@jonathanBieler 
I don't have an internet connection (down), I answer from my smarthphone.
No differences in this build (cmake...) with other branches or dev, otherwise additional commits without new libraries.*

All checks (7) are passed in the last 5f786'c...

Jacques

jonathanBieler:
@Desmis yes it's not an issue specific to this branch, not sure why I can't build anymore, I reinstalled everything from scratch but no success. I was just wondering if there's no binaries available somewhere in the checks.

**Edit :**  I merged the branch on my fork and it generated the binaries, I'll play with it !

Lawrence37:
@jonathanBieler FYI there are automated builds for Windows and Linux for all pull requests targeting `dev`. To access them, expand the "checks" near the bottom of the page, click on "Details" for your the build you want, navigate to the "Summary" on the left panel, and find the artifact at the bottom of the page.

jonathanBieler:
I tried it on a couple of images and overall I'm a bit confused what the purpose of the module is. In cam 16 mode there's not so many controls (I like the contrast/chroma/luminosity in the main color appearance module but they are not available here). You can use it for chromatic adaptation but it doesn't seem to be much different from warm/cool. Otherwise it can be used to handle some high dynamic image but I find Log encoding easier to use for that. The JzCzHz gave me some strange colors (looks to have differnt white balance)

The contrast control in the sigmoid seems to mainly increase the brightness and not the contrast. 

Overall it seems to work fine but I haven't found a very obvious use for it.

Log :
0V2A7339-log2
Cam16 :
0V2A7339-cam16
JzCzHz :
0V2A7339-JzCzHz

Desmis:
@jonathanBieler 
Thank you for trying.

When you say « I like the contrast/chroma/luminosity in the main color appearance module but they are not available here ». They are available but to lighten/simplify the interface, you need to deploy the **"CAM16 image adjustments"** expander.

**As far as the Log encoding part of this module and the JzCzHz part are concerned, they are unchanged from the LA version located in Dev.**

The Sigmoid part also exists in Dev, and the basic algorithm is unchanged.

What has changed in this PR:
* the possible options for Mask
* the way to try to make Sigmoid and Log encoding settings automatic, and their possible association.

Of course, you may find that "Log encoding" is simpler, that's probably true. Integrating such tools (Log encoding, sigmoid) into Cam16, as I wrote above and in the related issue is a real challenge.
JzczHz (which has not evolved) is also very complex, but has been tested. I'm not saying it's perfect - far from it - but it should be taken as an experimental module.


Thanks again for testing.

Jacques

jonathanBieler:
Hoo, I've used it for quite a while but never had the idea to click on "CAM16 image adjustments", it looks like the heading for the "Conditions de visionnage" part. It certainly makes the module much more useful, but yeah I think these controls should be a bit more in the front :)

rtdata/languages/default
```diff
@@ -2815,7 +2815,8 @@ TP_LOCALLAB_BUTTON_DUPL;Duplicate
 TP_LOCALLAB_BUTTON_REN;Rename
 TP_LOCALLAB_BUTTON_VIS;Show/Hide
 TP_LOCALLAB_BWEVNONE;None
-TP_LOCALLAB_BWEVSIG;Sigmoid only
+//TP_LOCALLAB_BWEVSIG;Sigmoid only
```
Beep6581: Entries should be deleted, not commented out.

rtdata/languages/default
```diff
@@ -2815,7 +2815,8 @@ TP_LOCALLAB_BUTTON_DUPL;Duplicate
...
+TP_LOCALLAB_BWEVSIG;Activated
```
Beep6581, line 2824: You could use GENERAL_ENABLED instead of creating a new key.

Desmis: Hello @Beep6581 
OK for delete instead of commented out.
For the new key,  I want to keep the possibility of a third choice (possible in locallabtools2)

Jacques

Issue #6663, opened 2023-01-11
RT crush when using before/after + abstruct Profile


Foxtailgh:
RT crushes when using before/after + abstruct Profile

1. Go to tab - color 
2. Click on  color managment
3. Scroll down to abstruct Profile 
4. Change abstruct Profile to custom and make changes in gamma and slope
5. Click before/after
6. Rt will crush.

RawTherapee 5. 9
Windows 10

Thanatomanic:
See #6638. It is very likely the same issue.

matei-georged:
Hey, 

I tested this on windows with a recent dev build (8047aded3ba4756233be0d4e833e535bdd6293f0). This doesn't seem to be related to #6638 . It's a crash in the GUI, `primListener` at `rtengine/improccoordinator.cc:1865` (`rtengine::ImProcCoordinator::updatePreviewImage`) seems to be a deallocated pointer.  I will to look into it.

matei-georged:
Hey, 

I think I found the problem. I fixed it and MR will be on the way

Lawrence37:
Thanks for the fix @din99dany. Before the merge, the AppImage crashes. After, it no longer crashes.

PR #6686, opened 2023-02-18
RT crush when using before/after + abstruct Profile #6663

matei-georged:
Hey, 

As described in the bug, there was a crash when using the before and after function after doing some adjustments to the photo. 
I managed to track the problem to a bad pointer in `rtengine/improccoordinator.cc:1865 ` (`rtengine::ImProcCoordinator::updatePreviewImage`). What was happening was that is that when using the before/after function a new `ImProcCoordinator` object is created. The problem was that `primListener` was never initialized  with `NULL` and it contained garbage and the member wasn't set by `ToolPanelCoordinator`. The logic inside `updatePreviewImage` seems good as the member was checked for `NULL` before using it. 

As for the fix, just initializing the member with `NULL` seems to have solved the problem. If you think any other changes or analysis is needed please let me know but it's clear that we should initialize the variable with `NULL`. 

Thank you,
Daniel

### DIFF
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -163,6 +163,7 @@ ImProcCoordinator::ImProcCoordinator() :
     imageTypeListener(nullptr),
     filmNegListener(nullptr),
     actListener(nullptr),
+    primListener(nullptr),
     adnListener(nullptr),
     awavListener(nullptr),
     dehaListener(nullptr),


matei-georged:
Link to the original issue: #6663

Floessie:
@din99dany Thanks for the fix. As this is a no-brainer, I merged it without much ado. :+1:

birbacher:
Should we update the code to use default member initializers (C++11), particularly for trivial built-in types such as pointers or integers?  That would make it less likely to miss them in the constructors, see https://en.cppreference.com/w/cpp/language/data_members

matei-georged:
That sounds good. Maybe we can also add `-Winit-self` to the compilation flags to make the uninitialized members warnings be treated as errors.

PR #6687, opened 2023-02-18
PDAF lines filter crashes with RAWs from Nikon Z 50

### LINKED ISSUE #6630: PDAF lines filter crashes with RAWs from Nikon Z 50
**This bug has been fixed in ART some days ago:** https://bitbucket.org/agriggio/art/commits/365a76ed78e36dac7d7797510b750aa92276da03

**Steps to reproduce**

-> Raw-Tab
-> Sensor Matrix
-> Preprocessing
-> Activate *PDAF lines filter*
-> Crash

**Expected behavior**

Not to crash.

**Additional information**

Up-to-date Windows 11, latest automated dev build as of today.

I tried RAWs from about 10 - 12 cameras, including Z fc, other Nikons of the Z series and some Sonys, Canons, but the bug is very picky and only crashes with Z 50 so far.

Screencast:
https://user-images.githubusercontent.com/96216629/205400968-8f6d899f-06a5-4bc8-abc0-f35c306fb0b9.mp4

Sample raw: 
https://share.mailbox.org/ajax/share/0731300e0495c64d7237702495c64b13a6008e2447669246/1/8/Mzg/MzgvNg

BT:
[log.txt](https://github.com/Beep6581/RawTherapee/files/10144784/log.txt)

matei-georged:
Hey, 

Link to the original issue #6630 . I tested it and it was crashing in windows 10 as well. 

There seems to be a problem in the logic of the pdfa filter, `PDAFLinesFilter::markLine` might get out of the bounds of the image as `PDAFLinesFilter::mark` will try to mark the adjacent lines as well for a specific row and  `PDAFLinesFilter::markLine` it's looking at the adjacent lines as well. So, this will give an effective offset of 2 lines ... 

The crash was happening on the last row but it should have happened for the first row as well. I think this was not the case because the pattern in `PDAFLinesFilter::mark` might have prevented that. I saw that this pattern in static and loaded from some file ... so maybe we could make an update on that pattern to exclude the second to last line as well ? 

For the moment, I added some extra checks for boundaries in `PDAFLinesFilter::mark`. This solved the crash and I belief it's side effect free. I tough that maybe changing the logic in `PDAFLinesFilter::markLine` to do some bound check might be better, but then we should give `g2` and `g4` some dummy values when they get out of image's bounds. For example, trying `0` would result in the line not being marked, same if the dummy value is some big negative float. So I thought it might be better to skip the call to `markLine` altogether. 

On the [Rawpedia ](https://rawpedia.rawtherapee.com/Preprocessing) I found this in the documentation: 
```
RawTherapee 5.5 can fix Nikon PDAF banding artifacts for the following cameras:
- Nikon Z 6
- Nikon Z 7
```
So maybe the Nikon Z 5 is not supported and it's expected not to work ?

Please let me know your opinion about what would be the best course of action.

Thank you,
Daniel

### DIFF
--- a/rtengine/pdaflinesfilter.cc
+++ b/rtengine/pdaflinesfilter.cc
@@ -275,7 +275,11 @@ int PDAFLinesFilter::mark(const array2D<float> &rawData, PixelsMap &bpMap)
     for (int y = 1; y < H_-1; ++y) {
         int yy = pattern_[idx] + off;
         if (y == yy) {
-            int n = markLine(rawData, bpMap, y) + markLine(rawData, bpMap, y-1) + markLine(rawData, bpMap, y+1);
+            int n = 0;
+            n += markLine(rawData, bpMap, y);
+            n += (y-1 <= 0   ) ? 0 : markLine(rawData, bpMap, y-1);
+            n += (y+1 >= H_-1) ? 0 : markLine(rawData, bpMap, y+1);
+            
             if (n) {
                 found += n;
                 if (settings->verbose) {


Lawrence37:
What do you think of the solution used by Alberto in https://bitbucket.org/agriggio/art/commits/365a76ed78e36dac7d7797510b750aa92276da03? He is the one who originally wrote the PDFA filter code.

matei-georged:
Well, I think it's better to skip just the oob adjencent lines. 

Lines 0 and 1 will not be marked directly in his version. With my version we will semi mark line 1 (same thing for `h-2`). I think there should be some logic for the troubled lines as well. 
More or less it's the same thing.  

What do you think?

Lawrence37:
I guess this is fine.

Lawrence37:
Thanks for the fix @matei-georged

Issue #1843, opened 2015-08-11
Mapping/reprojection of perspective


Beep6581:
Originally reported on Google Code with ID 1859
```
Wish :) for ability to map/reproject images, e.g. from fisheye to rectilinear or panini.
http://en.wikipedia.org/wiki/Fisheye_lens
```

Reported by `entertheyoni` on 2013-05-02 02:36:39

Beep6581:
Handled by Hugin, closing.

PR #6689, opened 2023-02-21
De-fish

### LINKED ISSUE #6676: Feature request: Add tool for de-fishing images taken with fisheye lenses
I suggest adding de-fishing capabilities to RT. There already is a feature "Distortion correction" but that seems to be limited to the distortion exhibited by common rectilinear lenses.

This was taken with a 6.5mm f/2 lens:
DSC02590-s1

And I want to have the option to convert it to this:
DSC02590 jpgundistorted-s

I implemented a demo how to defish an image using OpenCV to create the above example: https://github.com/abrock/defishdemo

I could give it a shot to implement it in RT, but I'd like to get a bit of feedback first if this feature would be accepted.

When undistorting a fisheye image there is one problem: The source pixels will be stretched more the further away from the center they are. This means that if I want to preserve the resolution in the center and (most of) the field of view I have to increase the size of the destination image. In the example above I scaled width and height by a factor of 3.45. Is this possible in RT? I think there also would need to be controls for that so the user can choose between increasing image size and reducing the resolution in the center.

abrock:
This pull-request introduces two changes to the perspective correction tool, previous discussion was in #6676 

1. Add support for de-fishing by clicking a checkbox.
The code uses the focal length provided by the user (already existing feature) and assumes a perfect equidistant fisheye model for computing the un-distortion operation. It works well together with the other perspective corrections.

Here's a perfect checkerboard rendered using blender with a equidistant camera model (and also a cat):
https://drive.google.com/drive/folders/1_DOu7Zm-_YjaRgid8okSnyvQ6rP6Gtm8?usp=share_link

The sidecar file already has the correct focal length set. If you change any perspective parameter it will always look like a checkerboard captured using a rectilinear lens, even though the image file shows a circular fisheye.

2. Add a scale option to the perspective adjustments.
This is needed since de-fishing of an extreme fisheye lens means scaling the corners of the fisheye image. Without down-scaling we lose a large portion of the field of view.

An alternative (or addition) would be to allow the user to increase the resulting image size. This would have the benefit that the center doesn't have to be scaled down (or not as much) so the optical resolution in the center is preserved. I think the current state is quite useful already, so increasing the image size should be tackled separately later.


rtengine/iptransform.cc
```diff
@@ -519,6 +519,21 @@ bool ImProcFunctions::transCoord (int W, int H, const std::vector<Coord2D> &src,
             pLCPMap->correctDistortion(x_d, y_d, w2, h2);
         }
 
+        x_d /= params->perspective.camera_scale;
+        y_d /= params->perspective.camera_scale;
+        if (params->perspective.camera_defish) {
+            double const focal_source = params->perspective.camera_focal_length * maxRadius * 0.0462;
```
Floessie: We use `const` on the left throughout the code. Please change `double const` to `const double`. Thanks.

rtengine/iptransform.cc
```diff
@@ -519,6 +519,21 @@ bool ImProcFunctions::transCoord (int W, int H, const std::vector<Coord2D> &src,
...
+            double focal_dst = focal_source;
+            x_d /= focal_dst;
+            y_d /= focal_dst;
+
+            double const r = std::sqrt(x_d*x_d + y_d*y_d);
+            double const factor = focal_source * std::atan(r)/r;
```
Floessie: Only nitpicking, but please add spaces to binary operators independent of the precedence. So this should read `x_d * x_d + y_d * y_d` and `focal_source * std::atan(r) / r`.

rtengine/iptransform.cc
```diff
@@ -1244,6 +1259,21 @@ void ImProcFunctions::transformGeneral(bool highQuality, Imagefloat *original, I
                 pLCPMap->correctDistortion(x_d, y_d, w2, h2);
             }
 
+            x_d /= params->perspective.camera_scale;
+            y_d /= params->perspective.camera_scale;
+            if (params->perspective.camera_defish) {
+                double const focal_source = params->perspective.camera_focal_length * maxRadius * 0.0462;
```
Floessie: Maybe `0.0462` isn't that magical, but I can't tell from reading the code how you came up with it. A probably good idea was to put a `constexpr double reasonable_name = 0.0462;` into an anonymous namespace at the beginning of the file, or when used only locally, to the start of the function.

rtengine/iptransform.cc
```diff
@@ -1244,6 +1259,21 @@ void ImProcFunctions::transformGeneral(bool highQuality, Imagefloat *original, I
...
+                double focal_dst = focal_source;
```
Floessie: Why this copy?

rtgui/batchtoolpanelcoord.cc
```diff
@@ -197,7 +197,7 @@ void BatchToolPanelCoordinator::initSession ()
             rotate->setAdjusterBehavior (options.baBehav[ADDSET_ROTATE_DEGREE]);
             resize->setAdjusterBehavior (options.baBehav[ADDSET_RESIZE_SCALE]);
             distortion->setAdjusterBehavior (options.baBehav[ADDSET_DIST_AMOUNT]);
-            perspective->setAdjusterBehavior (options.baBehav[ADDSET_PERSPECTIVE], options.baBehav[ADDSET_PERSP_CAM_FOCAL_LENGTH], options.baBehav[ADDSET_PERSP_CAM_SHIFT], options.baBehav[ADDSET_PERSP_CAM_ANGLE], options.baBehav[ADDSET_PERSP_PROJ_ANGLE], options.baBehav[ADDSET_PERSP_PROJ_SHIFT], options.baBehav[ADDSET_PERSP_PROJ_ROTATE]);
+            perspective->setAdjusterBehavior (options.baBehav[ADDSET_PERSPECTIVE], options.baBehav[ADDSET_PERSP_CAM_FOCAL_LENGTH], options.baBehav[ADDSET_PERSP_CAM_SHIFT], options.baBehav[ADDSET_PERSP_CAM_ANGLE], options.baBehav[ADDSET_PERSP_PROJ_ANGLE], options.baBehav[ADDSET_PERSP_PROJ_SHIFT], options.baBehav[ADDSET_PERSP_PROJ_ROTATE], options.baBehav[ADDSET_PERSP_CAM_SCALE]);
```
Floessie: A long line getting even longer. I propose to break it like so:

```c++
perspective->setAdjusterBehavior (
    options.baBehav[ADDSET_PERSPECTIVE],
    options.baBehav[ADDSET_PERSP_CAM_FOCAL_LENGTH],
    options.baBehav[ADDSET_PERSP_CAM_SHIFT],
    options.baBehav[ADDSET_PERSP_CAM_ANGLE],
    options.baBehav[ADDSET_PERSP_PROJ_ANGLE],
    options.baBehav[ADDSET_PERSP_PROJ_SHIFT],
    options.baBehav[ADDSET_PERSP_PROJ_ROTATE],
    options.baBehav[ADDSET_PERSP_CAM_SCALE]
);
```

rtgui/paramsedited.cc
```diff
@@ -3303,6 +3307,14 @@ void ParamsEdited::combine(rtengine::procparams::ProcParams& toEdit, const rteng
         toEdit.perspective.camera_crop_factor = dontforceSet && options.baBehav[ADDSET_PERSP_CAM_FOCAL_LENGTH] ? toEdit.perspective.camera_crop_factor + mods.perspective.camera_crop_factor : mods.perspective.camera_crop_factor;
     }
 
+    if (perspective.camera_defish) {
+        toEdit.perspective.camera_defish = mods.perspective.camera_defish;
+    }
+
+    if (perspective.camera_scale) {
+        toEdit.perspective.camera_scale = dontforceSet && options.baBehav[ADDSET_PERSP_CAM_FOCAL_LENGTH] ? toEdit.perspective.camera_scale + mods.perspective.camera_scale : mods.perspective.camera_scale;
```
Floessie:
```c++
toEdit.perspective.camera_scale =
    dontforceSet && options.baBehav[ADDSET_PERSP_CAM_FOCAL_LENGTH]
        ? toEdit.perspective.camera_scale + mods.perspective.camera_scale
        : mods.perspective.camera_scale;
```

rtgui/perspective.cc
```diff
@@ -247,6 +251,13 @@ PerspCorrection::PerspCorrection () : FoldableToolPanel(this, "perspective", M("
     projection_rotate = Gtk::manage (new Adjuster (M("TP_PERSPECTIVE_PROJECTION_ROTATE"), -45, 45, 0.01, 0, ipers_rotate_left, ipers_rotate_right));
     projection_rotate->setAdjusterListener (this);
 
+    camera_scale= Gtk::manage (new Adjuster (M("TP_PERSPECTIVE_CAMERA_SCALE"), .1, 10, 0.01, 0, ipers_rotate_left, ipers_rotate_right));
```
Floessie: Why `.1` and `0.01` on the same line?

rtgui/perspective.cc
```diff
@@ -452,6 +474,7 @@ void PerspCorrection::setDefaults (const ProcParams* defParams, const ParamsEdit
         camera_crop_factor->setDefaultEditedState (pedited->perspective.camera_crop_factor ? Edited : UnEdited);
         camera_focal_length->setDefaultEditedState (pedited->perspective.camera_focal_length ? Edited : UnEdited);
         camera_pitch->setDefaultEditedState (pedited->perspective.camera_pitch ? Edited : UnEdited);
+        camera_scale->setDefaultEditedState (std::abs(pedited->perspective.camera_scale-1) > 1e-6 ? Edited : UnEdited);
```
Floessie: `camera_scale - 1`

rtgui/perspective.h
```diff
@@ -64,29 +64,37 @@ class PerspCorrection final :
     Gtk::Button* lines_button_erase;
     Adjuster* projection_pitch;
     Adjuster* projection_rotate;
+    Adjuster* camera_scale;
+    Gtk::CheckButton *camera_defish;
```
Floessie: `Gtk::CheckButton* camera_defish;`

rtgui/perspective.h
```diff
@@ -119,8 +127,9 @@ class PerspCorrection final :
     void linesEditButtonPressed (void);
     void linesEraseButtonPressed (void);
     void methodChanged (void);
+    void defishChanged (void);
     void requestApplyControlLines(void);
-    void setAdjusterBehavior (bool badd, bool camera_focal_length_add, bool camera_shift_add, bool camera_angle_add, bool projection_angle_add, bool projection_shift_add, bool projection_rotate_add);
+    void setAdjusterBehavior (bool badd, bool camera_focal_length_add, bool camera_shift_add, bool camera_angle_add, bool projection_angle_add, bool projection_shift_add, bool projection_rotate_add, bool projection_scale_add);
```
Floessie: That should be re-formatted to:

```c++
void setAdjusterBehavior(
    bool badd,
    bool camera_focal_length_add,
    bool camera_shift_add,
    bool camera_angle_add,
    bool projection_angle_add,
    bool projection_shift_add,
    bool projection_rotate_add,
    bool projection_scale_add
);
```

Change the definition accordingly.

Floessie (changes requested):
@abrock Thanks for your contribution! :+1: 

I've left some comments for you to streamline the code. They are just some minor formatting issues that should be easy to fix.

Best,
Flössie

rtengine/iptransform.cc
abrock: The whole geometric transformation code was already twice in this file, once for the actual transformation and once for figuring out which source pixels are needed when showing some zoomed in portion of the image. If I don't add the fisheye code in both places the preview doesn't work when zooming in.

I think the code duplication should be avoided but I didn't look into it so I don't know if it's easy to put it into a single function and call it in both places.

rtgui/perspective.cc
abrock: 0.1 is the minimum value for the scale and 0.01 is the  step. Does that answer the question?

rtengine/iptransform.cc
abrock: Fixed.

rtengine/iptransform.cc
abrock: It's kind of magical as in I don't understand it and figured it out by experiment:
It converts the focal length set by the user into a focal length that makes sense for the current scale of x_d and y_d.
I adjusted it so it works for the rendered checkerboard which has a "physical focal length" of ~8.09mm on a small frame sensor.
It also works for my real 6.5mm lens

rtengine/iptransform.cc
Floessie: Sorry, my thinking was more mundane: `focal_dst` isn't changed in this scope, if I'm not mistaken, so why not use `focal_source` right away?

rtgui/perspective.cc
Floessie: No. :grin: 

If you look at the lines nearby, there are more `0\.[0-9]+` than `\.[0-9]+`. So the `.1` should read `0.1`, that's all.

rtengine/iptransform.cc
abrock: Ah, that was a relic from my early experiments, I fixed it.

rtgui/perspective.cc
abrock: Fixed.

Floessie (approved):
@abrock @Lawrence37 @Thanatomanic I'm fine code-wise.

@Lawrence37 @Thanatomanic Please review functionality.

abrock:
I found one bug I'm not sure what's causing it:
When I open an image I haven't opened before and enable "De-fish" the preview gets 100% black. As soon as I change the focal length it suddenly works as expected.

Lawrence37:
De-fishing is not related to perspective, so it makes more sense to place the check box in the upper level or inside Distortion Correction. 
image

Similarly, Scale applies to more than just perspective. I suggest combining it with Auto-fill, like how the radius in Capture Sharpening can be calculated automatically or set manually.

rtengine/iptransform.cc
```diff
@@ -1244,6 +1264,24 @@ void ImProcFunctions::transformGeneral(bool highQuality, Imagefloat *original, I
...
             }
 
+            x_d /= params->perspective.camera_scale;
+            y_d /= params->perspective.camera_scale;
+            if (params->perspective.camera_defish) {
+                const double focal =
+                        params->perspective.camera_focal_length *
+                        maxRadius *
+                        focal_length_in_px_factor *
+                        params->perspective.camera_crop_factor;
```
Lawrence37: Use `f` from here https://github.com/Beep6581/RawTherapee/blob/e0b72af60efe0616e38d9d656fc1dfae080a9ac3/rtengine/iptransform.cc#L1184-L1187

That may fix the initial black image issue and you can get rid of the magic number (`1 / √( (36/2)² + (24/2)² )` == Inverse radius of sensor with crop factor of 1).

rtengine/iptransform.cc
```diff
@@ -1244,6 +1264,24 @@ void ImProcFunctions::transformGeneral(bool highQuality, Imagefloat *original, I
                 pLCPMap->correctDistortion(x_d, y_d, w2, h2);
             }
 
+            x_d /= params->perspective.camera_scale;
```
Lawrence37: Is this the correct order for de-fishing? I vaguely remember that LCP distortion correction for fish-eye lenses results in an ideal fish-eye image. Therefore de-fishing comes after.

abrock:
I agree that de-fishing is unrelated to perspective. At the moment there's a problem with moving the checkbox to distortion: The effect of de-fishing depends to a very high degree on the focal length setting in perspective.
Should I add a second focal length setting in the distortion correction section?

rtgui/perspective.cc
```diff
@@ -247,6 +251,13 @@ PerspCorrection::PerspCorrection () : FoldableToolPanel(this, "perspective", M("
     projection_rotate = Gtk::manage (new Adjuster (M("TP_PERSPECTIVE_PROJECTION_ROTATE"), -45, 45, 0.01, 0, ipers_rotate_left, ipers_rotate_right));
     projection_rotate->setAdjusterListener (this);
 
+    camera_scale= Gtk::manage (new Adjuster (M("TP_PERSPECTIVE_CAMERA_SCALE"), 0.1, 10, 0.01, 0, ipers_rotate_left, ipers_rotate_right));
```
Lawrence37:
```suggestion
    camera_scale= Gtk::manage (new Adjuster (M("TP_PERSPECTIVE_CAMERA_SCALE"), 0.1, 10, 0.01, 0));
```

Also, for some reason the reset button is missing.

Lawrence37 (changes requested):

Lawrence37:
> Should I add a second focal length setting in the distortion correction section?

That is one option, or you can move the focal length to the upper level. The choice depends on if there is ever a situation where the focal lengths need to be different.

rtengine/iptransform.cc
abrock: Nice find with the explanation of the magic number, I removed it and use `f` as suggested.

abrock:
> That is one option, or you can move the focal length to the upper level. The choice depends on if there is ever a situation where the focal lengths need to be different.

In theory the correct focal length for de-fishing is also the correct focal length for perspective, assuming a perfect fisheye. In practice I don't use the correct focal length, I use something that looks good.
After de-fishing a user might want to change the perspective so something that looks good with all the degrees of freedom provided by the perspective tool.

So it might make more sense to have separate tools without shared parameters.

rtengine/iptransform.cc
abrock: Changed as requested.

Beep6581:
I'm very happy to see de-fishing capability in RT!

A thought regarding the discussion on where to put this tool in the GUI and pipeline. De-fishing is a form of re-projection - you are changing the projection from equidistant fisheye to rectilinear. It is not a form of distortion correction. While in lay terms anything wonky is called distorted, there's more to it. Definitions vary, but one which serves well to clarify the matter is that a distortion is a deviation from the perfect projection, whatever that projection might be. A fisheye lens can itself suffer from distortion if the image produced by the lens is not perfectly projected (nothing is). Theoretically, a user might want to perform distortion correction on a fisheye image without re-projecting it into rectilinear.

With the tool the way it currently is, I cannot re-project a photo and fix perspective automatically - would be nice if that was possible.

screenshot_20230302_174122

screenshot_20230302_174204

screenshot_20230302_174221

Test image:
http://rawtherapee.com/shared/test_images/fisheye_samyang_8mm_shed.pef

abrock:
I did the following test with you image:
1. de-fish it, export it
2. auto-correct the perspective.

Result:
image

So I stepped through PerspectiveCorrection::autocompute after hitting the auto-correct button to see what's happening when I try auto-correction with the fisheye.
Here's my understanding of what's happening:
The perspective auto-correction is searching for straight lines in the original image and using them to auto-correct the perspective. Since the image contains mostly curved lines this just fails.
The solution would be (I guess) to de-fish the image, then detect the lines, compute the perspective correction.
I tried copying the parameters from the de-fished and then auto-corrected image and it almost works:

image

I think I fixed it by copying the relevant parameters to the `neutral` variable in `perspectivecorrection.cc`. I commented out 
`if (ipf.needsTransform...` since it would always return false and I don't understand why.
Now it works with your image, this is what I get from automatic vertical correction:

image

Regarding your comments on distortion vs. projection: If I understand it correctly it is already possible in RT to un-distort a imperfect fisheye image to get a perfect fisheye image given the right lensfun profile, right?

abrock:
When I disable auto-fill and scale down the image I can get automatic perspective correction for both vertical and horizontal:

image

Lawrence37:
> I commented out `if (ipf.needsTransform...` since it would always return false and I don't understand why.

I'm guessing it is because de-fish is the only active transform, but it is currently part of perspective transformation and perspective is disabled when performing the automatic calculation.

abrock:
I figured it out, now it works as I would expect.

Lawrence37:
Converting to a draft as the UI is not finalized.

abrock:
I'm happy to change the UI, at the moment I just don't know how it should look like, is there a consensus?

Lawrence37:
I don't know if there will be a consensus, but my opinion is the de-fish should be in a more appropriate location (upper level, distortion, or its own section, it doesn't matter) and the scale be tied to the auto-fill.

abrock:
I moved it to distortion.

abrock:
I merged the dev branch into my branch, I hope I did it right. Still compiles and works for me.

abrock:
> [...] and the scale be tied to the auto-fill.

How does that work / which behavior do you want to see?

Lawrence37:
I envision something like the the contrast threshold or radius in capture sharpening. The user can adjust it manually or choose automatic. In the current stable release, there is only an automatic option which scales the image by the smallest amount possible while filling the entire frame up with the source image. The scale also automatically updates as needed when the user tweaks the transform parameters. When combined with a slider, the user can override the scale. At that point, automatic selection of the scale is disabled until the user enables it again.

Beep6581:
Great work @abrock  Would be great to have this in 5.10.
I tested [87e96b3e9 ](https://github.com/abrock/RawTherapee/commit/87e96b3e965c6be4f53cdf5e80132e62e8e97175) on the same image and it seems to work fine. Am I correct in understanding that the "Focal length" in the "Distortion Correction" tool must be specified manually?

Beep6581:
@Lawrence37 what are your thoughts on merging this? I'm leaning towards merging and leaving improvements for the future.

abrock:
> Great work @abrock Would be great to have this in 5.10. I tested [87e96b3e9 ](https://github.com/abrock/RawTherapee/commit/87e96b3e965c6be4f53cdf5e80132e62e8e97175) on the same image and it seems to work fine. Am I correct in understanding that the "Focal length" in the "Distortion Correction" tool must be specified manually?

Yes. In theory this could be automatically set from the focal length of the lens if that information is included in the exif data, but I only own 100% manual fisheye lenses without any electronics inside them.
Another possibility would be to have some smart detection of lines that are straight in reality within a fisheye image and then automatically adjust the focal length to make the de-fished lines as straight as possible.
When working with the tool I found that eye-balling the focal length is quick and easy. Usually I re-use the  setting of one image for a whole series of images taken with the same focus setting.

Lawrence37:
I will take a look a this soon. Just scanning the code, I see the scale is inside the camera-based perspective tool. That should be moved outside to make it more accessible. The scale is also independent of the auto-scale (or does auto-scale override it?). It could complicate backward compatibility if we decide to integrate it with the auto-scale before 6.0.

abrock:
> I will take a look a this soon. Just scanning the code, I see the scale is inside the camera-based perspective tool. That should be moved outside to make it more accessible.

I'm happy to change it if someone could make a mockup, or draw some arrows in a screenshot of what should go where. 

> The scale is also independent of the auto-scale (or does auto-scale override it?). It could complicate backward compatibility if we decide to integrate it with the auto-scale before 6.0.

With de-fishing disabled:
Setting a scale smaller than 1 does nothing. Setting a scale larger than 2 zooms in (I'm confused, I would have expected a scale larger than 1 to already zoom in). With Auto-Fill disabled, it works as I expected it, scale < 1 means zooming out and black borders appearing, scale > 1 means zooming in.

With de-fishing enabled:
Setting a scale smaller than 1 suddenly makes sense, it allows for a larger field of view to be visible in the rectified image. With auto-fill disabled and vertical field of view of the input < 180° Auto-fill prevents very small scales which would lead to some areas of the resulting image to be black. Without Auto-fill it's the user's job to crop it correctly

I think the current behavior makes sense.

Lawrence37:
I think putting it at the first arrow shown in [this comment](https://github.com/Beep6581/RawTherapee/pull/6689#issuecomment-1445186359) makes sense.

The Auto-Fill behavior is due to it fighting the scale. There's probably a limit to how much Auto-Fill scales the image which explains why scale starts working at 2x. Disabling the scale when Auto-Fill is selected should make the behavior more intuitive. Making the slider inactive when disabled will show the user that it has no effect.

abrock:
I moved the slider to the suggested place and made it inactive when autofill is turned on:

image

I also changed the key name in the sidecar file from TP_PERSPECTIVE_CAMERA_SCALE to TP_LENSGEOM_SCALE which breaks all my existing sidecar files, next time I'll try to finalize my MR before using it in production for a year :-)

Lawrence37:
Thanks @abrock! I will take a look once I'm done with the 5.11 release tasks.

rtengine/iptransform.cc
```diff
@@ -471,6 +471,10 @@ bool ImProcFunctions::transCoord (int W, int H, const std::vector<Coord2D> &src,
             ((params->perspective.camera_focal_length > 0) ? params->perspective.camera_focal_length : PerspectiveParams::DEFAULT_CAMERA_FOCAL_LENGTH)
             * ((params->perspective.camera_crop_factor > 0) ? params->perspective.camera_crop_factor : PerspectiveParams::DEFAULT_CAMERA_CROP_FACTOR)
             * (maxRadius / sqrt(18.0*18.0 + 12.0*12.0));
+    const double f_defish =
+            ((params->distortion.focal_length > 0) ? params->distortion.focal_length : DistortionParams::DEFAULT_FOCAL_LENGTH)
+            * ((params->perspective.camera_crop_factor > 0) ? params->perspective.camera_crop_factor : PerspectiveParams::DEFAULT_CAMERA_CROP_FACTOR)
+            * (maxRadius / sqrt(18.0*18.0 + 12.0*12.0));
```
Lawrence37: Remove this dependency on the perspective parameter

rtgui/addsetids.h
```diff
@@ -20,8 +20,10 @@ enum {
     ADDSET_WB_GREEN,
     ADDSET_ROTATE_DEGREE,
     ADDSET_DIST_AMOUNT,
+    ADDSET_DIST_FOCAL_LENGTH,
     ADDSET_PERSPECTIVE,
     ADDSET_PERSP_CAM_ANGLE,
+    ADDSET_PERSP_CAM_DEFISH,
```
Lawrence37: `ADDSET_DIST_DEFISH`?

rtgui/batchtoolpanelcoord.cc
```diff
@@ -197,7 +197,15 @@ void BatchToolPanelCoordinator::initSession ()
             rotate->setAdjusterBehavior (options.baBehav[ADDSET_ROTATE_DEGREE]);
             resize->setAdjusterBehavior (options.baBehav[ADDSET_RESIZE_SCALE]);
             distortion->setAdjusterBehavior (options.baBehav[ADDSET_DIST_AMOUNT]);
```
Lawrence37: This needs the adjuster behavior for the focal length and scale.

rtgui/distortion.h
```diff
@@ -54,4 +64,6 @@ class Distortion final :
     {
         rlistener = l;
     }
+    void defishChanged (void);
+    void focalLengthChanged(Adjuster* a, double const newval);
```
Lawrence37: This function is unused

rtgui/lensgeom.cc
```diff
@@ -113,6 +127,19 @@ void LensGeometry::autoCropPressed ()
...
 
+void LensGeometry::adjusterChanged(Adjuster *a, double newval)
+{
+    if (listener) {
+        if (a == scale) {
+            listener->panelChanged (EvTransScale,
+                    Glib::ustring::format(scale->getValue()));
+        }
+        else {
+            std::cout << "Unknown adjuster given in LensGeometry::adjusterChanged, file " << __FILE__ << " line " << __LINE__ << std::endl;
```
Lawrence37: Print this only if verbose mode is true

rtgui/paramsedited.cc
```diff
@@ -3450,6 +3460,14 @@ void ParamsEdited::combine(rtengine::procparams::ProcParams& toEdit, const rteng
         toEdit.distortion.amount = dontforceSet && options.baBehav[ADDSET_DIST_AMOUNT] ? toEdit.distortion.amount + mods.distortion.amount : mods.distortion.amount;
     }
 
+    if (distortion.defish) {
+        toEdit.distortion.defish = mods.distortion.defish;
+    }
+
+    if (distortion.focal_length != DistortionParams::DEFAULT_FOCAL_LENGTH) {
+        toEdit.distortion.focal_length = dontforceSet && options.baBehav[ADDSET_DIST_FOCAL_LENGTH] ? toEdit.distortion.focal_length + mods.distortion.focal_length : mods.distortion.focal_length;
+    }
```
Lawrence37: Similar code is needed for the scale parameter.

Lawrence37 (commented):
Sorry for taking so long to review @abrock! I left a few comments.

One other suggestion. In the GUI, the focal length adjuster is not tied closely with de-fish. Wrapping focal length in a frame with de-fish would make it clearer, like how this log encoding is:
image

rtgui/paramsedited.cc
abrock: I thought the similar code is already present in lines 3447-3450
```
    if (commonTrans.scale) {
        toEdit.commonTrans.scale = mods.commonTrans.scale;
    }
```

abrock:
> One other suggestion. In the GUI, the focal length adjuster is not tied closely with de-fish. Wrapping focal length in a frame with de-fish would make it clearer, like how this log encoding is: 

Like this?
image

abrock:
Ah, the "Log Encoding" has the checkbox in the frame, I'll try to replicate that

abrock:
> One other suggestion. In the GUI, the focal length adjuster is not tied closely with de-fish. Wrapping focal length in a frame with de-fish would make it clearer, like how this log encoding is

I think I figured it out, is this how you meant it?

image

I think I implemented your suggestions / change requests except those where I commented that I'm not sure how to do it.

rtgui/paramsedited.cc
Lawrence37: Oh, I selected too much code for this comment. I mean the scale should handle the add/set behavior like what's done to the focal length here.

Lawrence37:
Yes, that looks good.

Lawrence37:
I did some more extensive testing and found these issues.
1. The history message for the de-fish checkbox just shows HISTORY_MSG_DISTORTION_DEFISH. This needs to be added to the `default` file.
2. The history message for the focal length shows up as the distortion correction amount. The ProcEvent for focal length should be created and a history message string added.
3. A manually-set scale behaves differently from the auto-fill scale. It's noticeable when using perspective correction. When toggling auto-fill, the resulting image "zooms" in and out as expected. When using the manual scale, the perspective also changes, which is not expected. The code performs the manual scale between the de-fish and perspective correction. It should be applied after all transformations (before all calculations, since the code does the inverse transformation) just like the auto-fill scale.

I think these are the last remaining issues. I finished testing all the functionality. Thanks for addressing my comments!

abrock:
@Lawrence37 I think I fixed all three problems.

Lawrence37 (approved):
Nice! Looks good to me.

FloBEAUG:
Looks like a good feature, but it would have been great if we had a geometry selector ; something like in darktable :
image
When the lens type is correctly defined in lensfun database :
`<type>fisheye</type>`
Switching from a projection to another would better than just ticking a 'de-fish' checkbox and provides more flexibility.

Lawrence37:
@FloBEAUG Such a feature was deemed out-of-scope for RawTherapee: #1843

FloBEAUG:
@Lawrence37 Yes, that was back in 2016. The argument of "Hugin already does that" isn't very convincing because switching between different software is not ideal.

Since you're working on the de-fish feature (which is essentially remapping), would it be possible to add support for other remapping transformations as well?

Is this too complex to implement? If possible, I'd like to build a version with your de-fish code to review it and perhaps add these features myself. I'd be happy to contribute to the community!

Lawrence37:
My understanding of the Hugin comment is that the choice of projection is a niche feature that does not belong in a raw image developer and is better suited for tools like panorama creators, which Hugin has already done. I do understand @Beep6581's point. Everything in RawTherapee is either used to develop an image in a parametric, non-destructive way (all the adjustments in the editor), or facilitates the development (basic file manager, histogram/scopes, inspector, etc.). Other tasks are better left for programs that specialize in them, like GIMP for pixel editing, digiKam for asset management, and Hugin for panaorama creation.

Adding various projections is a simple task, so complexity is not the problem. The question is, does it belong in a raw image developer? Feel free to make an argument for it. I had no problems with supporting a de-fish feature because it's a fairly common thing to people to do. Fish-eye lenses are common and people are used to seeing rectilinear images.

abrock:
I can implement the projection code for
"I have a lens with projection type X and want a rectilinear image"
This would probably be nice for de-fishing non-equidistant fisheye lenses

I can also implement the code for
"I have a lens with projection type X but I want the image to have projection type Y"
That is probably a much more rare use case, i typically don't think "man, this equidistant lens would look so much better as stereographic"

I think the important questions are:
- Which ob the above do we want (if any)?
- How should the UI look like

FloBEAUG:
@Lawrence37 Ok fine I understand the argument.
@abrock That's the spirit ! If the actual code can take into account the real projection of the lense (`<type>` definition in lensfun database) when selecting the de-fish option it could use the correct algorithm to de-fish this particular lens.
For example ; my Samyang 8mm is stereographic, not equidistant. I have created a lensfun profile with a custom Lens ID in ~/.exiv2 to make RT automatically select the right lens from the database. Selecting the de-fish option doesn't make the outcome rectilinear.
Rules in lensfun documentation says "if type is not mentioned, assume it's rectilinear"
For the GUI I can suggest a new drop down menu "lens projection type" that can be automatically set  if found in lensfun database or manually if not (maybe) ?

PR #6691, opened 2023-02-26
Fix the buffer-size computation flow of exif marker of JPEG

elliptic-shiho:
Hello, 

This pull-request aims to fix the computation flow of the size of JPEG marker buffer. The "buffer" is at line 1230 of `rtengine/imageio.cc`( `ImageIO::saveJPEG` ). 

https://github.com/Beep6581/RawTherapee/blob/23408bfcb3117404af376d24bad07e011e8f0fb2/rtengine/imageio.cc#L1230

In previously, it initialized as fixed length array that length is `165535` so RawTherapee will crash when that render a jpeg image with large exif data (Actually, I encountered this bug). This PR will fix this bug.

The crash bug had caused by the (fixed-size) buffer overflowed by large exif data. so I modified three files as follows:

* `rtexif/rtexif.cc`: Fix `ExifManager::createJPEGMarker` as like as `ExifManager::createPNGMarker`. 
  - `ExifManager::createJPEGMarker` assumed the given buffer has been initialized as an array that have sufficient size. 
  - In `ExifManager::createPNGMarker`, It'll initialize the given buffer by computed buffer size. 
  - After this PR, `ExifManager::createJPEGMarker` will not trust the given buffer. It compute the size of buffer needs, then initialize the buffer. 
* `rtexif/rtexif.h`: Follows the change of the signature of `ExifManager::createJPEGMarker`.
* `rtengine/imageio.cc`: Follows the change of the signature, and separate buffer variable.
  - Previously, the buffer uses as marker buffer of exif and iptc.
  - After this PR, "the buffer for exif" and "the buffer for iptc" will separate. 

Thanks and sorry for the bad English.
Shiho

### DIFF
--- a/rtexif/rtexif.cc
+++ b/rtexif/rtexif.cc
@@ -3243,7 +3243,7 @@ std::vector<Tag*> ExifManager::getDefaultTIFFTags (TagDirectory* forthis)
 
 
 
-int ExifManager::createJPEGMarker (const TagDirectory* root, const rtengine::procparams::ExifPairs& changeList, int W, int H, unsigned char *&buffer, unsigned &bufferSize)
+void ExifManager::createJPEGMarker (const TagDirectory* root, const rtengine::procparams::ExifPairs& changeList, int W, int H, unsigned char *&buffer, unsigned &bufferSize)
 {
 
     // write tiff header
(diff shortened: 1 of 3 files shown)


rtexif/rtexif.cc
```diff
@@ -3322,11 +3315,20 @@ int ExifManager::createJPEGMarker (const TagDirectory* root, const rtengine::pro
...
+    sset2 (42, buffer + offs, order);
+    offs += 2;
+    sset4 (8, buffer + offs, order);
+
+    int endOffs = cl->write (8, buffer + 6);
 
     delete cl;
 
-    return size + 6;
+    return endOffs;
```
Lawrence37: Why does this not have `+ 6` while the old code does? Or maybe this should be removed because the return value is not used anywhere anymore.

elliptic-shiho: > Why does this not have + 6 while the old code does?

It's my bug, correct value is `endOffs + 6`... thanks for pointing out.

> maybe this should be removed because the return value is not used anywhere anymore

I agree. At the time in wrote this code, I was trying to keep the changes to a minimum. 

I think to make `ExifManager::createJPEGMarker` does not return a value, in next commit.

Lawrence37 (approved):

Lawrence37:
This will be obsolete with the Exiv2 integration, but it's good to have until then. Thanks @elliptic-shiho

PR #6695, opened 2023-02-28
increase readme logo legibility in dark mode

Benitoite:
by adding an invisible white drop shadow behind lettering.

Dark mode showing Diff:
Screenshot 2023-02-27 at 8 40 17 PM

The change does not affect the default appearance:
Screenshot 2023-02-27 at 8 42 45 PM


Beep6581:
@Benitoite I don't like the drop-shadow. How about one of these?

rt-logo-text-black-bgwhite

rt-logo-text-black-bgwhite2

rt-logo-text-black-bgwhite3

Benitoite:
They all have a white background.  That also is not compatible with dark-mode viewing...
Screenshot 2023-03-02 at 4 51 58 PM

Floessie:
@Benitoite I think that's what @Beep6581 intended. Like having a banner-like appearance with a white background in dark-mode. I like the lowest most.

Thanatomanic:
I do not like the white background, because it is added for the wrong reasons.

~~Please see https://www.stefanjudis.com/notes/how-to-define-dark-light-mode-images-in-github-markdown/~~

I would prefer to make a version that has the RawTherapee text in white to accommodate dark mode users.

Edit: this is what GitHub says in their official documentation: https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax#specifying-the-theme-an-image-is-shown-to

Beep6581:
@Thanatomanic that solves it then.

PR #6696, opened 2023-03-01
Support for saving TIFFs as BigTIFF (#6690)

### LINKED ISSUE #6690: Saving gigapixel image to tiff >= 16bit fails
I want to save an image with ~55000x27000 Pixel (1522MP). As JPEG 8 Bit works, as TIFF 8 bit works, but starting with 16 bit it fails.

Under Windows, there is no rawtherapee-debug.exe.

**How can i get log-files or debug prints of whats happening?**

RT 5.9
Windows 10 Pro 64bit 22H2

RAM usage is only one third, NVME 328GB left. So resources are free.

Thank you

Floessie:
Hello RawTherapists,

as requested in #6690 I've extended saving TIFFs in [BigTIFF](https://en.wikipedia.org/wiki/TIFF#BigTIFF) format, which was straightforward. Please check if I missed something.

Best,
Flössie

### DIFF
--- a/rtgui/saveformatpanel.cc
+++ b/rtgui/saveformatpanel.cc
@@ -71,7 +71,7 @@ SaveFormatPanel::SaveFormatPanel () : listener (nullptr)
     jpegOpts->set_row_spacing(5);
     setExpandAlignProperties(jpegOpts, true, false, Gtk::ALIGN_FILL, Gtk::ALIGN_CENTER);
 
-    jpegQual = new Adjuster (M("SAVEDLG_JPEGQUAL"), 0, 100, 1, 100);
+    jpegQual = Gtk::manage (new Adjuster (M("SAVEDLG_JPEGQUAL"), 0, 100, 1, 100) );
     setExpandAlignProperties(jpegQual, true, false, Gtk::ALIGN_FILL, Gtk::ALIGN_CENTER);
     jpegQual->setAdjusterListener (this);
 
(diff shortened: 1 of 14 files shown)


rtgui/saveformatpanel.cc
```diff
@@ -100,6 +100,11 @@ SaveFormatPanel::SaveFormatPanel () : listener (nullptr)
     tiffUncompressed->signal_toggled().connect( sigc::mem_fun(*this, &SaveFormatPanel::formatChanged));
     tiffUncompressed->show_all();
 
+    bigTiff = new Gtk::CheckButton (M("SAVEDLG_BIGTIFF"));
```
Lawrence37: Gtk::manage

Lawrence37 (changes requested):

Lawrence37:
Functionality-wise, everything looks good. I did notice some of the EXIF is missing when read with exiftool. It also shows this message
`Warning                         : Truncated ExifOffset directory`
I don't know if this is a RawTherapee problem, a BigTIFF problem, or an exiftool problem. Exiv2 is not able to read BigTIFF.

rtgui/saveformatpanel.cc
Floessie: @Lawrence37 Thanks for review! :+1: 

In `dev` both `jpegQual` and `tiffUncompressed` aren't managed. You seem to indicate there's no reason for this, so I'll change them to `Gtk::manage()` as well. Okay?

Floessie:
@Lawrence37 I'll have a look at the way the EXIF is patched in in `saveTIFF()`...

rtgui/saveformatpanel.cc
Floessie: @Lawrence37 It's quite revealing to search for ` = new Gtk::`. :raised_eyebrow:

Floessie:
@Lawrence37 I'm afraid we won't be able to support BigTIFF EXIF with our handmade `rtexif` without a lot of work. See the technical details [here](http://www.bigtiff.org/), which has a comparison about the directory formats. In RT `TagDirectory::write()` etc. make assumptions about sizes that only fit for classic TIFF. All those assumptions must be revised.

You're more into the Exiv2 toppic than I. Would it be easier to support large offset EXIFs with Exiv2?

kmilos:
Exiv2 does not currently support BigTIFF and a substantial rewrite is also needed unfortunately.

Floessie:
@kmilos Thanks for the info. I did some research in parallel and came to the [same conclusion](https://github.com/Exiv2/exiv2/issues/1329).

Floessie:
@Beep6581 @Thanatomanic How do we proceed?

1. Add BigTIFF support as is without EXIF export and add that ("without metadata") to the checkbox label.
2. Pimp our `rtexif` for large offsets so BigTIFF is fully supported. It is doable but takes some hours of work.
3. Drop this PR as we won't be able to support metadata with the Exiv2 branch anyway.
4. Support the Exiv2 folks with their BigTIFF implementation.
5. Do something else.

Best,
Flössie

Beep6581:
@Floessie I think my reply will come as no surprise :)

Personally, I would only use BigTIFF (probably unknowingly) when working with stitched panoramas. Having metadata would be preferable, but not having it would still be better than not being able to save the panorama at all.

RT should not put any resources into the current `rtexif` code and finally fully focus on Exiv2. If possible for the Exiv2 branch to support BigTIFF with some work, great. Adding BigTIFF support without metadata is entirely acceptable, as long as it's documented (and either visible via a label in the GUI somewhere, or via a tooltip).

Best regards from a thawing Sweden

Lawrence37 (approved):

Floessie:
@Beep6581 Well, I feared an "all or nothing" reply from you, so I'm positively surprised. :wink:

I'll fix the label and prevent the EXIF hack in `saveTIFF()` from overwriting the header ASAP and merge the PR (if there are no objections) on Sunday.

PR #6705, opened 2023-03-11
Local adjustments - denoise improvments

Desmis:
This PR is an attempt  to address some of the usability issues raised by a user on Pixls.us.

The user interface has been re-organised and a panel has been added to provide an indication of the residual noise levels when carrying out denoise adjustments.

Jacques


.github/workflows/windows.yml
```diff
@@ -96,8 +96,6 @@ jobs:
 
       - name: Bundle dependencies
         run: |
-          echo "Listing shared library dependencies."
-          ldd "./build/${{matrix.build_type}}/rawtherapee.exe"
```
Lawrence37: Restore this too before merging

Desmis: I think you have made this change.

Thank you

Jacques

Lawrence37:
@Desmis In this comment https://discuss.pixls.us/t/chrominance-noise-reduction-causes-size-dependent-desaturation/35582/8 you indicated the automatic chrominance noise reduction is reduced when the local adjustments denoise is enabled. I think there should be no change otherwise it can be confusing and we would have to ensure reproducibility of older edits. Anyone who uses both noise reduction tools together should already be aware that their effects stack and can make adjustments while taking that into consideration.

Desmis:
@Lawrence37 

OK, i will change

Jacques

Desmis:
If nobody objects, I will merge this PR in the next days

Jacques

Issue #6708, opened 2023-03-13
dcraw crash on Fuji Finepix HS10 with address sanitizer.


hfiguiere:
**Short description**
dcraw cause address sanitizer errors with Fuji Finepix HS10 files.

**Steps to reproduce**

Download the RAW file from https://raw.pixls.us/getfile.php/1652/nice/Fujifilm%20-%20FinePix%20HS10%20HS11%20-%2012bit%20uncompressed%20(4:3).RAF  

1. Build rawtherapee with `-DWITH_ASAN=address`
2. Go to the directory with the file above
3. Double click on the file

Crash. See below for the output.

**Expected behavior**

Doesn't crash.

**Additional information**

- Version of RawTherapee: `dev` branch. Reproduced in 5.9 or commit ca08a279d50b55d278ca1b6659a8892fac44d663
- Version of your operating system: Linux Fedora 37
- If you built from source, the contents of the `AboutThisBuild.txt` file in the installation folder:
```
Version: 5.9-126-gca08a279d
Branch: dev
Commit: ca08a279d
Commit date: 2023-03-10
Compiler: cc 12.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.7
Lensfun: V0.3.3.0
Build type: Debug
Build flags:  -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -fsanitize=address -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -g -ftree-vectorize
Link flags:   -fsanitize=address
OpenMP support: ON
MMAP support: ON
Build OS: Linux 6.1.14-200.fc37.x86_64 x86_64
Build date: Mon, 13 Mar 2023 18:28:38 +0000 UTC
Build epoch: 1678732118
Build UUID: 5a5319d8-e2be-4d9b-b304-5701a5c354fc
```

Other useful information:

This is clear cut case a `strcpy` misuse. The documentation explicit says that source and dest shall not overlap.

The address sanitizer output:
```
=================================================================
==461684==ERROR: AddressSanitizer: strcpy-param-overlap: memory ranges [0x7f038d5e0b9d,0x7f038d5e0ba7) and [0x7f038d5e0ba5, 0x7f038d5e0baf) overlap
[...]

PR #6709, opened 2023-03-13
Issue #6708 - fix overlapping buffer strcpy

hfiguiere:
- address sanitizer triggered an error on Fujifilm Finepix HS10


Lawrence37 (approved):

Lawrence37:
Thanks for the fix @hfiguiere

PR #6711, opened 2023-03-15
Mac: add CR3 defalut handling

Benitoite:
App lets system know it can open CR3 files.


Lawrence37:
I don't know anything about this, but it looks right. Thanks @Benitoite

PR #6715, opened 2023-03-17
Fixed typo "lominosité" -> "luminosité"

RubisetCie:
Maybe this is intended... anyway "lominosité" is certainly not a french word...


Benitoite (approved):
Yes, the `o` appears incorrect... especially given that lom means dirt in latin.

Hombre57 (approved):

PR #6722, opened 2023-03-20
Issue #6721 - Fix memory leak in Crop

### LINKED ISSUE #6721: Memory leak in `rtengine::Crop`
**Short description**
Memory leak in `rtengine::Crop`

**Steps to reproduce**

Download the RAW file from https://raw.pixls.us/getfile.php/1018/nice/Canon%20-%20EOS%2020D%20-%20RAW%20(3:2).CR2
You can also use the file listed in #6709 (with the patch). Or almost any other file.

- Build rawtherapee with -DWITH_ASAN=address
- Go to the directory with the file above
- Double click on the file
- Quit RawTherapee

asan report the following leak:

```
Direct leak of 3666016 byte(s) in 1 object(s) allocated from:
    #0 0x7fb91e2bb308 in operator new[](unsigned long) (/lib64/libasan.so.8+0xbb308)
    #1 0x18ecb23 in rtengine::Crop::setCropSizes(int, int, int, int, int, bool) /home/hub/source/RawTherapee/rtengine/dcrop.cc:1922
    #2 0x18d0333 in rtengine::Crop::update(int) /home/hub/source/RawTherapee/rtengine/dcrop.cc:168
    #3 0x18ed445 in rtengine::Crop::fullUpdate() /home/hub/source/RawTherapee/rtengine/dcrop.cc:1999
    #4 0x734ce1 in sigc::bound_mem_functor0<void, rtengine::DetailedCrop>::operator()() const /usr/include/sigc++-2.0/sigc++/functors/mem_fun.h:1991
    #5 0x734a6b in sigc::adaptor_functor<sigc::bound_mem_functor0<void, rtengine::DetailedCrop> >::operator()() const /usr/include/sigc++-2.0/sigc++/adaptors/adaptor_trait.h:256
    #6 0x7343c0 in sigc::internal::slot_call0<sigc::bound_mem_functor0<void, rtengine::DetailedCrop>, void>::call_it(sigc::internal::slot_rep*) /usr/include/sigc++-2.0/sigc++/functors/slot.h:136
    #7 0x7fb91d173aa9 in call_thread_entry_slot.lto_priv.0 (/lib64/libglibmm-2.4.so.1+0x5aaa9)

```

**Expected behavior**
No leak (there are others smaller still, bit the one above is gone.

**Additional information**
```
Version: 5.9-128-g5481a8bbe
Branch: dev
Commit: 5481a8bbe
Commit date: 2023-03-20
Compiler: cc 12.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.7
Lensfun: V0.3.3.0
Build type: Debug
Build flags:  -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -fsanitize=address -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -g -ftree-vectorize
Link flags:   -fsanitize=address
OpenMP support: ON
MMAP support: ON
Build OS: Linux 6.1.14-200.fc37.x86_64 x86_64
Build date: Mon, 20 Mar 2023 21:16:15 +0000 UTC
Build epoch: 1679346975
Build UUID: 1aa7869e-e061-40d2-8944-151ff8f32d18
```

hfiguiere:
- Crop::freeAll() didn't free the memory

### DIFF
--- a/rtengine/dcrop.cc
+++ b/rtengine/dcrop.cc
@@ -1735,6 +1735,11 @@ void Crop::freeAll()
             shbuffer = nullptr;
         }
 
+        if (shbuf_real) {
+            delete [] shbuf_real;
+            shbuf_real = nullptr;
+        }
+
         PipetteBuffer::flush();
     }
 


Floessie:
@hfiguiere Good catch. :+1: This is definitely one source of the memory fragmentation RT is suffering from.

Those `if (...Crop) { delete ...Crop; ... }` are crap. Deleting a `nullptr` is not an error. Could you tidy that up? In the long run the `Crop` memory handling should be changed to smart pointers or `std::` containers.

Note to the team: if we had C++14, `delete item; item = nullptr;` should be replaced by the `delete std::exchange(item, nullptr);` idiom.

Best,
Flössie

hfiguiere:
From https://en.cppreference.com/w/cpp/language/delete

> If expression evaluates to a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer. 

In short `delete nullptr` is UB.

Floessie:
No, it's undefined whether the deallocation function is called for a `nullptr` (which might be interesting for a custom global `operator delete`, but I don't think that killing a `nullptr` is of any practical interest :wink:) but everything else is well defined (same is true for good old `free()`.

In short `delete nullptr` is NOP.

HTH,
Flössie

Floessie:
`clang-tidy` even has an automated fix for this:

https://clang.llvm.org/extra/clang-tidy/checks/readability/delete-null-pointer.html

Best,
Flössie

hfiguiere:
The patch was done to follow the current code style.

Lawrence37 (approved):
This is ok but would be better with @Floessie's suggestion applied to the new and existing code.

Lawrence37:
OK to merge, or enforce stricter coding standards?

hfiguiere:
As the original submitter my answer is:

1. Is this "code style" change part of the problem. NO
2. Is the patch inconsistent with the existing style by mixing a different one? NO
3. Does the code work better after change the code style? NO

So it's good to merge.

Lawrence37:
Just to clarify, my question is directed at the "RawTherapee team".

hfiguiere:
But as the OP waiting for this to get merged or rejected, I just wanted to make sure that the "request" was unreasonable, just in case that was the blocker.

ValeZAA:
>Deleting a nullptr is not an error. 

There is also protective **programming against double free (double delete here)**, because you will not free any memory if you execute free(NULL), while free of pointer to already freed (and thus can be allocated again) memory can already corrupt everything. Yes, free(NULL) is allowed because implementation of free() must check against NULL and just return if true. Though there are some implementations that specifically remove that check to save an extra 0.5 or 0.2 CPU cycle.

>handling should be changed to smart pointers or

You cannot double delete smart pointer, it is no-op and does not cause problems. So... Do you really think they are zero overhead? No.


I do think this is better though. Just two lines.


delete [] shbuf_real;
shbuf_real = nullptr;

ValeZAA:
Oops, I actually did not notice all file is like that, checking pointer before delete. Then you may as well just apply it.

hfiguiere:
> Oops, I actually did not notice all file is like that, checking pointer before delete. Then you may as well just apply it.

As I wrote:
https://github.com/Beep6581/RawTherapee/pull/6722#issuecomment-1501020116

I did follow the original code style. Whether I like it or not wasn't the point. Consistency was. Minimal patch was. Getting this merged quickly was.

ValeZAA:
@Lawrence37 LGTM. Please merge. Ideally then fix all the file.

Hombre57:
This is a serious issue with an easy fix, merged.

Issue #6735, opened 2023-04-01
Leak in rtexif with DNGPrivate and Fujifilm makernote


hfiguiere:
**Short description**

Parsing a DNG converted Fujifilm X-Pro1 leak memory.

**Steps to reproduce**

You need RT build with address sanitizer. Same as #6733 

1. Get a file that is a RAF (Fujifilm X-Pro1) converted to DNG. The condition is DNGPrivate tag with FUJIFILM MakerNote. Might work with a RICOH or a PENTAX converted file given the code path.
2. Run `./_build/rtgui/rawtherapee-cli -d -Y -c DSCF4230.dng `
3. AddressSanitizer will report a leak.

**Expected behavior**

No leak.

**Additional information**

```
Version: 5.9-132-g351c9d821
Branch: dev
Commit: 351c9d821
Commit date: 2023-03-25
Compiler: cc 12.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.6
Lensfun: V0.3.3.0
Build type: Debug
Build flags:  -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -fsanitize=address -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -g -ftree-vectorize
Link flags:   -fsanitize=address
OpenMP support: ON
MMAP support: ON
Build OS: Linux 6.2.7-200.fc37.x86_64 x86_64
Build date: Sat, 01 Apr 2023 17:57:12 +0000 UTC
Build epoch: 1680371832
Build UUID: 40bba686-143e-452a-93ce-a082c0b05bac
```

The asan output.

```
Direct leak of 12 byte(s) in 1 object(s) allocated from:
    #0 0x7fd0a62bb308 in operator new[](unsigned long) (/lib64/libasan.so.8+0xbb308)
    #1 0x14af5d3 in rtexif::Tag::parseMakerNote(_IO_FILE*, int, rtexif::ByteOrder) /var/home/hub/source/RawTherapee/rtexif/rtexif.cc:1331
    #2 0x14a9fc6 in rtexif::Tag::Tag(rtexif::TagDirectory*, _IO_FILE*, int) /var/home/hub/source/RawTherapee/rtexif/rtexif.cc:956
[...]

PR #6742, opened 2023-04-08
Issue #6735 - Remove memory leak when processing Fuji RAF converted to DNG

hfiguiere:


Lawrence37 (approved):

Lawrence37:
Merged, thanks @hfiguiere

ValeZAA:
"In some circumstances, value may have been allocated, so"

What do you mean in some of them?? If it is not allocated you cannot delete it again! That is double free.

PR #6744, opened 2023-04-09
External editor: native commands

Lawrence37:
The multiple external editors feature changed the way applications are launched by using Gio instead of Glib or the operating system's APIs. Some applications, Photoshop in particular, have issues with the Gio launcher. Here I introduce a new attribute for external editors which marks an editor as a "native command". Native commands are launched with Glib / the OS. All existing editors which have been configured using the old interface are automatically marked as native commands which guarantees they will work as before in versions 5.9 and before.

Here you can see an example of configuring the external editors in preferences. The first two entries are converted from the old interface (GIMP and a custom command line). The third entry is added with the new interface. Users can manually toggle the "Native command" check box.
image


Lawrence37:
@AndyAstbury The workaround I had for launching Photoshop on Windows was already applied to macOS, so I had to make this change. Photoshop should be marked as a native command for it to work properly.

Lawrence37:
I will merge this soon if there are no objections.

Lawrence37:
@Benitoite, this is in the Exiv2 branch. Once a build is ready, @AndyAstbury can try it out.

PR #6755, opened 2023-04-30
Edit Lensfun database directory from preferences

Lawrence37:
[RawPedia directs users](http://rawpedia.rawtherapee.com/Lens/Geometry#Updating_your_Lensfun_database_in_Linux) to edit their `options` file if they want to use a Lensfun database in a non-standard location. This pull request allows users to open the preferences to view and change the database directory without needing to directly edit `options`.


Lawrence37:
Screenshot of the new section:
image

It is an editable text entry because an empty path (representing default locations) and relative paths are acceptable.

Hombre57:
I can't say it works fine since I don't have lensfun installed (not the correct version), but few things about the UI :
1. nothing tells that an empty field = automatic default place (they won't RTFM, I'm pretty sure). Could you add a warning icon after the label with this kind of hint : "Leave this field empty to use the default installation place"
2. a hint would also be nice over the "Lens profiles directory" label and "Lensfun database directory", what's their difference ?

Lawrence37:
I've added tooltips to the labels. By the way, the text entry will show "(Automatic)" if it is left blank.

Hombre57 (approved):

Issue #6357, opened 2021-09-11
Crash in rtengine::Color::rgb2lab01, "wprof" nullptr deref of rtengine::ICCStore::getInstance()->workingSpaceMatrix(profileCalc);


mandree:
A SIGSEGV crash (nullptr) dereference has been reported against FreeBSD's port of RawTherapee 5.8 as [FreeBSD's Bugzilla #257255](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=257255). 
Looking at color.cc and iccstore.cc git logs, I don't see obvious related code changes since release 5.8.   (I am the maintainer of FreeBSD's port and package of Rawtherapee.)

I have not fully established how the reporter triggers this (he claims zooming in, possibly moving the frame) and asked for more information, however what happens in the code is in this rtengine/color.cc fragment:

```
   const TMatrix wprof = rtengine::ICCStore::getInstance()->workingSpaceMatrix(profileCalc);

    const float xyz_rgb[3][3] = { {static_cast<float>(wprof[0][0]), static_cast<float>(wprof[0][1]), static_cast<float>(wprof[0][2])},
                                  {static_cast<float>(wprof[1][0]), static_cast<float>(wprof[1][1]), static_cast<float>(wprof[1][2])},
                                  {static_cast<float>(wprof[2][0]), static_cast<float>(wprof[2][1]), static_cast<float>(wprof[2][2])}
                                };
```

These are the arguments and locals of the relevant stack frame. 
Note `wprof` is `0`, and the crash is on the `xyz_rgb[3][3]` initialization.
If I understand the code correctly, this means that in the wMatrices, either sRGB is not found, or its `second` member itself is null.

```
#0  0x0000000000ddb930 in rtengine::Color::rgb2lab01(Glib::ustring const&, Glib::ustring const&, float, float, float, float&, float&, float&, bool)
    (profile=..., profileW=..., r=1, g=1, b=0, LAB_l=@0x7fffffffbe28: -nan(0x7fcab0), LAB_a=@0x7fffffffbe30: -nan(0x7fcab0), LAB_b=@0x7fffffffbe2c: 4.59163468e-41, workingSpace=false)
    at /usr/ports/graphics/rawtherapee/work/rawtherapee-5.8/rtengine/color.cc:483
        profileCalc = {static npos = 18446744073709551615, string_ = "sRGB"}
        wprof = 0x0
        xyz_rgb = 
[...]

PR #6765, opened 2023-05-29
Fix L*a*b* values in the lockable color picker and navigator

### LINKED ISSUE #5408: Color picker Lab readouts depend on the output profile (and are almost always wrong))
Version: 5.6
Branch: releases
Commit: 5a4ed7317
Commit date: 2019-04-20
Compiler: x86_64-pc-linux-gnu-gcc 7.3.0
Processor: Intel(R)\ Core(TM)\ i7-4790K\ CPU\ @\ 4.00GHz
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.1
Lensfun: V0.3.2.0
Build type: Gentoo
Build flags: -march=haswell -O2 -pipe -std=c++11  -Werror=unused-label -Wall -Wuninitialized -Wno-deprecated-declarations -Wno-unused-result -fopenmp -Werror=unknown-pragmas  -ftree-vectorize
Link flags: -Wl,-O1 -Wl,--as-needed
OpenMP support: yes
MMAP support: ON

Linux gentoo 4.9.34-gentoo #1 SMP Fri Aug 4 18:25:06 EDT 2017 x86_64 Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz GenuineIntel GNU/Linux

Here's the problem: 

In RT-5.6 from gentoo portage, and also RT-locallab updated and built last week, the color picker Lab readouts depend on the output profile, producing noticeably variable and wrong results especially when choosing linear TRC output profiles. Eg a reading of around L=14 that really should be around L=45.

For example, after opening an sRGB image (gimp built-in sRGB) the only selected output profile in RT that shows correct results is RTv4_sRGB . RT_sRGB (a V2 profile?) is slightly off, as is using the input profile as the output profile. Similarly with images in any other RGB color space.

Here's an example pp3 and image file - change the output profile in RT and watch the Lab values change when color-picking: 
[rt-lab-readout-bug.zip](https://github.com/Beep6581/RawTherapee/files/3490207/rt-lab-readout-bug.zip)

My apologies, I'm not going to supply a screenshot or a video. This bug is easy to reproduce. In case there is any doubt as to what part of the user interface I'm referring to, the Lab readouts are in the leftside panel. The place to change the output profile is in the right side "Color" panel.

Lawrence37:
The function used by the lockable color picker and navigator does not correctly convert RGB values to L\*a\*b\*. The pipette buffer contains RGB values that have been converted with the working or output profile using LCMS (or by a simple matrix + TRC if it is an internal profile). There is an existing function that does the correct inverse transformation. This pull request makes use of that function.

Closes #5408.
Also may fix the crash in #6357 because the problematic function is no longer used.

### DIFF
--- a/rtengine/improcfun.cc
+++ b/rtengine/improcfun.cc
@@ -5668,6 +5668,11 @@ void ImProcFunctions::rgb2lab(const Imagefloat &src, LabImage &dst, const Glib::
 
 void ImProcFunctions::rgb2lab(const Image8 &src, int x, int y, int w, int h, float L[], float a[], float b[], const procparams::ColorManagementParams &icm, bool consider_histogram_settings) const
 {
+    rgb2lab(src, x, y, w, h, L, a, b, icm, consider_histogram_settings, multiThread);
+}
+
+void ImProcFunctions::rgb2lab(const Image8 &src, int x, int y, int w, int h, float L[], float a[], float b[], const procparams::ColorManagementParams &icm, bool consider_histogram_settings, bool multiThread)
+{
     // Adapted from ImProcFunctions::lab2rgb
     const int src_width = src.getWidth();
     const int src_height = src.getHeight();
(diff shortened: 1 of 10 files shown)


Lawrence37:
I will merge this if there are no objections

PR #6766, opened 2023-05-29
Fix for L*a*b* curves histogram scaling and pipette values

### LINKED ISSUE #6519: L*a*b CC curve histogram scaled incorrectly
**Short description**
The vertical indicator line is in the wrong spot

**Steps to reproduce**
1. Create an image with the Gradient black to blue to white. Tested with tiff. Not allowed to upload it here.
blue gradient
2. Open image and look at the vertical slider in any of the chromaticity based curve editors

**Observed behavior**
The vertical indicator line only goes from 0 to grid line 9 in the most saturated part.
CC  curve max saturation

**Expected behavior**
The vertical indicator line goes from 0 to 10 (assuming that the grid is 10% steps, left is 0% aka 0, right is 100% aka 10)
The image shows the most saturated part of the base image with `L*a*b` chroma at +12
CC  curve max saturation with chroma +12

**Additional information**
RawTherapee is on the latest dev build

According to the Navigator the values for saturation, blue channel of RGB as well as the b channel are maxed out. This can not be more saturated unless it displays information outside the colour space. Why would it though? The L curve has not 10% reserved for out areas out of the luminance range.

This is the "chroma drop". Numbers refer to grid lines.
CC  curve chroma drop

Image with chroma drop at 9:
default gradient with chroma drop at 9

Image with `L*a*b` chroma +12 and chroma drop at 9:
default gradient with with chroma +12 and chroma drop at 9

Image with `L*a*b` chroma +2 and chroma drop at 8:
The CC curve of this image is the one above. Even more parts of the image should be effected by the position of the vertical indicator, they are not.
default gradient with with chroma +2 and chroma drop at 8

Lawrence37:
Fixes the x-axis alignment between the chroma histogram and curves of the CC and LC curves. Also fixes the pipette sampling values for the a*, b*, CC, and LC curves.

Closes #6519.

### DIFF
--- a/rtengine/improcfun.cc
+++ b/rtengine/improcfun.cc
@@ -4176,9 +4176,6 @@ void ImProcFunctions::chromiLuminanceCurve(PipetteBuffer *pipetteBuffer, int pW,
     }
 
 
-    const float histLFactor = pW != 1 ? histLCurve.getSize() / 100.f : 1.f;
-    const float histCFactor = pW != 1 ? histCCurve.getSize() / 48000.f : 1.f;
-
     float adjustr = 1.0f;
 
 //  if(params->labCurve.avoidclip ){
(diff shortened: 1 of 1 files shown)


Lawrence37:

PR #6702, opened 2023-03-07
Add missing fftw libraries for Windows build

Lawrence37:
Attempt to fix the missing DLL error. In the future, we should use something like ldd to include all the required libraries.

Closes #6701.

PR #6767, opened 2023-05-30
Automatically detect DLLs for Windows packages

Lawrence37:
Uses `ldd` to get all the DLLs that need to be included in the Windows packages instead of hard-coding the DLLs in the workflow file.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -98,6 +98,10 @@ jobs:
         run: |
           echo "Listing shared library dependencies."
           ldd "./build/${{matrix.build_type}}/rawtherapee.exe"
+          echo "Finding DLLs to include."
+          DLLS=($(ldd "./build/${{matrix.build_type}}/rawtherapee.exe" | grep '/mingw64/bin/' | awk '{print($1)'}))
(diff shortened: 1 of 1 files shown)


kmilos:
Please [consider again](https://github.com/Beep6581/RawTherapee/pull/6702#issuecomment-1457949458) using [`ntldd`](https://packages.msys2.org/base/mingw-w64-ntldd) instead:

> Pros:
>  * Knows about delayed loading.
>  * Can list dependencies recursively (like depends.exe does, but without
duplicating entries)

Lawrence37:
I considered ntldd, but seeing that ldd works fine for the current `dev`, I paused there. I would like to figure out what advantages ntldd may bring in the future (compared to the flavor of ldd in this pull request) and whether those bugs mentioned in the "cons" list are relevant to our use case.

kmilos:
Yeah, perhaps you're not having recursive deps... In any case, if it works, it works!

Lawrence37:
I thought I saw a DLL that ldd missed in the Exiv2 branch a while back, but the latest build shows no problems. ldd should be fine for now. I can try ntldd if we encounter an issue in the future.

PR #6774, opened 2023-06-18
Fix for HSV equalizer pipette crash

### LINKED ISSUE #6772: Crash when choosing the HSVeq adj.point tool during image processing
RT 5.9 on Linux (Debian 12) crashes when choosing the adjustment point selector tool in the HSV equalizer if the image is not yet done processing from any previous interactions. The crash is more likely to occur when the image is heavily processed (a lot of stuff is applied, see the left column). I don't think that the issue is related to any specific processing option and will just occur when many settings are applied at once.

A video of the crash
https://github.com/Beep6581/RawTherapee/assets/98033774/280f37ef-928c-4d3e-bfb7-f95d2834c933

A pp3 file
[test2.pp3.txt](https://github.com/Beep6581/RawTherapee/files/11772677/test2.pp3.txt)

Lawrence37:
Checks if the pipette buffer is valid before filling it in `rgbProc`.

Closes #6772.

### DIFF
--- a/rtengine/improcfun.cc
+++ b/rtengine/improcfun.cc
@@ -1985,7 +1985,7 @@ void ImProcFunctions::rgbProc(Imagefloat* working, LabImage* lab, PipetteBuffer 
 
     Imagefloat* editImgFloat = nullptr;
     PlanarWhateverData<float>* editWhatever = nullptr;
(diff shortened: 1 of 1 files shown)


Lawrence37:
Merging this if there are no objections.

PR #6775, opened 2023-06-18
Add .idea to .gitignore.

ghost:
Just adds .idea folder of clion (jetbrains ide) to be ignored by gitignore.


Lawrence37 (approved):
I guess it doesn't hurt to have these kinds of ignore patterns.

PR #6776, opened 2023-06-19
Bug #5912: Set dummy width and height for ThumbBrowserEntry, cause height will b…

### LINKED ISSUE #5912: scrolling with the mouse wheel in the file browser is disabled when using filters on colour labels and change folder.
When I activate the colour filters on the file browser and change folders while keeping the same filters, it is no longer possible to scroll through the photos with the mouse wheel. If I delete all the colours filters, the scrolling returns and remains if I reactivate the filters. If I change the folder again and go back to the previous one, the problem repeats itself until I deactivate them again.

Version: 5.8
Branch: releases
Commit: 9a9e0acbf
Commit date: 2020-02-04
on Ubuntu 20.04

ghost:
…e used to calculate the height of the scroll pane. Positions also need to be set as a dummy, cause no method allows only width or height.

For calculation of scroll step height, the height of a thumb will be used. If the first thumb is not drawn because it gets filtered out, the used effective height is not set and therefore zero. Because the height step will be used from the first thumb element, which might not be drawn, it takes zero for height step and therefore scroll is not possible.

This fix is just minimal-invasiv to the code. There might be better a better solution which needs way more resources.

### DIFF
--- a/rtgui/thumbbrowserbase.cc
+++ b/rtgui/thumbbrowserbase.cc
@@ -781,6 +781,7 @@ void ThumbBrowserBase::arrangeFiles(ThumbBrowserEntryBase* entry)
 
             for (int i = 0; ct < fd.size() && i < numOfCols; ++i, ++ct) {
                 for (; ct < fd.size() && fd[ct]->filtered; ++ct) {
+                    fd[ct]->setPosition(0, 0, colWidths[i], rowHeight);
                     fd[ct]->drawable = false;
                 }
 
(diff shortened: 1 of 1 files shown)


Lawrence37:
Seems promising. I will take a closer look in the coming days.

Lawrence37:
I went through the code and understand where the issue comes from. While this solution works, it's unclear at first glance why the line is needed. I suggest adding a comment in the code explaining how that line supplies the required information for scrolling. Even better would be adding new fields in `ThumbBrowserBase` to hold the row height and column width. `arrangeFiles()` sets these fields and `configScrollBars()` reads them instead of getting the values from `fd[0]->getEffective*()`

ghost:
I just added a comment. I think there are way more changes needed following the paradigm, that those values should be placed centrally. I would find that useful, but it causes a lot of changes and might also impact the logic. For me, this is too much for now.

Btw. `arrangeFiles()` gets called very often. I don't think that this is necessary. This could also be cleaned up in bigger and future changes to the logic.

Hopefully, the single change with the comment is ok for you.

Greetings.

Lawrence37:
A comment is ok for me. It's just that I saw an opportunity to improve the existing code. Very little logic change is required, only moving the width and height storage location from the entry to the browser.

Lawrence37 (approved):

PR #6777, opened 2023-06-19
Bug #6329: Variable done should be set to true inside loop, otherwise one fail w…

### LINKED ISSUE #6329: Saving profile to directory without permission results in blocked UI
System:
Windows 10 Home, Version 20H2, Build 19042.1110

RawTherapee:
Version: 5.8-3014-g94b5f7232
Branch: dev
Commit: 94b5f7232
Commit date: 2021-07-27

**Steps to reproduce**
- open a photo (raw)
- move any slider or just
- press save profile
- **NOW:** choose a folder in which you dont have the rights to save a file
(here preset i found: c:\Program Files\RawTherapee\RawTherapee_dev_5.8-3014-g94b5f7232_W64_Skylake_210727)
- Click "ok"
- Get informed by an infobox you dont have rights to save a file here
- Close this infobox by clicking „ok“
(From now on, it doesnt matter, if you save the profile to another folder or not)
- Choose a different folder and save the file anywhere or
- Just click „cancel“
- The Box „save a profile“ wont close anymore there is no chance to get back to the editor
- Finally shut down Raw Therapee by using the taskmanager.

**Expected behavior**
After saving a profile, the box „save profile“ should close.
After clicking „cancel“ the box „save profile“ should close.
Raw Therapee should come back to the editor

**Additional information**
Saving the profile in a different folder and overwriting an existing file will show the buttons YES / NO in english, seems they are not getting translated by the default-file.

ghost:
…ill cause to stay in loop forever.

Bug will be caused by setting done to false and never resetting it to be true.

### DIFF
--- a/rtgui/profilepanel.cc
+++ b/rtgui/profilepanel.cc
@@ -339,6 +339,8 @@ void ProfilePanel::save_clicked (GdkEventButton* event)
     bool done = true;
 
     do {
+        done = true;
+
         if (dialog.run() == Gtk::RESPONSE_OK) {
 
             std::string fname = dialog.get_filename();
--- a/rtgui/profilepanel.cc
+++ b/rtgui/profilepanel.cc
@@ -288,6 +288,12 @@ void ProfilePanel::save_clicked (GdkEventButton* event)
         toSave = entry ? ProfileStore::getInstance()->getProfile(entry) : nullptr;
     }
 
+    // If no entry has been selected or anything unpredictable happened, toSave
+    // can be nullptr.
+    if (toSave == nullptr) {
+        return;
+    }
+
     // If it's a partial profile, it's more intuitive to first allow the user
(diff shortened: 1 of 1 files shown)


Lawrence37:
Nice catch. The loop could be refactored to make the intention more clear and reduce the scope of the exit condition variable, but of course that's optional.

ghost:
> Nice catch. The loop could be refactored to make the intention more clear and reduce the scope of the exit condition variable, but of course that's optional.

I refactored the code just to be more readable. Buuuut, more readable to my prefs. And sometimes I like to write a little but more code, but make it better readable for me to let the code explain what it does. That is subjective for sure.

So please review whether the change is ok for everyone here.

Lawrence37 (approved):
There's always more to refactor, but this is ok.

Lawrence37:
Thanks for these pull requests @Christian-Kr!

PR #6779, opened 2023-06-23
LA Color and Light - Merge file - Screen : improve behavior

Desmis:
When we activated the "screen" blending mode, the image had strange colors. This PR solves this problem.

Jacques


Lawrence37 (approved):
Looks good. This is technically a bug-fix because the screen math was wrong, right?

Desmis:
@Lawrence37 

Yes, the math was wrong

Thank you 

Jacques

PR #6782, opened 2023-06-26
Switch from Glib threads and mutex to std threads and mutex.

ghost:
Hello to everyone,

while switching to gtk4 I realized that we are using the Glib::Threads, which has been marked as deprecated in 2015/2016. (https://www.murrayc.com/permalink/2015/11/30/glibmm-deprecated-glibthreads/)

I switched the own rt MyMutex version from Glib to std. I decided to still use the own implementation instead of removing it in general - although I don't know whether the disadvantage of Glib mutex still exist with std mutex - just because I wanted to leave the rest of the existing code in rt unchanged.

Now the most important part: I tested the code so far and it compiles and the applications runs very well, but I am not an expert for threads or mutex at all. The main task of replacing seem to be easy at a first glance, but the devil is in the details. So please have a good look and test, so we can be sure that everything is still working the same as with the Glib::Threads implementation.

I am aware that this merge request may lead to a lot of further edits, so don't worry. :-)

Thank you in advance.

Greetings

### DIFF
--- a/rtgui/threadutils.cc
+++ b/rtgui/threadutils.cc
@@ -67,7 +67,7 @@ void MyReaderLock::acquire ()
         return;
     }
 
-    Glib::Threads::Mutex::Lock lock (mutex.mutex);
+    std::unique_lock<std::mutex> lock (mutex.mutex);
 
     if (mutex.writerCount == 0) {
         // There's no writer operating, we can increment the writer count which will lock writers.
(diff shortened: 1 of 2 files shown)


Floessie:
Hi Christian,

First of all: Thank you for your contribution(s)! :+1:

Concerning this PR: I guess you were unaware of #5435. Maybe you can again join with @Hombre57 and finalize this topic quickly? Your solution is very similar to @Hombre57's and should work as well. If you find the time to look into the possible `MyReaderLock::acquire()` imbalance, all the better but not mandatory.

Best,
Flössie

Issue #101, opened 2015-08-11
Update COMPILE.txt


Beep6581:
Originally reported on Google Code with ID 110

```
Hello, I just saw that the file COMPILE.txt, Debian/Ubuntu section, says
one should install subversion (line 27). Change this to mercurial. 

Regards, Paul. 
```

Reported by `paul.matthijsse4` on 2010-05-13 13:21:33

PR #6789, opened 2023-07-05
Language names

### LINKED ISSUE #6657: Displaying correct, untranslated languages in preferences
As discussed in #6591, the language selection drop-down in preferences can be improved by showing the languages exactly as they would appear in their respective language, e.g. "Español" instead of "Espanol". Currently, the name of the language file is used as the display name (and also for storing the language in `options`). This is not good design because from a user's perspective, there is no relationship between languages and files. @Entropy512 has suggested adding the language names as entries in the language files. The file name to language display name mapping could also be in its own configuration file or in the source code itself.

What are your thoughts on how to define the mapping?

Lawrence37:
See #6657 for context.

This displays language names in their own language, e.g. Castellano Spanish is "Español (Castellano)". The name is defined in the language's respective file as a metadata variable called `LANGUAGE_DISPLAY_NAME`. I describe metadata variables in the commit message:
><pre>Metadata variables are defined in the first section of a language file,
>i.e. the comments at the beginning of the file. At most one variable can
>be defined per line. The definition takes the following format:
>    #&lt;header>&lt;whitespace>@&lt;key>=&lt;value>
>where:
>    &lt;header> is any sequence of non-whitespace characters
>    &lt;whitespace> is any non-empty sequence of whitespace characters
>    &lt;key> is the metadata key name
>    &lt;value> is the metadata value
>For example,
>    #101 @LANGUAGE_DISPLAY_NAME=English (US)
>defines a metadata key "LANGUAGE_DISPLAY_NAME" with value
>"English (US)". Lines that do not match the format are ignored.</pre>

These changes only affect the names displayed in the drop-down for selecting the language. `options` still saves the language file's name. All language names have been added to the best of my ability.

Screenshot:
image
Excuse the tofu for Chinese. Something is wrong with the fonts on my system.

Closes #6657.


rtgui/multilangmgr.cc
```diff
@@ -219,6 +242,77 @@ Glib::ustring MultiLangMgr::getStr (const std::string& key) const
...
+            std::cerr << "Unable to open language file " << fname << " to get metadata." << std::endl;
+        }
+        return nullptr;
+    }
+
+    if (rtengine::settings->verbose) {
+        std::cout << "Reading metadata from language file " << fname << std::endl;
+    }
+    std::map<std::string, std::string> raw_metadata;
+    const auto read_key_value = [&raw_metadata](const std::string &meta_line) {
```
Hombre57, line 270: @Lawrence37 The following comment (or similar) should also be in the `languages/REAME` file IMHO.

Hombre57:
@Lawrence37 PR tested on my side, it's working fine :+1: . This PR should be merged once the README has been updated.

Lawrence37:
Thanks for reviewing, @Hombre57. I agree with the README.

kmilos:
> All language names have been added to the best of my ability.

Just in case you were not aware, you can pull these from https://cldr.unicode.org/

Lawrence37:
Checking against CLDR, I confirmed all languages except for Chinese (Simplified) (not present in CLDR), Czech ([čeština in CLDR](https://github.com/unicode-org/cldr/blob/42e33f86d88066cee629687a79a42ccfd18532be/common/main/cs.xml#L143), [Český](https://invent.kde.org/frameworks/kconfigwidgets/-/blob/4f88cfa812abb46dfdb7d1092cdd393cdd116d84/po/cs/kf6_entry.desktop#L17) on my KDE system), and Brazilian Portuguese (not present in CLDR). Čeština seems correct based on a quick internet search.

PR #6794, opened 2023-07-16
Fix for Auto-matched camera profile detection when using the English (US) language

### LINKED ISSUE #5511: Auto-matched camera profile no longer working in 5.7
Hi, I've been a very happy user of Rawtherapee since about 5.0, so first let me say a huge thank you to all involved in its development!
Unfortunately my custom profile for my Panasonic camera stopped "auto-matching" in the last version. The "Auto-matched camera profile" is greyed out. I have the profile stored in /.config/RawTherapee/dcpprofiles/ and until 5.6 it always matched my Panasonic photos. (The full name is "Panasonic DMC-G80.dcp", which matches the camera detected in Rawtherapee). It can be manually selected but that would be a nuisance with every photo.
Any idea why it stopped working? Maybe it's looking for the profile in a different folder? The behaviour is the same whether I use a packaged version or an appimage, 5.6 works, 5.7 doesn't.
Thanks!
Screenshot_20191028_102438
Screenshot_20191028_102457

Lawrence37:
I was able to reproduce #5511 by setting the language to "English (US)" and launching RawTherapee with the environment variables `LANG=es_ES.UTF-8` and `LANGUAGE=es`. The issue is caused by using a locale-dependent function to refer to the DCP files. When RawTherapee starts, it finds the DCPs and uses the locale in the environment variable. Later in the program initialization, it sets the language and also updates the `LANG` environment variable. As a result, opening an image will cause RawTherapee to look up the DCP using a different locale which is problematic for certain locales.

This pull request makes the DCP lookup use a locale-independent function.

Closes #5511.

### DIFF
--- a/rtengine/dcp.cc
+++ b/rtengine/dcp.cc
@@ -430,6 +430,16 @@ std::map<std::string, std::string> getAliases(const Glib::ustring& profile_dir)
     return res;
 }
 
+/**
+ * Returns a locale-independent case-insensitive collate key. Differs from
+ * Glib::ustring::casefold_collate_key() in that the Glib method may return
+ * different results depending on the current locale.
+ */
+std::string casefold_collate_key(const Glib::ustring &str)
+{
+    return str.casefold().raw();
+}
+
 }
 
 struct DCPProfileApplyState::Data {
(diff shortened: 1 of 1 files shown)


Lawrence37:
Merging this because it is a simple fix.

Issue #6178, opened 2021-03-20
film negative: add features to spot picker


nomar500:
Hi everyone,

> Environment: RT-5.8 dev, module "Film Negative", by @rom9
> See forum discussion [here](https://discuss.pixls.us/t/any-interest-in-a-film-negative-feature-in-rt/12569)
> posted here following my sad attempt 😆 in #6175 

I suggest two evolutions to the spot picker in the Film Negative tool. Let me know if I should file two separate "Feature Request" cases.

A. Spot size:
- current behaviour: size is fixed (32*32 pixels square?)
- suggested behaviour: 
  - size can be set from a drop-down list, just as in the current White Balance pipette tool.
  - spot can be made into an arbitrary rectangle, just as a selection rectangle (see darktable's negadoctor selection tools).

B. What the spot samples:
- current behaviour: spot area average pixel value is evaluated (_to be confirmed?_).
- suggested behaviour: due to the nature of the signal (scanned film, with grain, dust and scratches, crudely compared to salt-and-pepper noise), spot needs to be smooth:
  - (quick win?) spot returns the median pixel value; 
  - (more complicated) to smoothen the spot area, its content first goes through a median _filter_ (on the spot selection, with a certain [kernel size](https://pixinsight.com/doc/legacy/LE/19_morphological/median_filter/median_filter.html), say 5 or 9) and then the average pixel value is sampled.

Let me know if something needs to be rephrased 😁

nomar500:
Hmm, what if I filed a new feature request for the WB module spot picker (the pipette): it already has adjustable size, but I don't think I've found a reference explaining how it processes the square (Rectangle, if I read some of the code).

What do you think? Would that make it easier to derive from this class, further down the line, in Film Negative?

should the user be able to choose from several sampling algorithms, when using the WB spot picker?

rom9:
Hi! Sorry for the delay on this, it's been a super busy year at work :_(

> Hmm, what if I filed a new feature request for the WB module spot picker (the pipette)

There's already issue #709 somewhat related to this for normal, non-filmnegative WB (look at the last comment). That issue suggests using an arbitrary rectangular selection in _Auto WB_ mode, which makes sense: this way, the user could choose any of the auto wb methods on the region.

> it already has adjustable size, but I don't think I've found a reference explaining how it processes the square (Rectangle, if I read some of the code).

[...]

PR #6798, opened 2023-07-20
Selectable negative pick size

LoKolbasz:
When picking _neutral_ and white _balance spots_ in the _Film Negative_ section the size of the picking rectangle can be changed. The default value is 8. Resolves feature requests in #6178.


LoKolbasz:
I've just realized, that `generateTranslationDiffs` generated changes unrelated to the picker. I can prepare a version with only the relevant changes, or one where only `rtdata/languages/default` is changed.

Lawrence37:
Usually `generateTranslationDiffs` is run in preparation for a release, so it's ok to omit it for pull requests like this one.

rtgui/filmnegative.cc
```diff
@@ -565,7 +542,7 @@ bool FilmNegative::mouseOver(int modifierKey)
 {
     EditDataProvider* const provider = getEditProvider();
     Rectangle* const spotRect = static_cast<Rectangle*>(visibleGeometry.at(0));
-    spotRect->setXYWH(provider->posImage.x - 16, provider->posImage.y - 16, 32, 32);
+    spotRect->setXYWH(provider->posImage.x - *displayRectWidth, provider->posImage.y - *displayRectWidth, *displayRectWidth * 2, *displayRectWidth * 2);
```
Lawrence37: Is `displayRectWidth` the width or radius (half width)? Here it looks like it's treated as a radius.

rtgui/filmnegative.cc
```diff
@@ -727,10 +707,13 @@ void FilmNegative::editToggled()
 
 void FilmNegative::refSpotToggled()
 {
-    if (refSpotButton->get_active()) {
+    if (refPicker.get_active()) {
 
-        spotButton->set_active(false);
+        picker.set_active(false);
         refSpotCoords.clear();
+        displayRectWidth = &(refPicker._associatedVar);
```
Lawrence37: This pointer to an `int` is a bit weird. Maybe create a function to get the active picker and return its `_associatedVar`?

rtgui/filmnegative.h
```diff
@@ -120,10 +135,14 @@ class FilmNegative final :
     Adjuster* const redRatio;
     Adjuster* const blueRatio;
 
-    Gtk::ToggleButton* const spotButton;
+    #define DEFAULT_SPOT_WIDTH 8
```
Lawrence37: Replace this macro with a `constexpr int`.

rtgui/guiutils.h
```diff
@@ -689,7 +689,36 @@ class BackBuffer : public RefCount
...
-inline void setActiveTextOrIndex (Gtk::ComboBoxText& comboBox, const Glib::ustring& text, int index)
+/** 
+ * @brief A gui element for picking spots on an image
+ */ 
+class SpotPicker : public Gtk::Grid
+{
+    public:
+        int _associatedVar;
+        Gtk::Label _spotLabel;
+        MyComboBoxText* const _spotSizeSetter;
```
Lawrence37: This can be a `MyComboBoxText` itself. If it's required to be a pointer, use `std::unique_ptr<MyComboBoxText>` instead so you don't have to explicitly `delete` it. Either way, an explicit destructor won't be necessary anymore.

rtgui/guiutils.cc
```diff
@@ -1923,3 +1923,88 @@ void BackBuffer::copySurface(Cairo::RefPtr<Cairo::Context> crDest, Gdk::Rectangl
...
+
+SpotPicker::SpotPicker(int const defaultValue, Glib::ustring const &buttonKey, Glib::ustring const &buttonTooltip, Glib::ustring const &labelKey) :
+    Gtk::Grid(),
+    _associatedVar(defaultValue),
+    _spotLabel(labelSetup(labelKey)),
+    _spotSizeSetter(new MyComboBoxText(selecterSetup())),
+    _spotButton(spotButtonTemplate(buttonKey, buttonTooltip))
+
+{
+    Gtk::Grid* spotSizeHelper = new Gtk::Grid();
```
Lawrence37: `Gtk::manage` this to avoid leaking memory, but is `spotSizeHelper` necessary? It only holds one item.

rtgui/guiutils.cc
```diff
@@ -1923,3 +1923,88 @@ void BackBuffer::copySurface(Cairo::RefPtr<Cairo::Context> crDest, Gdk::Rectangl
...
+    setExpandAlignProperties(spotSizeHelper, false, false, Gtk::ALIGN_START, Gtk::ALIGN_CENTER);
+
+    this->get_style_context()->add_class("grid-spacing");
+    setExpandAlignProperties(this, true, false, Gtk::ALIGN_FILL, Gtk::ALIGN_CENTER);
+
+    spotSizeHelper->attach (*_spotSizeSetter, 0, 0, 1, 1);
+
+    this->attach (_spotButton, 0, 0, 1, 1);
+    this->attach (_spotLabel, 1, 0, 1, 1);
+    this->attach (*spotSizeHelper, 2, 0, 1, 1);
```
Lawrence37: Hmm... This is a one row grid. Could it be a horizontal `Gtk::Box` instead?

rtgui/guiutils.h
```diff
@@ -689,7 +689,36 @@ class BackBuffer : public RefCount
...
+        Gtk::ToggleButton _spotButton;
```
Lawrence37: Make member variables private and provide getters/setters if needed.

rtgui/guiutils.h
```diff
@@ -689,7 +689,36 @@ class BackBuffer : public RefCount
...
 };
 
-inline void setActiveTextOrIndex (Gtk::ComboBoxText& comboBox, const Glib::ustring& text, int index)
+/** 
+ * @brief A gui element for picking spots on an image
+ */ 
+class SpotPicker : public Gtk::Grid
+{
+    public:
+        int _associatedVar;
```
Lawrence37: This name is vague. Do I understand correctly this will always refer to the picker size?

rtgui/filmnegative.cc
```diff
@@ -265,9 +241,10 @@ FilmNegative::FilmNegative() :
     pack_start(*greenExp, Gtk::PACK_SHRINK, 0);
     pack_start(*redRatio, Gtk::PACK_SHRINK, 0);
     pack_start(*blueRatio, Gtk::PACK_SHRINK, 0);
-    pack_start(*spotButton, Gtk::PACK_SHRINK, 0);
+    pack_start(picker, Gtk::PACK_SHRINK, 0);
+    //    pack_start(*spotButton, Gtk::PACK_SHRINK, 0);
```
Lawrence37: There's a couple of commented code throughout this pull request that can be removed.

rtgui/guiutils.h
```diff
@@ -689,7 +689,36 @@ class BackBuffer : public RefCount
...
+            return _spotButton.get_active();
+        }
+        void set_active(bool b)
+        {
+            _spotButton.set_active(b);
+        }
+
+    protected:
+        static Gtk::Label labelSetup(Glib::ustring const &key);
+        MyComboBoxText selecterSetup();
```
Lawrence37: This method can be `const`.

Lawrence37 (changes requested):
Hi @LoKolbasz, thanks for your work. This is a nice improvement on the film negative tool's usability, and creating a class for the picker widget is a good idea. I left some review comments for you.

rtgui/guiutils.h
LoKolbasz: Essentially. I"ll change the name to `_spotSize` or `_spotSideLength`.

rtgui/guiutils.cc
LoKolbasz: I cannot get the `Gtk::Box` to replicate the behaviour of the `Gtk::Grid` version. The label always seems to be reacting as if it were set to `ALIGN_FILL` and `hExpand` was `true` (it expands), while the first one is set to `ALIGN_END` and the latter one is `false`.

Example:
https://github.com/Beep6581/RawTherapee/assets/86350313/6a2908f7-4812-47a7-b175-630b31ef4966

So I'll leave the `Gtk::Grid` solution in, if that's OK.

rtgui/filmnegative.cc
LoKolbasz: It is half the length of the rectangle. This is what the user  selects as the size of the spot. It could be changed to the full length, but then the size wouldn't match the corresponding size of the WB picker. I'm renaming the variable to `_spotSideLength` but `_spotSideHalfLength` might be more appropriate. What do you think?

Lawrence37: Apparently, it's called an "apothem". I don't think many people know that word, so size or half width is better.

rtgui/guiutils.cc
Lawrence37: A grid is still fine. It's just more complex than it needs to be and therefore can make the code harder to read. In this case, there's a bunch of integers and it's not immediately clear what they mean.

rtgui/filmnegative.cc
```diff
@@ -587,8 +560,8 @@ bool FilmNegative::button1Pressed(int modifierKey)
 
                 RGB ref1, ref2, dummy;
 
-                if (fnp->getFilmNegativeSpot(refSpotCoords[0], 32, ref1, dummy) &&
-                        fnp->getFilmNegativeSpot(refSpotCoords[1], 32, ref2, dummy)) {
+                if (fnp->getFilmNegativeSpot(refSpotCoords[0], picker.get_spot_half_width(), ref1, dummy) &&
+                        fnp->getFilmNegativeSpot(refSpotCoords[1], picker.get_spot_half_width(), ref2, dummy)) {
```
Lawrence37: Should this be the full spot width?

rtgui/filmnegative.cc
```diff
@@ -634,7 +607,7 @@ bool FilmNegative::button1Pressed(int modifierKey)
             }
 
             RGB refOut;
-            fnp->getFilmNegativeSpot(provider->posImage, 32, refInputValues, refOut);
+            fnp->getFilmNegativeSpot(provider->posImage, refPicker.get_spot_half_width(), refInputValues, refOut);
```
Lawrence37: Same here

rtgui/guiutils.cc
```diff
@@ -1964,3 +1964,81 @@ void BackBuffer::copySurface(Cairo::RefPtr<Cairo::Context> crDest, Gdk::Rectangl
...
+    setExpandAlignProperties(this, true, false, Gtk::ALIGN_FILL, Gtk::ALIGN_CENTER);
+
+    this->attach (_spotButton, 0, 0, 1, 1);
+    this->attach (_spotLabel, 1, 0, 1, 1);
+    this->attach (_spotSizeSetter, 2, 0, 1, 1);
+    _spotSizeSetter.signal_changed().connect( sigc::mem_fun(*this, &SpotPicker::spotSizeChanged));
+}
+SpotPicker::~SpotPicker()
+{
+}
```
Lawrence37: This empty destructor is not needed.

rtgui/guiutils.h
Lawrence37: `selecterSetup()` can be `const`

Lawrence37 (approved):

PR #6806, opened 2023-07-30
Enable black level and abstract profile preview in file browser

### LINKED ISSUE #6628: Raw Black Points and Abstract Profile are ignored for the File Browser thumbnails
For the thumbnails in the File Browser, RawTherapee 5.9 ignores:
- The Raw Black Points
- The Abstract Profile

Steps to reproduce the problems:
1. Open a RAW file in RawTherapee 5.9.
2. Open the top panel.
3. Change substantially one of the Raw Black Points in the Raw Tab and observe the lack of effect on the image thumbnail in the top panel.
4. Reset the Raw Black Points.
5. In the Color Tab activate a linear profile (Linear g=1) and again observe the lack of effect on the image thumbnail in the top panel.

Tested on two Windows 10 computers.

Lawrence37:
- Use the black level adjustments in the thumbnail. Changing the black levels will initiate a full re-processing of the thumbnail because the cached base image already has the black level applied. The alternative is to cache the thumbnail even earlier in the pipeline so the black level adjustment can be applied without the full re-processing. This would have required changing the cache behavior. The possible side-effects are not worth the (rarely realized) performance advantage in my opinion.
- Add abstract profile to the thumbnail. This was probably forgotten because RawTherapee has three processing pipelines.
- Update the thumbnail when the white level is changed. The white level is factored into creating the thumbnail, but the thumbnail would not update until another adjustment was made.

Closes #6628.


Beep6581:
@Lawrence37 do you think you could resolve the merge conflicts? Else I can do it in a few days.

Lawrence37:
Yes, I'll resolve the conflicts when I get the chance.

Lawrence37:
@Beep6581 Conflicts are resolved.

Beep6581:
Great, +1 for merge.

PR #6809, opened 2023-08-01
Fix external editor on macOS

Lawrence37:
Hopefully this finally fixes the external editor on macOS by avoiding the use of `Gio::AppInfo`.


Benitoite:
👍🏽 Working for me on intel and arm64.

Lawrence37:
There is another bug related to the external editor that I spent some time trying to find a fix for and hoped to include in this pull request. I will merge this now so macOS users will have a usable external editor feature for the beta version.

PR #6814, opened 2023-08-06
External editor button: synchronize in multiple editors layout

Lawrence37:
This makes sure that after updating the external editors in preferences, the external editor button updates to reflect the new settings even when using one of the multiple editors layouts (Multiple Editor Tabs or Multiple Editor Tabs In Own Window).


Lawrence37:
I'll merge this if there are no objections.

PR #6817, opened 2023-08-07
Use `_WIN32` instead of `WIN32` to detect Windows

ss23:
`WIN32` is not defined when building a 64-bit executable on Windows with Clang. `_WIN32` is the more appropriate option here.

http://web.archive.org/web/20191012035921/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system contains a handy table/matrix of the options and why this is best.

### DIFF
--- a/rtengine/dcraw.c
+++ b/rtengine/dcraw.c
@@ -51,7 +51,7 @@
 #ifdef __CYGWIN__
 #include <io.h>
 #endif
-#ifdef WIN32
+#ifdef _WIN32
 #include <sys/utime.h>
 #include <winsock2.h>
 #pragma comment(lib, "ws2_32.lib")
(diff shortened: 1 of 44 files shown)


Lawrence37:
`WIN32` is used in many other parts of the code besides `rtengine/mytime.h`. For example:
https://github.com/ss23/RawTherapee/blob/54fc69afe34989b4ec4ebae26b305e1142adadbb/rtengine/dynamicprofile.cc#L224

ss23:
Right you are, @Lawrence37. I have gone through and updated all the cases where I believe it makes sense to do so now.

Lawrence37:
I'm curious which ones don't make sense to update. Grep shows 141 instances of `WIN32`. About 10 of those are not relevant, so that leaves roughly 130. So far you've changed 96 of them.

ss23:
Sorry @Lawrence37, I missed a bunch of them still. I did a grep for `ifdef` and `defined(` but not `ifndef`. I've now searched for just `[^_]WIN32`, so hopefully got all of them this time... The main one I was thinking of as not relevant is the check in `cJSON.h` and the cmake related files.

It's times like these I'm glad that other people can review my changes and I don't have direct commit access :)

Lawrence37:
Ok, that makes sense.

Lawrence37 (approved):
Looks good to me

PR #6818, opened 2023-08-08
Replace Inpaint-opposed by Blend - hlrecovery in all pp3 profiles and default

Desmis:
To ensure a faster refresh, I replace "Inpaint opposed" by "Blend" in "Highlight reconstruction".

Of course this is accompanied by a less good treatment of highlights


Hombre57:
Thanks @Desmis . As discussed, I'll try to come up with a solution on the WB event side so we don't have to apply this PR. Please hold it for 3-4 days.

Hombre57:
Now that the problem is fixed by PR #6822, we can ignore this PR and delete this branch.

PR #6822, opened 2023-08-14
Better use of cache with inpaint opposed highight reconstruction

### LINKED ISSUE #6815: Inpaint opposed highlight reconstruction cache strategy
If inpaint opposed highlight reconstruction is activated, the editor becomes much slower when applying other adjustments because the image is reprocessed completely. The full reprocessing may have been added to ensure the white balance is correctly taken into account for the highlight reconstruction. There should be a more efficient way to do this.

Lawrence37:
Closes #6815.


Hombre57 (approved):
:+1:  to merge

PR #6830, opened 2023-08-28
"Real" Hi-DPI support for RawTherapee GUI - Part 1

Pandagrapher:
This PR brings "real" Hi-DPI support for RawTherapee GUI. Except for image related custom widgets (ex: thumbnail, image area, ...), all the RT GUI has been migrated to handle Hi-DPI.

Example 1:
- Before (RT 5.9):
image
- After (this PR):
image

Example 2:
- Before (RT 5.9):
image
- After (this PR):
image

This PR has been tested with the following environnements:
- Windows 11, 2K screen
- MacOS Ventura, MacBook Retina screen
- Debian Testing (VM), MacBook Retina screen

In a second PR, I will work on the image related custom widgets.

Detailed list of changes:
- Icons are now managed as Gtk Icon Theme. Surfaces are directly generated from .svg files and takes into account the DPI of the screen to avoid "blur" effect. To avoid reaching GDI limit on Windows, surfaces are buffered and reused (I hope it is handled differently in Gtk4...). Large parts of GUI code have been updated.
- Custom widget `on_draw` functions have been updated to handle Hi-DPI during drawing.
- Bug fix with Histogram panel widget: When loading an image, raw histogram data were incorectly not set (but provided)
- Bug fix with Histogram panel widget: Fixes incorrect curve scaling in raw histogram mode (was incorrectly scaled using luma and chroma data)
- Bug fix loading icon cache data: With some versions of Glib/Glibmm library on MacOS, Glib::KeyFile "load_from_file" function does not raise a Glib::Error but another exception. This causes cached data not saved on disk (and thumbnails not displayed in explorer). Adds robustness testing if file is present first before calling the function
- Bug fix loading dynamic profiles data: Same issue with Glib::KeyFile "load_from_file" funtion
- Cleanup of RT theme management
- Simplification of font size management as Pango/Gtk directly manage pt/px conversion based on screen DPI
- Bug fix in Preference panel: If panel is closed with "cancel" button, font size was not reinitialized to previously saved one
- Bug fix in `RawTherapee.css` theme: Font size was sprecified in css file and so user choice could be ignored


Benitoite:
I noticed in step https://github.com/Beep6581/RawTherapee/actions/runs/6001775737/job/16276688899#step:5:82
the requested version of RSVG-2 was lower than provided by the runner, which is Ubuntu 20.04.  Ubuntu 22.04 https://packages.ubuntu.com/search?suite=jammy&searchon=names&keywords=librsvg2-dev seems to provide that minimum version, with 22.10 and 23.04 providing even more recent RSVG-2.

Pandagrapher:
> I noticed in step https://github.com/Beep6581/RawTherapee/actions/runs/6001775737/job/16276688899#step:5:82 the requested version of RSVG-2 was lower than provided by the runner, which is Ubuntu 20.04. Ubuntu 22.04 https://packages.ubuntu.com/search?suite=jammy&searchon=names&keywords=librsvg2-dev seems to provide that minimum version, with 22.10 and 23.04 providing even more recent RSVG-2.

Hi @Benitoite 
Yes I noticed that. I don't know why we don't use Ubuntu 22.04 LTS for the AppImage build (CodeQL build uses this version and correct librsvg library version is available). Using Ubuntu 22.04 LTS will also allow using a recent exiv2 package (instead of building it in the AppImage build). If it's ok I can update the AppImage workflow in a commit

BTW. I don't think this PR impacts your script for the MacOS build. Do not hesitate to check :)

Pierre

Lawrence37:
The reason to use Ubuntu 20.04 is to maximize compatibility. Essentially, new OSs have backwards compatibility with old AppImage components, but old OSs don't necessarily have forwards compatibility with new AppImage components. See this comment for more information:
https://github.com/Beep6581/RawTherapee/issues/6567#issuecomment-1229066766

Pandagrapher:
@Lawrence37 Thanks for the details
So, if I have understood well, if we want to keep Ubuntu 20.04 and solve the build issue, we have to build from source a most recent version of librsvg2 and install it (as it is done for exiv2)?

Lawrence37:
We just need to install a version of the library which works with this branch. It doesn't have to be the latest version and we can use pre-built binaries from a trustworthy source.

Benitoite:
Screenshot 2023-08-30 015702
Running the generated Windows artifact, 4k/UHD screenshot, Full display resolution w/ 200% UI scale

Lawrence37:
Just wondering how much the icons and text are scaled. Since GTK is handling it, am I correct to think the icons are scaled using the GDK scale (corresponding to the GDK_SCALE environment variable) and text is scaled according to the DPI (adjustable with the GDK_DPI_SCALE environment variable)?

Trying it briefly on Linux, hi-dpi seems to be working well. I noticed just one problem with the Favorites tab settings in preferences. The widgets are very slim. When I designed them, I used the DPI to set their widths. I hindsight, it would have been better to use one of those two panel resizable widgets.

Pandagrapher:
@Lawrence37 
How it works:
Gtk icon sizes correponds to a pixel size given for a 96 DPI reference (i.e. default Cairo DPI). Text sizes in pt corresponds to a pixel size given for a 72 DPI reference (i.e. default Pango DPI). GDK retrieves DPI data based on OS parameters and so scales up/down the pixel size of GUI elements based on their corresponding DPI reference (refer to notes in `rtscalable.h`).

Thanks for the testing :). I have fixed the Favorite tab settings issue: tab widgets are now dynamically sized based on Preferences window one.

Pierre

Lawrence37:
The AppImage build failed because it couldn't find the application icon:
`ERROR: Could not find icon executable for Icon entry: rawtherapee`
The icon file is picked up from the .desktop file:
`Icon=rawtherapee`

A `dev` build installs these files with the word "rawtherapee" in it (`find . -name 'rawtherapee*'`):
```
./rawtherapee
./rawtherapee-cli
./share/man/man1/rawtherapee.1
./share/icons/hicolor/256x256/apps/rawtherapee.png
./share/icons/hicolor/64x64/apps/rawtherapee.png
./share/icons/hicolor/48x48/apps/rawtherapee.png
./share/icons/hicolor/24x24/apps/rawtherapee.png
./share/icons/hicolor/128x128/apps/rawtherapee.png
./share/icons/hicolor/16x16/apps/rawtherapee.png
./share/icons/hicolor/32x32/apps/rawtherapee.png
./share/icons/hicolor/scalable/apps/rawtherapee.svg
./share/applications/rawtherapee.desktop
./images/rawtherapee-logo-48.png
./images/rawtherapee-logo-256.png
./images/rawtherapee.ico
./images/rawtherapee-logo-128.png
./images/rawtherapee-logo-64.png
./images/rawtherapee-logo-32.png
./images/rawtherapee-logo-16.png
./images/rawtherapee-logo-24.png
```

This branch installs the following:
```
./rawtherapee
./icons/rawtherapee
./icons/rawtherapee/256x256/app_icons/rawtherapee.png
./icons/rawtherapee/64x64/app_icons/rawtherapee.png
./icons/rawtherapee/48x48/app_icons/rawtherapee.png
./icons/rawtherapee/24x24/app_icons/rawtherapee.png
./icons/rawtherapee/128x128/app_icons/rawtherapee.png
./icons/rawtherapee/16x16/app_icons/rawtherapee.png
./icons/rawtherapee/32x32/app_icons/rawtherapee.png
./icons/rawtherapee/scalable/app_icons/rawtherapee.svg
./rawtherapee-cli
./share/man/man1/rawtherapee.1
./share/applications/rawtherapee.desktop
./images/rawtherapee.ico
```

Looks like it recognizes icons in `./share/icons/hicolor/*/apps/` but not `./icons/rawtherapee/*/app_icons/`. [This page](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html) recommends having at least a 48x48 hicolor PNG icon:
> So, you're an application author, and want to install application icons so that they work in the KDE and Gnome menus. Minimally you should install a 48x48 icon in the hicolor theme. This means installing a PNG file in $prefix/share/icons/hicolor/48x48/apps.

Pandagrapher:
@Lawrence37 
Thanks for your investigations. AppImage build is now fixed.

I have performed some tests: AppImage may crash at startup with Ubuntu because this distro does not provide the `image-missing` icon with the `adwaita-icon-theme` package (installed by default). We need to install the `gnome-icon-theme` one to obtain the missing icon. This behavior does not happen with other distro as Debian (here `Adwaita` icon theme already has the icon).

IMHO we should package the `Adwaita` and `hicolor` icon theme in the AppImage to be robust to any distro/desktop. What do you think about it?

Lawrence37:
Do you think all the icons are necessary? Checking on my computer, Adwaita is 21 MB and hicolor is 14 MB. They will be smaller in the AppImage because of compression, but they will still increase the package size by a significant percentage.

Pandagrapher:
> Do you think all the icons are necessary? Checking on my computer, Adwaita is 21 MB and hicolor is 14 MB. They will be smaller in the AppImage because of compression, but they will still increase the package size by a significant percentage.

It can be complicated to know what icons Gtk is internally using (ex: for its widgets). On a clean Ubuntu 20.04 LTS install (such as Github host), Adwaita folder is only ~4MB and hicolor one is ~3MB. It seems to be acceptable for an AppImage package.

Lawrence37:
Ok, that's a more reasonable size. If it's necessary to include the icon themes in the AppImage, then we have no choice but to do so. I do wonder why the AppImage packaging tool doesn't automatically do it though. For GTK support, it uses [this plugin](https://github.com/linuxdeploy/linuxdeploy-plugin-gtk) which doesn't include any icons. What's the exact error that happens in Ubuntu? I looked up image-missing icon errors on the internet and all the results are problems with the OS itself.

Pandagrapher:
@Lawrence37 
Here the error message I get when running the AppImage on Ubuntu 20.04 LTS without `gnome-icon-theme` package installed:

> Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): L'icône "image-missing" n'est pas présente dans le thème rawtherapee (gtk-icon-theme-error-quark, 0)
Bail out! Gtk:ERROR:../../../../gtk/gtkiconhelper.c:494:ensure_surface_for_gicon: assertion failed (error == NULL): L'icône "image-missing" n'est pas présente dans le thème rawtherapee rawtherapee (gtk-icon-theme-error-quark, 0)

This is normal that the `image-missing` icon is not present in the `rawtherapee` theme because Gtk is expecting it to be in the `Adwaita` icon theme (`rawtherapee` icon theme is configured to use the `Adwaita` one if an icon is not present).

When searching for this error on the Internet, there is two kinds of solution:
- For a package: add dependency for the `gnome-icon-theme` package.
- For an AppImage (ex: Inkscape), packaging the `Adwaita` icon theme in the AppImage. But I agree with you, I am surprised the AppImage Gtk plugin does not manage it directly.

Edit: I pushed a commit: Only a part of the Adwaita full icon theme is now bundled in the AppImage (about ~4MB).

rtengine/dynamicprofile.cc
```diff
@@ -174,9 +175,12 @@ bool DynamicProfileRules::loadRules()
...
     Glib::KeyFile kf;
+    const Glib::ustring fileName = Glib::build_filename (Options::rtdir, "dynamicprofile.cfg");
 
     try {
-        if (!kf.load_from_file (Glib::build_filename (Options::rtdir, "dynamicprofile.cfg"))) {
+        if (Glib::file_test(fileName, Glib::FILE_TEST_EXISTS)) {
+            kf.load_from_file (fileName);
+        } else {
             return false;
         }
```
Lawrence37: This should return false if the load fails. Something like
```c++
if (!(Glib::file_test(fileName, Glib::FILE_TEST_EXISTS) && kf.load_from_file(fileName))) {
    return false;
}
```

rtgui/batchqueuebuttonset.cc
```diff
@@ -25,9 +25,9 @@
 
 bool BatchQueueButtonSet::iconsLoaded = false;
 
-Cairo::RefPtr<RTSurface> BatchQueueButtonSet::cancelIcon;
-Cairo::RefPtr<RTSurface> BatchQueueButtonSet::headIcon;
-Cairo::RefPtr<RTSurface> BatchQueueButtonSet::tailIcon;
+std::shared_ptr<RTSurface> BatchQueueButtonSet::cancelIcon = std::shared_ptr<RTSurface>(nullptr);
+std::shared_ptr<RTSurface> BatchQueueButtonSet::headIcon = std::shared_ptr<RTSurface>(nullptr);
+std::shared_ptr<RTSurface> BatchQueueButtonSet::tailIcon = std::shared_ptr<RTSurface>(nullptr);
```
Lawrence37: No need to change anything, but just an FYI, the assignment is not necessary to create an empty shared pointer.

rtgui/batchqueuepanel.cc
```diff
@@ -228,13 +228,13 @@ void BatchQueuePanel::updateTab (int qsize, int forceOrientation)
...
-            grid->attach_next_to(*Gtk::manage (new RTImage ("gears.png")), Gtk::POS_RIGHT, 1, 1);
+            grid->attach_next_to(*Gtk::manage (new RTImage ("gears")), Gtk::POS_RIGHT, 1, 1);
             grid->attach_next_to(*Gtk::manage (new Gtk::Label (M("MAIN_FRAME_QUEUE") )), Gtk::POS_RIGHT, 1, 1);
         } else if (qStartStop->get_active()) {
-            grid->attach_next_to(*Gtk::manage (new RTImage ("gears-play.png")), Gtk::POS_RIGHT, 1, 1);
+            grid->attach_next_to(*Gtk::manage (new RTImage ("gears-play")), Gtk::POS_RIGHT, 1, 1);
             grid->attach_next_to(*Gtk::manage (new Gtk::Label (M("MAIN_FRAME_QUEUE") + " [" + Glib::ustring::format( qsize ) + "]" )), Gtk::POS_RIGHT, 1, 1);
         } else {
-            grid->attach_next_to(*Gtk::manage (new RTImage ("gears-pause.png")), Gtk::POS_RIGHT, 1, 1);
+            grid->attach_next_to(*Gtk::manage (new RTImage ("gears-pause")), Gtk::POS_RIGHT, 1, 1);
```
Lawrence37: Aren't these icons supposed to be sized with `Gtk::ICON_SIZE_LARGE_TOOLBAR`? They look a little small when using horizontal tab layouts.

rtgui/controlspotpanel.cc
```diff
@@ -74,14 +74,14 @@ ControlSpotPanel::ControlSpotPanel():
     struc_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_THRES"), 1.0, 12.0, 0.1, 4.0))),
     thresh_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_THRESDELTAE"), 0.0, 10.0, 0.1, 2.0))),
     iter_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_PROXI"), 0.2, 10.0, 0.1, 2.0))),
-    balan_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_BALAN"), 0.2, 2.5, 0.1, 1.0, Gtk::manage(new RTImage("rawtherapee-logo-16.png")),  Gtk::manage(new RTImage("circle-white-small.png"))))),
-    balanh_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_BALANH"), 0.2, 2.5, 0.1, 1.0, Gtk::manage(new RTImage("rawtherapee-logo-16.png")), Gtk::manage(new RTImage("circle-red-green-small.png"))))),
-    colorde_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_COLORDE"), -15, 15, 2, 5, Gtk::manage(new RTImage("circle-blue-yellow-small.png")), Gtk::manage(new RTImage("circle-gray-green-small.png"))))),
+    balan_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_BALAN"), 0.2, 2.5, 0.1, 1.0, Gtk::manage(new RTImage("rawtherapee")),  Gtk::manage(new RTImage("circle-white-small"))))),
+    balanh_(Gtk::manage(new Adjuster(M("TP_LOCALLAB_BALANH"), 0.2, 2.5, 0.1, 1.0, Gtk::manage(new RTImage("rawtherapee")), Gtk::manage(new RTImage("circle-red-green-small"))))),
```
Lawrence37: The left icons don't show up (Local Adjustments > Settings > (make sure "Show additional settings" is checked) Shape detection > ab/L balance *and* C-H balance).

rtgui/cursormanager.cc
```diff
@@ -35,74 +34,52 @@ void CursorManager::init (Glib::RefPtr<Gdk::Window> mainWindow)
...
+        // Cursor is retrieved from theme thanks to an RTSurface
+        // - By default, cursor hotspot is located at middle of surface.
+        // Use (offX, offY) between -1 and 0.99 to move cursor hotspot
+        auto cursor_surf = RTSurface(name, Gtk::ICON_SIZE_MENU);
+        double offXb = std::min(std::max(-1., offX), 0.99); // offX should belong to (-1; 0.99)
+        double offYb = std::min(std::max(-1., offY), 0.99); // offY should belong to (-1; 0.99)
+        auto cursor = Gdk::Cursor::create(this->display,
+            cursor_surf.get(),
+            cursor_surf.getWidth() / 2 * (1. + offXb),
+            cursor_surf.getHeight() / 2 * (1. + offYb));
```
Lawrence37: `2.`, or is it supposed to be integer division?

rtgui/cursormanager.cc
```diff
@@ -35,74 +34,52 @@ void CursorManager::init (Glib::RefPtr<Gdk::Window> mainWindow)
...
+        const double offX = 0., const double offY = 0.) -> Glib::RefPtr<Gdk::Cursor>
+    {
+        // Notes:
+        // - Gdk Cursor Theme is not supported on some OS (ex : MacOS).
+        // Cursor is retrieved from theme thanks to an RTSurface
+        // - By default, cursor hotspot is located at middle of surface.
+        // Use (offX, offY) between -1 and 0.99 to move cursor hotspot
+        auto cursor_surf = RTSurface(name, Gtk::ICON_SIZE_MENU);
+        double offXb = std::min(std::max(-1., offX), 0.99); // offX should belong to (-1; 0.99)
+        double offYb = std::min(std::max(-1., offY), 0.99); // offY should belong to (-1; 0.99)
```
Lawrence37: It might be better to clamp the pixel coordinates instead.

rtgui/cursormanager.cc
```diff
@@ -35,74 +34,52 @@ void CursorManager::init (Glib::RefPtr<Gdk::Window> mainWindow)
...
+        if (!cursor) {
+            cursor = Gdk::Cursor::create(this->display, fb_cursor);
+        }
+
+        return cursor;
+    };
+
+    cAdd        = createCursor("crosshair-hicontrast", Gdk::PLUS);
+    cAddPicker  = createCursor("color-picker-add-hicontrast", Gdk::PLUS, -0.333, 0.75);
+    cCropDraw   = createCursor("crop-point-hicontrast", Gdk::DIAMOND_CROSS);
```
Lawrence37: Offset at -0.75, -0.75?

rtgui/filterpanel.h
```diff
@@ -61,6 +63,8 @@ class FilterPanel : public Gtk::Box
     ExifFilterSettings curefs;
     FilterPanelListener* listener;
 
+    std::shared_ptr<RTSurface> ornamentSurface;
```
Lawrence37: The surface is not shared anywhere, so it can be a `std::unique_ptr`.

rtgui/histogrampanel.cc
```diff
@@ -1522,23 +1471,21 @@ void HistogramArea::drawCurve(Cairo::RefPtr<Cairo::Context> &cr,
     cr->line_to (hsize - padding, vsize - 1);
 }
 
-void HistogramArea::drawMarks(Cairo::RefPtr<Cairo::Context> &cr,
-                              const LUTu & data, double scale, int hsize, int & ui, int & oi)
+void HistogramArea::drawMarks(const Cairo::RefPtr<Cairo::Context> &cr,
+                              const LUTu & data, const double scale, const int wsize, int & ui, int & oi)
 {
-    int s = 8 * RTScalable::getScale();
```
Lawrence37, line 1528: Keep this variable but with a better name to avoid magic numbers.

rtgui/lensprofile.cc
```diff
@@ -64,7 +64,7 @@ LensProfilePanel::LensProfilePanel() :
     lensfunCameras(Gtk::manage((new MyComboBox()))),
     lensfunLensesLbl(Gtk::manage((new Gtk::Label(M("EXIFFILTER_LENS"))))),
     lensfunLenses(Gtk::manage((new MyComboBox()))),
-    warning(Gtk::manage(new RTImage("warning.png"))),
+    warning(Gtk::manage(new RTImage("warning"))),
```
Lawrence37: This image looks a bit small to me.

rtgui/rtimage.cc
```diff
@@ -20,214 +21,139 @@
...
-    if (surf) {
-        set(surf);
-        surface = surf;
-    }
+    // Configure RTImage based on g_icon
+    set(this->g_icon, this->size);
 }
 
-RTImage::RTImage (Glib::RefPtr<RTImage> &other)
+void RTImage::set_from_icon_name(const Glib::ustring& iconName)
```
Lawrence37, line 83: This looks like it can be implemented with `set_from_icon_name(iconName, this->size)`.

rtgui/rtimage.cc
```diff
@@ -20,214 +21,139 @@
...
-        scaleBack = getScale();
-        updateImages();
+    // Unset Gio::Icon if previously chosen
+    if (this->g_icon) {
+        g_icon = Glib::RefPtr<const Gio::Icon>();
     }
 }
 
-void RTImage::changeImage(const Glib::RefPtr<const Gio::Icon> &gIcon, int size)
+void RTImage::set_from_gicon(const Glib::RefPtr<const Gio::Icon>& gIcon)
```
Lawrence37, line 107: Similarly, this looks like `set_from_gicon(gIcon, this->size)`.

rtgui/rtscalable.cc
```diff
@@ -18,236 +19,258 @@
...
-        if (scale == 1) {
-            if (dpi >= baseHiDPI) {
-                scale = 2;
-            }
-        }
-        else if (scale == 2) {
-            if (dpi < baseHiDPI) {
-                dpi *= 2.;
-            }
+void RTScalable::updateDPInScale(const Gtk::Window* window, double &newDPI, int &newScale)
```
Lawrence37: `getDPInScale` seems like a more appropriate name.

rtgui/rtscalable.cc
```diff
@@ -18,236 +19,258 @@
...
-    if (!Glib::file_test(imagesCacheFolderDPI.c_str(), Glib::FILE_TEST_EXISTS)) {
-        auto error = g_mkdir_with_parents (imagesCacheFolderDPI.c_str(), 0777);
-        if (error != 0) {
-            std::cerr << "ERROR: Can't create \"" << imagesCacheFolderDPI << "\" cache folder: " << g_strerror(error)  << std::endl;
-            Cairo::RefPtr<Cairo::ImageSurface> surf = Cairo::ImageSurface::create(Cairo::FORMAT_RGB24, 10, 10);
+        if (error) {
+            std::cerr << "Failed to load SVG file \"" << fname << "\": " << std::endl
+                      << Glib::ustring(error->message) << std::endl;
             return surf;
         }
```
Lawrence37: Free the `GError`.
From https://gnome.pages.gitlab.gnome.org/librsvg/Rsvg-2.0/ctor.Handle.new_from_data.html:

> In case of error, the argument will be set to a newly allocated GError; the caller will take ownership of the data, and be responsible for freeing it.

rtgui/rtscalable.cc
```diff
@@ -18,236 +19,258 @@
...
+            .height = static_cast<double>(h * RTScalable::getScale())
+        };
+        c->set_operator (Cairo::OPERATOR_OVER);
+        const bool success = rsvg_handle_render_document(handle, c->cobj(), &rect, &error);
+
+        if (!success && error) {
+            std::cerr << "Failed to load SVG file \"" << fname << "\": " << std::endl
+                      << Glib::ustring(error->message) << std::endl;
+            return surf;
+        }
```
Lawrence37: Free the `GError`.

rtgui/rtsurface.h
```diff
@@ -18,37 +19,38 @@
...
-class RTSurface :
-    public RTScalable
+class RTSurface final : public RTScalable
 {
 public:
-    RTSurface();
-    explicit RTSurface(const Glib::ustring& fileName, const Glib::ustring& rtlFileName = {});
-
-    void setImage(const Glib::ustring& fileName, const Glib::ustring& rtlFileName = {});
+    enum RTSurfaceType {
```
Lawrence37: Make this an enum class.

rtgui/shcselector.cc
```diff
@@ -198,42 +181,47 @@ void SHCSelector::updateBackBuffer()
...
+        const double fontScale = RTScalable::getDPI() / RTScalable::pangoDPI; // Refer to notes in rtscalable.h
+#else
+        // On MacOS, font is already scaled by the System library
+        // Refer to https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/gdk/quartz/gdkscreen-quartz.c
+        const double fontScale = 1.;
+#endif
+        const double absoluteFontSize = static_cast<double>(fontSize) * fontScale; // px
+        // Absolute size is defined in "Pango units" and shall be multiplied by
+        // Pango::SCALE from "px":
+        fontd.set_absolute_size (absoluteFontSize * static_cast<double>(Pango::SCALE));
```
Lawrence37: The text is too large and too low especially with a large DPI and 1x scale.
image

Lawrence37 (changes requested):
Finally managed to review all the code. I can see you put a lot of work into this. Thanks!

rtgui/cursormanager.cc
Pandagrapher: Yes, it is supposed to be integer division

rtgui/cursormanager.cc
Pandagrapher: Original code:
> cAddPicker = colPickAdd     ? Gdk::Cursor::create(display, colPickAdd, (int)(4.*s), (int)(21.*s))   : Gdk::Cursor::create(display, Gdk::PLUS);

So, from a 24x24 cursor size, it needs a (-0.666, 0.75) offset to get (4, 21) coordonates (fixed it from -0.333 to -0.666)

rtgui/histogrampanel.cc
Pandagrapher: This variable is not needed anymore and has been removed

Pandagrapher:
@Lawrence37 Thanks for the code review :). Took them into account

rtgui/cursormanager.cc
Lawrence37: I hadn't noticed that! However, I was referring to the crop cursor which was:
`    cCropDraw = cropDraw        ? Gdk::Cursor::create(display, cropDraw, (int)(3.*s), (int)(3.*s))      : Gdk::Cursor::create(display, Gdk::DIAMOND_CROSS);`

Pandagrapher: Fixed with commit `e3062e2`

Lawrence37 (approved):

Lawrence37:
Looks good!

Lawrence37:
@Pandagrapher Will this have any negative effects if part 2 is not completed before the next release? If not, then now is a good time to merge.

Pandagrapher:
> @Pandagrapher Will this have any negative effects if part 2 is not completed before the next release? If not, then now is a good time to merge.

No, there is no side effect on merging part 1 without part 2. The GUI widgets that have not been updated to handle Hi-DPI will remain "blur" (as they currently are). I gonna merge with dev during the week.

Pandagrapher:
@Lawrence37 Merge done. Feel free to accept the Pull Request when you think it's the right moment :)

Lawrence37:
Wonderful! I see no reason to wait. 🙂

Issue #6821, opened 2023-08-13
Failure to compile with Clang under MSYS2


ss23:
**Short description**
The `dev` branch builds but fails to link when compiled with Clang under MSYS2. The error is caused by the `-latomic` linker flag:
```
cmd.exe /C "cd . && C:\msys64\clang64\bin\c++.exe -std=c++11 -march=native -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp=libomp -Werror=unknown-pragmas -O3 -DNDEBUG -march=native rtgui/CMakeFiles/rth-cli.dir/myicon.rc.obj r
tgui/CMakeFiles/rth-cli.dir/alignedmalloc.cc.obj rtgui/CMakeFiles/rth-cli.dir/editcallbacks.cc.obj rtgui/CMakeFiles/rth-cli.dir/main-cli.cc.obj rtgui/CMakeFiles/rth-cli.dir/multilangmgr.cc.obj rtgui/CMakeFiles/rth-cli.dir/options.cc.obj rtgui/CMakeFiles/rth-cli.dir/paramsedited.cc.obj rtgui/CMakeFiles/rth-cli.dir/pathutils.cc.obj rtgui/CMakeFiles/rth-cli.dir/th
readutils.cc.obj -o rtgui\rawtherapee-cli.exe -Wl,--out-implib,rtgui\librawtherapee-cli.dll.a -Wl,--major-image-version,0,--minor-image-version,0 -LC:/msys64/home/ss23/RawTherapee/rtgui/. rtengine/librtengine.a  -lcairomm-1.0  -lsigc-2.0  -lcairo  -lexpat  -lwinmm  -latomic  -lfftw3f  C:/msys64/clang64/lib/libfftw3f_omp.dll.a  -lgiomm-2.4  -lgio-2.0  -lglibmm-2
[...]

PR #6831, opened 2023-08-28
detects libatomic flags in a cmake module

Benitoite:
Detect libatomic flags by cmake module rather than providing for their exclusions, see #6821


CMakeLists.txt
```diff
@@ -565,6 +565,12 @@ if(WITH_MYFILE_MMAP)
     add_definitions(-DMYFILE_MMAP)
 endif()
 
+# Atomic is required in some builds: #6821
+find_package(ATOMIC)
+if(ATOMIC_FOUND)
+	target_link_libraries(executable ${ATOMIC_LIBRARIES})
```
Floessie, line 571: @Benitoite `executable` wasn't meant literally. It should rather be `rawtherapee` and another line with `rawtherapee-cli` or completely different by adding `${ATOMIC_LIBRARIES}` to `EXTRA_LIB_RTGUI` and the corresponding variable for RT.

I'm surprised this works as is...

kmilos:
FYI, see also how CMake itself does it for more ideas (although the principle is pretty much the same): https://gitlab.kitware.com/cmake/cmake/-/blob/4cd207b6f3b2b262cf4cc8eb46a2320906542d1e/Source/Checks/cm_cxx_features.cmake#L100-105

CMakeLists.txt
Benitoite: I don't have a system requiring Atomic in a gnu build, so I couldn't trigger any error on that line... I think the targets we need are `rth` and `rth-cli` as set in `rtgui/CMakeLists.txt`

https://github.com/Beep6581/RawTherapee/blob/dev/rtgui/CMakeLists.txt#L330C1-L330C1

Benitoite: #6832 has the properly named targets.

Issue #6826, opened 2023-08-19
DNG files from a X-T3 file have strong purple cast


Alex-Vasile:
**Short description**
Opening a X-T3 raw file that was converted to DNG will result in an extreme purple color cast.

**Steps to reproduce**
1. Convert RAF to DNG
2. Open DNG in RawTherapee

**Expected behavior**
Looks identical to the .RAF version.

**Additional information**
- Using RawTherapee 5.9 (arm native version)
- Using Adobe DNG Converter 14.3.0.1072
- m1 mpb on macOS 13.4.1
- Other raw editors read the file fine

Other useful information:
- Reproducible with https://raw.pixls.us/getfile.php/2783/nice/Fujifilm%20-%20X-T3%20-%2014bit%20compressed%20(3:2).RAF
- Screenshot of opening the DNG in RawTherapee:
![Purple](https://github.com/Beep6581/RawTherapee/assets/48962821/7b4e8ee1-ce63-49bb-99c3-6fd555d0210d)

Lawrence37:
The dng is a different resolution which is causing RawTherapee to incorrectly interpret the color filter array pattern. A quick hack is to set the raw crop in the `camconst.json` like so:
```json
"raw_crop" : [
   { "frame" : [6384, 4182], "crop": [0, 5, 6252, 4176] }, // RAF
   { "frame" : [6240, 4160], "crop": [3, 2, 6237, 4158] } // DNG from ADC
]
```
A proper solution is to fix the CFA interpretation.

hedwig2013:
For SONY ARW it is the same IF taken in RAW with lossless compression (S/M/L) only. Images taken with uncompressed RAW works fine without any issues. My camera model is ILCE-7M4. Like @Alex-Vasile I convert ARW into DNG with Adobe DNG Converter too.

Darktable can show and edit without issues.
Digikam can show and edit without issues.
Other Viewers like IrfanView, XNViewMP and others can show and edit without issues.

Is there a workaround without programming knowledge? I am a photographer.

Adobe DNG Converter: 15.5.0.1595
OS: Windows 10 22H2
RawTherapee: 5.9

Something more I observed. The resolutions are different:
SONY ARW uncompressed 7028 x 4688 -> DNG 7040 x 4688
SONY ARW lossless compressed (S) 3504 x 2336 -> DNG 3514 x 2344

kmilos:
[...]

PR #6833, opened 2023-08-30
Segfault fixed when opening Fuji cropped raw. Fixes #6312

### LINKED ISSUE #6312: Segfault when opening cropped Fujifilm X-Pro3 raw
Reported here with a file to test: https://discuss.pixls.us/t/crash-when-opening-fuji-crop-raf-files/26055
Seems very similar to https://github.com/Beep6581/RawTherapee/issues/6279 (open issue) and https://github.com/Beep6581/RawTherapee/issues/6255 (closed issue).

I'm on holidays the coming week so cannot investigate immediately.

LoKolbasz:
Fixes #6312 where opening a Fuji cropped raw RAF file would result in a Segmentation Fault. Now RT should read the metadata of cropped Fuji files and debayer the image accordingly. For non cropped files the behavior remains unchanged. I only tested it with the image referenced in the issue, and a few other (not cropped) [files](https://drive.google.com/file/d/13IbwuOPCrGMKV-H3BXqBVnGTPVrMRjGL/view?usp=sharing) I've found online.

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -5775,6 +5775,8 @@ nf: order = 0x4949;
       cam_mul[0] = get2() / 256.0;
     if (tag == 0x1018 || tag == 0x20400100)
       cam_mul[2] = get2() / 256.0;
+    if (tag == 0x104D)
+        read_crop.crop_mode = (CropMode)get2();
     if (tag == 0x2011 && len == 2) {
 get2_256:
       order = 0x4d4d;
@@ -7697,7 +7699,6 @@ void CLASS parse_fuji (int offset)
   }
   height <<= fuji_layout;
   width  >>= fuji_layout;
-  read_crop.complete = read_crop_c == 4;                                // RT
 }
 
 int CLASS parse_jpeg (int offset)
@@ -10153,12 +10154,12 @@ canon_a5:
         width = raw_width = 6016;
         height = raw_height = 4014;
     } else if (!strcmp(model, "X-Pro3") || !strcmp(model, "X-T3") || !strcmp(model, "X-T30") || !strcmp(model, "X-T4") || !strcmp(model, "X100V") || !strcmp(model, "X-S10")) {
-        constexpr std::uint_fast16_t x_width = 6384, x_height = 4182;                                                               // RT
-        is_cropped = read_crop.complete && (x_width - raw_width > is_cropped_margin || x_height - raw_height > is_cropped_margin);  // RT
-        if (!is_cropped) {                                                                                                          // RT
-            width = raw_width = x_width;                                                                                            // RT
-            height = raw_height = x_height;                                                                                         // RT
-        }                                                                                                                           // RT
+        raw_width = 6384;           // RT
+        raw_height = 4182;          // RT
+        if (!read_crop.crop_mode) { // RT
+            width = 6384;           // RT
+            height = 4182;          // RT
+        }                           // RT
     } else if (!strcmp(model, "DBP for GX680")) { // Special case for #4204
         width = raw_width = 5504;
         height = raw_height = 3856;
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -7658,6 +7658,15 @@ void CLASS parse_fuji (int offset)
     if (tag == 0x100) {
       raw_height = get2();
       raw_width  = get2();
+    // RawImageCroppedSize 0x111 = 273 	(including borders)                RT
+    } else if (tag == 0x111) {                                          // RT
+      read_crop.height = get2();                                        // RT
+      read_crop.width = get2();                                         // RT
+      // RawImageTopLeft 0x110 = 272 (top margin first, then left margin)  RT
+    } else if (tag == 0x110){                                           // RT
+      read_crop.top_margin = get2();                                    // RT
+      read_crop.left_margin = get2();                                   // RT
+    // 0x115 = 277 RawImageAspectRatio                                     RT
     } else if (tag == 0x121) {
       height = get2();
       if ((width = get2()) == 4284) width += 3;
@@ -7678,16 +7687,7 @@ void CLASS parse_fuji (int offset)
       width = tag;
       height = get4();
       order = c;
-    // RawImageCroppedSize 0x111 = 273 	(including borders)                RT
-    } else if (tag == 0x111) {                                          // RT
-      read_crop.height = get2();                                        // RT
-      read_crop.width = get2();                                         // RT
-      // RawImageTopLeft 0x110 = 272 (top margin first, then left margin)  RT
-    } else if (tag == 0x110){                                           // RT
-      read_crop.top_margin = get2();                                    // RT
-      read_crop.left_margin = get2();                                   // RT
-    }                                                                   // RT
-    // 0x115 = 277 RawImageAspectRatio                                     RT
+    }
     // 0x9650 = 38480 RawExposureBias                                      RT
 
     fseek (ifp, save+len, SEEK_SET);
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -7113,8 +7113,8 @@ void CLASS apply_tiff()
     (unsigned)tiff_ifd[i].bps < 33 && (unsigned)tiff_ifd[i].samples < 13 && // RT
 	 ns && ((ns > os && (ties = 1)) ||
 		(ns == os && shot_select == ties++))) {
-      raw_width     = read_crop.complete ? read_crop.width : tiff_ifd[i].width;
-      raw_height    = read_crop.complete ? read_crop.height : tiff_ifd[i].height;
+      raw_width     = read_crop.complete ? read_crop.width : tiff_ifd[i].width;   // RT
+      raw_height    = read_crop.complete ? read_crop.height : tiff_ifd[i].height; // RT
       tiff_bps      = tiff_ifd[i].bps;
       tiff_compress = tiff_ifd[i].comp;
       data_offset   = tiff_ifd[i].offset;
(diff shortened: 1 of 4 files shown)


Lawrence37:
I'm interested in why the crash occurs, so I checked the raw width and height and both are correct. Then I added an additional raw crop like so:
```diff
diff --git a/rtengine/camconst.json b/rtengine/camconst.json
index e71cfbb90..c7eae985e 100644
--- a/rtengine/camconst.json
+++ b/rtengine/camconst.json
@@ -1571,7 +1571,10 @@ Camera constants:
     { // Quality A, samples provided by Daniel Catalina (#5839) and pi99y (#5860)
         "make_model": [ "FUJIFILM X-T3", "FUJIFILM X-PRO3" ],
         "dcraw_matrix": [ 13426,-6334,-1177,-4244,12136,2371,-580,1303,5980 ], // DNG_v11, standard_v2 d65
-        "raw_crop": [ 0, 5, 6252, 4176],
+        "raw_crop" : [
+           { "frame" : [6384, 4182], "crop": [ 0, 5, 6252, 4176] },
+           { "frame" : [6384, 3348], "crop": [624, 0, 5004, 3348] }
+        ],
         "white": [ 16170, 16275, 16170 ] // typical safe-margins with LENR
         // negligible aperture scaling effect
     },
```

This also fixes the crash. Something in the image margins is causing the crash and may be worth a look.

This does not mean this pull request has no merit. It is useful for defining the default crop in case there is no raw crop in `camconst.json`.

LoKolbasz:
Ok, I hope it helped in finding the correct resolution at least. But I'm not sure if the corrected data will not be applied due to the (frankly a bit janky) crop detection. Also when tested with non cropped raws the margins from the metadata seemed to be off. The processed versions ended up with a magenta tint. So this isn't necessarily a reliable way to load every Fuji file.

rtengine/dcraw.cc
```diff
@@ -7113,8 +7113,8 @@ void CLASS apply_tiff()
     (unsigned)tiff_ifd[i].bps < 33 && (unsigned)tiff_ifd[i].samples < 13 && // RT
 	 ns && ((ns > os && (ties = 1)) ||
 		(ns == os && shot_select == ties++))) {
-      raw_width     = tiff_ifd[i].width;
-      raw_height    = tiff_ifd[i].height;
+      raw_width     = read_crop.complete ? read_crop.width : tiff_ifd[i].width;   // RT
+      raw_height    = read_crop.complete ? read_crop.height : tiff_ifd[i].height; // RT
```
Lawrence37: I think the raw dimensions are supposed to exclude the crop, which would make this unnecessary.

rtengine/dcraw.cc
```diff
@@ -7647,37 +7647,57 @@ void CLASS parse_fuji (int offset)
...
+      width = raw_width = read_crop.width = get2();                     // RT
+      read_crop_dimensions = true;                                      // RT
+      read_crop_c += 2;                                                 // RT
+      // RawImageTopLeft 0x110 = 272 (top margin first, then left margin)  RT
+    } else if (tag == 0x110){                                           // RT
+      read_crop.top_margin = get2();                                    // RT
+      read_crop.left_margin = get2();                                   // RT
+      read_crop_c += 2;                                                 // RT
+    }                                                                   // RT
+    // 0x115 = 277 RawImageAspectRatio                                     RT
```
Lawrence37: Coding style: These should be moved up so that the tags are in numerical order.

rtengine/dcraw.cc
```diff
@@ -7647,37 +7647,57 @@ void CLASS parse_fuji (int offset)
...
       order = 0x4949;
       while ((tag = get4()) > raw_width);
       width = tag;
       height = get4();
       order = c;
-    }
+    // RawImageCroppedSize 0x111 = 273 	(including borders)                RT
+    } else if (tag == 0x111) {                                          // RT
+      height = raw_height = read_crop.height = get2();                  // RT
+      width = raw_width = read_crop.width = get2();                     // RT
```
Lawrence37: Like the other comment, the raw dimensions probably should not be set here.

rtengine/dcraw.cc
```diff
@@ -10133,8 +10153,12 @@ void CLASS identify()
         width = raw_width = 6016;
         height = raw_height = 4014;
     } else if (!strcmp(model, "X-Pro3") || !strcmp(model, "X-T3") || !strcmp(model, "X-T30") || !strcmp(model, "X-T4") || !strcmp(model, "X100V") || !strcmp(model, "X-S10")) {
-        width = raw_width = 6384;
-        height = raw_height = 4182;
+        constexpr std::uint_fast16_t x_width = 6384, x_height = 4182;                                                               // RT
+        is_cropped = read_crop.complete && (x_width - raw_width > is_cropped_margin || x_height - raw_height > is_cropped_margin);  // RT
```
Lawrence37: Perhaps determine this from the crop mode (tag 0x104d, [reference](https://exiftool.org/TagNames/FujiFilm.html))?

rtengine/rawimage.cc
```diff
@@ -560,6 +560,13 @@ int RawImage::loadRaw(bool loadData, unsigned int imageNum, bool closeFile, Prog
...
+                is_cropped = read_crop.complete && (raw_width - read_crop.width > is_cropped_margin || raw_height - read_crop.height > is_cropped_margin);  // RT
+                if (is_cropped){                                                                                                                            // RT
+                    left_margin = read_crop.left_margin;                                                                                                    // RT
+                    top_margin = read_crop.top_margin;                                                                                                      // RT
+                    tm = 0;                                                                                                                                 // RT
+                    lm = 0;                                                                                                                                 // RT
+                }                                                                                                                                           // RT
```
Lawrence37: This looks like it overrides the values in `camconst.json`, but `camconst.json` should take precedence over values read from the metadata in dcraw. There should be a way to fix the bug without touching this file.

rtengine/dcraw.cc
```diff
@@ -7647,37 +7647,57 @@ void CLASS parse_fuji (int offset)
...
       fuji_layout = fgetc(ifp) >> 7;
       fuji_width = !(fgetc(ifp) & 8);
+      // tag 0x131 = 305 XTransLayout                                      RT
     } else if (tag == 0x131) {
       filters = 9;
       FORC(36) xtrans_abs[0][35-c] = fgetc(ifp) & 3;
     } else if (tag == 0x2ff0) {
       FORC4 cam_mul[c ^ 1] = get2();
-    } else if (tag == 0xc000 && len > 20000) {
+    } else if (tag == 0xc000 && len > 20000 && !read_crop_dimensions) { // RT
```
Lawrence37: Since you're reading the crop dimensions into its own struct, I think it would be better to to update the width and height after the loop with the crop's values if a crop mode is detected (tag 0x104d, [reference](https://exiftool.org/TagNames/FujiFilm.html)). That will also be a good place to set the `left_margin` and `right_margin`.

Lawrence37 (changes requested):
I don't feel particularly comfortable with dcraw code, but I tried to review this code to the best of my understanding. I hope my comments are helpful.

rtengine/rawimage.cc
LoKolbasz: I added that part because the incorrect data from `camconst.json` overrode the more applicable metadata. But with your correction to `camconst.json` this part might indeed be unnecessary.

rtengine/dcraw.cc
LoKolbasz: Thanks! I didn't know of this flag.

rtengine/rawimage.cc
LoKolbasz: By removing this overwrite, I'm not sure when the metadata crop would be used.

rtengine/dcraw.cc
```diff
@@ -7674,8 +7688,19 @@ void CLASS parse_fuji (int offset)
       height = get4();
       order = c;
     }
+    // 0x9650 = 38480 RawExposureBias                                      RT
+
     fseek (ifp, save+len, SEEK_SET);
   }
+
+  if (read_crop.crop_mode)                // RT
+  {                                       // RT
```
Floessie: To stay a bit more consistent, the `{` should be on the same line as the if.

rtengine/dcraw.cc
```diff
@@ -10133,8 +10158,12 @@ void CLASS identify()
         width = raw_width = 6016;
         height = raw_height = 4014;
     } else if (!strcmp(model, "X-Pro3") || !strcmp(model, "X-T3") || !strcmp(model, "X-T30") || !strcmp(model, "X-T4") || !strcmp(model, "X100V") || !strcmp(model, "X-S10")) {
-        width = raw_width = 6384;
-        height = raw_height = 4182;
+        raw_width = 6384;           // RT
+        raw_height = 4182;          // RT
+        if (!read_crop.crop_mode) { // RT
+            width = 6384;           // RT
```
Floessie: Though semantically the same, the more expressive term would be `width = raw_width;`, wouldn't it?

rtengine/dcraw.h
```diff
@@ -76,6 +76,20 @@ class DCraw
     }
 
 protected:
+    enum CropMode : std::uint_fast16_t {                                         // RT
```
Floessie: Don't import short symbols like `NA` into the surrounding namespace. Better use `enum class`.

Floessie (changes requested):

rtengine/dcraw.cc
```diff
@@ -7693,13 +7693,12 @@ void CLASS parse_fuji (int offset)
...
   }
 
-  if (read_crop.crop_mode)                // RT
-  {                                       // RT
-    height = read_crop.height;            // RT
-    width = read_crop.width;              // RT
-    top_margin = read_crop.top_margin;    // RT
-    left_margin = read_crop.left_margin;  // RT
-  }                                       // RT
+  if ((std::uint_fast16_t)read_crop.crop_mode) {  // RT
```
Floessie: Hm, shouldn't that be `if (read_crop.crop_mode != CropMode::NA) {`? No need to do a C-cast...

rtengine/dcraw.cc
```diff
@@ -10160,10 +10159,10 @@ void CLASS identify()
     } else if (!strcmp(model, "X-Pro3") || !strcmp(model, "X-T3") || !strcmp(model, "X-T30") || !strcmp(model, "X-T4") || !strcmp(model, "X100V") || !strcmp(model, "X-S10")) {
         raw_width = 6384;           // RT
         raw_height = 4182;          // RT
-        if (!read_crop.crop_mode) { // RT
-            width = 6384;           // RT
-            height = 4182;          // RT
-        }                           // RT
+        if (!(std::uint_fast16_t)read_crop.crop_mode) { // RT
```
Floessie: Same here: `if (read_crop.crop_mode == CropMode::NA) {`.

Floessie:
@LoKolbasz Nearly done. Unfortunately 1a0b7f45620350cba94e68aa56fab68fc9a5a840 contains an error. :slightly_frowning_face:

LoKolbasz:
I negated where I wasn't supposed to :rage1:

Floessie (approved):
I'm fine now code-wise.

rtengine/rawimage.cc
Lawrence37: It will be used if there is no matching raw crop in `camconst.json`. There may be some cameras which have a single raw crop defined for all resolutions even though the camera has a crop mode. We will have to fix those raw crops, either by adding the correct crops or limiting the crops to the full resolution images.

Lawrence37:
> The processed versions ended up with a magenta tint.

I think RawTherapee expects the X-Trans pattern to start where the left and top margins end, which makes it sensitive to margins that are not multiples of the pattern size. For reference, the crop detected in the metadata using this branch is `[696, 20, 4992, 3328]` (20 is not a multiple of 6) while the well-behaved crop in `camconst.json` is `[624, 0, 5004, 3348]`. This problem can be addressed separately (#6826 probably).

Lawrence37 (approved):

Lawrence37:
@LoKolbasz I believe your name is not in `AUTHORS.txt`. Could you please add yourself? :slightly_smiling_face:

Lawrence37:
I merged the pull request. Thanks for your help @LoKolbasz! Feel free to open a pull request to add your name to `AUTHORS.txt` any time.

LoKolbasz:
Thanks! Had a lot going on in these past few weeks, so I forgot about this.

PR #6836, opened 2023-09-03
Raw crop for Sony ILCE-7M4

Lawrence37:
Remove 10 columns of bad pixels from the right border.
See https://discuss.pixls.us/t/resolution-and-pixel-right-side-edge-pixel-stretch-bug-in-rawtherapee-with-sony-a7-ivs-arw-files/39191


Beep6581:
Please merge ASAP if for 5.10.

PR #6842, opened 2023-09-09
Fix mutex try lock for debug builds

Lawrence37:
The debug build crashes because failing to acquire a lock with `trylock` throws an error. `trylock` should not throw an error upon failing, but just return false. The release build works properly. This fixes the bug for debug builds.

### DIFF
--- a/rtgui/threadutils.h
+++ b/rtgui/threadutils.h
@@ -59,7 +59,7 @@ public:
 
 private:
     bool locked;
-    bool checkLock (bool noError=false);
+    bool checkLock (bool noError = false);
     void checkUnlock ();
 #endif
 };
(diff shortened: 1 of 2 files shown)


rtgui/threadutils.h
```diff
@@ -59,7 +59,7 @@ class MyMutex :
 
 private:
     bool locked;
-    void checkLock ();
+    bool checkLock (bool noError=false);
```
Floessie: @Lawrence37 Good catch (as always)! :+1:

I'd rather read that with a bit more of whitespace: `bool noError = false`. :slightly_smiling_face:

Floessie:
@Lawrence37 Thanks! :+1:

PR #6843, opened 2023-09-11
modal quit dialog

Benitoite:
Exiting RawTherapee is costly because of the loss of the detailed processing history. In order to prevent accidental deletion of the main RT window, this PR presents a modal exit confirmation dialog on exit.
exit2

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -1677,6 +1677,7 @@ MAIN_MSG_NAVIGATOR;Navigator
 MAIN_MSG_OPERATIONCANCELLED;Operation cancelled
 MAIN_MSG_PATHDOESNTEXIST;The path\n\n<b>%1</b>\n\ndoes not exist. Please set a correct path in Preferences.
 MAIN_MSG_QOVERWRITE;Do you want to overwrite it?
+MAIN_MSG_QUIT;Do you really want to quit RawTherapee?
 MAIN_MSG_SETPATHFIRST;You first have to set a target path in Preferences in order to use this function!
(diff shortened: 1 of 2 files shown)


Floessie:
Hi Richard,

Thanks for digging into the code and your contribution! :+1: 

Here is my very personal opinion on this topic: In more than 99% of the cases I'm sure, I want to quit. I don't like being treated as if I don't know what I'm doing. Same here on GitHub when they introduced the "Are you sure you want to log out?" dialog after explicitly clicking "Logout". As if I were nuts...

Ages ago I tested a DOS program which was a fake warehouse system to show teachers the (then) future of digitization. Of course, it was keyboard driven. So when you pressed 'q' it asked: "Are you sure you want to quit? ('y'/'n')" Then, after the 'y' keystroke: "Do you really want to quit or cancel? ('r'/'c')" This says something about the programmer's opinion of teachers. :wink: 

While I understand your reasoning, I don't think this dialog should be the default. IMHO it should not even be opt-out, but that's probably just me.

Best,
Flössie

Benitoite:
The program is too easy to quit. This adds a layer of safety to that. Sorry you don’t see the need.

Lawrence37:
Some programs, Firefox for example, have this feature and have an option to enable or disable it. I share the same opinion as @Floessie that a confirmation dialog is unnecessary personally. Given that no one has brought up this issue before, I think it should be disabled by default, or have a checkbox that says "Do not ask again".

Benitoite:
As a non-programmer this PR demonstrates what I need in this software as a user.  Clearly I will not recieve any assistance in this matter.

PR #6850, opened 2023-09-15
Fix UB in `ProcParams::load()` and the resulting Clang 15+ miscompilation

### LINKED ISSUE #6847: Miscompilation with Clang 15+ and optimization (SEGV at startup)
Hello fellow RawTherapists,

This issue is a follow up on #6846, where an unrelated SEGV on FreeBSD was reported. Quoting the stack trace from there:

```
Process 17289 stopped
* thread #10, name = 'pool-rawtherape', stop reason = signal SIGSEGV: invalid address (fault address: 0x1)
    frame #0: 0x0000000000d1731c rawtherapee`rtengine::procparams::ProcParams::load(Glib::ustring const&, ParamsEdited*) [inlined] bool (anonymous namespace)::assignFromKeyfile<int>(keyfile=0x000000089f03dda0, group_name=0x000000089f03e120, key=0x000000089f03de58, has_params_edited=true, value=0x00000008b156497c, params_edited_value=0x0000000000000001) at procparams.cc:260:33
   257          getFromKeyfile(keyfile, group_name, key, value);
   258 
   259          if (has_params_edited) {
-> 260              params_edited_value = true;
   261          }
   262 
   263          return true;
(lldb) bt
* thread #10, name = 'pool-rawtherape', stop reason = signal SIGSEGV: invalid address (fault address: 0x1)
  * frame #0: 0x0000000000d1731c rawtherapee`rtengine::procparams::ProcParams::load(Glib::ustring const&, ParamsEdited*) [inlined] bool (anonymous namespace)::assignFromKeyfile<int>(keyfile=0x000000089f03dda0, group_name=0x000000089f03e120, key=0x000000089f03de58, has_params_edited=true, value=0x00000008b156497c, params_edited_value=0x0000000000000001) at procparams.cc:260:33
    frame #1: 0x0000000000d172e9 rawtherapee`rtengine::procparams::ProcParams::load(this=0x00000008b1563000, fname=0x000000089f03fbf8, pedited=0x0000000000000000) at procparams.cc:7895:13
    frame #2: 0x00000000009310f5 rawtherapee`Thumbnail::loadProcParams(this=0x00000008b155e000) at thumbnail.cc:410:36
    frame #3: 0x000000000056fcd3 rawtherapee`CacheManager::getEntry(Glib::ustring const&) [inlined] Thumbnail::Thumbnail(this=0x00000008b155e000, cm=<unavailable>, fname=0x000000089f03fea0, cf=0x000000089f03fd00) at thumbnail.cc:129:5
    frame #4: 0x000000000056fac6 rawtherapee`CacheManager::getEntry(this=<unavailable>, fname=0x000000089f03fea0) at cachemanager.cc:118:34
    frame #5: 0x00000000008c4dc3 rawtherapee`PreviewLoader::Impl::processNextJob(this=0x000000086fe86440) at previewloader.cc:130:37
    frame #6: 0x00000000008c4f46 rawtherapee`sigc::internal::slot_call0<sigc::bound_mem_functor0<void, PreviewLoader::Impl>, void>::call_it(sigc::internal::slot_rep*) [inlined] sigc::bound_mem_functor0<void, PreviewLoader::Impl>::operator(this=<unavailable>)() const at mem_fun.h:1991:14
    frame #7: 0x00000000008c4f2b rawtherapee`sigc::internal::slot_call0<sigc::bound_mem_functor0<void, PreviewLoader::Impl>, void>::call_it(sigc::internal::slot_rep*) [inlined] sigc::adaptor_functor<sigc::bound_mem_functor0<void, PreviewLoader::Impl> >::operator(this=<unavailable>)() const at adaptor_trait.h:256:12
    frame #8: 0x00000000008c4f2b rawtherapee`sigc::internal::slot_call0<sigc::bound_mem_functor0<void, PreviewLoader::Impl>, void>::call_it(rep=<unavailable>) at slot.h:136:14
    frame #9: 0x00000008304a6711 libglibmm-2.4.so.1`___lldb_unnamed_symbol4301 + 65
    frame #10: 0x000000082bc82870 libglib-2.0.so.0`___lldb_unnamed_symbol2613 + 144
    frame #11: 0x000000082bc81672 libglib-2.0.so.0`___lldb_unnamed_symbol2606 + 66
    frame #12: 0x00000008400a1a7a libthr.so.3`thread_start(curthread=0x0000000876a81700) at thr_create.c:292:16
(lldb) 
```

The interesting part is `pedited=0x0000000000000000` in frame 1 and the `has_params_edited=true` in frame 0.

I was able to reproduce the bug on Debian (in a i386 VM I had at hand) and here are my findings so far:

* This is a Clang-only problem.
* It is only visible with optimization turned on. In a debug build, I had to add `-Og` to trigger the bug.
* It starts with Clang 15. I tested Clang 14, 15, and 16 on that VM.
* Turning the `bool` [here](https://github.com/Beep6581/RawTherapee/blob/c380044a29515700fe3c7907886642f801e28913/rtengine/procparams.cc#L251) into a `void*` "fixes" the problem.

My wild guess is that Clang 15+ somehow "fuses" the implicitly bool-converted `pedited` pointer in `has_params_edited` and the `bool&` pointer into the structure pointed to. This is why the first call to `assignFromKeyfile()` for "General"/"Rank" succeeds (&pedited->general.rank == 0), while the second call for "General"/"ColorLabel" (offset 1) fails. And while the access to an absent element surely is undefined, the conversion of a pointer (protecting against this undefined behavior) to a `bool` is absolutely within bounds.

How shall we proceed?

Best,
Flössie

Floessie:
Hi,

This is a fix for #6847 based on my idea there:

> Easiest would be to create a dummy instance and discard it (ideally automatically) at the end of the function.

HTH,
Flössie

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -17,7 +17,7 @@
  *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
  */
 
-#include <map>
+#include <memory>
 
 #include <locale.h>
 
@@ -248,7 +248,6 @@ bool assignFromKeyfile(
     const Glib::KeyFile& keyfile,
     const Glib::ustring& group_name,
     const Glib::ustring& key,
-    bool has_params_edited,
     T& value,
     bool& params_edited_value
 )
@@ -256,9 +255,7 @@ bool assignFromKeyfile(
     if (keyfile.has_key(group_name, key)) {
         getFromKeyfile(keyfile, group_name, key, value);
 
-        if (has_params_edited) {
-            params_edited_value = true;
-        }
+        params_edited_value = true;
 
         return true;
     }
@@ -271,7 +268,6 @@ bool assignFromKeyfile(
     const Glib::KeyFile& keyfile,
     const Glib::ustring& group_name,
     const Glib::ustring& key,
-    bool has_params_edited,
     const std::map<std::string, T>& mapping,
     T& value,
     bool& params_edited_value
@@ -289,9 +285,7 @@ bool assignFromKeyfile(
             return false;
         }
 
-        if (has_params_edited) {
-            params_edited_value = true;
-        }
+        params_edited_value = true;
 
         return true;
     }
@@ -7854,8 +7848,14 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
     Glib::KeyFile keyFile;
 
     try {
+        std::unique_ptr<ParamsEdited> dummy_pedited;
+
         if (pedited) {
             pedited->set(false);
+        } else {
+            dummy_pedited.reset(new ParamsEdited());
+
+            pedited = dummy_pedited.get();
         }
 
         if (!Glib::file_test(fname, Glib::FILE_TEST_EXISTS) ||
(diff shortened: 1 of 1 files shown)


Lawrence37 (approved):

Floessie:
@Lawrence37 Thanks for the review! :+1:

PR #6851, opened 2023-09-15
Batch edit of IPTC tags

Bezierr:
Hi there,

another proposal. As always, feel free to reject if you don't like it, or don't think it's suitable for general RawTherapee.

This adds the functionality to batch edit IPTC tags -- something that I was used to in Lightroom but missing from RawTherapee.

The fact that IPTC tags are, unlike other parameters, not individual values but contained in a single map, gave me some headache. For everything I have tested, my solution works, but it may well be that, for a more experienced RawTherapee developer, there would have been a better way to do this.

Logic is as follows:
- For everything except keywords and supplemental categories: A tag value entered for the first selected picture will replace existing values in all other selected pictures. If the new tag value is empty, tag values will be replaced by the empty string, i.e. deleted.
- For keywords and supplemental categories: An entry added to the first selected picture will be added to all other selected pictures unless it already exists. An entry deleted from the first selected picture will be deleted from all other selected pictures provided it was there. Empty strings will be treated just like any other entries.

This logic is, I believe, in line with the behavior one would expect, and with the way how batch editing works for other parameters. Since values are not numeric, the distinction between "add" and "replace" does not make sense; therefore, no changes to Preferences/Batch Processing.

Tested quite extensively for Linux, not for the other OS.

Side note: When working on this, I found that I needed some automatism for initializing all pictures before batch editing them (setting IPTC tags is usually the very first thing I do), since I kept forgetting to do it manually. Which was the reason for my "initialization" proposal some time ago. Thanks for your comments there; I'll look into that one again as soon as I find the time.


Lawrence37:
Nice work!

> Since values are not numeric, the distinction between "add" and "replace" does not make sense; therefore, no changes to Preferences/Batch Processing.

For keywords and supplemental categories, a set mode could do a replacement, though I'm not sure how useful that would be.

Bezierr:
I had thought about this replacement functionality myself, but given that (a) I personally have never missed it, and (b) all the programs I know don't provide it, I thought it would be better to not unnecessarily complicate things. So I'd propose to leave it as it is (provided nobody finds any bugs, of course) for the time being.

Bezierr:
Hi there,

for personal reasons, it's been quite a while since I last looked into RawTherapee (apart from using it for my pictures). Given that you seemed to like this proposal, I was a bit amazed to find that it neither made its way into the dev branch nor into the new releases.

As I always emphasize, it is, of course, entirely up to you what you include into main RawTherapee, and what not. Still, may I ask for the reason?

Lawrence37:
For me, it dropped off my radar after waiting for others to comment. Then there were higher priorities for 5.11 so I decided to leave it for 5.12 unless someone else reviewed it sooner.

rtgui/batchtoolpanelcoord.cc
```diff
@@ -562,11 +563,250 @@ void BatchToolPanelCoordinator::panelChanged(const rtengine::ProcEvent& event, c
...
+				initialPP[i].metadata.iptc[SUPPLEMENTAL_CATEGORIES].clear();
+				initialPP[i].metadata.iptc.insert(TITLE, "");
+				initialPP[i].metadata.iptc.insert(TRANS_REFERENCE, "");
+			}
+		}
+		
+		// determine changes to IPTC tags of first selected picture
+		originalParams = initialPP[0];
+
+		if (pparams.metadata.iptc[CAPTION].at(0).compare(originalParams.metadata.iptc[CAPTION].at(0)) != 0) {
```
Lawrence37: This if-else-if chain belongs in `IPTCPanel::write`. The booleans like `captionChangedFlag` can be placed in `ParamsEdited`. The `Glib::ustring`s like `captionChanged` are not needed since they can be found in `pparams`. The exceptions are the added/deleted keyword/category, which can also go in `pparams`.

rtgui/batchtoolpanelcoord.cc
```diff
@@ -562,11 +563,250 @@ void BatchToolPanelCoordinator::panelChanged(const rtengine::ProcEvent& event, c
...
+	}
+
     // combine with initial parameters and set
-    ProcParams newParams;
 
     for (size_t i = 0; i < selected.size(); i++) {
-        newParams = initialPP[i];
+		newParams = initialPP[i];
+
+		if (event == rtengine::EvIPTC) {
```
Lawrence37: Move this block into `ParamsEdited::combine`.

Lawrence37 (commented):
I noticed your code uses tabs for indentation, but the surrounding code uses spaces. Could you change the tabs to spaces for consistency? GitHub shows the tabs with a width of 8 characters which makes your code look over-indented. I think your editor is set to display tabs with a width of 4.

Function-wise, this works perfectly.

Bezierr:
Hi,
thanks for the feedback. I'll look into this as soon as I find the time, which unfortunately won't be during the next couple of weeks.

Bezierr:
Hi there,
just to give you an update.

I had implemented most of your requests, but found that when merged into the current dev branch, my functionality does not work at all anymore.

Given that I currently have only very limited time outside of my family and my day job, I can't say anything about when, or if at all, I'll have the time to continue working on this. So please don't count on it to be available for any upcoming release.

Lawrence37:
Understandable. It's interesting that it doesn't work after merging with `dev`. Nothing immediately comes to mind about what could break it.

Bezierr:
Found it :-)

A change was introduced to rtgui/metadatapanel.cc in pull request #7242 of 2024-11-25 that needs to be partially reverted -- it is immediately clear that my stuff can't work with that. Since I don't know why this was changed in the first place, I can't say whether reverting it has any undesired side effects, though I would assume that this is not the case.

I have implemented your proposals from September 28, with the exception of the last one, "Move this block into ParamsEdited::combine", which I haven't been able to get to work in all cases: When adding more than one keyword without changing the picture inbetween, only the last one is retained. I suspect that it has something to do with not changing the "initialPP[i]" directly anymore, and with the keywords subsequently being lost in BatchToolPanelCoordinator::closeSession. May be far off, though; if you have any idea, please let me know.

Everything included in the updated pull request is tested and seems to work.

I may look into the "ParamsEdited::combine" request again, but for the reasons given above, can't say whether that may be next week... or next year.

Lawrence37:
PR #7242 fixed an edge case where EXIF and IPTC data were reset to their default values when going back to the first history entry in the batch editor. The metadata editor was not available for the batch editor (and it wasn't designed to support batch editing), so it was sufficient to disable the metadata editor's reading and writing of the EXIF and IPTC data.

Lawrence37:
I have confirmed the EXIF reset bug has not returned in this PR.

PR #6864, opened 2023-10-12
Streamline Windows build action

kmilos:
Use pacboy for msystem independent install (easier to switch/test more [environments](https://www.msys2.org/docs/environments/) later)
Use build directory directly from CMake
Remove lensfun patch (included upstream)

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -35,21 +35,22 @@ jobs:
         with:
           location: C:\msys2
           update: true
+          msystem: MINGW64
           install: |
             gzip
             git
             intltool
-            mingw-w64-x86_64-gcc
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -35,21 +35,22 @@ jobs:
         with:
           location: C:\msys2
           update: true
+          msystem: MINGW64
           install: |
             gzip
```
kmilos, line 40: gzip can also be removed btw, it's already part of msys2 [base](https://packages.msys2.org/package/base?repo=msys&variant=x86_64) install...

Beep6581:
Will merge tomorrow if no one objects, including the removal of gzip.

PR #6865, opened 2023-10-14
Fix camconst.json white levels for 1DxII

CarVac:
I recently bought a Canon EOS 1D X Mark II and I had white level issues affecting highlight recovery specifically at ISO 160, 320, 640, and 1250.

I found that making these changes to implement per-color white levels fixed things.

Note: I've been testing in Filmulator, not in RawTherapee. Here's a filebin for example raws having clipped highlights at these ISOs: https://filebin.net/a4zbnfz7bdpvevsi (link will expire 2023-10-20)


Lawrence37:
I'm going to trust you on the white levels. Thanks for sharing @CarVac!

PR #6878, opened 2023-11-04
Fix for crash when moving cursor over loading raw image

### LINKED ISSUE #6875: RawTherapee crashes often on different actions in
Quite often, RT chrashes when working in Single Editor Tab. The actions leading to a crash differ; to me it appears random. The crash of which gdb's log file is attached happened:

1. I started RT and opened a folder which contains DNG-images
This opened the Single Editor Tab.
2. I pressed several times <F4> to browse forward the images
3. I pressed<F3> to backwards.
This caused the crash

# Version and OS details
- dev-branch self build with on following commit
` 72bf27121 - (HEAD -> dev, origin/dev, origin/HEAD) Removed gzip from windows.yml workflow #6864 (Fri, 13 Oct 2023 23:04:33 +0200) <Morgan Hardwood>`
But the release 5.9 crashes as well.
- NixOS 23.05
`uname -a`
`Linux maxiTower 6.5.9 #1-NixOS SMP PREEMPT_DYNAMIC Wed Oct 25 10:16:30 UTC 2023 x86_64 GNU/Linux` 

# GDB Log File
[log.txt](https://github.com/Beep6581/RawTherapee/files/13199136/log.txt)

Lawrence37:
If the image has the demosic method set to None, the navigator requests the raw value of the pixel under the cursor. When loading an image, the raw data may not be completely loaded when the request occurs. It causes a crash due to an out of bound access. I added a check that returns zero values for the pixel if the data is not loaded and also asserts for debug builds to catch other use-before-initialization errors. The data loading is asynchronous which makes it difficult to have a "proper" fix. That's why I opted for this simple fix.

Closes #6875.

### DIFF
--- a/rtengine/rawimagesource.cc
+++ b/rtengine/rawimagesource.cc
@@ -16,6 +16,7 @@
  *  You should have received a copy of the GNU General Public License
  *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
  */
+#include <cassert>
 #include <cmath>
 #include <cstdlib>
 #include <iostream>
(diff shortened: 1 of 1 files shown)


kaelte:
Locally, I have done the merge and tested: RT does not crash anymore and issue https://github.com/Beep6581/RawTherapee/issues/6875 would be closed :+1: 
But as I cannot program in C, I cannot judge the code changes.

Lawrence37:
Thanks for testing @kaelte!

PR #6881, opened 2023-11-10
Fix focus distance from metadata

### LINKED ISSUE #6879: Lensfun's vignetting correction does not work with CR2 files
While the necessary data are present in the lens profile, activating vignetting correction does nothing. Devignetting works in darktable and ART (1.20.1 or later).

Steps to reproduce: all CR2 files with corresponding lens profile, that I have tried, are affected.

The issue looks to be introduced, when the exiv2 branch has been merged into dev.

I have noticed this glitch in ART some time ago, but it was fixed in the meantime; first version, that let me devignette CR2 photos, was `ART_master_1.20-16-g9565d00c0_W64_Skylake_230710`.

Last RT version without the problem, that I have been and still am using, is `RawTherapee_dev_5.9-198-g1d9225dfc_W64_Skylake_230630`.

Two PlayRaw samples can be found here:

https://discuss.pixls.us/t/loch-ard-boat-shed/32842

https://discuss.pixls.us/t/maman-et-son-chiot/34177

Windows 10/11: `RT_dev_5.9-198-g1d9225dfc` vs `RT_dev_5.9-461-g0e835e8c0`
Ubuntu Studio 23.10 virtual machine: `RT 5.9-1 build1` vs `5.9-461-g0e835e8c0` (appimage)

Lawrence37:
Cherry-picked commit from ART that fixes the focus distance calculation for many cameras. Fixes Lensfun vignette correction which was not being applied to certain images.

Closes #6879.

### DIFF
--- a/rtengine/imagedata.cc
+++ b/rtengine/imagedata.cc
@@ -315,7 +315,29 @@ FramesData::FramesData(const Glib::ustring &fname, time_t ts) :
         /*
          * Get the focus distance in meters.
          */
-        if (find_exif_tag("Exif.NikonLd2.FocusDistance")
+        if (Exiv2::testVersion(0, 27, 4) && find_exif_tag("Exif.NikonLd4.LensID") && to_long(pos) != 0) {
+            // Z lens, need to specifically look for the second instance of
+            // Exif.NikonLd4.FocusDistance unless using Exiv2 0.28.x and later
+            // (also expanded to 2 bytes of precision since 0.28.1).
+#if EXIV2_TEST_VERSION(0, 28, 0)
+            if (find_exif_tag("Exif.NikonLd4.FocusDistance2")) {
+                float value = pos->toFloat();
+                if (Exiv2::testVersion(0, 28, 1)) {
+                    value /= 256.f;
+                }
+#else
+            pos = exif.end();
+            for (auto it = exif.begin(); it != exif.end(); it++) {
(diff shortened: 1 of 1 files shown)


kmilos:
There are also some new gotchas for Nikon Z bodies, see https://github.com/darktable-org/darktable/pull/15030 @agriggio

Lawrence37:
Thanks for bringing that to my attention.

Lawrence37:
https://github.com/kmilos/darktable/blob/c458f7e23d24a85d49c4bfb1cbe10f10c8cf282a/src/common/exif.cc#L1186

> `Z lens, need to specifically look for the second instance of Exif.NikonLd4.FocusDistance`

The comment says the second instance is needed, but the code looks for the last instance.
https://github.com/kmilos/darktable/blob/c458f7e23d24a85d49c4bfb1cbe10f10c8cf282a/src/common/exif.cc#L1195-L1198
```c++
      for(auto it = exifData.begin(); it != exifData.end(); it++)
      {
        if(it->key() == "Exif.NikonLd4.FocusDistance") pos = it;
      }
```

Is there always two instances of the tag if it exists?

kmilos:
> Is there always two instances of the tag if it exists?

Yes, in Exif.NikonLd4 group there can be exactly 2 instances of this tag (in exiv2 0.28 this was fixed and renamed to a unique one w/ suffix 2).

PR #6885, opened 2023-11-12
Fixes for memory leaks and memcpy overlap found by AddressSanitizer

Lawrence37:
AddressSanitizer flagged a couple of minor problems while I was fixing a noise reduction crash. This fixes all of the direct leaks and memory overlaps found within RawTherapee.

### DIFF
--- a/rtgui/controlspotpanel.cc
+++ b/rtgui/controlspotpanel.cc
@@ -660,7 +660,7 @@ void ControlSpotPanel::on_button_delete()
     nbSpotChanged_ = true;
     selSpotChanged_ = true;
     eventType = SpotDeletion;
-    std::unique_ptr<SpotRow> const delSpotRow = getSpot(selIndex);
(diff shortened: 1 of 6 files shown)


rtgui/controlspotpanel.cc
```diff
@@ -660,7 +660,7 @@ void ControlSpotPanel::on_button_delete()
     nbSpotChanged_ = true;
     selSpotChanged_ = true;
     eventType = SpotDeletion;
-    SpotRow* const delSpotRow = getSpot(selIndex);
+    std::unique_ptr<SpotRow> const delSpotRow = getSpot(selIndex);
```
Floessie: @Lawrence37 This is a nice cleanup. Would you mind to change `std::unique_ptr<SpotRow> const` to `const std::unique_ptr<SpotRow>` to align it with the rest of the code?

Lawrence37: Updated

Desmis:
I tested, and found no problem :)

PR #6890, opened 2023-11-21
Localization German Update

marter001:
Contains the german translation update for the current dev-branch.
Also fixes 2 typos in the default file.


rtdata/languages/Deutsch
```diff
@@ -85,8 +85,7 @@
 #084 06.10.2019 Erweiterung (TooWaBoo) RT 5.7
 #084 18.07.2019 Erweiterung (TooWaBoo) RT 5.6
 #085 29.07.2022 Erweiterung (marter, mozzihh) RT 5.9 
-#100
-#101 @LANGUAGE_DISPLAY_NAME=Deutsch
```
Lawrence37: Keep these lines (at the bottom of all the comments). Although it doesn't have an effect because the language name is the same as the file name, it helps people see that the language name is indeed "Deutsch".

Beep6581:
Seems to have been made obsolete by #6891, closing.

PR #6892, opened 2023-11-26
LA exposure hide show sliders when expcomp different from zero 

Desmis:
See https://discuss.pixls.us/t/problems-with-local-adjustment-function-in-rt/40615/3


Lawrence37 (approved):
Looks good to me. I do have a question though. What is the reason why the exposure compensation must be non-zero before the other parameters can be used?

Desmis:
@Lawrence37 

Thank you, I will merge soon.

it's a choice...I consider (personal opinion) that "exposure" and its associated settings are not very efficient - when used with LA. Hence this choice so that the user uses other features like Equalizer or TRC

Jacques

PR #6643, opened 2022-12-14
Research and development - WB auto -temperature correlation - tests - finding the optimum settings

Desmis:
I propose to modify "Preferences" following the change of observer 10° by observer 2°.
observer 10° is newer (1964) than observer 2° (1931) and is recommended by the CIE. Nevertheless we made the choice of observer 2° by default (what there was in 5.8).

In "Color Management" (Preferences) you now have the choice between observer 2° and observer 10°.

In addition, I take this opportunity to make available 3 of the 9 possible options of "Temperature correlation" (White balance auto).
* 0bserver 2° instead of observer 10° (default) - best results are with observer 10° (from my tests)
* Sort values in chroma order instead of histogram
* Maximum number of colors used in picture (default = 34).
 
 There are no rules to say such or such of these 3 parameters is better or less well it depends on each image...

Jacques

PR #6903, opened 2023-12-19
White balance auto temperature correlation - improve behavior with non raw files

### LINKED ISSUE #6902: Temperature Correlation bias sliders have no effect on JPGs
Currently using Dev Version: 5.9-472, Windows 10.
The 'AWB temperature bias' and 'Green refinement' sliders have no effect on jpg images but are still enabled.

(As a side point, the capillarization of the the title "**A**uto **T**emperature **c**orrelation **Setting**" is inconsistent. Recommend only 'Auto' be capitalized)

@Desmis

### LINKED ISSUE #6911: Export via Que gives different white balance
Using the latest version, Dev 5.9-474 I get a white balance that is is different than displayed if I export via the que on this image and pp3. Exporting via the save function gives the correct output.
RAW file: https://drive.google.com/file/d/1fOpHDtlEbLccLF3JifZUify-wp8duZxc/view?usp=drive_link
save output pp3: https://drive.google.com/file/d/1oVw0HVQkU1ATMsD1iVk9U7-lV64vzPVc/view?usp=drive_link
que output pp3: https://drive.google.com/file/d/1FEQu-kOJfxn7N_RaSbmiJbK-W-oQScy3/view?usp=drive_link
DSCF9184-save
DSCF9184-1
save vs que

Desmis:


Lawrence37:
@Desmis I see the temperature and tint are fixed to 6501 and 1 for non-raw files which is different behavior compared to 5.9. Is there a reason for disabling ITCWB?

Desmis:
@Lawrence37 
You are true, I forgot the Observer conversion

For 2° - Temp=6501 green=1
For 10° - Temp = 6473 green = 0.967

Lawrence37:
5.9 performs automatic white balance for non-raw images, but this pull request disables it so you get different results with the same pp3. Sorry for not being clear in my previous comment.

Desmis:
@Lawrence37 
Ok, it's clearer. However, I think I don't know how to do both keep automatic detection for TIF/JPG and take into account the particularities of "Temperature correlation".

If you have a solution do not hesitate to make the modification

Thank you

Jacques

Lawrence37:
I think I'm still not able to explain my thoughts. In 5.9, auto temperature correlation works for all images. The only problem is the green refinement and temperature bias has no effect for non-raw images. In this PR, auto temperature correlation is unchanged for raw images, but doesn't do any automatic correction for non-raw images.

I *think* the correct behavior for non-raw files is to allow the automatic correction. The only change required for dev is disabling the green refinement and temperature bias sliders (this PR already does that). However, I don't know the algorithm well so I don't know if this is the right thing to do. Is there a reason why the algorithm shouldn't be used for non-raw images?

@chaimav Am I correct about what needs to be done, or am I misunderstanding the issue?

Desmis:
@Lawrence37  @chaimav

As I designed "Temperature correlation" (Itcwb), it should only be able to work (including 5.9) with Raw files.
Certainly 5.9 displays the menu, but the data is completely wrong.

The latest version (the one for the PR) which only makes the interface and the Temperature value "acceptable" supports even less (so to speak) anything other than Raw files. I reported this during the "Research and development" update this spring.

But, if it is possible to make it work (TIF, JPG), then I am interested.

jacques

chaimav:
> I think I'm still not able to explain my thoughts. In 5.9, auto temperature correlation works for all images. The only problem is the green refinement and temperature bias has no effect for non-raw images. In this PR, auto temperature correlation is unchanged for raw images, but doesn't do any automatic correction for non-raw images.
> 
> I _think_ the correct behavior for non-raw files is to allow the automatic correction. The only change required for dev is disabling the green refinement and temperature bias sliders (this PR already does that). However, I don't know the algorithm well so I don't know if this is the right thing to do. Is there a reason why the algorithm shouldn't be used for non-raw images?
> 
> @chaimav Am I correct about what needs to be done, or am I misunderstanding the issue?

You are right, I missed the fact that Temperature Correlation and RGB Grey stopped working on jpgs entirely. I was only asking for the sliders (which have no effect) to be disabled.

Here's a thought - could these new updated white balance methods be moved to the Raw Tab and have the old ones put back on the color tab? Or would that cause too many other issues?

Lawrence37:
@Desmis I looked over that thread briefly and saw that you mentioned both AWB methods only work on raw images. It surprised me that there is no way to automatically white balance non-raw images. What do the algorithms require that only raw images have?

If we remove the AWB from non-raw files, I'll have to think about how to preserve existing edits.

Desmis:
@Lawrence37 
for AWB - RGB grey - the behavior (before the modification I just made) is identical to 5.9.
I reinstalled 5.8, and there AWB -RGB grey works in TIF / JPG.

I just edited the PR to separate the code from RGBgrey and Temperature corellation. Now AWB RGB gray behaves like in 5.8.

For "Temperature correlation" the current behavior is similar to Camera, while in 5.9 - where non-raw was not taken into account the results were bad.
Maybe it is possible to make it work in TIF/JPG, but I don't see how? The current solution does not bring (identical to Camera) heretic behavior.

It seems to me, I am not an IT specialist, that this part of code is not JPG/TIF compatible
Improccordinator.cc line 499
        if (todo & (M_INIT | M_LINDENOISE | M_HDR)) {
            if (params->wb.method == "autitcgreen") {
                imgsrc->getrgbloc(0, 0, fh, fw, 0, 0, fh, fw, params->wb);
            }
        }

 I added the missing references (at least I hope I do) for getrgbloc

But perhaps ?

chaimav:
> @Desmis I looked over that thread briefly and saw that you mentioned both AWB methods only work on raw images. It surprised me that there is no way to automatically white balance non-raw images. What do the algorithms require that only raw images have?
> 
> If we remove the AWB from non-raw files, I'll have to think about how to preserve existing edits.

As I mentioned before, I don't think it should be removed from non-raw. Non-raws and legacy edits can still use the old code. The new code can be added to the raw tab and only work on raw images.

Lawrence37:
The problem is that temperature correlation AWB was always supposed to be raw-only, so it will have to be removed for non-raw images.

Lawrence37:
> It seems to me, I am not an IT specialist, that this part of code is not JPG/TIF compatible

@Desmis I think you can move `getrgbloc` (and `redloc`, `greenloc`, and `blueloc`) from `RawImageSource` to `ImageSource`.

Desmis:
@Lawrence37

getrgbloc is already in imagesource.h
 Excuse my incompetence, but what do you mean by move redloc, greenloc, blueloc to imagesource.h, and how to do ?

You say:
> The problem is that temperature correlation AWB was always supposed to be raw-only, so it will have to be removed for non-raw images.

So, what do we do ? The current PR solution which makes AWB rgbgrey work (quite poorly) like 5.8 ensures "compatibility", and "Temperature correlation" does not behave badly

Lawrence37:
You can move the entire code for `getrgbloc` to `ImageSource` (`imagesource.h`). Currently it is in `RawImageSource` (`rawimagesource.cc`):
https://github.com/Beep6581/RawTherapee/blob/ac15cb4a5c3d302080439177f6e99242b5fbfe2c/rtengine/rawimagesource.cc#L7403-L7479
It uses the variables `redloc`, `greenloc`, and `blueloc`, so those will also have to be moved from `rawimagesource.h` to `imagesource.h`:
https://github.com/Beep6581/RawTherapee/blob/ac15cb4a5c3d302080439177f6e99242b5fbfe2c/rtengine/rawimagesource.h#L91-L97

> So, what do we do ?

I was talking to @chaimav , but if `getrgbloc` is the only thing preventing ITCWB from working on non-raw images, then moving it to `ImageSource` will be the best and easiest solution.

Desmis:
@Lawrence37 
Happy new year.

As you suggested I copied the code, adapting it a little from rawimagesource.cc to imagesource.h

    virtual void getrgbloc(int begx, int begy, int yEn, int xEn, int cx, int cy, int bf_h, int bf_w, const procparams::WBParams & wbpar)
   
 and all code after...

and also

protected:
    double redAWBMul, greenAWBMul, blueAWBMul; // local copy of the multipliers, to avoid recomputing the values
    cmsHPROFILE embProfile;
    Glib::ustring fileName;
    FramesData* idata;
    ImageMatrices imatrices;
    double dirpyrdenoiseExpComp;
    array2D<float> green;
    array2D<float> greenloc;
    // the interpolated red plane:
    array2D<float> red;
    array2D<float> redloc;
    // the interpolated blue plane:
    array2D<float> blue;
    array2D<float> blueloc;
    int W, H;
public:


I added W and H
But compiler says...
In file included from E:/code/repo-rt1/rtgui/editorpanel.cc:25:
E:/code/repo-rt1/rtengine/imagesource.h:87:20: error: field 'green' has incomplete type 'rtengine::array2D<float>'
   87 |     array2D<float> green;
      |                    ^~~~~
In file included from E:/code/repo-rt1/rtgui/editorpanel.cc:20:
E:/code/repo-rt1/rtgui/editorpanel.h:39:7: note: declaration of 'class rtengine::array2D<float>'
   39 | class array2D;
      |       ^~~~~~~
E:/code/repo-rt1/rtengine/imagesource.h:88:20: error: field 'greenloc' has incomplete type 'rtengine::array2D<float>'
   88 |     array2D<float> greenloc;
      |                    ^~~~~~~~
E:/code/repo-rt1/rtgui/editorpanel.h:39:7: note: declaration of 'class rtengine::array2D<float>'
   39 | class array2D;
      |       ^~~~~~~

what am I forgetting ?
Probably something here or elsewhere
 ImageSource () : references (1), redAWBMul(-1.), greenAWBMul(-1.), blueAWBMul(-1.),
        embProfile(nullptr), idata(nullptr), dirpyrdenoiseExpComp(RT_INFINITY)  {}

But assuming that it works for getrgbloc, won't we have the same problem with the other functions Wbauto and Itcwb? And again there is nothing to suggest that this will work well?

Of course you can directly modify this PR.

Thank you

Desmis:
@Lawrence37 

I push a change with #6911

But when we know what to do for the auto balances, we may need to modify "simpleprocess", to insert the same modifications as in "improcoordinator" for TIF/JPG

Lawrence37:
Happy new year @Desmis 🙂
I think adding `#include "array2D.h"` will fix the error. If the other functions are necessary for the AWB, then they'll have to be moved too. Of course, I'm not familiar with the algorithms so I'm not exactly sure what needs to be done to enable AWB for non raw files.

Desmis:
@Lawrence37 
#include "array2D.h" is already in imagesource.h

I tried this
    array2D<float> green;
    array2D<float> greenloc;
    // the interpolated red plane:
    array2D<float> red;
    array2D<float> redloc;
    // the interpolated blue plane:
    array2D<float> blue;
    array2D<float> blueloc;
    int W;
    int H;
public:
    ImageSource () : references (1), redAWBMul(-1.), greenAWBMul(-1.), blueAWBMul(-1.),
        embProfile(nullptr), idata(nullptr), dirpyrdenoiseExpComp(RT_INFINITY), green(0.f, 0.f), greenloc(0.f, 0.f), red(0.f, 0.f), redloc(0.f, 0.f), blue(0.f, 0.f), blueloc(0.f, 0.f), W(-1), H(-1) {}

But always the same compilation error...
error: field 'green' has incomplete type 'rtengine::array2D<float>'
etc.

Thank you
Jacques

Desmis:
@Lawrence37 

I searched a lot, tinkered, tinkered...
In fact imagesource.h is not used for this case (WB procedures), but it is stdimagesource.cc and .h

I set benchmarks for AWB RGB grey and the automatic calculation functions are never called for TIF/JPG, and yet the code (at least part) is in stdimagesource...this apparently doesn't work.
I have to say that I don't know how the results are obtained (perhaps an operating bias) , in any case, not by the Raw modes or similar.

I don't think we should continue trying to make Itcwb work with TIF/JPG as well as AWB RGB grey. I don't think this is possible unless someone else has a solution.
In addition, duplicating (very) large quantities of code from rawimagesource to stdimagesource risks causing maintenance problems.

What do you think ?

Desmis:
@Lawrence37 
 I have just modified simpleprocess to make it compatible with the "Temperature correlation" logic for TIF/Jpg. This version preserve also "compatibility" AWB RGB grey with 5.8.

Of course we can delete if it is possible to make "Temperature correlation" work (as well as Autowb rgb grey ) with TIF/JPG

Lawrence37:
@Desmis I'll have a look maybe this weekend.

Lawrence37:
I got distracted by the LibRaw integration which is taking longer than expected. I will try to find some time after it's done (it's almost complete!).

Desmis:
@Lawrence37 
No problem at all... since 5.10 ???????

Lawrence37:
@Desmis I looked at the AWB behavior for 5.9 and here is my understanding.

- When ITCWB was introduced, it always used temperature 4999 and tint 1 (observer 2°) for non-raw files.
- In 5.9, RGB grey always used temperature 6473 and tint 0.967 (observer 10°) for non-raw files (equivalent to temperature 6501 and tint 1 for observer 2°). Before 5.9, RGB grey worked for all files.
- In 5.9, ITCWB used observer 10°, but still used temperature 4999 and tint 1.

Is this correct? If so, then I think it is best if we

- For pp3s created by 5.9, if the WB method is RGB grey, change it to custom with temperature 6473, tint 0.967, and observer 10°.
- For pp3s created by 5.8, if the WB method is Temperature correlation, change it to custom with temperature 4999, tint 1, and observer 2°.
- For pp3s created by 5.9, if the WB method is Temperature correlation, change it to custom with temperature 4999, tint 1, and observer 10°.
- Re-enable non-raw RGB gray for 5.10.
- Disable non-raw Temperature correlation for 5.10.

The first 3 will preserve edits from 5.8 and 5.9. The 4th one is important because RGB grey worked in 5.8 and older. The last one removes confusion because Temperature correlation is not implemented for non-raw files. Does this sound good?

Desmis:
@Lawrence37 

Version 5.9 was wrong for all no-raw files.
The settings obtained by 5.9 (tif, jpg) corresponded to the starting settings for the raw, then not calculated. So inconsistent.
I made an error for "AWB rgb grey" for TIF and Jpg to "simplify" the code. So wrong value.
In summary for TIF/JPG, 5.9 is still bad. I don't think we should recover values from pp3s that are bad...

In the current version (that of the PR):
* Itcwb cannot be calculated for TIFF and JPG, so I chose the Camera settings
* for AWB grey - I reproduced exactly the same code as for 5.8, which for this PR, gives the same results as 5.8. What we can notice (5.8 and the PR) is that the Raw code is never used. The results are erratic - like 5.8.

Jacques

chaimav:
> I don't think we should recover values from pp3s that are bad...

I am not sure I agree. It is likely the user already compensated for the error elsewhere in the editing pipeline. My understanding is that all 5.x should be backwards compatible. When 6.0 gets released, then we break backwards compatibility and fixe errors.

Lawrence37:
@Desmis I think it is not hard to preserve 5.9 edits *IF* the behavior of the AWB algorithms is well-defined. Similar to what @chaimav wrote, some people may have used AWB and were satisfied with the results even though they were wrong. If it's not hard to preserve edits, then I think it should be done.

> So inconsistent.

This part makes me wonder if it is possible to keep the 5.9 edits. However, looking at the code in 5.9:
https://github.com/Beep6581/RawTherapee/blob/9b85839884f920de0c7cd3a150f865f3c5e47732/rtengine/improccoordinator.cc#L531-L566
It seems to confirm my observations about 5.9 with non-raw files. For ITCWB, the initial temperature and tint are 5000 and 1. The function `getAutoWBMultipliersitc` simply sets `rm`, `gm`, and `bm` to 1, so it does not change the temperature and tint. Due to rounding errors, the final temperature and tint become 4999 and 1. For RGB grey, `rm`, `gm`, and `bm` are used. Since they are all 1, the temperature and tint become 6473 and 0.967.

The AWB behavior for non-raw files in 5.9 looks simple.

Desmis:
@chaimav  I understand :)

@Lawrence37
Can you make this modification to maintain compatibility - handling pp3 is not my strong point.
Of course by directly pushing the change into the PR.
Thank you very much  :)

Jacques

Lawrence37:
@Desmis Is my analysis of the 5.9 behavior correct? If so, I will start implementing the compatibility.

Desmis:
@Lawrence37 

The part of Itcwb you mentionned fixe temp and green, at the beginning of the algorithm.
I think you can.

Thank you
Jacques

Lawrence37:
@Desmis I added the code to maintain compatibility for 5.9 edits. I found some bugs for raw files.
- In the editor, after using Temperature correlation, RGB grey gives the wrong results. I think it's because RGB grey is caching the multipliers and the ITCWB function overwrites the cached multipliers. Strangely, it is ok in 5.9 even though the is code is the same.
- Temperature bias gives different results for Temperature correlation compared to 5.9.

These probably aren't caused by this pull request, but the second and maybe first one should be addressed before the release.

Desmis:
@Lawrence37 @chaimav 

I just did a first test of the change you made.
Overall it works better, even if there are some quirks as you point out, and in some cases (maybe I'm doing something wrong), I get the values "Camera".
The operation of the 2 "Auto" modules is already not simple in "Raw" mode, especially Itcwb. 
I understand the desire to maintain compatibility - even with poor 5.9 results - for TIF/JPG, but this should not result in software and update complexity. I see you used the variable "compat_version"  - interesting for other compatibility cases.

I see the modification to "rthumbnail" , a good thing :)

For me, it is OK.

Jacques

Lawrence37:
@Desmis Is the change in the Temperature bias behavior intentional?

Desmis:
@Lawrence37 
"Is the change in the Temperature bias behavior intentional?" -  What are you talking about ?
* For AWB grey, I haven't changed anything at all
* for Itcwb, it works for RAW files, and not for TIF/JPG...
I will verify a second time.

Thank you

Other thing. I see in your last comment this morning - News about Rawtherapee : " I’d like to update the instructions too as I have the relevant knowledge. Unfortunately, I don’t have permission to do so (yet) "
I have this permission"
For me at least the lib "exifv2" is missing for Windows.
For me I used : pacman -S mingw-w64-x86_64-exiv2
Is it enough to add "exiv2" in the list ?

$ pacman -S mingw-w64-x86_64-gcc mingw-w64-x86_64-gdb mingw-w64-x86_64-make mingw-w64-x86_64-pkg-config mingw-w64-x86_64-cmake mingw-w64-x86_64-ninja mingw-w64-x86_64-gtkmm3 mingw-w64-x86_64-lcms2 mingw-w64-x86_64-fftw mingw-w64-x86_64-lensfun mingw-w64-x86_64-libiptcdata

Jacques

Lawrence37:
For ITCWB on raw files, there is a slight difference in output compared to 5.9 when using the Temperature bias. If Temperature bias is 0, the results are identical.

Desmis:
@Lawrence37 
If you compare to 5.9, the algo since july 2023, is completely new for some parts... and of course leads to difference...
In rare cases new algo and 5.9 give the same results.. (temp, green) but rare.

Jacques

Lawrence37:
Pull request #6643 preserved 5.9 edits. At least it was supposed to do that. The Temperature bias may have been missed. Now I wonder how much work it is to make it compatible.

Desmis:
Honestly, I don't think it's worth it.
Jacques

Desmis:
@Lawrence37 
I even think that would be a mistake.
Compatibility with 5.9 in terms of colorimetry is ensured by Temerature and green. Which is perfect. Introducing “bias” amounts to (re)running the old algorithm and comparing it to the new one. Considerable work for little or no interest.

As a reminder (of course you know), the "Research" branch and PR, carried out in spring 2023 made it possible - thanks to the participation and requests of users  - to very clearly improve the algorithm which now works very well in the vast majority case, but which at the code level is quite incompatible with 5.9. Maintaining Temp and Green compatibility is necessary, but we should not go further.

Thank you again

Jacques

Lawrence37:
> Is it enough to add "exiv2" in the list ?

Yes, I think this is the only new requirement.

Desmis:
@Lawrence37 

Do you want due to the fact that you do not have the rights - which is surprising - that I modify this part (addition of exiv2) in Rawpedia

Lawrence37:
That would be great Jacques. Thanks!

Desmis:
@Lawrence37 

Done. I just add "mingw-w64-x86_64-exiv2"

Jacques

Lawrence37:
@Desmis I'm going to do some code cleanup and merge it into `release-5.10` since it has important fixes to preserve older edits.

Desmis:
@Lawrence37 

OK

Jacques

Lawrence37:
@Desmis I completed the cleanup and fixed the RGB grey cache bug. If everything is fine this should be good to go in 5.10.

Desmis:
@Lawrence37 
I just looked at your work, impressive for a "clean-up". From the tests I'm doing it works.

But, incidentally, I still have a lot of trouble understanding how the pipeline for TIF/JPG works - that is to say the architecture rawimagesource, imagesource, stdimagesource, and so on.

Okay for me, for merge.

Thank you very much, impressive work :)

Jacques

Lawrence37:
The cleanup was mostly combining the duplicate code in improccoordinator and simpleprocess. It should make it easier to maintain and reduce the likelihood of bugs due to changing the code in one place but forgetting to do the same in the other place.

I think the non-raw AWB requires a couple of functions in the classes in iimage.h.

Desmis:
@Lawrence37 
This is a simplification of the code, which will facilitate updates.

My remark on the pipeline is general for raw and non-raw (not specifically WB) and also concerns the articulation of the process in several pieces "thumbnail", "dcrop", "improccoordinator", "simpleprocess"... which does not facilitate either the process or the updates. This all dates back to the design of the system in 2006  :)

Question: who makes the merge ? 

Thank you again

Jacques

Lawrence37:
@Desmis I will let you have the honor of merging. I will rebase the commits for `release-5.10`.

Desmis:
OK i will merge soon :)
I apprecied this symbolique gesture.
Thank you
Jacques

PR #6920, opened 2024-01-07
Queue template improvements

### LINKED ISSUE #6915: Queue template improvement for better mirroring of directory structure
I propose providing improved support for the following use-case:

- Raw files are under a base directory
- Place exported files under a different base directory
- Retain the directory structure from the base directory when exporting

For example suppose the raw images are in the base folder `D:\raw_photos\forProcessing`, and you want to export into `E:\exported_photos`, so that a raw image in `D:\raw_photos\forProcessing\some\more\directory\terms` is exported to `E:\exported_photos\some\more\directory\terms`. The existing `%p` and `%dN` template terms defined in [https://rawpedia.rawtherapee.com/Queue](https://rawpedia.rawtherapee.com/Queue) can't do that unless every image is at the same depth under the base directory.

I propose adding a variant of the `%dN` term with a minus sign, so `%d-N` means "the entire directory, but with the first N terms removed". That means `%d-1` would remove just the drive letter (`D:` in the example above). In the example, `%d-2` would remove `D:\raw_photos\' and `%d-3` would remove `D:\raw_photos\forProcessing`, which is what desired in this case.

In the queue tab you would enter `E:/exported_photos/%d-3/%f` to accomplish the stated goal.

The attached patch implements this enhancement in a branch taken from the 5.9 tag. To issue it as a pull request I should target the `release-5.10` branch, correct?

[0001-Added-queue-template-term-d-N-to-insert-the-full-dir.patch](https://github.com/Beep6581/RawTherapee/files/13806096/0001-Added-queue-template-term-d-N-to-insert-the-full-dir.patch)

sgilbertson:
Resolves #6915 

This pull request includes the following changes related to location templates and destination paths in the Queue panel:

- New format specifier `%PN` in location templates, meaning path from given position to the end (whereas `%pN` is from the start to the given position)
- `%p`, `%P` and `%d` specifiers can all accept either
  - `N` = position from the end, so input path is for example `5/4/3/2/1/filename`
  - `-N` = position from the start, so input path is for example `-1/-2/-3/-4/-5/filename`
- Revision of `BatchQueue::calcAutoFileNameBase`, eliminating the `pa` vector, instead using `da` for d, p and P specifiers
- Updated tooltip describing those features
- New "destination preview" label, initially loaded with "Select a thumbnail to preview its destination path here". If you select a thumbnail in the file list, the destination path and filename are shown, and:
  - If multiple files are selected, the path shown is for the one that would be processed earliest (i.e., the one nearest the upper-left corner). I anticipate people will normally select a single image when using this feature.
  - The previewed path reflects events that would change it's value, so you can for example immediately see the effect of any change you make to the template text.
  - The field is selectable, so you can copy its contents to the clipboard, for example to be pasted into File Explorer.
  - The field is intentionally not reverted back to its default text when no image is selected, because you might want to copy a path after processing the queue.
- On Windows, the template-driven path now includes backslashes rather than forward slashes. On other operating systems forward slashes are inserted as path separators. The template text may include either character, and those are faithfully copied to the output path.
- Previously, the template character following `%pN` was always omitted, I assume because users would commonly enter something like `%p1/...` and you would wind up with two consecutive slashes (which is a real problem on Windows). The character is now skipped only if it's a slash or backslash, so for example the `A` in `%p2A` is not skipped.

Here is the field in use, with its tooltip (and Yours Truly at Point Pelee):
image

The text changes when, for example, you select the "Save to folder" radio button:
image

The tooltip for the location template entry box is getting very large, and will likely continue to grow.
image

 It seems to me a better approach would be to put a help button beside the field, perhaps like this:
image

The "?" icon shown would be a toggle button, which when in the on state causes help text to appear in a panel perhaps inserted between the control panel and the file thumbnail panel. That means you would be able to keep the help on the screen while you're editing the template. I don't propose including that change here, but it might be appropriate as a separate task.


Floessie (approved):
@Lawrence37 I think this is okay now, code-wise.

Lawrence37:
Nice work @sgilbertson. The code looks good. While testing (in Linux), I found a few issues and one item to ponder.

Issues
1. `%p` no longer starts with a `/`, so the default template `%p1/converted/%f` yields something like `home/me/Pictures/photo.jpg` instead of `/home/me/Pictures/photo.jpg`
2. The tool-tip for `%P4`/`%P-1` erroneously shows the path starting with `/`.
3. A minor complaint, but the width of the Output Location frame keeps changing as you type the template. Maybe the path preview needs a different expand/fill option or be wrapped in a `Gtk::Viewport`?

Previously, an invalid value of `%p#`, e.g. `%p5` for the file `/only/three/directories/photo.raw`, was replaced with an empty string. For example, `%p5/converted/%f` would have become `converted/photo.jpg`. With this pull request, it is replaced by the full path, e.g. `/only/three/directories/converted/photo.jpg`. Since it's an invalid value, I don't care much if the behavior changes, but it would be nice to have some intuitive default. An empty value or `/` makes sense for positive numbers since the path gets shorter as the number increases. The behavior of invalid `%P#` values can also be considered. In this pull request, they become empty strings.

sgilbertson:
@Lawrence37 I believe I have addressed all of your comments.

> %p no longer starts with a /, so the default template %p1/converted/%f yields something like home/me/Pictures/photo.jpg instead of /home/me/Pictures/photo.jpg

The dev-branch code matches these start-of-path strings, and the changes I've made do likewise:

- `\\` for UNC paths like `\\server\share\directory` (although I have not figured out how to test that)
  - I have added `//` for Samba UNC paths like `//server/share/directory` (again I do not know how to test those)
- `/` for Unix/Linux absolute paths
- (nothing) for Windows paths, the first term of which is letter+colon (such as `C:`)
- `/` for a path that starts with `\` but not `\\` (in this case my code now inserts `\` rather than `/`)

Do you know of a scenario in which the last one can occur? At a Windows command prompt, you can type `\a` when referring to directory `a` at the root of the current drive, but I think the path passed to `BatchQueue::calcAutoFileNameBase` is always canonical and absolute, so it will never start with a single backslash.

BTW This time around I tested on both Windows 10 and Ubuntu 22.04 (using WSL and GWSL, so I can build in the same source directory, and so both can access my photos).

> The tool-tip for %P4/%P-1 erroneously shows the path starting with /.

The new code adds leading characters whenever a `%p` or `%P` result includes the start of the path, which means the tooltip now matches what the code does on Linux and macOS. It doesn't match what happens on Windows, but that's not new.

> Previously, an invalid value of %p#,...

As `%p` and `%P` are now implemented, if you specify a range where either end is outside the available limits, you get the full directory path. So if you say "give me the first 5 elements" or "give me the last 5 elements" and there are only four elements, it will give you all four elements.

For the same 4-element path, if you say "give me the last N elements starting with the 5th one" or "give me the first N elements starting with the one before the first one", the inserted string is empty. Those are "zero element" cases.

The `%d` specifier inserts nothing if you specify a path element that does not exist.

rtgui/batchqueue.cc
```diff
@@ -939,10 +967,9 @@ Glib::ustring BatchQueue::calcAutoFileNameBase (const Glib::ustring& origFileNam
                     }
                 } else if (options.savePathTemplate[ix] == 'd') {
                     // insert a single directory name from the file's path
-                    // da.size()-1 omits the last element, which is the filename
                     ix++;
                     unsigned int n = decodePathIndex(ix, options.savePathTemplate, da.size());
```
Lawrence37: `n` should be an `int`

rtgui/batchqueue.cc
```diff
@@ -928,8 +953,11 @@ Glib::ustring BatchQueue::calcAutoFileNameBase (const Glib::ustring& origFileNam
                 } else if (options.savePathTemplate[ix] == 'p') {
                     // insert path elements from the start of the path up to the given index
                     ix++;
-                    unsigned int n = decodePathIndex(ix, options.savePathTemplate, da.size());
-                    for (unsigned int i=0; i <= n && i < da.size(); i++) {
+                    int n = decodePathIndex(ix, options.savePathTemplate, da.size());
+                    if (n > 0) {
```
Lawrence37: I think this should be `n >= 0` for the `%p4`/`%p-1` case shown in the tool-tip.

rtgui/batchqueue.cc
```diff
@@ -43,6 +43,59 @@
...
+
+namespace   // local helper functions
+{
+    // Look for N or -N in templateText at position ix, meaning "index from end" and "index from start".
+    // For N, return Nth index from the end, and for -N return the Nth index from the start.
+    // N is a digit 1 through 9. The returned value is not range-checked, so it may be >=numPathElements.
+    // or negative. The caller performs any required range-checking.
+    int decodePathIndex(unsigned int& ix, Glib::ustring& templateText, size_t numPathElements)
+    {
+        int pathIndex = (int)numPathElements;    // a value that means input was invalid
```
Lawrence37: Could you change these to `static_cast` instead of C-style casts?

Lawrence37:
> Do you know of a scenario in which the last one can occur?

I don't know. It seems like it will never happen.

rtgui/batchqueue.cc
sgilbertson: Yikes! In my "testing" on Ubuntu I totally missed the slash disappearing for `%p-1` (note to self: get more sleep). I'm glad you caught it.

Lawrence37 (approved):
Looks good!

sgilbertson:
Thanks, @Floessie and @Lawrence37. It's my first pull request on this project, and I don't see a "merge" button here. Am I waiting for someone else to do that?

Floessie:
@sgilbertson 
> Am I waiting for someone else to do that?

Exactly. And I vote for @Lawrence37 to do that, as he also tested the code.

Thanks for your contribution!

Best,
Flössie

Issue #5410, opened 2019-08-12
Locallab: Improvements of GUI


Pandagrapher:
Hello All,

Lately I have started a new session of improvements for Locallab GUI. The main objectives are:

- Rewrite some part of Locallab GUI to make it more generic: this will improve robustness, maintainability and simplify further evolution
- Remove Batch Mode support for Locallab: local treatments cannot really be managed in Batch Mode...
- By default, a spot won't have linked Locallab tool. New GUI will allow to add/remove Locallab tool for each tool (only linked tool parameters will be saved in .pp3 file, will fix #5255). Preview:
![image](https://user-images.githubusercontent.com/40305292/62855753-59223680-bcf3-11e9-865c-68890604947c.png)
- Improve copy/paste mechanism for Locallab spot
- Add more mouse/keyboard controls on spot (to reduce number of widgets to manage), maybe add new spot shapes...
- Bugfixes...

Here a detailed schedule I will edit regularly:

- [x] Locallab tools management
- - [x] Creation of generic class to manage Locallab tool individually
- - [x] Conversion to all Locallab tool to this new generic class
- - [x] Adaptation of actual GUI
- - [x] Remove Batch Mode support for Locallab
- - [x] Add possibility add/remove Locallab tool
- - [x] Fix #5255
- - [x] Improve copy/paste of spot
- [ ] Locallab spot management
- - [ ] Creation of generic class to manage several type of spot (rectangle, ellipse, others?)
- - [ ] Adaptation of actual GUI
- - [x] Avoid reprocessing when switching between spots when not necessary (fix #5268)
- - [ ] Add more mouse/keyboard controls
- - [ ] Add possibility to deactivate a specific spot to see its effects on image
- - [x] Fix #5354 (thanks to Hombre57)

For the moment, modifications are managed in my fork (I have started from commit `afd5d5d` of newlocallab branch):
https://github.com/Pandagrapher/RawTherapee

List to be completed with your suggestions ;)
Pierre

@Desmis

Desmis:
@Pandagrapher 
It's ok for me.
When you say "Creation of generic class to manage several type of spot (rectangle, ellipse, others?)", does this incorporate the drawing of curves, polygons .. ?

jacques

Pandagrapher:
@Desmis 
> When you say "Creation of generic class to manage several type of spot (rectangle, ellipse, others?)", does this incorporate the drawing of curves, polygons .. ?

[...]

PR #6928, opened 2024-01-15
Local adjustments tools used globally  - Laspotmain

### LINKED ISSUE #6922: Discussion about using local-adjustment tools globally
The Local Adjustments tab contains a number of tools that are also useful as global adjustments (e.g. Log Encoding). Currently they can be used as such by creating a full-image spot and setting the Scope to 100 to disable the deltaE detection. While this works well in most cases, it requires a number of clicks and often considerable scrolling to set up. From a workflow standpoint it is cumbersome and could deter some users. 
One solution to this would be to add the possibility of choosing between global or local adjustments at the top of the current local-adjustments menu. Selecting the global option would ideally hide the spot set-up menu along with the Preview deltaE and Scope options and let the user add the appropriate tool as the next step. Although this would help de-clutter the local-adjustments menu, having to go into the Local Adjustments tab to select global-adjustment tools isn't optimal.
Another alternative, which would no doubt require considerably more work (I speak from a position of blissful ignorance about what this might entail), would be to have access to selected Local Adjustments modules (set up as global adjustments) in the Advanced tab. There is already some duplication of functionality between the Advanced and Local Adjustment tabs (Wavelets, CIECAM, Retinex) so conceptually it wouldn’t be inconsistent with what exists already. Access to any of the local-adjustments modules from the advanced tab would also presumably give access to the C, L, L(ch) masks associated with the particular tool thereby enhancing RawTherapee's global masking capabilities, which are currently restricted to Color Correction Regions.

Desmis:
A first try... It's 90% a GUI problem.

In this solution I added an extra choice "main".

It sets all the settings as "full image", but systematically sets "scope" to 100, all manipulations on deltaE are canceled as well as all those on transitions.
The corresponding settings are hidden.

Of course you can duplicate, use masks, use if necessary "exclude", etc.

Jacques


Desmis:
I added (not without difficulty...I'm a poor computer scientist,  maybe there is a simpler method ?) the possibility in "Preferences" to choose the Spot startup method (Normal spot, Excluding spot, Full image, Global). 
* "Define Spot method for Local Adjustments"

Note that in the new "Global" (main) option, the center of the RT-spot remains visible and movable. This is particularly useful if you want to make "Graduated Filters"

Jacques

waynesutton50:
This certainly does a good job of de-cluttering the interface. I haven't tested it in any detail but on the face of it, everything seems to work as it should. I did find one anomaly though that can be reproduced as follows:
Select Full-image and a particular tool and then activate the Preview DeltaE button. If you then switch to Global instead of Full-image without de-activating the Preview deltaE first, the deltaE indication will remain. 

For local-adjustment spots (Normal, Excluding & Full-image) there are a couple of things which might help also:
1. Move the "Show additional settings" checkbox down after the spot set-up . To me this would be more logical and it would also mean that you don't have to scroll right to the top of the menu if you want to add additional settings when a particular tool has been opened.
2. Remove the common Scope (color tools) slider and equip all tools with their own Scope slider. Having a common slider for some tools only is not only confusing but it also means you have to scroll up to adjust it when using several tools on the same spot. A deltaE preview button on a per-tool basis would also be a help and would avoid the confusion that sometimes arises because someone has pressed the deltaE preview button not realising that a tool has to be added before it does anything.

Desmis:
A summary of the current version of this PR.

1) The system now works with 4 possible RT-spot choices : 
* Normal : the basic setup for local editing.
* Exclude : authorizes to cancel a correction on a defined area.
* Full-image : uses the tools on the whole image, but retains the functionalities of LA - deltaE, transitions,...
* Global :  uses the tools on the whole image without deltaE, transitions.


2) in the Global case:
* the GUI is reduced to the strict minimum, allowing the execution of several sessions (several spots)
* you retain the possibility of LA, by associating several spots to correct the image with spots in "normal", "exclude", or even "full image" mode
* the center of the spot remains visible to allow, for example, multiple Graduated filters of variable “center”
image



3) "Preferences" allows you to choose the default RT-spot type : Normal,...., Global.
*  Of course this is only the default choice, you can change this choice for each spot created
image



4) LA of this PR has just had modifications
* bugs and bad behavior in some cases when you want "Preview deltaE" in Settings  or in "Mask and modifications" - Preview deltaE
* Most tools now have a Preview built into the tool
image


Thanks to @waynesutton50 for his cooperation and suggestions.

Jacques

waynesutton50:
Thank you @jdc, this makes a big difference to the usability of Local Adjustments. Using any of the tools globally now only requires one extra click compared to the tools in RawTherapee's main menu. The inclusion of the Preview button in the individual tool menus also saves a lot of scrolling back and forth for the majority of use cases.

Desmis:
@Lawrence37 

Could you take a look at what I did.
This works, but I'm a (very) bad computer scientist. My solutions (which work) are probably not very conventional.

The GUI, like the rtengine code, is complex due to managing:
* multiple Spot
* multiple Tools
* multiple levels of complexity
* multiple levels of Spot type.

Main changed are:
Rtengine : improccoordinator.cc , iplocallab.cc, rtengine.h, procparams.cc
Rgui : controlspotpanel.cc (and h), locallab.cc (and h), locallabtools.cc (and h), locallabtools2.cc
Thank you

Jacques

epadepa:
Lots of nice changes above! I posted a suggestion in #6922 to rename Local Adjustments (now that they are sometimes global) and move the module to the Advanced tab.

But seeing the above work I also would like to suggest the following changes to the handling of spots. I'm fully aware these are complex requests that I'm myself unable to program. I do think they would be another significant step for this module.

1. Make Global the default and enable Local Ajustments via a checkbox.
2. Replace the add button and spot type dropdown with icons for Ellipse, Rectangle and Full image.
3. Change the behaviour so that the buttons at 2 above work as the spot removal tool. Click icon, click position in image.
4. Remove the rename button and enable long press rename on spot name
5. Replace enable button with eye icon toggle next to the spot name following the typical convention of layers/outliners.
6. Remove the hide spot feature and instead use the pattern of the Spot Removal tool. A global button to hide all spots.
7. Place the delete and duplicate button under the list of spots.

By and large the Local Adjustments tools could fit well into established ui patterns but are currently solved in unusual ways. In particular clicking Add to "randomly" place a spot that you then have to move and resize feels like a bad pattern. ART has a lot of this type of ui in its masking etc.

Some examples
[Blender Outliner](https://docs.blender.org/manual/en/2.80/_images/editors_outliner_interface.png)
[Gimp Layers](https://docs.gimp.org/2.10/en/images/dialogs/link-visibility.png)

Desmis:
@epadepa 

Thank you for your suggestions and your participation.
As I wrote in issue #6922, I am not at all a GUI specialist. 

All GUI specialist players (@Pandagrapher -Pierre  @Hombre57 - Jean Christophe) have not been active for a while, perhaps this is an opportunity to make their contribution

Your proposed modifications
1. it's easy to do, with Preferences". If everybody agree I can change the default setting in 'options'. But originally the concept is "Local", multiplicity of Spots, possibility of canceling (exclude)...So no checkbox, it's going down a path that absolutely doesn't suit me and is very complex to do (at least for me)
2. ...7 : It's GUI work... I don't know how to do it, and I think that's what Pierre had planned.

Jacques

waynesutton50:
> * Make Global the default and enable Local Ajustments via a checkbox.
> *
Currently you can choose to make Global the default in Preferences
image

Changing to local within the local-adjustments menu is simply a matter of selecting another type of Spot using the "Spot method" drop-down menu . 
image

Unless I've missed something , I don't see what advantage there would be having a checkbox. In fact it would introduce an extra step (Checkbox + spot selection).

epadepa:
@Desmis, yes it's all gui work and I'm aware of the lack of devs. When/if someone looks at it it might help to have a considered workflow/ui idea. 

@waynesutton50 my list is a proposal for a whole workflow/ui logic. Individual points may or may not make sense without the others. The key to my proposal was:
1. For the user global isn't a spot type, it's just the tools. A kind of default state.
2. The pattern for local edit spots should have the normal click > place > size logic that is the convention for such tools. Even full image needs to be placed, but not sized.

The above means a different order and different ui elements to match conventional ui patterns. I understand that current proposals match the logic of the code but I think that for users a different approach makes sense.

Global shouldn't have any of the spot handling ui. When your needs go beyond global you switch to local and once spots have been added you can't go back to global without deleting the spots. (greyed out checkbox).

My proposal is less flexible than the one you have made but flexibility is complexity. To really maximize the straightforwardness of global it makes sense to limit how it interacts with local edits. In the code global would probably be a "spot" and when tools have been attached and the users enables local edits said global spot will appear in the spot list as a global spot but no further globals can be added. You have to use full image after that.

I can try to mockup a ui image if people want.  Perhaps just wireframe but gkt look would be better. For some background i spend all my days in design software and has done so since the nineties. CAD, desktop publishing, image editing, photo editing, 3d modelling. It my dayjob and my hobby. (perhaps not a good mix :)

Desmis:
@epadepa

Excuse my bad english. 

May I say (while putting myself in Wayne's shoes) that I completely disagree with you :)
I'm not trying to be unpleasant with you, but this overall comment is what I've been seeing since I started LA in 2014....Some went as far as saying that it was bullshit... 

It's absolutely not a matter of code, but of logic (for example @Lawrence37  wants several sessions), and of process. Have you ever been in the situation where the overall setting is good, but one area is wrong ? Of course you are going to tell me that we can make masks, etc. but it's complicated...and not in LA's logic.

Following your logic, which I respect, the solution is to duplicate each of "LA's" tools. This makes no sense to me..

Of course, it is entirely respectable to think differently from me. I remind you, since the start in 2014/2015, I have had a lot of opposition (it's an understatement). The logic from "Photoshop" with mask and layers is totally different, it is also relevant, although it is not at all the one I chose.

And excuse me if the tone is not "very pleasant", but I appreciate your intervention, because you have a point of view and you defend it.
:)

But this debate must not become heated, it makes no sense.

Jacques

waynesutton50:
> I think that for users a different approach makes sense.

I understand the logic and may well have agreed with you before having spent quite some time using a commercial alternative that incorporates the U-Point technology (as well as using it as a separate plug-in). It includes a lot of the principles that you have outlined but in my experience the workflow isn't any more efficient than the current RawTherapee solution. That coupled with the fact that RawTherapee's Spot technology, even in the Basic mode, is far more powerful than its commercial equivalent, means that RawTherapee can achieve better results more quickly in a lot of cases.

epadepa:
Sorry there has been some misunderstanding! I'm not proposing any changes to how local edits work. It's only some changes to types of buttons and the ordering of things in the interface. The spot concept works great imho. I don't know what made you think I proposed such changes.

Ui changes have large and positive effects on how the tools work as you have already demonstrated with the changes in the laspotmain branch (which I have compiled and tested). This cleanup is great.

So I made some quick mockups. Please ignore the bad icons, suitable ones should be available from existing icon sets but I did quick ones myself instead so save time. It's just a rearrangement of icons and the global mode as a simpler mode without spot management in the ui. Naturally the center spot has to be visible and movable on the 'canvas' but all other spot settings are hidden.

The transition gradient etc are missing below on the right but this is just an omission not a proposal. The three icons in the right pane are ellipse, rectangle, full image and exluding spots. Click spot icon, click canvas location resize. This is more efficient than select type, click add, click canvas location, resize spot. "Composed tools" is not a real name just a stand in for a good one.

So below the left pane shows what we call global mode which would be the default. The right shows the tool after checking local edit and adding a spot.

240123_rawtherapee-laui
Note the "Global" spot in the list above would only be there if you have added one more more tools in the global mode. That is before checking the Local Edit box.

Desmis:
@epadepa 

I understand better.

It's entirely GUI that I don't know how to do, even the solution you propose with the checkbox. The system is extremely complex.

If someone can modify the GUI before 6.0 it will cause big compatibility problems. 
But I'm not convinced, the solution with "checkbox"  is better.

I hope I didn't hurt you. This is not the kind of behavior I share.

Thank you again for your participation.

Jacques

Lawrence37:
@epadepa To avoid complicating this pull request, could you add your suggestions to a relevant issue, creating one if necessary? This particular comment may interest you. https://github.com/Beep6581/RawTherapee/issues/5410#issuecomment-644651339

Lawrence37:
@Desmis Since 5.10 is around the corner and this pull request has over 1000 new lines, I'm bumping the milestone to 6.0. I hope you are ok with it.

Desmis:
@Lawrence37 

OK, no problem

Lawrence37:
So far in my tests, it looks good. There is one minor issue with the Preview ΔE toggle buttons. If I enable one button, then enable another one, the first button is still toggled. That means this is possible:
1. Enable Preview ΔE in tool 1. It shows the ΔE. Good.
2. Enable Preview ΔE in tool 2. It shows the ΔE for tool 2. Good. The Preview ΔE for tool 1 is still enabled. Strange.
3. Disable Preview ΔE in tool 2. The ΔE turns off. Bad, because the Preview ΔE for tool 1 is still enabled.
4. Disable Preview ΔE in tool 1. The ΔE is still off. Good.
5. Enable Preview ΔE in tool 1. It shows the ΔE. Good, but I had to press the button twice for it to show the ΔE.

Ideally, enabling a Preview ΔE button turns off the previously active one, if it exists.

Desmis:
@Lawrence37 

Thank you very much for this review :)

This behavior (bad) is known and documented here.
https://rawpedia.rawtherapee.com/Local_Adjustments#Previewing_the_adjustment_area_using_deltaE_(%CE%94E)

I don't think there is a solution with the "button". But using the same command in "Mask and Modifications" everything works correctly (at least I hope).

This entire code was designed, in principle, by Pierre  @Pandagrapher , whom I thank infinitely. What remains now, and this is not an easy task, is to create a more intuitive GUI interface for “Settings”.

Thank you again

Jacques

Lawrence37:
I read the documentation you linked, but I can't find the section that talks about the issue I described. I see it says there is a limitation when using the button in Settings, but the issue I see is with the buttons in the individual tools. Can you copy and paste the part of the documentation here?

Desmis:
@Lawrence37 

Here the copy of this part of Rawpedia.

But I don't really understand what you mean, by "buttons in the individual tools" (my english is bad!).

Perhaps the documentation is not precise enough or not up to date (from this PR obviously), but **I added these buttons in some (not all) modules** to **avoid going back and forth** with Settings....But their operation is identical to that of Settings, **we can only use one at a time**. Otherwise to use several and reset, you must use **"Mask and Modifications" => Preview deltaE, which is not accessible in "Basic" mode**

=====
Previewing the adjustment area using deltaE (ΔE)
You can preview the areas of the image that will be affected by any changes. The preview does not show the changes themselves or the transitions, but allows you to set the scope of any adjustments.

There are two possibilities:
* Use the Preview ΔE button located in Settings. **This will only work properly if you have activated one (and only one) of the color tools in "Add tool to current spot" menu.**
* Use the Preview ΔE option in the "Mask and modifications" menu associated with a particular tool (standard and advanced modes only). In this case the GUI takes into account any adjustments made with the tool and works regardless of the number of activated tools.
Jacques

Desmis:
@Lawrence37 
The modules which are not equipped with it correspond, at the LA level, to different processing algorithms for the management of the deltaE 
* Contrast by detail levels
* Sharpening
* Blur grain & denoise
* Soft light & Original Retinex

Jacques

Lawrence37:
Now there are multiple Preview ΔE buttons, and you write that only one can be used at a time. That makes sense to me. However, if you click on two of the buttons, you can see that both are pressed. I think if you click on the second button, the first button should be deactivated automatically.

Desmis:
@Lawrence37 
I think you can't have everything :)

The operation of the GUI is complex when you have several tools in the same Spot and/or several Spots. And I don't know how to do. Pierre made a system work with "Mask & modifications", but this button added "for convenience" always had this restriction. It seems to me that this should be specified in the documentation.
I added these buttons, this avoids going back and forth, but their management remains the same.

Making Scope work for all tools - with a separate slider, whereas previously the "color" tools were grouped together is a notable plus. But the management of the Preview remains with this handicap

Thank you

Desmis:
@Lawrence37 
But of course if someone: you, Pierre  @Pandagrapher   or another volunteer tries to improve I only see advantages.

Jacques

Desmis:
@Lawrence37 
I know that there is a small problem with "Preview deltaE", but I think that to my knowledge there are no solutions, without a complete overhaul of the GUI. This problem seems secondary to me compared to the interest of Global mode.

I propose to merge in the coming days, if no one objects.

Jacques

Lawrence37:
There is a simple solution. Allow me some time to fix it. I am sure someone will bring this up as a usability problem if it is not fixed.

Desmis:
@Lawrence37 

No problem, I wait :)

Jacques

Lawrence37:
@Desmis I finally had enough time to finish the button toggling code. I think this is good now.

Desmis:
@Lawrence37 @waynesutton50 

I just compiled and tried. Everything seems to work fine.

Many thanks. This is the kind of change I don't know how to make.
I'm going to try it every which way, but I think it will confirm the correct behavior.

If nothing stands in the way, we'll be able to merge within the week.

Thank you again

Jacques

rtdata/languages/default
```diff
@@ -1,4 +1,5 @@
 #00 default translation file
+#00 default translation file
```
Lawrence37: There's a duplicate line here.

Desmis: OK, I will change in next commit

PR #6861, opened 2023-10-02
Local adjustements - Color appearance Cam16 and Log encoding - Harmonize Abstract Profile

Desmis:
I had some problems with the lacam16 branch merging with dev, which led me to change the pull request

I put the previous PR in draft mode.

The content is the same

PR #6931, opened 2024-01-24
Improve Dynamic Range Compression - for some images with very high DR - Main and LA

Desmis:
When I tried very high dynamic images, more than 15Ev (artificial TIF), I noticed that "Dynamic Range compression" both in "main" and "LA" behaved badly (very).
I re-examined the code from ART (which came from RT...) and found anomalies.
Thanks to Alberto @agriggio 

https://drive.google.com/file/d/1vAzFY7Qh8MdJ882J_4JeO_cnpGELzD8D/view?usp=sharing
Jacques


Lawrence37:
@Desmis Is this intended for 5.10? I see there is a fix and also a new feature.

Desmis:
@Lawrence37 

Yes, for 5.10
Same code as in ART, a fix and add a checkbox in LA "saturation control"

rtdata/languages/default
```diff
@@ -1450,6 +1450,7 @@ HISTORY_MSG_LOCALCONTRAST_RADIUS;Local Contrast - Radius
 HISTORY_MSG_LOCALLAB_TE_PIVOT;Local - Equalizer pivot
 HISTORY_MSG_LOCAL_DEHAZE_BLACK;Local - Dehaze - black
 HISTORY_MSG_LOCAL_GAMUTMUNSEL;Local - SC - Avoid Color Shift
+HISTORY_MSG_LOCAL_TMO_SATUR;Local Exp Fattal Satur
```
Lawrence37: "Satur" should be "Sat" or "Saturation"

rtengine/tmo_fattal02.cc
```diff
@@ -1242,15 +1244,33 @@ void ImProcFunctions::ToneMapFattal02(Imagefloat *rgb, const FattalToneMappingPa
                 float &r = rgb->r(y, x);
                 float &g = rgb->g(y, x);
                 float &b = rgb->b(y, x);
+                float s = 1.f;
                 if(l > 1.f) {
                     r = max(r * l - offset, r);
                     g = max(g * l - offset, g);
                     b = max(b * l - offset, b);
+                    if (satcontrol) {
+                        s = pow_F(1.f / l, 0.3f);
```
Lawrence37, line 1253: Is there an explanation of how this function and the 0.3 value was chosen?

Lawrence37 (changes requested):

rtengine/tmo_fattal02.cc
Desmis: This value 0.3 (same as in ART), brings about the change in behavior when the values RGB  are very high. In the majority of cases this has no effect (or little).
The change brings a reasonable weakening, we can see it on the TIF test image that I gave

Note: all tone-mapping tools introduce, depending on their design, a change in saturation (more or less). This value aims to significantly restore the original “saturation” values. This allows, among other things, to make the colors more natural and avoid (in this case) an increase in chromatic noise.

This is partly why I tried (then gave up) to work on the Q variable without change to colors (Q = brightness in LA - Cam16 - ColorApperance). I gave up because it was very difficult. Now with the new PR this is done upstream of the Ciecam process in RGB mode

https://github.com/Beep6581/RawTherapee/pull/6861

Jacques

rtdata/languages/default
Desmis: OK, I will change soon

Thank you

Lawrence37 (approved):

rtengine/tmo_fattal02.cc
Lawrence37: Ok, I just wanted to know if it is empirical or has a logical derivation. It is the first, it seems.

Desmis:
@Lawrence37 

OK, I merge

Thank you very much

Jacques

rtengine/tmo_fattal02.cc
Desmis: I think it is empirical. It is Alberto's work
But there are places in the code where I put similar functions which take into account that we are in the interval [0..1]. I sometimes used SQRT (0.5) or CBRT (0.33) because they can be faster.

Thank you again

Jacques

PR #6934, opened 2024-01-28
Defringe tool banding & desaturation fix

### LINKED ISSUE #6929: Defringe banding
**Short description**
Cross-post from [here](https://discuss.pixls.us/t/defringe-banding/41564), the defringe tool creates banding visible in cases of extreme purple fringes. CA correction tools do not help, neither does changing the defringe settings. I usually set the threshold to zero to get the smoothest result. The image below is the difference of the demosaiced images, with and without the defringe tool applied. A periodic banding is clearly present.

diff-screen-scaled

**Steps to reproduce**
1. Open the raw image I provided
2. Load the included profile
3. Banding caused by the defringe is most noticeable on the left side of the left bird.

**Expected behavior**
The defringe tool should create smooth results (at least with zero threshold), without creating any visible artifacts or banding.

**Additional information**
- Version of RawTherapee: 5.9
- Version of your operating system: Arch

Other useful information:
[Raw, pp3, diff](https://discuss.pixls.us/uploads/short-url/lMCsMSEWGxArXyyyzfjKDetaZXR.7z)

Lawrence37:
Because Defringe reads and writes to the same buffer as it defringes, it progressively desaturates the image as it process the image from top to bottom. With multithreading, this causes visible horizontal bands. This pull request separates the reading to a copy of the buffer. In addition to Defringe, similar bugs affect
- Color Appearance & Lighting > Hot/bad pixel filter
- Contrast by Detail Levels > Reduce artifacts

These are fixed too. The CbDL bug causes a similar desaturation and banding problem. The CAL bug does not have noticeable effects because only a small percentage of pixels are modified, resulting in a much smaller feedback effect.

Fixes #6929.

### DIFF
--- a/rtengine/PF_correct_RT.cc
+++ b/rtengine/PF_correct_RT.cc
@@ -27,6 +27,8 @@
 //
 ////////////////////////////////////////////////////////////////
 
+#include <vector>
+
 #include "gauss.h"
 #include "improcfun.h"
 #include "cieimage.h"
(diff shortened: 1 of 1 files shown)


Lawrence37:
I will merge it this weekend of there are no objections.

PR #6948, opened 2024-02-07
Change default behavior Local Adjustments settings Avoid Color Shift to XYZ Absolute

Desmis:
I propose this small change of 2 files - Procparams.cc and Controlspotpanel.cc - both seem necessary to ensure proper functioning.

This change aims to avoid bad behavior when the image contains strong highlights (artifacts, bad command responses, etc.).

This choice executes a gamut check in XYZ mode, at the end of the various actions on an RT-spot, whatever the number of tools used in this RT-spot.

Of course it is followed by “Munsell correction”, which corrects color drift specific to Lab mode (blue/purple, red/orange, green, etc.)

The increase in processing time is reasonable on average 200ms (full image) and concerns an RT-spot.

Of course the user can always select another choice ("none", "Munsell only", etc.)

Jacques

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -2972,7 +2972,7 @@ LocallabParams::LocallabSpot::LocallabSpot() :
     structexclu(0),
     struc(4.0),
     shapeMethod("IND"),
-    avoidgamutMethod("MUNS"),
+    avoidgamutMethod("XYZ"),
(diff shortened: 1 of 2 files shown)


Lawrence37 (approved):

Desmis:
@Lawrence37 
Thank you

PR #6951, opened 2024-02-11
Issue 4765 queue template time and date

### LINKED ISSUE #4765: Allow date as output template formatting string
I've recently switched from LR to RT and I am really impressed.
However there are two things which I really miss in the batch queue.

1. Batch resize option
If you are reducing the size of a big collection of pictures, resizing every single of them individually is quite tedious. It would be really nice to have the possibility to resize pictures in batch queue by defining width, height or Megapixel limit (while keeping the ratio) and also an option to avoid enlarging.

2. More formatting options for naming the exported files
I use the date as a prefix to make the exported files sortable by the original date of the picture. There are surely other formatting options which other people miss.

sgilbertson:
Resolves #4765

This pull request adds support for a new time/date format specifier in queue output templates. The following text, copied from the on-screen help, describes how the new specifier works:

-----

**Date and time**

Three different date/time values may be used in templates:

-   `%tE"%Y-%m-%d"` = when export started
-   `%tF"%Y-%m-%d"` = when file was last saved
-   `%tP"%Y-%m-%d"` = when photo was taken

The quoted string defines the format of the resulting date and/or time. The format string %tF"%Y-%m-%d" is just one example. The string can use all conversion specifiers defined for the g_date_time_format function (see https://docs.gtk.org/glib/method.DateTime.format.html).

Example format strings:

-   `%tE"%Y-%m-%d"` = 2001-02-03
-   `%tE"%Y%m%d_%H%M%S"` = 20010203_040506
-   `%tE"%y/%b/%-d/"` = 01/Feb/3/

-----

In addition to the feature described above, this change also changes the way template help is presented. Previously, a large and growing tooltip appeared when you hovered over the template entry field, describing the feature in detail. As soon as you started to type, the tooltip would of course disappear. I wanted the help text to be persistent, so you can refer to it while editing the template, and also scrollable, so that we don't need to artificially constrain its size as new template features are implemented.

Here is what I did to improve the display of help information for the output template format:

- I added a "?" button to toggle display of help for output templates (click once to show help, again to hide it):
  - Help appears on the left side of a split pane, with thumbnails appearing in the right-hand pane
  - The center split can be dragged to resize the two child panes
  - It's a scrolling window
  - The help pane is selectable, so you can copy its contents to the clipboard and paste somewhere else. In particular that may be useful for the URL that's provided for g_date_time_format documentation.
- I split help text into multiple language-specific strings, with code to combine them
  - Some parts of the help are OS-specific, such as the `%d`, `%p` and `%P` examples being based in `D:\tom\photos\2010-10-31\photo1.raw` on Windows and `/home/tom/photos/2010-10-31/photo1.raw` on Linux (and others)
  - Some examples in the help are generated in the code, so the result strings are guaranteed to match what the code actually does. Directory and path examples are based on a sample path that can be different for different languages. Time and date examples are based on the following "1,2,3" timestamp: `2001-02-03T04:05:06.123456`
- I added help for the specified time/date mechanism, with generated examples
- I reduced the tooltip text for the output template entry field, including an instruction to click the "?" button for detailed help.

I contemplated whether `""` should be a special-case date/time template, equivalent to `"%Y-%m-%d"`. That way you could insert for example `%tE""` if you just want a directory per day and don't really care about the format. I have not included that feature, but could easily do so if reviewers think I should.

Here is what it looks like with the help toggled on using the "?" button. This is the Windows version. I was hovering over the template entry field, so you can see the reduced tooltip for that control (because the details are now in the help window). The help pane is resizeable and scrollable. It's also selectable, so you can copy text to the clipboard from it.

image

Here's part of the screen doing the same thing in Linux, showing the OS-dependent difference in the help text:
image


rtgui/batchqueue.cc
```diff
@@ -18,8 +18,10 @@
  */
 #include <glibmm/ustring.h>
 #include <glib/gstdio.h>
+#include <sys/stat.h>
```
Lawrence37: Is this portable? It looks like it works on Windows only because we're using MinGW.

rtgui/batchqueue.cc
```diff
@@ -94,6 +96,21 @@ namespace   // local helper functions
             }
         }
     }
+
+    // Look in templateText at index ix for quoted string containing a time format string, and
+    // use that string to format dateTime. Append the formatted time to path.
+    void appendFormattedTime(Glib::ustring& path, unsigned int& ix, Glib::ustring& templateText, const Glib::DateTime& dateTime)
```
Lawrence37: `const` `templateText`?

rtgui/batchqueue.cc
sgilbertson: I'm learning GTK+ as I go along... Found https://developer-old.gnome.org/glibmm/stable/classGio_1_1FileInfo.html
I think that'll make it platform-independent. (UPDATE: Done)

rtgui/batchqueue.cc
sgilbertson: It looks like I don't actually need `dateTimeIsValid`. I can just make sure `dateTime` is either valid or `nullptr`. I'll make that change. (UPDATE: Done) (UPDATE2: I meant to add that as a new comment, not a reply. My reply to "const templateText?" is "done").

rtgui/batchqueue.cc
```diff
@@ -1022,20 +1021,19 @@ Glib::ustring BatchQueue::calcAutoFileNameBase (const Glib::ustring& origFileNam
...
                             {
-                                struct stat fileStat;
-                                if (stat(origFileName.c_str(), &fileStat) == 0) {
-                                    time_t timestamp = fileStat.st_mtime;
-                                    dateTime = Glib::DateTime::create_now_local(timestamp);
-                                } else {
-                                    dateTimeIsValid = false; // file not found, access denied, etc.
+                                Glib::RefPtr<Gio::File> file = Gio::File::create_for_path(origFileName);
+                                if (file) {
+                                    Glib::RefPtr<Gio::FileInfo> info = file->query_info("time::modified");
```
Lawrence37: [`G_FILE_ATTRIBUTE_TIME_MODIFIED`](https://docs.gtk.org/gio/const.FILE_ATTRIBUTE_TIME_MODIFIED.html) is defined as `"time::modified"`.

Lawrence37 (commented):
I noticed the help text has different amounts of indentation for the various template specifiers.
image

sgilbertson:
@Lawrence37 My original intent was that "The meaning of..." was part of the sentence that starts with "Using this pathname" (so maybe "The" should have been "the"), but your way is probably clearer. I went with your suggested and added a newline. I have also made the other change involving `G_FILE_ATTRIBUTE_TIME_MODIFIED`.

Lawrence37:
For the spacing, I mean the indentation is different. The first red line is shorter than the second one.

sgilbertson:
@Lawrence37 My colourblindness is my excuse. I didn't see the red lines at first, only the arrows. I can see them, they just don't particularly stand out. Thanks for the clarification. Indents are used in four places [UPDATE: Looks like I missed the fifth one for the example path], and I have made them all the same.
image

sgilbertson:
It was only after I posted that last comment that I noticed I had missed the extra indent (3 spaces rather than 2) on the example path. The two indents illustrated below are now equal too.
image

Lawrence37 (approved):
Looks good now!

rtgui/batchqueue.cc
```diff
@@ -94,6 +95,21 @@ namespace   // local helper functions
             }
         }
     }
+
+    // Look in templateText at index ix for quoted string containing a time format string, and
+    // use that string to format dateTime. Append the formatted time to path.
+    void appendFormattedTime(Glib::ustring& path, unsigned int& ix, const Glib::ustring& templateText, const Glib::DateTime& dateTime)
+    {
+        constexpr unsigned int quoteMark = (unsigned int)'"';
```
Floessie: This looks wrong. The C-cast is wrong on its own, but `unsigned int` is even more wrong.

rtgui/batchqueue.cc
```diff
@@ -1001,6 +1017,39 @@ Glib::ustring BatchQueue::calcAutoFileNameBase (const Glib::ustring& origFileNam
 
                     seqstr << sequence;
                     path += seqstr.str ();
+                } else if (options.savePathTemplate[ix] == 't') {
+                    // Insert formatted date/time value. Character after 't' defines time source
+                    if (++ix < options.savePathTemplate.size()) {
+                        Glib::DateTime dateTime;
+                        switch(options.savePathTemplate[ix++])
+                        {
+                            case 'E':   // (approximate) time when export started
```
Floessie, line 1026: It's always a good idea to scope `case`es, and if so, to include the `break` in that scope.

rtgui/batchqueuepanel.cc
```diff
@@ -322,6 +337,122 @@ void BatchQueuePanel::setGuiFromBatchState(bool queueRunning, int qsize)
...
+            pathElementCount = n;
+            break;
+        }
+    }
+    // Function inserts examples for a particular specifier, with every valid N value for the
+    // number of elements in the path.
+    auto insertPathExamples = [&buffer, &pos, pathElementCount, exampleFilePath](char letter, int offset1, int mult1, int offset2, int mult2)
+    {
+        for (int n=0; n<pathElementCount; n++) {
+            auto path1 = Glib::ustring::format("%",letter,offset1+n*mult1);
```
Floessie: Space after `,`, please. There are other places in your new code, too, that you should pad a bit more.

rtgui/batchqueuepanel.cc
```diff
@@ -322,6 +337,122 @@ void BatchQueuePanel::setGuiFromBatchState(bool queueRunning, int qsize)
...
+    insertTopicHeading(M("QUEUE_LOCATION_TEMPLATE_HELP_TIMESTAMP_TITLE"));
+    pos = buffer->insert_markup(pos, M("QUEUE_LOCATION_TEMPLATE_HELP_TIMESTAMP_BODY"));
+    Glib::ustring dateTimeFormatExamples[] = {
+        "%Y-%m-%d",
+        "%Y%m%d_%H%M%S",
+        "%y/%b/%-d/"
+    };
+    auto timezone = Glib::DateTime::create_now_local().get_timezone();
+    auto timeForExamples = Glib::DateTime::create_from_iso8601("2001-02-03T04:05:06.123456", timezone);
+    for (auto fmt : dateTimeFormatExamples) {
```
Floessie: I don't think you want to take a copy here.

rtgui/batchqueuepanel.cc
```diff
@@ -322,6 +337,122 @@ void BatchQueuePanel::setGuiFromBatchState(bool queueRunning, int qsize)
...
+
+    insertTopicHeading(M("QUEUE_LOCATION_TEMPLATE_HELP_RANK_TITLE"));
+    pos = buffer->insert_markup(pos, M("QUEUE_LOCATION_TEMPLATE_HELP_RANK_BODY"));
+
+    insertTopicHeading(M("QUEUE_LOCATION_TEMPLATE_HELP_SEQUENCE_TITLE"));
+    pos = buffer->insert_markup(pos, M("QUEUE_LOCATION_TEMPLATE_HELP_SEQUENCE_BODY"));
+
+    insertTopicHeading(M("QUEUE_LOCATION_TEMPLATE_HELP_TIMESTAMP_TITLE"));
+    pos = buffer->insert_markup(pos, M("QUEUE_LOCATION_TEMPLATE_HELP_TIMESTAMP_BODY"));
+    Glib::ustring dateTimeFormatExamples[] = {
```
Floessie: As so many things, this should be `const`.

Floessie (commented):
Some comments on style.

Floessie:
@Lawrence37 Feel free to merge even with my comments unaddressed.

rtgui/batchqueue.cc
sgilbertson: Somehow in about four decades of C programming nobody's suggested that approach, and I've never thought of it. It makes perfect sense in terms of reducing the likelihood of a subsequent error going unnoticed, and moving `break` inside the `{}` is also nicer looking. It's yet another tip I'll be carrying over to my day job. Thanks!

rtgui/batchqueue.cc
sgilbertson: Changing to `gunichar('"')`

rtgui/batchqueuepanel.cc
sgilbertson: I came up with some searches to find a bunch of them to fix. Will do.

sgilbertson:
@Lawrence37 and @Floessie -- all done, including the style comments.

Lawrence37:
Thanks for contributing again!

PR #6952, opened 2024-02-12
Sony ILCE-7CR / ILCE-7RM5 camconst

### LINKED ISSUE #6938: Sony ILCE-7CR PDAF lines pattern
Sample image provided by clockworknapkin on the forum:
https://discuss.pixls.us/t/pdaf-lines-filter-no-effect-on-a7cr/41789/3

Lawrence37:
PDAF pattern, raw crop, and dcraw matrix for the ILCE-7CR and ILCE-7RM5. The PDAF pattern does not perfectly repeat, but is very close. A better sample image would be needed to determine if the pattern is larger. So far, it looks like it does not repeat perfectly every 840 pixels either (the current pattern is 420 pixels). The two or three extra false rows per pattern is unlikely to have a noticeable effect on the image quality.

Closes #6938.

### DIFF
--- a/rtengine/camconst.json
+++ b/rtengine/camconst.json
@@ -3076,7 +3076,17 @@ Camera constants:
     { // Quality C,
         "make_model": "Sony ILCE-7RM4",
         "dcraw_matrix": [ 7662, -2686, -660, -5240, 12965, 2530, -796, 1508, 6167 ],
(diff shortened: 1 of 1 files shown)


Lawrence37:
I confirmed the pattern is not 1260 nor 1680 pixels. This should be good enough until there is a better image.

PR #6962, opened 2024-02-27
Read embedded dcp profile from raw files

sgotti:
raw files, usually dng files, can provide a dcp profile by providing the related tags.

This patch will make the "Use embedded" options in the color management panel selectable if the source raw file embeds a dcp profile. In this case the DCP panel will be enabled like when reading an external dcp profile.

Additional changes:

* Rename "Use embedded, if possible" to just "Use embedded" since the option is already disabled when no embedded profile is found.
* Update the "Use embedded" options tooltip as it's now not related to only non raw images.

## NOTES

* This a port from ART.

* We try to read dcp tags from every raw file, not only dng. Is this ok or should be limited to only some file formats?

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -2634,8 +2634,8 @@ TP_ICM_INPUTCAMERA_TOOLTIP;Use a simple color matrix from dcraw, an enhanced Raw
 TP_ICM_INPUTCUSTOM;Custom
 TP_ICM_INPUTCUSTOM_TOOLTIP;Select your own DCP/ICC color profile file for the camera.
 TP_ICM_INPUTDLGLABEL;Select Input DCP/ICC Profile...
-TP_ICM_INPUTEMBEDDED;Use embedded, if possible
-TP_ICM_INPUTEMBEDDED_TOOLTIP;Use color profile embedded in non-raw files.
+TP_ICM_INPUTEMBEDDED;Use embedded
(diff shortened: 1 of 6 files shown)


Lawrence37:
Works well in my tests. The DCP profile is in the EXIF, right? If so, then it's hard to know in advance which kinds of raw files don't have an embedded DCP.

sgotti:
> Works well in my tests.

Thanks.

> The DCP profile is in the EXIF, right?
> If so, then it's hard to know in advance which kinds of raw files don't have an embedded DCP.

Yes. Let's keep it as is.

kmilos:
> The DCP profile is in the EXIF, right?

It's in the the file metadata, but not part of the Exif specification. DCP is a DNG thing only AFAIK.

sgotti:
> DCP is a DNG thing only AFAIK.

Yes, that's why I asked if we should filter on file type. But after some though what the DCPStore::getProfile method does to check dcp validity is enough. Obviously if someone adds to a tiff like file the dcp tags then they'll probably be used but that could be ok.

Entropy512:
> Obviously if someone adds to a tiff like file the dcp tags then they'll probably be used but that could be ok.

At that point it basically becomes a DNG file with a different extension, as long as the DCP tags are appropriate for the file.  Otherwise it's bad metadata, and there's not really anything we can do if someone adds bogus tags to a file.

Lawrence37 (approved):

PR #6964, opened 2024-02-28
Local adjustments - Chromaticity curve C(h)

Desmis:
Fixed bad behavior.
https://discuss.pixls.us/t/problem-color-local-adjustment-cromaticity-curves/42345

Jacques

### DIFF
--- a/.github/workflows/appimage.yml
+++ b/.github/workflows/appimage.yml
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
 
 env:
-  publish_pre_dev_labels: '["Beep6581:wbrefinement"]'
+  publish_pre_dev_labels: '["Beep6581:lachroma"]'
 
 jobs:
   build:
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
 
 env:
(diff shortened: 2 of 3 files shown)


Desmis:
@Lawrence37 
For the sake of principled rigor, can you validate this PR. The error which must date from more than 5 years ago is just one coefficient too many (2.f)

Thank you

Jacques

Lawrence37:
Why was it multiplied by 2? Could the problem be missing parentheses? I think it makes sense to have the expression be `2 * (x - 0.5)`. If `x` is in the range [0, 1], then the expression evaluates to the range [-1, 1] which corresponds to a chroma adjustment from -100% (complete desaturation) to +100% (double the chroma).

Desmis:
@Lawrence37 
This is simply an error on my part. This code must have been around 9 years old. 
As you said elsewhere, the evaluation of Local Adjustments  (it is only today that someone spotted this anomaly) and its improvements are only at the beginning

Thank you

Jacques

Lawrence37:
I'm thinking there was a reason for multiplying by 2. It makes sense to have an adjustment range from -100% to +100%. In this PR, the range is -50% to +50%. There is no problem with that. I just wonder what the original intent was.

Lawrence37:
@Desmis I'm ok with the current pull request, or you can add the 2x multiplication. Your choice.

PR #6965, opened 2024-02-29
Local adjustments - set method Avoid Gamut to Munsell - better processing of highlights and gamut

Desmis:
See:

https://discuss.pixls.us/t/another-tutorial-color-appearance-truck-under-a-tunnel/41947/17

https://discuss.pixls.us/t/local-adjustments-highlight-recovery/42354/4

Jacques

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -2972,7 +2972,7 @@ LocallabParams::LocallabSpot::LocallabSpot() :
     structexclu(0),
     struc(4.0),
     shapeMethod("IND"),
-    avoidgamutMethod("XYZ"),
+    avoidgamutMethod("MUNS"),
     loc{150, 150, 150, 150},
     centerX(0),
     centerY(0),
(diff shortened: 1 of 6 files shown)


Desmis:
@Lawrence37 
I put 2 links at the start of this PR which reflect the difficulty I had in finding this bug.
A few years ago (more than 5), I put some CLIP functions in some places in the code... There was probably a reason...
Since then the code has changed profoundly and these Clips sometimes lead to bad behavior.

I think that in the majority of cases, modifying as I did either by increasing the value limits or by overriding the function preserves the future.

I also made a modification of ppversion.h and procparams so that the pp3s are "converted"

This PR, in summary:
* set default value (new spot) to “Munsell only”
* change behavior in “Local Adjustments” when the algorithm finds a “clip” function (Luminance, chroma…)
* mark in pp3 if the value is set (5.10) to “XYZ absolute” and switch to “Munsell only”

What bothers me is that 5.10 is not the right choice? If you release version 5.10.1, maybe it should be included?

Thank you

Jacques

rtengine/procparams.cc
```diff
@@ -8742,6 +8742,11 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
                 assignFromKeyfile(keyFile, "Locallab", "ShapeMethod_" + index_str, spot.shapeMethod, spotEdited.shapeMethod);
                 if (keyFile.has_key("Locallab", "AvoidgamutMethod_" + index_str)) {
                     assignFromKeyfile(keyFile, "Locallab", "AvoidgamutMethod_" + index_str, spot.avoidgamutMethod, spotEdited.avoidgamutMethod);
+                    if (ppVersion < 351) {
+                       if(spot.avoidgamutMethod == "XYZ") {//5.10 default value
+                           spot.avoidgamutMethod = "MUNS";//set to Munsell only
+                       }
+                    }
```
Lawrence37: XYZ is a valid choice, so I think the method should not be changed.

Desmis: Excuse my bad english but I don't understand...

Yes "XYZ" (XYZ absolute) - it's now 5.10 default -  is a valid choice, but in some cases it leads to artifacts.

Rearranging the "clips" and choosing "Munsell only" is I think in 95% of cases a better (the best?) solution.

Jacques

Lawrence37: Even if XYZ causes artifacts in some cases, it is not a good idea for us to change parameters of existing edits. Those edits have been made while taking into account the effect of the avoid color shift method. Some people may have explicitly set the method to XYZ. The only time we should change edits is if it is a bug-fix.

Desmis: OK 

Jacques

epadepa:
It seems this PR changes the behaviour of local edit log encoding module. 5.10 release handles Sweep_sRGB_Linear_Half (as referenced on pixls.us) very well but this PR creates very surprising and colourful results. I may not be in control of all settings but I can't get remotely reasonable results. Do have a look and see if it's expected.

epadepa:
Recent build 5.10-360-gc535015ad now solves the issue I mention above.

Lawrence37 (approved):

PR #6976, opened 2024-03-07
Windows/innosetup: comment out deleted font

Benitoite:
bf988ad2744865f151611a2761a91c9d90f7ba83 removes the default font in favor of the OS default. `tools/win/InnoSetup/WindowsInnoSetup.iss.in` Line 122 causes the win build fail trying to load the deleted font.


Lawrence37:
Thanks @Benitoite! Since the installer is not built for pull requests, I'll merge this and check the dev build for any lingering issues.

PR #6977, opened 2024-03-08
denoise gui: set luminance detail adjuster default as procparams default

sgotti:
The rtgui denoise luminance detail adjuster widget default value is set to 50, but the procparams default is 0 causing some confusing behavior: when enabling the denoise tool in the gui the luminance detail value is set to the proc params default value (0), when resetting the adjuster the values is set to 50.

This patch sets the adjuster default value to the procparams default value like the other adjusters.


Lawrence37 (approved):
Thanks @sgotti!

PR #6543, opened 2022-08-08
Add Canon EOS R3, R7 and R10 support

blastrock:
I've added support for these camera in RawTherapee. This is all taken from LibRaw.

I don't know what to set for raw_crop, masked_areas and ranges, I couldn't find these values in LibRaw.

For some reason, EXIF data isn't extracted by RawTherapee.

PR #6983, opened 2024-03-10
Read camera white balance for Canon EOS R6m2, R7, and more

Lawrence37:
PR #6543 added support for these cameras, but the code for reading the camera white balance multipliers was incorrectly ported from LibRaw. It was reading the wrong tag, the daylight white balance instead of the as-shot white balance. Here's the LibRaw code where the tags were mixed up:
https://github.com/LibRaw/LibRaw/blob/a4c9b1981ee4ac2a144e7a290988428cc5bb7e85/src/metadata/canon.cpp#L1253-L1259

Test images can be found in [this comment on the forum](https://discuss.pixls.us/t/questions-how-to-choose-a-dcp-profile-if-at-all/42404/46).

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -6124,7 +6124,7 @@ get2_256:
       // imCanon.ColorDataVer = 11;
       imCanon.ColorDataSubVer = get2();
 
-      fseek(ifp, save1 + ((0x0069+0x0064) << 1), SEEK_SET);
+      fseek(ifp, save1 + (0x0069 << 1), SEEK_SET);
       FORC4 cam_mul[c ^ (c >> 1)] = (float)get2();
 
       offsetChannelBlackLevel2 = save1 + ((0x0069+0x0102) << 1);


alice-mm:
🥳

There’s still something I don’t understand: Why do I get different white balance values for R6m2 sample shots from other sources than I do for mine?

* The vegetable plate from https://discuss.pixls.us/t/canon-eos-r6-ii/39647 gives me 5295 temp 1.041 tint.
* The `Canon - Canon EOS R6 Mark II - 3 2 dual pixel C-raw full frame.cr3` street view from [raw.pixls.us](https://raw.pixls.us/) gives 5481 temp 1.053 tint.
* The `5M9A0093.CR3` test shot [from DPReview](https://www.dpreview.com/reviews/image-comparison/download-image?s3Key=9dc4402898fe4882950f3b1ec600daa0.cr3) gives 5340 temp 1.049 tint.
* I keep getting 5257 temp 1.043 tint for _all of my own shots_.

My limited knowledge only allows me to formulate two hypotheses, none of which really convinces me (one even got virtually killed by the further checks I did while writing it):

* Canon introduced a breaking change in the R6m2’s firmware, shifting metadata tags around, at some point between the firmware version used by the examples cited above and mine (1.1.4).
  * → Sure, DPReview used 1.0.0, but the vegetable plate uses the same version as I do, so…
* Each of these people’s R6m2s have a different notion of “daylight white balance”.
  * → Can that even be customized or changed in any way? While I don’t quite see what the point of this would be, the fact that the daylight temperature (along with other temperatures) gets written to the metadata gives the impression that it _can_ indeed change.

Lawrence37:
I'm not sure why they are different, but the metadata does show that the daylight levels have different values. Who knows, besides Canon itself, how Canon decides what the daylight levels are. Maybe it is influenced by the JPEG color settings or changes with the ISO.

alice-mm:
> I'm not sure why they are different, but the metadata does show that the daylight levels have different values.

OK 🤷🏻‍♀️ At least that part is confirmed.

> Maybe it is influenced by the JPEG color settings or changes with the ISO.

I can at least guarantee that ISO is not enough to make that change: I kept it on auto, it went from 100 to 20000, and same RT balance. xD Unless it’s the “Auto ISO” field, which is stuck at 100 on my pics. Not sure what it represents.

As for JPEG color settings, not sure what they are either, but that’s possible, though I don’t see why the DPReview guys would use non-default values for this during their test shot. I personally don’t recall having changed anything like that.

Well, that doesn’t change anything regarding the fix, though, I suppose. Just curious. 😆

PR #6988, opened 2024-03-13
Add optional image rank/color load/save from/to xmp sidecar

### LINKED ISSUE #7036: XMP Metadata Synchronization not Writing XMP Sidecars
**Short description**

XMP metadata synchronization is not writing to XMP sidecars when settings are changed.

**Steps to reproduce**
1. Go to settings > Image Processing
2. Metadata Synchronization with XMP Sidecars: Bidirectional
3. XMP-Sidecar style: darktable-like
4. Go to file browser
5. Apply a "star" rating to an image
6. Look in the folder where the image is stored. Notice how a .pp3 file has been written, but a .xmp has not

**Expected behavior**

An XMP sidecar file has been written containing the `xmp:Rating` tag.

**Additional information**
- Version of RawTherapee: 5.10
- Version of your operating system: Void Linux - `Void 6.6.25_1 x86_64 GenuineIntel uptodate rrrdFFFFFFFFFFF`
- `AboutThisBuild.txt`
```
Version: 5.10
Branch: 5.10
Commit: 8adb5439d
Commit date: 2024-02-15
Compiler: cc 13.2.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.8
Lensfun: V0.3.4.0
Build type: None
Build flags: -DNDEBUG -fstack-clash-protection -D_FORTIFY_SOURCE=2 -mtune=generic -O2  -O3 -g -ffile-prefix-map=/builddir/rawtherapee-5.10/build=. -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas  -ftree-vectorize
Link flags: -Wl,-z,relro -Wl,-z,now -Wl,--as-needed
OpenMP support: ON
MMAP support: ON
Build OS: 
Build date:  UTC
Build epoch: 
Build UUID: 
```

image

sgotti:
## TLDR

Add optional image rank/color load/save from/to xmp sidecar

Why only rank and color and not trash?
* Rank is mappable to `xmp.Rating` (but no support for -1 rating)
* Colors can be mapped to `xmp.Label` with some downsides due to a non standardized use of a label string.
* trash doesn't clearly map to -1 rating. If we really want to do this there's the need to change the current rawtherapee rank/trash logic.

**A detailed explanation is in the NOTES below.**

* Also color could be confusing (see NOTES below) so I'm open to just keep the rating if preferred.

* If this will be merged I'll add related docs to rawpedia for integrating this with other tools like digikam.

* Should Fix different opened issues (need to reference them here....)

## Description

These are 3 (self contained) patches. They're split to better explain the various changes, so I suggest to look at them one by one.

Add optional ability to load/save image rank property from/to xmp sidecar "xmp.Rating" and color property from xmp "xmp.Label" ignoring the ones provided in the processing params file.

This behavior is disabled by default and an option under settings -> file browser has been added to enable it.

When enabled:

* On load:
  * rank and color are not read from processing params.
  * rank is mapped from xmp sidecar file rating entry.
  * color is mapped from xmp sidecar file label entry.

* On save:
  * rank and color are saved to the xmp sidecar
  * rank and color are also saved to the processing param (pp3) files to keep them in sync

Rating mapping:

Since rating can be also -1 but rank only goes from 0 to 5, the -1 value is ignored like already done when importing from embedded xmp data.

Color mapping:

XMP has no color concept, usually programs like digikam uses the label field to write a color name ("Red", "Orange"). The problem is that this isn't standardized and label can be any string. Additionally Rawtherapee has 5 specific colors while other programs can have different colors with different name so they won't be shown if they don't map to the 5
color names supported by rawtherapee. On save only the 5 color supported by rawtherapee wil be saved.

## NOTES

* The main use case is to do culling by rating using another program like digikam, but be able to see and change the same rating also in rawtherapee. So a bidirectional sync. This can be accomplished by **correctly** configuring the "culling" program to read/write the rating to an xmp sidecar file (choosing one of the two xmp extension modes supported by rawtherapee).

* This is something already implemented in ART but with also trash mapped to xmp rating -1.

## Trash mapping issues

When testing the ART implementation, the mapping of trash to rating = -1 caused some bad/confusing behavior.
So I felt that the better way was to map only the rating and colors (also with some downsides) and leave trash in the processing parameters.

Here some reason of this choice:

* There's no trash concept in xmp, there's the rejected concept assigned
  to a rating == -1.
* We could map rejected to trash but in rawtherapee rank and trash are two different values and an image can have both rank >= 0 and trashed set to true.
  Using an unique value like rating for rank and trash (when -1) will require changing the current rawtherapee logic.
* Also digikam only handles ratings from 0 to 5 (no -1) and handles trash in its own internal way without reflecting it in the xmp sidecar.

## Color mapping issues

* XMP has no color concept, usually programs like digikam uses the xmp label field to write a color name ("Red", "Orange"). The problem is that this isn't standardized and label can be any string. Additionally Rawtherapee has 5 specific colors while other programs can have different colors with different name so they won't be shown if they don't map to the 5 color names supported by rawtherapee. 
* Every app has custom support for color and uses additional fields. This patch only uses the xmp.Label field.
  * As an example "digikam" by default uses an additional custom field (`digiKam.ColorLabel`) to save color labels. This has the priority over the xmp label field. So if the xmp has already the `digiKam.ColorLabel` (because you set the color with digikam), the changes to colors done by rawtherapee to the xmp.Label fields are ignored. For a working two way sync Digikam must be configured to prefer the xml.Label field.

### DIFF
--- a/rtgui/thumbnail.cc
+++ b/rtgui/thumbnail.cc
@@ -133,7 +133,7 @@ Thumbnail::Thumbnail(CacheManager* cm, const Glib::ustring& fname, CacheImageDat
     generateExifDateTimeStrings ();
 
     if (cfs.rankOld >= 0) {
-        // rank and inTrash were found in cache (old style), move them over to pparams
+        // rank and inTrash were found in cache (old style), move them over to pparams or xmp sidecar
 
         // try to load the last saved parameters from the cache or from the paramfile file
         createProcParamsForUpdate(false, false); // this can execute customprofilebuilder to generate param file
@@ -143,6 +143,8 @@ Thumbnail::Thumbnail(CacheManager* cm, const Glib::ustring& fname, CacheImageDat
         setRank(cfs.rankOld);
         setTrashed(cfs.inTrashOld);
     }
+
+    loadProperties();
 
     delete tpp;
     tpp = nullptr;
@@ -172,6 +174,8 @@ Thumbnail::Thumbnail(CacheManager* cm, const Glib::ustring& fname, const std::st
     cfs.recentlySaved = false;
 
     initial_ = false;
+
+    loadProperties();
 
     delete tpp;
     tpp = nullptr;
@@ -306,7 +310,7 @@ const ProcParams& Thumbnail::getProcParamsU ()
  *  @param returnParams Ask to return a pointer to a ProcParams object if true
  *  @param force True if the profile has to be re-generated even if it already exists
  *  @param flaggingMode True if the ProcParams will be created because the file browser is being flagging an image
- *                      (rang, to trash, color labels). This parameter is passed to the CPB.
+ *                      (rank, to trash, color labels). This parameter is passed to the CPB.
  *
  *  @return Return a pointer to a ProcPamas structure to be updated if returnParams is true and if everything went fine, NULL otherwise.
  */
@@ -441,12 +445,6 @@ void Thumbnail::clearProcParams (int whoClearedIt)
     {
         MyMutex::MyLock lock(mutex);
 
-        // preserve rank, colorlabel and inTrash across clear
-        int rank = getRank();
-        int colorlabel = getColorLabel();
-        int inTrash = getTrashed();
-
-
         cfs.recentlySaved = false;
         pparamsValid = false;
 
@@ -456,13 +454,10 @@ void Thumbnail::clearProcParams (int whoClearedIt)
         // reset the params to defaults
         pparams->setDefaults();
 
-        // and restore rank and inTrash
-        setRank(rank);
-        pparamsValid = cfs.rating != rank;
-        setColorLabel(colorlabel);
-        setTrashed(inTrash);
+        // preserve rank, colorlabel and inTrash across clear
+        updateProcParamsProperties();
 
-        // params could get validated by rank/inTrash values restored above
+        // params could get validated by updateProcParamsProperties
         if (pparamsValid) {
             updateCache();
         } else {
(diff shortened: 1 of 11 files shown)


rtgui/thumbnail.cc
```diff
@@ -105,6 +105,47 @@ bool CPBDump(
...
+    std::map<int, std::string> indexLabelMap;
+    std::map<std::string, int> labelIndexMap;
+
+    std::map<int, std::string>::iterator it;
+    ColorMapper(std::map<int, std::string> colors) {
+        for (it = colors.begin(); it != colors.end(); it++) {
+            indexLabelMap.insert({it->first, it->second});
+            labelIndexMap.insert({it->second, it->first});
+        }
+    }
```
Lawrence37: Is the iterator supposed to be a local variable instead? Can the for loop be a range-based for loop?

rtgui/thumbnail.cc
```diff
@@ -105,6 +105,47 @@ bool CPBDump(
...
+
+    int index(const std::string &label)
```
Lawrence37: This method can be `const`.

rtgui/thumbnail.cc
```diff
@@ -105,6 +105,47 @@ bool CPBDump(
...
+    {
+        auto it = labelIndexMap.find(label);
+        if (it != labelIndexMap.end()) {
+            return it->second;
+        }
+        return 0;
+    }
+
+    std::string label(int index)
```
Lawrence37: This method can also be `const`.

Lawrence37:
Looks good. I did some brief testing and everything works as expected. I just have some trouble getting digiKam to read the rating set in RawTherapee. The color label works after reordering the tag priority. RawTherapee is also able to read the rating if updated by digiKam. It's just the other way around that I can't figure out, so I look forward to the documentation on digiKam integration.

Lawrence37 (approved):

PR #6990, opened 2024-03-14
camconst.json added Nikon Z f

lightwisephoto:
According to Wikipedia, Nikon Z f uses the same sensor as Nikon Z 6 II.

Until further improvement with new measurements, we can reuse the existing Z 6 II constants.

Tested with a RAW image from the Z f and looks suitable.

### DIFF
--- a/rtengine/camconst.json
+++ b/rtengine/camconst.json
@@ -2032,7 +2032,7 @@ Camera constants:
     },
 
     { // Quality C
-        "make_model" : "Nikon Z 6_2",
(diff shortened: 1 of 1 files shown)


Lawrence37:
I extracted the color matrix from Adobe DNG converter for the Z 6_2 and Z f and they are different. The Z 6_2 matrix matches what is in `camconst.json`, but the Z f matrix is `[ 11607, -4491, -977, -4522, 12460, 2304, -458, 1519, 7616 ]`.

lightwisephoto:
Even better. Thank you.

PR #6991, opened 2024-03-14
Handle flat field images with different black levels than raw image

### LINKED ISSUE #6986: Flat field is not applied
**Short description**
Applying a flat field correction to some RAW files has no effect.

**Steps to reproduce**
1 Open supplied RAW-File
2 Ensure no lens profile is applied
3 Apply supplied flat field
4 Observe vignetting correction: nothing happens

**Expected behavior**
The flat field is applied and the vignetting is corrected.

**Additional information**
- RawTherapee 5.10 on Windows 11
- https://discuss.pixls.us/t/bizarre-flat-field-effect-rawtherapee-5-9-5-8/33992/5 I think it is maybe this bug

[flat.zip](https://github.com/Beep6581/RawTherapee/files/14564366/flat.zip)
[raw_file.zip](https://github.com/Beep6581/RawTherapee/files/14564369/raw_file.zip)

sgotti:
For bayer images handle flat field images with different black levels than the raw image.

## NOTES

* This is a port from ART.

* Looks likes this is implemented only for bayer CFA or monochrome images, not sure if something similar should be done also for xtrans CFA

Fixes #6986

### DIFF
--- a/rtengine/rawflatfield.cc
+++ b/rtengine/rawflatfield.cc
@@ -263,12 +263,20 @@ void cfaboxblur(const float* const * riFlatFile, float* cfablur, int boxH, int b
 namespace rtengine
 {
 
-void RawImageSource::processFlatField(const procparams::RAWParams &raw, const RawImage *riFlatFile, array2D<float> &rawData, const float black[4])
+void RawImageSource::processFlatField(const procparams::RAWParams &raw, RawImage *riFlatFile, array2D<float> &rawData, const float black[4])
 {
 //    BENCHFUN
     std::unique_ptr<float[]> cfablur(new float[H * W]);
 
     const int BS = raw.ff_BlurRadius + (raw.ff_BlurRadius & 1);
+
+    std::array<float, 4> ffblack;
+    {
+        const auto tmpfilters = riFlatFile->get_filters();
+        riFlatFile->set_filters(riFlatFile->prefilters); // we need 4 blacks for bayer processing
+        riFlatFile->get_colorsCoeff(nullptr, nullptr, ffblack.data(), false);
+        riFlatFile->set_filters(tmpfilters);
+    }
 
     if (raw.ff_BlurType == procparams::RAWParams::getFlatFieldBlurTypeString(procparams::RAWParams::FlatFieldBlurType::V)) {
         cfaboxblur(riFlatFile->data, cfablur.get(), 2 * BS, 0, H, W);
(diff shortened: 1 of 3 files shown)


timsurber:
I checked out your fix and it works for me :+1:

Benitoite:
Application of flats also fixed in macOS with this.

rtengine/rawflatfield.cc
```diff
@@ -263,13 +263,21 @@ void cfaboxblur(const float* const * riFlatFile, float* cfablur, int boxH, int b
...
 
-void RawImageSource::processFlatField(const procparams::RAWParams &raw, const RawImage *riFlatFile, array2D<float> &rawData, const float black[4])
+void RawImageSource::processFlatField(const procparams::RAWParams &raw, RawImage *riFlatFile, array2D<float> &rawData, const float black[4])
 {
 //    BENCHFUN
     std::unique_ptr<float[]> cfablur(new float[H * W]);
 
     const int BS = raw.ff_BlurRadius + (raw.ff_BlurRadius & 1);
 
+    float ffblack[4];
```
Lawrence37: Replace this C-style array with `std::array`

rtengine/rawflatfield.cc
```diff
@@ -263,13 +263,21 @@ void cfaboxblur(const float* const * riFlatFile, float* cfablur, int boxH, int b
...
+    {
+        auto tmpfilters = riFlatFile->get_filters();
```
Lawrence37: Make `tmpfilters` `const`

rtengine/rawflatfield.cc
```diff
@@ -363,6 +371,9 @@ void RawImageSource::processFlatField(const procparams::RAWParams &raw, const Ra
         const vfloat blackv[2] = {_mm_set_ps(black[c4[0][1]], black[c4[0][0]], black[c4[0][1]], black[c4[0][0]]),
                                   _mm_set_ps(black[c4[1][1]], black[c4[1][0]], black[c4[1][1]], black[c4[1][0]])
                                  };
+        const vfloat ffblackv[2] = {_mm_set_ps(ffblack[c4[0][1]], ffblack[c4[0][0]], ffblack[c4[0][1]], ffblack[c4[0][0]]),
+                                    _mm_set_ps(ffblack[c4[1][1]], ffblack[c4[1][0]], ffblack[c4[1][1]], ffblack[c4[1][0]])
+                                   };
```
Lawrence37, line 376: There's another `blackv` declaration further down in the section specifically for the vertical + horizontal blur type. Doesn't that also need a `ffblackv`?

Lawrence37 (commented):
I verified this does not negatively impact the application of flat-fields which have worked before.

rtengine/rawflatfield.cc
sgotti: FWIW This is the same as the `const float black[4]` passed to this function and the same code done also by the caller https://github.com/Beep6581/RawTherapee/blob/7751019b6e03e300387563e85dcb3c49731fd503/rtengine/rawimagesource.cc#L2594-L2600

Anyway I updated it to be a std::array

rtengine/rawflatfield.cc
sgotti: I tried to do the same changes also for such case (only bayer). But I'm not sure how to test it.

Lawrence37: Before, vertical + horizontal produced nearly the same result as the area blur type. Now, there is a slight difference with similar magnitude as switching between the other blur types. The code looks right and the results are acceptable, so I believe everything is good.

Lawrence37 (approved):

PR #6992, opened 2024-03-14
Fix 'Hicolor' icon theme conflict when installing on Linux

### LINKED ISSUE #6980: RawTherapee git: conflict file on Arch system
Conditions:
Arch system
Compiled sources
hicolor-icon-theme 0.17-3 package https://archlinux.org/packages/extra/any/hicolor-icon-theme/
Due to conflict file with hicolor-icon-theme, can't install/upgrade RawTherapee git since recent commit
May be  this commit ? https://github.com/Beep6581/RawTherapee/commit/7f45af1182f77e5ff731a848e22f9788d0492cde
When I try to install:
```
error: failed to commit transaction (conflicting files)
rawtherapee-git: /usr/share/icons/hicolor/index.theme exists in filesystem (owned by hicolor-icon-theme)
Errors occurred, no packages were upgraded.

```

Pandagrapher:
Fix for issue #6980 

For `hicolor` icon theme (RT app icon), `index.theme` file is now only considered during installation if RawTherapee is built with `BUILD_BUNDLE` option at `OFF`

Fixes also RawTherapee icon not showing in app bar (example on linux gnome desktop):
image

Tested on:
- [X] MacOS (`BUILD_BUNDLE` option always at `ON`)
- [X] Linux (build install i.e. `BUILD_BUNDLE` option at `OFF`)
- [X] Linux (Appimage i.e. `BUILD_BUNDLE` option at `ON`)
- [X] Windows (`hicolor` icon theme not used nor installed)


Lawrence37:
I want to see if there is a simpler, more elegant solution, so I read https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html again. It says
> In order to have a place for third party applications to install their icons there should always exist a theme called "hicolor" [[1]](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#ftn.idm45507905808496). The data for the hicolor theme is available for download at: http://www.freedesktop.org/software/icon-theme/. Implementations are required to look in the "hicolor" theme if an icon was not found in the current theme.

Since the hicolor theme is always available, I think it is ok to leave the icons in hicolor. The only change is to remove the index.theme which should exist already in the system. Have you tried this?

Pandagrapher:
> I want to see if there is a simpler, more elegant solution, so I read https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html again. It says
> 
> > In order to have a place for third party applications to install their icons there should always exist a theme called "hicolor" [[1]](https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#ftn.idm45507905808496). The data for the hicolor theme is available for download at: http://www.freedesktop.org/software/icon-theme/. Implementations are required to look in the "hicolor" theme if an icon was not found in the current theme.
> 
> Since the hicolor theme is always available, I think it is ok to leave the icons in hicolor. The only change is to remove the index.theme which should exist already in the system. Have you tried this?

Tried and it works: commit has been updated accordingly :)

Lawrence37:
I tested the system-wide installation on Ubuntu 22.04 and Arch Linux (Docker images) and they install and run fine. Thanks @Pandagrapher!

PR #6997, opened 2024-03-17
Fix regression with demosaiced DNGs caused by 831a9bbd

### LINKED ISSUE #6996: Demosaiced DNGs fail to load since commit 831a9bbd
I have some demosaiced (aka Linear RAW) DNGs created by a rather oddball camera (I don't own it, but have put some effort into reverse engineering/supporting it).

One example is at https://filebin.net/a8zb1jja5rwb8dk5

These cause the asserts from 831a9bbd to fail because rawdata.getWidth() returns a width that is three times the image width since it has three samples per pixel

Is the value of DNG samplesPerPixel readily available where these asserts are?

Alternatively, maybe asserting (rawData.getWidth() % W) == 0 would work here, under the assumption that getWidth() won't ever be exactly zero if getHeight is nonzero?

Entropy512:
Demosaiced DNGs (aka LinearRAW PhotometricInterpretation) have 3 samples per pixel, so rawData.getWidth() returns triple the width Only throw an assert if getWidth is not a multiple of W.  This new implementation assumes that getWidth() will never be zero when getHeight() is nonzero

Fixes #6996

### DIFF
--- a/rtengine/rawimagesource.cc
+++ b/rtengine/rawimagesource.cc
@@ -742,9 +742,19 @@ void RawImageSource::getWBMults(const ColorTemp &ctemp, const RAWParams &raw, st
     autoGainComp = camInitialGain / initialGain;
 }
 
+bool checkRawDataDimensions(const array2D<float> &rawData, const RawImage &rawImage, int width, int height)
+{
+    const int colors = (rawImage.getSensorType() == rtengine::ST_BAYER ||
+                           rawImage.getSensorType() == rtengine::ST_FUJI_XTRANS ||
+                           rawImage.get_colors() == 1)
+                           ? 1
+                           : 3;
+    return rawData.getHeight() == height && rawData.getWidth() == colors * width;
+}
+
 void RawImageSource::getImage(const ColorTemp &ctemp, int tran, Imagefloat* image, const PreviewProps &pp, const ToneCurveParams &hrp, const RAWParams &raw)
 {
-    assert(rawData.getHeight() == H && rawData.getWidth() == W);
+    assert(checkRawDataDimensions(rawData, *ri, W, H));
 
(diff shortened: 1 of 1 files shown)


Lawrence37:
It's probably best to refactor these checks (including the first if-statement in `getRawValues`) to use a function that checks the validity of the dimensions.
```c++
bool checkRawDataDimensions(const array2D<float> &rawData, const RawImage &rawImage, int width, int height)
{
    const int colors = (rawImage.getSensorType() == ST_BAYER ||
                           rawImage.getSensorType() == ST_FUJI_XTRANS ||
                           rawImage.get_colors() == 1)
                           ? 1
                           : 3;
    return rawData.getHeight() == height && rawData.getWidth() == colors * width;
}
```

Entropy512:
Great! Thanks for the suggestions including where to get the equivalent of SamplesPerPixel.  I've had a few too many Tasty Beverages for tonight, I'll work on this refactor tomorrow.

rtengine/rawimagesource.cc
```diff
@@ -742,9 +742,19 @@ void RawImageSource::getWBMults(const ColorTemp &ctemp, const RAWParams &raw, st
...
+bool checkRawDataDimensions(const array2D<float> &rawData, const RawImage &rawImage, int width, int height)
+{
+    const int colors = (rawImage.getSensorType() == rtengine::ST_BAYER ||
+                           rawImage.getSensorType() == rtengine::ST_FUJI_XTRANS ||
+                           rawImage.get_colors() == 1)
+                           ? 1
+                           : 3;
+    return rawData.getHeight() == height && rawData.getWidth() == colors * width;
+}
+
```
Lawrence37: Better to put this in the anonymous namespace at the top of the file.

PR #7007, opened 2024-03-24
Fix Windows automated builds (Adwaita cursors)

Lawrence37:
This is an attempt to fix the Windows builds which stopped working after the upgrade to the latest Adwaita icon package. Some cursors were removed/renamed and all the ones that are (supposedly) required, as indicated in RawPedia, are affected. I simply removed the cursors completely and RawTherapee seems to run just fine. The caveat is that I tested through Wine, so I'd like confirmation on an actual Windows environment. At least the build succeeds, and a working pipeline with possibly bad artifacts is better than nothing.


Desmis:
@Lawrence37 

I just tested your change in "lacam16n"...

Thank you

Jacques

Issue #4129, opened 2017-10-03
RAW files and DNGs obtained from them give different results at identical settings.


daicehawk:
RawTherapee 5.3 ec0f7936 (older 4.2.1148 shows the same behaviour) downloaded from http://rawtherapee.com/downloads.
Win 7 professional x64, CPU intel pentium n 3700 1.6 GHz, 8 Gb RAM.
ORF file from Olympus E-500 and DNG file converted in Adobe DNG Converter 9.9.0.718:
https://drive.google.com/open?id=0B_gJzl0-HJ6RaFozbHFTRGNtaDQ
https://drive.google.com/open?id=0B_gJzl0-HJ6RZ1RXdldaR0haSzg
Even the Neutral profile already looks different for the ORF and DNG, then the Auto mode calculates different Exposure Compensation and Black\Shadow compression values. 
This difference also is observed with ARW Sony ILCA-68  and DNG files obtained from it. Cannot provide an ARW, but here the jpegs developed from ARW and DNG with the same settings and the same matrix-only DCP color profile:
 https://drive.google.com/open?id=0B_gJzl0-HJ6Rc2hvVDFiYjBCZEU
https://drive.google.com/open?id=0B_gJzl0-HJ6RZUJ6MFZOQlRXSkU
Considering DCP profiles are made in Adobe DNG profile Editor (and tweaked by eye) from DNGs which are supposed to work as good with native RAWs in RawTherapee I see this as a major bug.

daicehawk:
Under the same conditions as above, profile parameters cannot be copy-pasted between two instances of RT 5.3.

Benitoite:
Are there .pp3 files associated with the processing parameters?

daicehawk:
Should I share a one source Auto obtained either from RAW or DNG and copy-pasted to the other format to compare or two PP3 after Auto mode enabled for each format? 
Basically, as I posted, even the Neutral profile (default As Shot WB, camera profile and zeroed in sliders, i.e. no PP3 needed AFAIU) gives different results in both RT view and in render.

Benitoite:
@daicehawk Any generated .pp3's with matching associated input (RAW/DNG) and output (RT-generated jpg) would be good.

Could this be any relation? https://github.com/Beep6581/RawTherapee/issues/2809

daicehawk:
[...]

Issue #6498, opened 2022-06-09
Leic M8 DNGs not working in RT dev version 5.8-3095


Reiner1959:
Hi,

just discovered that my old Leica M8 DNGs not working anymore. Pictures look way overexposed.
Standard RT 5.8 is still working, darktable as well. 

best regards
Reiner

Here is the link to the file
https://filebin.net/0etmiu7dtibd0ni5

Here is the screenshot
![grafik](https://user-images.githubusercontent.com/68758094/172835527-44c22fb0-1254-4747-9b20-abd2a723bdb1.png)

Reiner1959:
I have just read this thread:
Leica M8 sample interpreted with wrong bit depth #6237 

In the discussion i have read that removing the Leica M8 entry in camconst.json could fix the issue.

I commented out the Leica M8 entry in camconst.json and it seems to work now.
The question is: Do we really need this entry, or must the fix be done somewhere else?

best regards 
Reiner

Lawrence37:
@Reiner1959 It turns out the fix (actually a workaround) for #6237 wasn't complete. I have removed the camconst entry from `dev`.

---

The DNGs report a bit depth of 8 and that causes the white level to be capped at 255. Maybe there should be an exception for these DNGs.
https://github.com/Beep6581/RawTherapee/blob/9bde8f18dc3ae7a49445c354cb4ab2d2ce32a158/rtengine/dcraw.cc#L9092-L9099

There's also some suspicious code here. It doesn't seem to have any effect on the black and white levels except when the white level exceeds the bit depth.
https://github.com/Beep6581/RawTherapee/blob/9bde8f18dc3ae7a49445c354cb4ab2d2ce32a158/rtengine/rawimage.cc#L1001-L1006

Issue #6237, opened 2021-05-13
Leica M8 sample interpreted with wrong bit depth


Thanatomanic:
When preparing https://github.com/Beep6581/RawTherapee/commit/6c96bf7507d631284c4c71134f12e09e3dbe9090 I checked this https://raw.pixls.us/getfile.php/2988/nice/Leica%20-%20M8%20-%208bit%20uncompressed%20(3:2).DNG sample and it seems like the file is being interpreted with a wrong bit-depth.

![image](https://user-images.githubusercontent.com/6567747/118166344-a762ed00-b425-11eb-8d83-45dd4d312dcc.png)

SimonSegerblomRex:
I'm probably missing something, but why did you introduce this regression in 6c96bf7 @Thanatomanic? Removing the Leica M8 entry (which looks like settings for a 14 bit mode) from `camconst.json` fixes this issue.

Thanatomanic:
Well, verbose mode tells me that `dcraw` provides a white level of 16383, when I load the raw data in my own analysis script, the values lie between 2 and 16256. So I wouldn't assume that setting a white level of 16383 in `camconst.json` would hurt, but apparently it does.
Judging from the regular gaps in the histogram, there seems to be 8-bit data in the file that gets expanded to 14-bit. Weird behavior to say the least...

Thanatomanic:
@SimonSegerblomRex I've pushed a fix https://github.com/Beep6581/RawTherapee/commit/7e36240f3af05bdb17e33db135c0e0ea0bef0bb3 
But there is something fishy going on with what `dcraw` does, what it reports and how the `camconst.json` override that seems okay actually screws things up. Something to figure out at some point...

PR #7016, opened 2024-03-29
dcraw: prefer raw crop and mask from dng metadata

sgotti:
Prefer raw crop and mask from dng metadata instead of camconst.json entries.

The unique exception, to keep backward compatibility, is from in camera Pentax DNGs. In such case, the preference goes to a camconst entry, if it exists.

This is based on discussion (see last comments) in #6826 

It uses the same logic already implemented for RT_whitelevel_from_constant and RT_blacklevel_from_constant

To sum up, from my understanding of the code and after reading some old discussions/issues:

* If black level, white level, and with this PR also raw crop and mask are provided by a DNG file with the exception of Pentax in camera DNG the DNG metadata values are preferred over camconst.
* The ColorMatrix1/2 entries from ADC generated DNG are ignored to keep constant whitebalancing between them and the source raw file (#4129)

## How to test
To test this PR use, for example, a Fuji XT3 file and convert it to dng with ADC. Change camconst to remove the source size filter from raw_crop:

```
--- a/rtengine/camconst.json
+++ b/rtengine/camconst.json
@@ -1596,10 +1596,7 @@ Camera constants:
     { // Quality A, samples provided by Daniel Catalina (#5839) and pi99y (#5860)
         "make_model": [ "FUJIFILM X-T3", "FUJIFILM X-PRO3" ],
         "dcraw_matrix": [ 13426,-6334,-1177,-4244,12136,2371,-580,1303,5980 ], // DNG_v11, standard_v2 d65
-        "raw_crop" : [
-           { "frame" : [6384, 4182], "crop": [ 0, 5, 6252, 4176] },
-           { "frame" : [6384, 3348], "crop": [624, 0, 5004, 3348] }
-        ],
+        "raw_crop": [ 0, 5, 6252, 4176],
         "white": [ 16170, 16275, 16170 ] // typical safe-margins with LENR
         // negligible aperture scaling effect
     },
```

Without this PR crop data is taken from camconst and the file is wrongly demosaiced, with this PR the crop data is taken from DNG metadata and correctly demosaiced.

@Lawrence37 @kmilos @Entropy512

### DIFF
--- a/rtengine/dcraw.cc
+++ b/rtengine/dcraw.cc
@@ -6931,11 +6931,13 @@ it under the terms of the one of two licenses as you choose:
 	left_margin = getint(type);
 	height = getint(type) - top_margin;
 	width = getint(type) - left_margin;
+	RT_raw_crop_mask_from_constant = ThreeValBool::F;
 	break;
       case 50830:			/* MaskedAreas */
         for (i=0; i < len && i < 32; i++)
 	  ((int *)mask)[i] = getint(type);
 	black = 0;
+	RT_raw_crop_mask_from_constant = ThreeValBool::F;
 	break;
       case 51008:           /* OpcodeList1 */
         {
@@ -9167,6 +9169,9 @@ void CLASS adobe_coeff (const char *make, const char *model)
   }
   if (RT_blacklevel_from_constant == ThreeValBool::X || is_pentax_dng) {
     RT_blacklevel_from_constant = ThreeValBool::T;
+  }
+  if (RT_raw_crop_mask_from_constant == ThreeValBool::X || is_pentax_dng) {
+    RT_raw_crop_mask_from_constant = ThreeValBool::T;
   }
   if (RT_matrix_from_constant == ThreeValBool::X) {
     RT_matrix_from_constant = ThreeValBool::T;
--- a/rtengine/dcraw.h
+++ b/rtengine/dcraw.h
@@ -173,6 +173,7 @@ protected:
     enum class ThreeValBool { X = -1, F, T };
     ThreeValBool RT_whitelevel_from_constant;
     ThreeValBool RT_blacklevel_from_constant;
+    ThreeValBool RT_raw_crop_mask_from_constant;
     ThreeValBool RT_matrix_from_constant;
     std::string RT_software;
     double RT_baseline_exposure;
--- a/rtengine/rawimage.cc
+++ b/rtengine/rawimage.cc
@@ -29,9 +29,10 @@ RawImage::RawImage(const Glib::ustring &name)
     , allocation(nullptr)
 {
     memset(maximum_c4, 0, sizeof(maximum_c4));
-    RT_matrix_from_constant = ThreeValBool::X;
-    RT_blacklevel_from_constant = ThreeValBool::X;
     RT_whitelevel_from_constant = ThreeValBool::X;
+    RT_blacklevel_from_constant = ThreeValBool::X;
+    RT_raw_crop_mask_from_constant = ThreeValBool::X;
+    RT_matrix_from_constant = ThreeValBool::X;
 }
 
 RawImage::~RawImage()
@@ -556,7 +557,7 @@ int RawImage::loadRaw(bool loadData, unsigned int imageNum, bool closeFile, Prog
             orig_raw_width = raw_width;
             orig_raw_height = raw_height;
 
-            if (cc && cc->has_rawCrop(raw_width, raw_height)) {
+            if (RT_raw_crop_mask_from_constant == ThreeValBool::T && cc && cc->has_rawCrop(raw_width, raw_height)) {
                 raw_crop_cc = true;
                 int lm, tm, w, h;
                 cc->get_rawCrop(raw_width, raw_height, lm, tm, w, h);
@@ -591,7 +592,7 @@ int RawImage::loadRaw(bool loadData, unsigned int imageNum, bool closeFile, Prog
                 }
             }
 
-            if (cc && cc->has_rawMask(orig_raw_width, orig_raw_height, 0)) {
+            if (RT_raw_crop_mask_from_constant == ThreeValBool::T && cc && cc->has_rawMask(orig_raw_width, orig_raw_height, 0)) {
                 for (int i = 0; i < 2 && cc->has_rawMask(orig_raw_width, orig_raw_height, i); i++) {
                     cc->get_rawMask(orig_raw_width, orig_raw_height, i, mask[i][0], mask[i][1], mask[i][2], mask[i][3]);
                 }


Lawrence37:
We should review `camconst.json` for any raw crops for dng-only cameras, aside from Pentax since it's already handled by your code.

sgotti:
> We should review `camconst.json` for any raw crops for dng-only cameras, aside from Pentax since it's already handled by your code.

I know only of leica SL cameras. The SL, SL2 and the just released SL3 creates only dng files. Not sure about other Leica models.

~~Since this will be just a cleanup (without side effects), do you prefer to do this in this PR or open an issue when this will be merged and do it in another PR?~~

Do you mean to check if the dng raw crop is different than the one in the camconst? In such case what should be done?

Lawrence37:
If there's a raw crop for a dng-only camera, that means the crop is meant to override the values in the dng metadata. The camconst crop should take precedence just like it is with Pentax dngs.

sgotti:
@Lawrence37 @kmilos 

For Leica SL cameras, their unique raw format is dng. I analyzed some samples for the three SL models and probably the same should also apply for the Q models

For example:

### Leica SL2-S

There'a a relative raw crop defined in camconst.json `[ 0, 2, 0, -4 ] `. Please note that the the DNG Active Area is used as an initial raw_crop, so the entries in camconst.json, if relative like in this case, are doing an additional subtraction from the dng ActiveArea.

See also issue #6390.

So for a **normal image**:

* raw dimensions: 6048 x 4048
* dng active area/row crop: 0 0 6024 4048
* camconst raw crop: 0 2 6024 4042


for a **high resolution image**:

* raw dimensions: 12096 x 8082
* dng active area/row crop: 0 0 12034 8082
* camconst raw crop: 0 2 12034 8076

Something similar also for the other SL models.

With the raw crop provided by the dng I don't see any garbage pixels so I don't see where they came from when the entries were added.

So looks like the raw crop is not required in camconst and the ones in camconst are greater/too conservative since, when relative, they are substracting from something already cropped (dng ActiveArea) and not the whole raw size.

Also darktable/rawspeed doesn't have any crop but uses the dng values.

### What should we do?
* Remove the entries from camconst also it will change a bit current edits (is this considered a backward compatibility issue?)
* keep the entries in camconst, document they aren't need but kept for bw compat, and perhaps remove them in 6.0?

kmilos:
> I analyzed some samples for the three SL models and probably the same should also apply for the Q models

M series as well I think...

On RPU also CL and X2 samples are DNGs, and also X-U and TL2 at [DPR](https://www.dpreview.com/sample-galleries?category=cameras&sort=byBrandsChronologically).

sgotti:
> M series as well I think...
> 
> On RPU also CL and X2 samples are DNGs, and also X-U and TL2 at [DPR](https://www.dpreview.com/sample-galleries?category=cameras&sort=byBrandsChronologically).

Yeah but many of these are not in camconst 

```bash
$ grep -i leica rtengine/camconst.json | grep make_model
        "make_model" : [ "LEICA C-LUX", "LEICA CAM-DC25" ],
        "make_model" : "LEICA D-LUX 7",
        "make_model": "LEICA Q (Typ 116)",
    //    "make_model": "LEICA M8",
        "make_model": "LEICA Q2",
        "make_model": "LEICA SL (Typ 601)",
        "make_model": "Leica SL2",
        "make_model": "Leica SL2-S",
        "make_model" : ["LEICA V-LUX 5","Panasonic DC-FZ1000M2"],
        "make_model": [ "Panasonic DMC-FZ1000", "Leica V-LUX (Typ 114)" ],
        "make_model": [ "Panasonic DMC-LF1", "Leica C (Typ 112)" ],
        "make_model": [ "Panasonic DMC-LX100", "Leica D-LUX (Typ 109)" ],
```
Leica m8 commented out since it was causing issues and with just dng metadata it works (#6237 and #6498).

But we can't know if they are supported or just not tested by anyone

@Lawrence37 probably a list of supported cameras, also without any overrides should be kept. Perhaps camconst.json is the right place (just add an entry without any overrides or also add a field reporting it as unsupported) like done by rawspeed with cameras.xml. If it was not already discussed I'd like to open a dedicated issue.

kmilos:
> Perhaps camconst.json is the right place (just add an entry without any overrides or also add a field reporting it as unsupported) like done by rawspeed with cameras.xml.

One caveat - dt/RawSpeed doesn't explicitly test nor mark as supported/unsupported DNG based models (except for Pentax maybe because of the dual PEF output), and many samples are missing on RPU indeed (Roman generally doesn't even want them unless they are to test some new functionality). DNGs are self-contained and should "just work" (assuming the implementation of various compression modes) w/o any additional metadata supplied or override. If that's not the case, it is probably an invalid/uncompliant file. Checking and listing all the possible devices outputting DNG is a fool's errand IMHO...

The presence of DNG based models in RawSpeed's cameras.xml is there just to normalize and prettify the Make/Model strings (and we pretty much draw the line at "real" camera bodies like Leica, Pentax/Ricoh, and Sigma; trying to cover smartphones and the myriad of other devices is not worth it IMHO...).

In other words, IMHO you should only use your camconst.json to exceptionally override some known problematic DNGs, not to track valid ones.

Lawrence37:
As far as Leica dngs go, these models have raw crops in `camconst.json`.
- LEICA Q (Typ 116)
- LEICA Q2
- LEICA SL (Typ 601)
- Leica SL2
- Leica SL2-S

For the Leica SL2-S, we know the original raw crop was `[ 0, 2, 6024, 4042 ]` with a comment explaining it:
> 2 rows at top and 4 rows at bottom are black

It is strange that the dng ActiveArea says otherwise, so I would like to confirm if there really are black rows.

The other models should be reviewed too, or just take the easy option of allowing Leica dng raw crops to be overridden by the camconst entries.

sgotti:
> It is strange that the dng ActiveArea says otherwise, so I would like to confirm if there really are black rows.

@Lawrence37 I did more tests and just found that until now there was the need to define custom raw crops also for dng files because dcraw isn't handling dng DefaultCropOrigin and DefaultCropSize (that should be applied to the dng ActiveArea) but only ActiveArea. For this reason there're additional raw crop for Leica DNGs in camconst. 

libraw parses and handles DefaultCrop (and also additional UserCrop) but doesn't apply them by default but only if the user calls `LibRaw::adjust_to_raw_inset_crops` to adjust the image dimensions to the cropped info provided by dng (or other vendors). So also the current libraw branch doesn't handle vendor crops like dcraw since it isn't calling that function.

Additionally there's a special case in libraw for leica sl2 (not sure the reason why): https://github.com/LibRaw/LibRaw/blob/d3cbbd0e9934898eb28e4963ee99b51928e2acaa/src/utils/open.cpp#L975-L978

From all of the above some open issues arise related to this issue and also the the libraw branch:

* dcraw and libraw images sizes con be different for the same file due to different handling (this may affect any raw file, not only dng).
* Sizes will change if we implement handling if DefaultCrop in dcraw.
* Sizes will change if we use `LibRaw::adjust_to_raw_inset_crops` and this may affect any raw file, not only dng, because it handles not only dng raw crop metadata but also all the other files (or it should be called only for dng files).
* The above image size changes won't affect camconst raw_crop with absolute values but will affect relative raw crops. (i.e. leica SL2 `"raw_crop": [ 0, 0, 0, -18 ]`
* The use of DNG DefaultCrop could lead to smaller image sizes. i.e. (Leica SL2 files have DefaultCropOrigin 12, 24 and DefaultCropSize  8368 5584, much smaller than current rawcrop that brings to `0, 0, 8392, 5613` )

Lawrence37:
The dng documentation doesn't explicitly say it, but it does imply that the DefaultCrop is not necessary to guarantee a good image. Using the whole ActiveArea should be acceptable. If the Leica dngs contain black pixels in the ActiveArea, then they do not follow the specification (masked pixels must not be in the ActiveArea). We should use the ActiveArea to crop images and have an exception for Leica dngs.

Lawrence37:
Converting to a draft because Leica camconst crops need to be considered.

PR #7018, opened 2024-03-30
Add Panasonic v8 decoder and support for DC-S5M2 and DC-S5M2X

sgotti:
Port Panasonic v8 decoder from libraw.

* Extract data required for the decoder from the panasonic custom exif
  tags
* Add panasonic v8 decoder
* dcraw: add Panasonic DC-S5M2 and DC-S5M2X to adobe_coeffs

Fixes #6771 #6909

## Tested images

* Tested with images from both cameras.
* Also tested in-camera pixelshift high resolution images

## NOTES
* Support for other cameras based on v8 (Panasonic GH6 and G9II) could be added after this PR.

@Lawrence37 I tried to adapt the libraw code to the current rawtherapee code practices but I may have lost a lot of things...

PR #6887, opened 2023-11-19
Internal LibRaw

Lawrence37:
As an experiment for #4462, I've added LibRaw directly into RawTherapee and borrowed heavily from ART for the implementation. It is functional and can potentially be adopted as the solution for replacing dcraw. I chose to go this route to familiarize myself with the raw image decoding part of the code. It didn't take much time or effort and at least I gained relevant knowledge if we decide to go with another route.

The LibRaw source code is included via [git-subtree](https://manpages.debian.org/bullseye/git-man/git-subtree.1.en.html) at [`rtengine/libraw`](https://github.com/Lawrence37/RawTherapee/tree/d8252ad89aa89864458dda0f98befdf68c23c37f/rtengine/libraw). Like git submodules, git-subtree facilitates git subprojects, but tracks the commits in the parent project. The first two commits here add the LibRaw code to RawTherapee. Having our own copy of LibRaw allows us to make tweaks to it just as we have done with dcraw. There are no modifications to the LibRaw code yet in this pull request.

LibRaw has its own configuration script to generate the Makefile. At some point, it may be worth it to directly configure and build LibRaw through CMake. For now, I simply have CMake run the configuration script and added a task to call `make`. It works seamlessly except for the clean target. There's no way to modify the clean target, so cleaning the LibRaw build requires a second command: `make clean-libraw`. That means a full clean requires `make clean` *and* `make clean-libraw`.

The integration is fairly simple with most of the changes contained within the `RawImage` class. LibRaw is the first decoder. If it fails or does not support decoding the image, it falls back to dcraw. Since the code is adapted from ART, it accumulated some edge-case handling during the time it lived in ART.

PR #7019, opened 2024-03-30
dcraw: add Panasonic DC-G9M2

sgotti:
dcraw: add Panasonic DC-G9M2 to adobe_coeff

Depends on v8 decoder provided by #7018 

## Tested images

* Tested normal images.
* Tested in-camera pixelshift high resolution images


Naastek:
hi boss~ i want to know when this will release? when i use g9m2 raw on 5.10, it's all black and i don't know how to make dmg from zero XD

Lawrence37:
@Naastek This will probably be included in 5.11. There is no set release date, but I'm going to guess early in the second half of this year. Until then, you can have a look at Adobe DNG Converter or a FOSS DNG converter (I think it was https://github.com/dnglab/dnglab). You could also try the pre-dev build for this if you are on Windows or Linux. See https://discuss.pixls.us/t/rawtherapee-pre-dev/35595. Once this gets merged into the development version, you can start using the development builds.

Naastek:
@Lawrence37 thanks for your info. unfortunately i work with macOS can't get access to pre-dev. maybe i will wait for the dev builds.

rtengine/camconst.json
```diff
@@ -2263,6 +2263,14 @@ Camera constants:
         }
     },
 
+    { // Quality C
+        "make_model": "Panasonic DC-G9M2",
+        "dcraw_matrix": [ 8325, -3456, -623, -4330, 12089, 2528, -860, 2646, 5984 ], // Adobe DNG Converter
+        "ranges": {
+            "white": 65535
+        }
```
Lawrence37: We should see why the decoder does not have the white level for this camera.

sgotti: The issue is in `dcraw.cc` `DCraw::adobe_coeff`. There's an entry for a `Panasonic DC-G9` camera and this function uses a prefix so it matches also `Panasonic DC-G9M2`.
https://github.com/Beep6581/RawTherapee/blob/5a25619bd50d42059abe9818420f639b6d10e3bf/rtengine/dcraw.cc#L8892-L8893

But this camera has more detailed information provided in camconst so this entry could just be removed from dcraw:adobe_coeffs since it's overridden by camconst entries. This will also fix this issue.

@Lawrence37 Is it ok for you? Should I create a new PR or a new commit in this PR to remove this entry from `DCraw::adobe_coeff`?

Lawrence37: I prefer to keep the dcraw value there because the decoder should provide a reasonable white level in the absence of a camconst entry. We should add the DC-G9M2 values from Libraw:
https://github.com/LibRaw/LibRaw/blob/158e635e5e80a95dfa402b0a79a7167922c1d4ec/src/tables/colordata.cpp#L1367-L1371

Edit: It also provides the color matrix, so a camconst entry would not be needed at all.

sgotti: @Lawrence37 I you think it'll be better I'll do this but I'm quite confused why `adobe_coeffs` is better than camconst in this scenario. I thought camconst was the new rawtherapee way to handle camera properties and adobe_coeffs the old dcraw way before camconst were introduced (I also thought that `adobe_coeffs` table could be removed and migrated to camconst but I may have lost something).

kmilos: I guess if RT is to [switch over](https://github.com/Beep6581/RawTherapee/pull/6887) to LibRaw, [it'll have its own adobe_coeff table](https://github.com/LibRaw/LibRaw/blob/master/src/tables/colordata.cpp) as well, so indeed no need to duplicate in camconst...

sgotti:
@Lawrence37 Updated the PR to update the dcraw adobe_coeffs table instead of a camconst entry.
 
> I prefer to keep the dcraw value there because the decoder should provide a reasonable white level in the absence of a camconst entry.

FWIW The white level is not provided by adobe_coeffs (the entry is 0 so ignored) but calculated from tiff_bps (16 in this case): 

https://github.com/Beep6581/RawTherapee/blob/5a25619bd50d42059abe9818420f639b6d10e3bf/rtengine/dcraw.cc#L10720

rtengine/camconst.json
Lawrence37: One is not necessarily better than the other. They serve different purposes. `adobe_coeff` provides constants straight out of the decoder. `camconst.json` is for overriding, fine-tuning, or supplementing the decoder's values. It's like coding style. There's no difference for the user, but it improves maintainability for developers.

PR #7020, opened 2024-03-30
dcraw: add Panasonic DC-GH6

### LINKED ISSUE #6469: Panasonic Lumix DC GH6 RW2 raw format support
Heya, I got my hands on a GH6 by Panasonic.

The model is branded as "DC-GH6E" on the Panasonic Nomenclature. (don't ask me why it is "DC" and not "DMC", I don't know, the GH5 is also named "DC", GH2 GH3 and GH4 were labelled as DMC ...)

I believe any version of rawtherapee (dev or released) will fail at loading the raws created by it.

The typical error message will be: (filename).RW2: Corrupt data near 0x48c200

From what I looked,
the file format changed, it is now using the following information:
RT_pana_info.encoding= 8
RT_pana_info.bpp= 16 

by looking at rtengine/panasonic_decoders.cc, I assume this new encoding version (8) is new and that there is a method needed that probably would be named as panasonicC8_load_raw() 
in rtengine/panasonic_decoders.cc

Now, this is not my field of expertise at all and I just don't know what is needed exactly.

The camera can also produce files having an enhanced resolution by merging several photos taken at the same time for a dimension up to 11560 by 8680 pixels, now the format data seems to be the same too (same version and bpp) for those files.

I am providing one raw of one photo.

https://filebin.net/1bh53vgllqhi2oy1/P1000265.RW2

I can also provide sets of white frames if needed, just give me the time to make these.

NB: mirror for the raw files - http://esby.free.fr/perso/GH6-RW2/

sgotti:
dcraw: add Panasonic DC-GH6 to adobe_coeffs

Depends on v8 decoder provided by #7018 

## Tested images

* Tested normal images.
* Tested in-camera pixelshift high resolution images


rtengine/camconst.json
```diff
@@ -2537,6 +2537,14 @@ Camera constants:
         }
     },
 
+    { // Quality C
+        "make_model": "Panasonic DC-GH6",
+        "dcraw_matrix": [ 7949, -3491, -710, -3435, 11681, 1977, -503, 1622, 5065 ], // Adobe DNG Converter
+        "ranges": {
+            "white": 65535
+        }
```
Lawrence37: The decoder already has the white level for this camera.
```white levels: R:65535 G1:65535 B:65535 G2:65535 (provided by dcraw)```

sgotti: Right.

If useful this isn't directly provided by any metadata but calculated from tiff_bps (16 in this case): https://github.com/Beep6581/RawTherapee/blob/5a25619bd50d42059abe9818420f639b6d10e3bf/rtengine/dcraw.cc#L10720

Lawrence37: Similar to the edit to my comment in the other Panasonic camconst PR, Libraw has the color matrix and white level. We should consider copying the new Libraw constants for all cameras into dcraw.cc.
https://github.com/LibRaw/LibRaw/blob/158e635e5e80a95dfa402b0a79a7167922c1d4ec/src/tables/colordata.cpp#L1387-L1388

sgotti:
@Lawrence37 Updated the PR to update the dcraw adobe_coeffs table instead of a camconst entry.
Note that the black level is set to 0, while in libraw is set to -2048 because the logic has changed a lot in libraw. If we set it to 2048 here we'll get a black level of 4096 since it's summed to the already provided single channel blacks levels (cblack) with value 2048 read from metadata.

PR #7023, opened 2024-04-01
Limits incompatibility between Ciecam and CBDL and others to Cam02   #7017

### LINKED ISSUE #7017: CDBL not always working
<
**Short description**
While editing pictures with CDBL and Colour Appearance & Lighting enabled CDBL works only if inside CDBL the process After Black and White is chosen. With Before Black and White there is no action. 

**Steps to reproduce**
1. Open a piucture
2. activate CDBL with Colour Appearance & Lighting deactivated. CDBL works in both modes after and before black and white.
4. Activate Colour Appearance &Lighting with CDBL activated. CDBL works only in the mode After Black and White.-->

**Expected behavior**
CDBL should work always in both modes.
If the described behavior is intended there should be a hint and the appropriate mode in CDBL should be set by activating Colour Appearance & Lighting automatically.

**Additional information**
Version: 5.10-179-g11240bc97
Branch: dev
Commit: 11240bc97
Commit date: 2024-03-28
Compiler: cc 11.4.0
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.5
Lensfun: V0.3.95.0
Build type: release

-->

Desmis:


Desmis:
@hholtm @Lawrence37

Same text as #7017 

For the first part, it will be very complicated to change this, which is still a marginal case.

For the second part, as I'm the author (oh sure not of the concept) of both the presentation style and the Ciecam code, I think it's important to maintain both (Cam16 and Cam02). On the one hand, for the sake of compatibility, and on the other, to keep track of the CAM (Color Appearance Model) history.

**I propose to merge next week, if no one objects**

Jacques

Desmis:
@Lawrence37 
if no objection I will merge tomorrow

Jacques

Issue #7026, opened 2024-04-03
Spot Removal not working properly


PDeb108:
The problem I am facing in RT is using the spot removal module. I am working with it at the finishing stage of the work. What is happening is that as soon as I click the button on the right to activate the spot (marked with a red circle in the bottom frame). The exposure of the entire picture increases to the point when I am unable to see the spot. The top frame is when I have not activated the spot creation button, and you can see the spot.
This is when I am using the most recent dev version of RT (RawTherapee_7023merge_5.10-182-g918ef5e5c_win64_release)

My system is Intel I5/16GB/Intel graphics processor/Windows 11
The situation is similar even when I use the stable release 5.10

Below is the nature of the problem

![a065cb07e48ccf9878335f8d1052380dbe3104b6](https://github.com/Beep6581/RawTherapee/assets/165793573/24ce3b6e-6e8a-4aef-985d-f0c6c0947e8e)

-

Desmis:
@Lawrence37 @PDeb108 

@Hombre57 - I believe it was you who led this change a few years ago, can you look. Thanks.

I confirm the bug found by @PDeb108 
The operation of "Spot Removal" seems to be linked to the Raw file and the lenses used. In many cases everything works fine. In this case it does not work, or poorly, or can cause a system crash.

The problem was found during the challenge to test "lacam16n2" but it seems that this malfunction occurs whatever the branch, and probably dates back a long time
https://discuss.pixls.us/t/rawtherapee-processing-challenge/42647/66

Here the file Raw, and the pp3
https://drive.google.com/file/d/1w6sn1BQ8FQ7uSTd3quceJcyNU5K_nCCZ/view?usp=sharing
https://drive.google.com/file/d/16HVp5w3_Oas9qzEj_2KHYSpKClpzsTWK/view?usp=sharing

I've tried several changes that make it work less badly.
In Dcrop.cc and Improcoordinator.cc - better (I think) cleaning of variables (copying part of the ART code, which is very little different), but I don't think that's it
In refreshmap.h #define M_spot change from 16 to 22
I also change but it's only cosmetic the icon in Tollbar.cc

It seems that: crop, Lens profile, Input profile have a strong impact.
It seems also that : Haze removal and Graduated Filter have an impact.
I try to disable all "expanders" without real effect.

It works only with "neutral".

[...]

PR #7029, opened 2024-04-06
Improve SpotRemoval 

Desmis:
Hello
As agreed with @Hombre57 , I added 2 lines in tweakParams "spot.cc"
    pparams.locallab.enabled = false;
   // pparams.toneCurve.hrenabled = false;  // not sure for this one, it could be useful for ExpComp w/o performance penalty
    pparams.toneCurve.histmatching = false;

And as some users pointed out that the icons in the toolbar were the same for "Local adjustments" and "Spot Removal", I changed the one for Spot Removal with "hand-open-hicontrast" ( I didn't find another acceptable icon) in "toolbar.cc"

see #7026 
I will be away today and in Paris since tomorrow.

Jacques

### DIFF
--- a/rtgui/spot.cc
+++ b/rtgui/spot.cc
@@ -893,4 +893,7 @@ void Spot::tweakParams(procparams::ProcParams& pparams)
     pparams.gradient.enabled = false;
     pparams.pcvignette.enabled = false;
     pparams.colorappearance.enabled = false;
+    pparams.locallab.enabled = false;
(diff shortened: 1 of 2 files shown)


Hombre57 (commented):
@Desmis
The patch looks good but I can't test since I did not installed any tool chain yet (and I'm not sure to do that soon), so I'll let you and other devs test the performance penalty of these changes.

Desmis:
@Hombre57 @Lawrence37 

I don't find any malfunctions in this change.
I propose to merge this week

Jacques

PR #7038, opened 2024-04-13
Fix for missing icons from the Windows installer

Lawrence37:
Updates the Inno setup to copy the icons folder which is new after merging the real hi-DPI pull request (#6830).


Desmis:
@Lawrence37 
I think this PR must solve the windows executable problem

Jacques

PR #7046, opened 2024-04-18
 Add correct DateTime (editing date) to Exif data

### LINKED ISSUE #7059: No DateTime (= editing date) in EXIF data of exported photo
Currently, if you export a photo, the EXIF tag DateTime is missing from the photo, so the date/time, when the photo was saved/edited is not present. Only DateTimeOriginal is in the file, which is the date when you captured the photo.

I fixed this a while ago with [this](https://github.com/Beep6581/RawTherapee/pull/7046) and would like the fix to be merged.

#7046

blitzgneisserin:
Currently, RT does not save the editing date (Exif tag DateTime) in the exif data of the exported files, only DateTimeOriginal (the date/time the photo was captured). This patch fixes this.
Screenshot_20240418_164957


rtengine/metadata.cc
```diff
@@ -312,6 +313,13 @@ void Exiv2Metadata::saveToImage(const Glib::ustring &path, bool preserve_all_tag
     }
 
     dst->exifData()["Exif.Image.Software"] = "RawTherapee " RTVERSION;
+
+    std::time_t t = std::time(nullptr);
+    char mbstr[100];
```
Lawrence37: This can be smaller

rtengine/metadata.cc
```diff
@@ -312,6 +313,13 @@ void Exiv2Metadata::saveToImage(const Glib::ustring &path, bool preserve_all_tag
...
+    if (std::strftime(mbstr, sizeof(mbstr), "%Y:%m:%d %H:%M:%S", std::localtime(&t))) {
+    dst->exifData()["Exif.Image.DateTime"] = mbstr;
```
Lawrence37: Indentation

blitzgneisserin:
```
if (std::strftime(mbstr, sizeof(mbstr), "%Y:%m:%d %H:%M:%S", std::localtime(&t))) {
    dst->exifData()["Exif.Image.DateTime"] = mbstr;
```

You mean like this? (Github automatically added the indentation when I pasted the code)

rtengine/metadata.cc
blitzgneisserin: `  char mbstr[20];`
  
  You mean like this, right?

Lawrence37:
Thanks Anna! Looks fine to me.

blitzgneisserin:
> Thanks Anna! Looks fine to me.

Thanks for the help. I am rather new to coding, especially C++.

PR #7049, opened 2024-04-19
Add some White Balance and add Russian Tranlation

ghost:


Lawrence37:
Thanks for the translations @Iliya5digital!

Regarding the white balance profiles, I don't think they should be bundled with RawTherapee. They are quite specific, affecting only the color temperature. There are alternatives to quickly setting the color temperature in most cases. For niche applications, users can add profiles to their own installations.

Lawrence37:
@Iliya5digital Could you please create separate pull requests for the translation and white balance additions? Also, I think the profiles are unnecessary (they also need some trimming because they total almost 25k lines), especially as the WB presets are in the WB tool.

ghost:
how do I make separate separate pull requests for the translation and white balance additions?

Lawrence37:
You need to separate your changes into different branches, then open a pull request for each branch.

PR #7053, opened 2024-04-22
dcp: avoid file descriptor leak on error.

### LINKED ISSUE #7045: File operations rename not working after file opened for editing. Development branch.
**Short description**
In file browser view, Mouse right click menu->File operations->Rename stops working after Canon EOS R10 generated CR3 raw file has been opened for editing. Nothing else seems to be wrong, only apparent problem is that the file renaming doesn't work for a file that has been opened for editing. Issue exists both in editor view and in file browser view. Files can be renamed before opening for editing and again after restarting the RawTherapee.

- Issue does not occur with DNG files generated from the EOS R10 CR3 files. 
- Issue does not occur with Sony A7II ARW files or with Canon 2000D CR2 files.
- Issue did not exists with 5.10 release version, it started after building with latest commits.
- You can still rename any other files including other CR3 files, but not those CR3 files that have been opened for editing.
- Renaming the files outside of RawTherapee, using Dolphin File Manager is also not possible until RawTherapee is closed. With the other filetypes, there doesn't seem to be any restrictions.

**Steps to reproduce**
1. Open RawTherapee and browse into a folder with a CR3 file from Canon EOS R10.
2. Double click file for editing it.
3. Using the mouse right click menu, try to rename file from the top panel or go back to the file browser view to attempt renaming.
4. Nothing happens after selecting OK.

**Expected behavior**
File should get renamed.

**Additional information**
<details><summary>About the system and the build:</summary>
<p>

Version: 5.10-179-g11240bc97
Branch: dev
Commit: 11240bc97
Commit date: 2024-03-28
Compiler: cc 13.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.8
Lensfun: V0.3.4.0
Build type: Release
Build flags:  -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:   
OpenMP support: ON
MMAP support: ON
Build OS: Linux 5.15.150-1-MANJARO x86_64
Build date: Wed, 10 Apr 2024 16:43:37 +0000 UTC
Build epoch: 1712767417
Build UUID: be55565b-c08a-4c2d-bc7d-f0dddb404715

</p>
</details>

sgotti:
If the provided file doesn't contain valid dcp metadata the DCPProfile constructor will return without closing the file descriptor.

Fixes #7045

### DIFF
--- a/rtengine/dcp.cc
+++ b/rtengine/dcp.cc
@@ -1134,6 +1134,7 @@ DCPProfile::DCPProfile(const Glib::ustring& filename) :
     DCPMetadata md(file);
     if (!md.parse()) {
         printf ("Unable to load DCP profile '%s'.", filename.c_str());
+        fclose(file);
         return;
     }
 


Floessie:
@sgotti, good catch!

The one-liner is so indisputable that I'll merge it right away.

BUT: The better solution was to wrap the `FILE*` into a `std::unique_ptr<>` like it is done [a few lines above](https://github.com/Beep6581/RawTherapee/blob/232c11c37c9e76cd8b1c76d982d5dcb55b137686/rtengine/dcp.cc#L383) and get rid of manually closing the file once and for all.

@sgotti Fancy a follow-up?

sgotti:
@Floessie Thanks!

> BUT: The better solution was to wrap the `FILE*` into a `std::unique_ptr<>` like it is done [a few lines above](https://github.com/Beep6581/RawTherapee/blob/232c11c37c9e76cd8b1c76d982d5dcb55b137686/rtengine/dcp.cc#L383) and get rid of manually closing the file once and for all.
> 
> @sgotti Fancy a follow-up?

Created PR #7056 .

PR #7056, opened 2024-04-24
dcp: use std::unique_ptr to automatically close file descriptor

sgotti:
Use std::unique_ptr to automatically close file descriptor instead of manually closing it at every return point.


Floessie:
@sgotti Perfect! :+1: Thanks a lot!

Issue #7055, opened 2024-04-24
History is getting spammed when Color Appearance & Lighting module is enabled Build 5.10-190 on dev branch.


kaesa:
**Short description**
When the Color Appearance & Lighting module is enabled and you modify any values in any modules, the history is getting spammed with CAL-VC-Tint and CAL-VC-Temperature values, which have not changed from the previous. Also, you need to double click snapshots to get them selected, as on the first click, the focus moves to the History.

**Steps to reproduce**
1. Enable Color Appearance & Lighting module from the Advanced tab.
2. Modify sliders anywhere or switch between snapshots.
3. History is filling up with lines of CAL-VC-Tint and CAL-VC-Temperature values, which have not changed from their previous values.

**Expected behavior**
History should stay cleaner to remain useful. One click on a snapshot should be enough to get it selected(for deleting).

**Additional information**
<details><summary>Details</summary>
<p>

Version: 5.10-190-gc72e67ae2
Branch: dev
Commit: c72e67ae2
Commit date: 2024-04-20
Compiler: cc 13.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.8
Lensfun: V0.3.4.0
Build type: Release
Build flags:  -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:   
OpenMP support: ON
MMAP support: ON
Build OS: Linux 5.15.154-1-MANJARO x86_64
Build date: Sat, 20 Apr 2024 15:59:44 +0000 UTC
Build epoch: 1713628784
Build UUID: a15bc86b-fbda-444d-bb2d-64476313b219

</p>
</details>

Desmis:
@kaesa 

This is a (very) complex event management problem...This concerns the "automatic" management of the 2 sliders "Temp" and "Tint" in "Viewing conditions". And that makes several “issues” on this subject directly or indirectly, since many years.
[...]

PR #7058, opened 2024-04-25
Ciecam - Change management viewing temperature and tint #7055

Desmis:
See issue #7055

### DIFF
--- a/rtengine/improccoordinator.cc
+++ b/rtengine/improccoordinator.cc
@@ -2308,8 +2308,8 @@ void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)
                     tempsym = params->colorappearance.tempout;
                     greensym = params->colorappearance.greenout;
                 }
-                if (params->colorappearance.enabled) {
-                    acListener->wbCamChanged(tempsym, greensym);    //real temp and tint.
+                if (params->colorappearance.enabled  && acListener) {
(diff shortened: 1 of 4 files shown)


kaesa:
Would this kind of comparison in the ColorAppearance::wbCamChanged function be considered unreliable or otherwise undesirable? This didn't spam the history values in my quick test and ticking the checkbox both moved the slider and changed the applied value. It also seemed to fix the issue with Automatic Symmetric mode, where the applied value was forced, but moving the slider printed wrong values to the history.

```c++
[this, temp, tin, autotemp]() -> bool
        {
            if (temp != tempout->getValue()) {
                disableListener();
                tempout->setValue(temp);
                enableListener();
                listener->panelChanged (EvCATtempout, tempout->getTextValue());
            }
            if (tin != greenout->getValue()) {
                disableListener();
                greenout->setValue(tin);
                enableListener();
                listener->panelChanged (EvCATgreenout, greenout->getTextValue());
            }
            /*
            disableListener();
            tempout->setValue(temp);
            greenout->setValue(tin);
            enableListener();

            if(!autotemp) {
                listener->panelChanged (EvCATgreenout, "");//greenout->getTextValue());
                listener->panelChanged (EvCATtempout, "");//tempout->getTextValue());
            }*/

            return false;
        }
```

Desmis:
@kaesa 

it works, it's another way - a little simpler - to achieve the result. I'm going to change it.
But, for the record, this "wbCamChanged" function, I have already disabled/enabled it several times.

In any case, thank you for this positive contribution.

Jacques

PR #7076, opened 2024-05-16
dcraw/camconst: add Sony ILCE-6700

### LINKED ISSUE #7075: Sony ILCE-6700 has invalid pixels on the right border
Currently, ARW files from Sony ILCE-6700 map to 6264x4160 image, which has rightmost 26 pixels without any meaningful data. The effect is visible on [this converted photo](https://flic.kr/p/2pPoKVe) (Flickr link), as well as on compressed raw sample from raw.pixls.us when processed by RawTherapee 5.10.

The following camconst.json fragment fixed it for me:

```json
    { // Quality C
        "make_model": "Sony ILCE-6700",
        "raw_crop": [ 20, 12, 6200, 4136 ] // columns after 6238 are b0rked; this crops the rest so that resulting image is 6192x4128, i.e. the same as camera JPEG
    },
```

sgotti:
* Add dcraw adobe_coeffs entry for Sony ILCE-6700
* Add camconst raw crop for Sony ILCE-6700

## Notes

* The raw crop is the same as the one provided by the raw converted to dng with ADC. (looks like the same done with previous models like ILCE-6500).

* Rawspeed uses a -28 on right that will lead to 6244 (6272 - 28) instead of 6240.

Fixes #7075


Lawrence37:
The comments at the top of camconst.json don't give specific guidelines for choosing the raw crop, but I think the RawTherapee philosophy of extracting the most out of images makes the Rawspeed crop the better option. It does seem to provide the largest usable crop. I'm not sure why the other camconst crops are more aggressive.

sgotti:
@Lawrence37 Given your comment https://github.com/Beep6581/RawTherapee/pull/7020#discussion_r1605945667 : 
> We should consider copying the new Libraw constants for all cameras into dcraw.cc.

In such case where a camconst entry is required to fix the raw_crop:
* should I also add an entry to dcraw.cc adobe_coeffs?
* should the camconst only contain the raw_crop entries without the dcraw_matrix since already provided by adobe_coeffs?

Lawrence37:
I would only add to `dcraw.cc` if dcraw or LibRaw has it. `camconst.json` entries should only be added if they make a difference, so yes to leaving out the matrix since it's already in the decoder.

sgotti:
> I would only add to `dcraw.cc` if dcraw or LibRaw has it. `camconst.json` entries should only be added if they make a difference, so yes to leaving out the matrix since it's already in the decoder.

@Lawrence37 Done. ~~The unique doubt is about using raw_crop `[ 0, 0, 6240, 4168 ]` (adc) vs `[ 0, 0, 6244, 4168 ]` (rawspeed). The latter provides more data and looks good also setting a demosaicer without border.~~ (sorry lost you first comment https://github.com/Beep6581/RawTherapee/pull/7076#issuecomment-2119106171)

Lawrence37 (approved):

PR #7079, opened 2024-05-19
Selective editing - adds the feather slider for each Graduated Filter

Desmis:
In the current version, the "feather" slider was common to all GFs (at least 8). This slider was in "Settings".

In the same approach as "Preview deltaE" this slider is now in each GF within its tool.

This also allows it to be available in "Global" mode.
99% it's only GUI


Lawrence37:
Code-wise, it looks good. I noticed the feather value does not carry over from old pp3s. They just get set to the default value of 25.

PR #7080, opened 2024-05-19
Fix linking with jpeg-turbo

### LINKED ISSUE #7074: Build fails with libjpeg-turbo 3.0.3: Undefined symbols for architecture x86_64: "_jpeg_std_message_table"
Looks that `libjpeg-turgo` changed API in 3.0.3, and this code now does not work:
https://github.com/Beep6581/RawTherapee/blob/5a25619bd50d42059abe9818420f639b6d10e3bf/rtengine/jdatasrc.cc#L251-L262

Windows build has this issue years ago: https://github.com/Beep6581/RawTherapee/issues/4713
Now identical linking failure on macOS: https://build.macports.org/builders/ports-14_x86_64-builder/builds/39199/steps/install-port/logs/stdio

This compiles though:
```
const char * const jpeg_std_message_table[] = {
#include "jerror.h"
  NULL
};
```
(Functionality not tested.)

Benitoite:
Fixes https://github.com/Beep6581/RawTherapee/issues/7074
This patch via Termux PR
https://github.com/termux-user-repository/tur/pull/1027

### DIFF
--- a/rtengine/jdatasrc.cc
+++ b/rtengine/jdatasrc.cc
@@ -247,20 +247,6 @@ my_error_exit (j_common_ptr cinfo)
 #endif
 }
 
-
-#ifdef _WIN32
-#define JVERSION	"6b  27-Mar-1998"
-#define JCOPYRIGHT_SHORT	"(C) 1998, Thomas G. Lane"
-#define JMESSAGE(code,string)	string ,
-
-const char * const jpeg_std_message_table[] = {
-#include "jerror.h"
-  NULL
-};
-#else
-extern const char * const jpeg_std_message_table[];
-#endif
-
 /*
  * Actual output of an error or trace message.
  * Applications may override this method to send JPEG messages somewhere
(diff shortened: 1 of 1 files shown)


rtengine/jdatasrc.cc
```diff
@@ -409,24 +395,14 @@ reset_error_mgr (j_common_ptr cinfo)
...
 my_jpeg_std_error (struct jpeg_error_mgr * err)
 {
+    err = jpeg_std_error(err);
 
+    /* override these functions */
     err->error_exit = my_error_exit;
     err->emit_message = emit_message;
     err->output_message = output_message;
     err->format_message = format_message;
     err->reset_error_mgr = reset_error_mgr;
```
Lawrence37: It seems to me `my_error_exit` is the only customized function. The other ones are exact copies of the ones in libjpeg-turbo and are already added to the `err` struct by `jpeg_std_error()`. Thus, the only thing this function (`my_jpeg_std_error`) needs to do is call `jpeg_std_error()` then override the `error_exit` function. The copied functions can be removed from this file.

```suggestion
    err = jpeg_std_error(err);

    /* override these functions */
    err->error_exit = my_error_exit;
```

Lawrence37:
Thank you @Benitoite.

PR #7092, opened 2024-06-02
Windows installer: Open images in existing RawTherapee window

### LINKED ISSUE #5738: Windows: gdbus.exe required to get -R option working
see https://discuss.pixls.us/t/how-do-i-switch-from-edit-window-to-browser-window/1723/33 for reference.

The -R option never worked satifactorily on windows. As I don't use this feature, I never worried until the thread above.

Can a dev explain what is the use of this exe and where it is called in RT and validate this addition?

@aferrero2707 when validated, can you add gdbus.exe to your windows builds?
Are there other builders providing official windows stable and nightly build?
I will update rawpedia/windows accordingly.

What about Linux and Mac?

Lawrence37:
Closes #5738.


Lawrence37:
Merging this to test the creation of the installer.

PR #7096, opened 2024-06-10
Enable extensions that are missing from config

xiota:
This PR adds and enables supported file extensions that are missing from config.  Feature is described in https://github.com/Beep6581/RawTherapee/pull/6367#issuecomment-2132712754

> One more possible improvement. Users who are upgrading will not see jxl images because their settings carry over. It would be good to automatically add the jxl extension to the list of file types to load if the version in the options file shows 5.10 or older. (@Lawrence37)

Notes:
* This PR changes default state of ori and png to enabled.  See [#6367 (comment)](https://github.com/Beep6581/RawTherapee/pull/6367#issuecomment-950320058).
* Known extensions are prevented from being removed from list.  Otherwise they would be added back in enabled state.  Disable by unchecking.
* Potential side effect I haven't checked for: List may be reordered during checks.

### DIFF
--- a/rtgui/preferences.cc
+++ b/rtgui/preferences.cc
@@ -1552,9 +1552,9 @@ Gtk::Widget* Preferences::getFileBrowserPanel()
 
     frmnu->add (*menuGrid);
 
-
     Gtk::Frame* fre = Gtk::manage(new Gtk::Frame(M("PREFERENCES_PARSEDEXT")));
     Gtk::Box* vbre = Gtk::manage(new Gtk::Box(Gtk::ORIENTATION_VERTICAL));
+
     Gtk::Box* hb0 = Gtk::manage(new Gtk::Box());
     Gtk::Label* elab = Gtk::manage (new Gtk::Label (M("PREFERENCES_PARSEDEXTADD") + ":", Gtk::ALIGN_START));
     hb0->pack_start(*elab, Gtk::PACK_SHRINK, 4);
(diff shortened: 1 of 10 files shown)


xiota:
Notes:
* I made a typo in `options.win`.  Will fix along with other improvement suggestions.
* ori and png are currently disabled by default.  This PR enables them by default.  Existing disabled configs are not altered.  Comment about enabling png at https://github.com/Beep6581/RawTherapee/pull/6367#issuecomment-950320058
* Current behavior is removing extension from list disables it.  After PR, known extensions will be added back in enabled state.  To disable requires unchecking.
    * Potential solution is to prevent removing known extensions from list.  I need a pointer to location of GUI in code.

xiota:
Force push to clean up commits.

Lawrence37:
Disabling the delete button when a "hard-coded" extension is selected would improve the user experience. The tooltip can explain why certain extensions cannot be removed from the list.

xiota:
@Lawrence37  Made some changes, somewhat based on your suggestion.  Left tooltip text alone because I don't know how it should be reworded.  The disabled/enabled buttons look the same on my computer with the default themes.  They are distinguishable with the TooWa themes.

* Disable addExt button when extension entry box is empty or matches existing extension
* Disable delExt button when selected extension is a known type
* Moved add-extension panel to top of frame to make accessible without scrolling
* Prepend new extensions to reduce need to scroll
    * I would like to insert at selected position, but don't know how.

Lawrence37:
Whatever the reason for the small difference for the disabled button's look, it is unique to these buttons. Other ones (favorites tab for example) have a clear distinction between the enabled and disabled states.

xiota:
Anymore changes needed for this?  I would like to rebase #7097 after this is merged.

I'm leaving tooltip rewording for someone else to do in another PR.

rtgui/preferences.cc
```diff
@@ -2668,25 +2675,69 @@ void Preferences::workflowUpdate()
...
+            delOkay = false;
+            break;
+        }
+    }
+    delExt->set_sensitive(delOkay);
+}
+
+void Preferences::extensionChanged()
 {
+    if (extension->get_text()[0] == '\0') {
```
Lawrence37: `!extension->get_text_length()` or `extension->get_text().empty()`

Lawrence37:
You can change `PREFERENCES_PARSEDEXTDELHINT` to `Delete selected extension from the list. Predefined extensions cannot be deleted.`. Deferring this change to another PR makes it likely that this will never be updated.

Issue #6773, opened 2023-06-16
JXL support


jurandfantom:
Hi there, I saw that similar request were made 3 times, but because topic ale lock im forced to make new one instead add my 3cents up there.

Apple recently announced their support for that format ( https://blog.desdelinux.net/apple-ya-admite-el-formato-jpeg-xl-en-safari-y-ademas-presento-un-port-de-wine/ ) with make more and more people to talk about it. Of course major player (Adobe) is still not supporting JXL, but I would suggest to place JXL support as at least low level feature (compare to not being considered at all). 
Darktable is supporting that format (8, 10, 12, 16 and 32 bit) so its possible to push it indeed. 

Other thing is, this is chance to at last push superior format into masses - all because of work from small but multiple communities that make people aware and willing to ask bigger players "hey why your product don't support that ? " List of softwares that would need implement support is long, but everything starts from processed RAW picture :)

Thank you for your work - happy to see open source projects Like RT, DarkT, Blender doing such great job in terms of promoting community driven solutions.

Lawrence37:
#6273 is still open.

kmilos:
> (Adobe) is still not supporting JXL

They do, in some way: https://helpx.adobe.com/camera-raw/using/hdr-output.html

PR #7097, opened 2024-06-11
Add ability to export JXL images

### LINKED ISSUE #6273: JPEG XL (*.jxl) file format support

xiota:
Closes #6273.  Related #6367, #6773.

Main problem is color management.  When saving with output profile other than sRGB (eg, P3):
* Thumbnails of the resulting images look wrong, but image is correct after import.
* Images look wrong when opened in other viewers or editors, like GIMP.

Potential improvements to add or leave for another PR...
* Force import/export profile to sRGB.  Is this possible?  How? **Need help.**
* Add Exif.  Need metadata as a binary blob to add as box.  Don't know how.  **Need help.**
    * Exiv2 doesn't support writing BMFF metadata.
* Update CLI for JXL conversion.  Currently seems to assume `-j*` options are for `jpeg`.

Following seem okay:
* Building with and without libjxl 0.10.2, and with `WITH_JXL=no`.
* Building with libjxl 0.7.x (via dispatch-only Ubuntu 24.04 workflow)
* Exporting with different quality settings.
* Batch processing.


rtengine/dcp.cc
```diff
@@ -1178,7 +1178,7 @@ DCPProfile::DCPProfile(const Glib::ustring& filename) :
     // Color Matrix (one is always there)
     if (!md.find(TAG_KEY_COLOR_MATRIX_1)) {
         if (settings->verbose) {
-            std::cerr << "DCP '" << filename.c_str() << "' is missing 'ColorMatrix1'. Skipped." << std::endl;
```
Floessie: @Lawrence37 Wasn't this a fix for something, lately?

rtengine/imageio.cc
```diff
@@ -1025,44 +1039,46 @@ int ImageIO::saveJXL(const Glib::ustring &fname, float quality) const
...
+    } else
+#endif
     if (JxlEncoderSetColorEncoding(enc.get(), &color_encoding) != JXL_ENC_SUCCESS) {
-        std::cerr << "Error: JxlEncoderSetColorEncoding failed" << std::endl;
-        return IMIO_CANNOTWRITEFILE;
+        std::cerr << "Warning: JxlEncoderSetColorEncoding failed" << std::endl;
     }
 
-    std::vector<std::uint8_t> imagebuffer(4 * 3 * width * height);
+    std::vector<std::uint8_t> imagebuffer((std::size_t) 3 * width * height * sizeof(float));
```
Floessie: No C-style casts, please.

Also, if this computation appears more than once, better create a new constant, like:

```c++
const std::size_t required_raw_line_size = sizeof(float) * 3 * width;
```

A `static_cast<std::size_t>` is not needed here, because the `sizeof` operator returns a `std::size_t` (C++11, §5.3.3 ¶6).

rtengine/dcp.cc
xiota: I had changed printf to cerr in jxl import, but didnt change any instances elsewhere. Changing them in this file now because missing newlines made output unreadable.  Should I open a separate PR to change all printf to cout/cerr ?

rtengine/imageio.cc
xiota: ~~CodeQL check fails without casts.  (See the new failure which includes `size_of()`)~~ Nevermind, will try reordering components.

rtengine/dcp.cc
Floessie: It's not the `printf()` conversion, that bothers me, but you killed the `.c_str()`...

rtengine/imageio.cc
Floessie: Yeah, but at the end.

rtengine/dcp.cc
xiota: Why do you need `.c_str()` with the stream operators?

Floessie: See 71f452aa1272eec0aa943ae579c6b5417fc204a6.

@Lawrence37 It would be better to add a comment to such odd changes, so they don't creep in again through well-intentioned "fixes".

rtengine/imageio.cc
Floessie: As I understand it, `3 * width` might already overflow `int`, so `int * int * size_t` doesn't help.  `size_t * int * int` -> `size_t * int` -> `size_t` on the other hand should solve the problem.

xiota:
Force push to rebase on `dev`, including #7104, #7096

Lawrence37:
Do you think the output profile issue is a problem with RawTherapee or GIMP? If it's a RawTherapee issue, then I think it should be addressed in this pull request unless the jxl specification says images without color profile information should be assumed to be in sRGB.

kmilos:
JPEG XL is very weird when it comes to color management (it takes on tasks it shouldn't IMHO) - if you do lossy compression, data is stored internally in float and "XYB" color space. The decoder can then decode it straight to any color space on the other end (sRGB usually), unless the "original" target profile is requested by the decoder (which had to be stored at encoding time of course, either as ICC or JPEG XL color space enum), and that is out of the control of the encoding app (which is the weird part, at least for me). For lossless compression, one can/should keep the "original" color profile throughout of course.

xiota:
The best solution at this time, for cross-application and cross-version libjxl, would probably be to force sRGB when exporting jxl, but I do not know how.  If images are intended only for use with RawTherapee with a single version of libjxl, the current state of this PR is very close to working as expected.

When exporting and importing across different versions of libjxl (0.7.x, 0.10.x), images may not display correctly.  A workaround for both RawTherapee and GIMP is to assign the correct profile after opening.

Cross-version problems may be related to 0.10.x converting ICC profiles to internal representation when saving, and not delivering an ICC profile when images use internal representation (color space enum).  So an image viewer, for instance, will display the image incorrectly because libjxl doesn't give it an ICC profile.

kmilos:
Why would you ever want to force sRGB?! It's a narrow gamut.

On export, you just need to supply libjxl w/ the profile of the data you're feeding it (ICC or enum).

> So an image viewer, for instance, will display the image incorrectly because libjxl doesn't give it an ICC profile.

No, it is because it didn't _ask_ for one from lbjxl upon decoding. That's why it's weird, it's not implicit.

xiota:
> Why would you ever want to force sRGB?! It's a narrow gamut.

For cross app compatibility.  Nothing I currently have installed works with non-sRGB jxl, except `djxl`.  jxl seems to use color completely differently from every other image format.

> On export, you just need to supply libjxl w/ the profile of the data you're feeding it (ICC or enum).

I'm feeding it ICC.  The images are coming out with enum set.

> No, it is because it didn't _ask_ for one from lbjxl upon decoding. That's why it's weird, it's not implicit.

What do I need to change to ask for one?  Right now, I'm expecting the decoder to tell me whether one is available.

https://github.com/Beep6581/RawTherapee/blob/13a07a58e2426791e5ea2cd6441ddd124731546d/rtengine/imageio.cc#L899-L936

kmilos:
Yes, that's the way. But in recent libjxl versions (can't remember when this behaviour was added exactly), that doesn't only mean an ICC profile was actually embedded at encoding - it just means the decoder can _generate_ and supply one on the fly, which it should do in most cases, as most internally stored enum combos are known:

https://github.com/libjxl/libjxl/blob/4a3b22d2600f92d8706fb72d85d52bfee2acbd54/lib/include/jxl/decode.h#L768-L791

xiota:
Using libjxl 0.10.x, sRGB and P3 JXLs exported with this PR seem to use JXL enums, no embedded ICC.  They are displayed correctly in RawTherapee with this PR (made adjustment to import), darktable, and GIMP 2.99/3.x.  The P3 JXL did not display correctly in GIMP 2.10.   Probably problem with the plugin or limitations of GIMP 2.10 libraries.  Workaround is to assign the correct profile after import.

Compiles with libjxl 0.7.x.  Can export and import images.  Issues mainly with non-sRGB profiles.  Not doing further testing because fixing 0.7.x seems to break 0.10.x, and vice versa.

Fixing thumbnail display could be left for another PR, but may be hopeless, especially with old libjxl.

To add metadata, Exif blobs are needed.  If someone tells me how to get them, I can add to this PR.  Otherwise, could leave for another PR.  Exiv2 does not support writing to JXL.

Updating CLI options, I could look at next, or leave for another PR.

0fbcb238c0:
> Compiles with libjxl 0.7.x. Can export and import images. Issues mainly with non-sRGB profiles. Not doing further testing because fixing 0.7.x seems to break 0.10.x, and vice versa.

0.11.x is expected in the near future according to https://github.com/libjxl/libjxl/issues/3700#issuecomment-2217476773.
Maybe that will work better?

xiota:
> 0.11.x is expected in the near future...  Maybe that will work better?

I tested RawTherapee built with libjxl 0.10.3, but run with libjxl-git 0.10.3.r227.g5ddb303a.  No difference in previously observed behavior.  Images open correctly with this PR, darktable, and GIMP 2.99/3.x.  I did not rebuild the applications because of how time consuming it would be, but don't think it matters.

If libjxl 11.x is released before this PR is ready for merge, should workflows be updated as part of this PR or in a separate PR? 

Wasn't paying close attention earlier, but darktable thumbnails are correct.  So it may be worth looking at for inspiration.

Somewhat related, had idea to use libjpegli to expand bit depth of imported jpegs. #7125

Lawrence37:
For adding EXIF, maybe [ExifParser::encode()](https://exiv2.org/doc/classExiv2_1_1ExifParser.html#a18fe7d067cddab8eb216125560c8459e) can help.

jonnyawsom3:
Thought I'd check in on this. If I'm reading it right, the main issues are color management for thumbnails and EXIF writing?

Lawrence37:
@xiota I saw your comment about needing pointers to solve the remaining issues. I suggested looking at `ExifParser::encode()` for metadata because that's what I saw in darktable.

Here's where they use exiv2 to write the metadata to bytes:
https://github.com/darktable-org/darktable/blob/f9951793d390f33599c0315615dd34d10d7774df/src/common/exif.cc#L2857

And here's where the bytes are attached to the JXL using libjxl:
https://github.com/darktable-org/darktable/blob/f9951793d390f33599c0315615dd34d10d7774df/src/imageio/format/jxl.c#L323

weaved-ja:
Is there any chance to see this completed some time soon? It would be great to be able to export JXL images for much smaller files and better image quality.

xiota:
Maybe AI assistance would help me figure out the remaining pieces.

kaesa:
@xiota I created new branch based on current dev, cherry picked your commits, execpt the Ubuntu workflow. I don't know anything about workflows. Added the metadata. I copied the technique from posted darktable links. I also added quite a few settings-verbose guards for printing, which of course also need a careful check.

https://github.com/kaesa/RawTherapee/tree/libjxl

You can rebase and cherry pick from me or what do you want to do?
Also, I didn't notice any problems opening the saved images with gimp or darktable even before the metadata was added.

xiota:
Thanks.  In the past there was something I didn't understand or had problems with.  I'll take a look, rebase, and cherry-pick if it works.

Alternately, I could rebase this PR, check it still works, request merge as-is... then @kaesa could open a new PR to fill the gaps, depending on your preference.

kaesa:
I pushed new commit. Unfortunately, implementing wasn't as easy as I yesterday thought. I hope it now saves all the metadata correctly. It seems to work well on my testing but unfortunately with my lack of complete understanding, I still have to say I'm hoping.

I did the initial work on Fedora 43 yesterday and today continued with Linux Mint 22. The libjxl differs on these systems and the images seem to open correctly with GIMP and darktable on both systems.

PR #7098, opened 2024-06-12
adobe_coeff constants from LibRaw

Lawrence37:
I pulled new constants from the LibRaw public snapshot 202403.

Wherever the camera naming seems to deviate from dcraw (Olympus cameras, for example), I used sample images to verify the correct name. Many Hasselblad cameras have very different names, so I did not verify them and left the constants commented out. 

LibRaw also has a different way of specifying the black level. For those that are not zero, I disabled `camconst.json` and opened sample images to see if the black level is reasonable. Panasonic cameras with a LibRaw black level of -15 become 15 in dcraw. Some had to be set to zero. A few Sony cameras with a LibRaw black level of 800 needed to be changed to 200 for dcraw.

The Leica M (Typ 262) and Leica TL color matrices are different now. The LibRaw constants for those cameras cover other cameras too (such as the M (Typ 240) and T (Typ 701)). I'm not sure what to make of this.

There are also other constants which have differences between dcraw and LibRaw. I will review those and determine if those need to be updated to align with LibRaw.


Lawrence37:
I'm going to trust the dcraw constants. I restored the color matrices for the Leica M (Type 262) and TL cameras. The difference between the dcraw and LibRaw matrices are small, and I don't want to make unnecessary changes.

Lawrence37:
I will merge this soon if there are no objections.

PR #5595, opened 2020-01-03
Move profiled distortion correction in pipeline

Lawrence37:
Resolves #5558 by applying distortion correction before auto-fill and perspective.

PR #7100, opened 2024-06-14
Lens correction based on metadata (Sony, Fuji, Olympus, Panasonic, DNG)

### LINKED ISSUE #2838: Support DNG WarpRectilinear opcode
Originally reported on Google Code with ID 2858

```
Hi,

I'm using RT for quite some time and I really like it. I'd like to post a feature request.


The idea is to support the WarpRectilinear opcode that some DNG files have. The opcode
specifies how the lens distortion should be corrected when demosaicing the DNG file.
Code for reading the opcode and correcting the distortion is available from Adobe as
part of the open source DNG SDK, in the utility dng_validate. Currently RT ignores
the opcode and therefore DNG files that contain it are uncorrected for distortion.

The ZIP file linked below includes the following files:
1_P7183225.ORF - Original RAW file in ORF format (taken with an Olympus E-M10 camera)
2_P7183225_OOC.jpg - Out of camera jpeg (downsized to 1MP)
3_P7183225.dng - DNG 1.4 file created with Adobe DNG Converter
4_P7183225_RT.jpg - Output of DNG file when converted to JPG in RT
5_P7183225_DNG_VALIDATE.tif - Output of DNG file when converted to TIFF using dng_validate.

ZIP (35MB):
https://dl.dropboxusercontent.com/u/21642925/RT_DNG_OPCODE.ZIP

As you can see, the out of camera jpeg and the file created by dng_validate are both
corrected for distortion, while the file created by RT is not. 

I tried to work with the DNG SDK in the past but it's a nightmare to get it to compile.
I didn't manage to build dng_validate locally. I'm also not familiar enough with RT's
code to implement such a feature by myself currently. I hope that this can be implemented
sometime in the future.

Thank you,
Assaf
```

Reported by `assaft` on 2015-07-26 20:03:33

### LINKED ISSUE #6076: Add an option to use lens corrections and flat fields embedded in image metadata
This would be great because many cameras store lens corrections in image metadata. Already present in ART

sgotti:
This PR implements lens correction based on file metadata for DNG and vendor (Sony, Fuji, Olympus, Panasonic) specific algorithms (many, if not all, reverse engineered).

This is based on my and other people same works done on darktable.

I created different self contained commits but due to the lack of github for dependant PRs i preferred to create an unique PR to easily tests all the work. 

**I warmly suggest to review the commits one by one to better understand every single change**

These are the commits titles in order:

* iptransform: move profile based CA with and after distortion
* iptransform: correct profile based distortion/CA before rotation
* iptransform: apply distortion and CA correction in a single pass
* lens profile: report availability of distortion and vignetting
* imagedata: report if file is DNG
* Add initial metadata lens correction handling
* lensmetadata: add abstract center radius helper class
* Add Sony metadata lens correction
* Add Fujifilm metadata lens correction
* Add Olympus metadata lens correction
* Add Panasonic metadata lens correction
* Add DNG metadata lens correction

### First commits
The first commit `iptransform: move profile based CA with and after distortion` tries to move CA together with distortion avoiding an additional step and related interpolation (this should also improve performances by removing an additional interpolation step).

The commit `iptransform: correct profile based distortion/CA before rotation` does the same fix done in #5595 for perspective correction also for rotation correction.
 
The commit `iptransform: apply distortion and CA correction in a single pass` is an enhancement useful for the corrections.

@Lawrence37 If you prefer I can open different PRs for these commits if you want to discuss them in deep and separately.

### Tested images

I tested all the various vendors with multiple images from raw.pixls.us or other images provided in the pixls.us forum in the past or from github issues that I already used with the darktable implementation. If you want I can provide some of them out of band.

### DNG

DNG OpcodesList1/2/3 are fully documented. Currently they aren't fully implemented in rawtherapee since they should be handled in different parts of the pipeline.
To add some story: OpcodeList1 is used in rawimagesource.cc to handle only GainMap, while OpcodeList2 is not handled.
Any OpcodeList1/2/3 can be an arbitrary list of mandatory or optional operations (also multiple GainMap, multiple WarpRectilinear). So this logic is not implemented by rawtherapee but only some specific cases.

The added lens correction uses OpcodeList3 (after demosaic). Like in DT only WarpRectilinear and FixVignetteRadial are implemented (WarpRectilinear2 and WarpFishEye are missing).
At the moment only the first WarpRectilinear and the first FixVignetteRadial is handled. This is enough to correct Leica DNGs.
I added some notes in the source.

### Future improvements and backward compatibility

Since the vendor ones are reverse engineered and DNG Opcode3 is not fully implemented in future there can be changes to the vendor algorithms (i.e. improved corrections, added Panasonic CA handling currently missing). This could cause some backward incompatible changes. This should be considered and handled when happening.
*If backward compatibility is very important some regressions tests like the DT integration tests will be really useful...*


sgotti:
@Lawrence37 Another big doubt is related to perspective correction done before lens correction introduced in #5595 . Logically lens correction should go before any other correction as it should work on the sensor output image (excluding 90xN degrees corrections).

For example if perspective rotation is applied, CA correction will be applied to a perspective corrected image (I avoided this for normal rotation in commit `iptransform: correct profile based CA right after distortion` but I left with the perspective correction doubt). The same for distortion, it should be applied to the out of camera image without any perspective correction.

So I'm not sure I totally understand the issue fixed by #5595 and if the cause was the order of the operations (distortion before perspective) or instead the cause was just the use of ascale during distortion correction that you removed.

During my tests I also tried moving distortion/ca correction before perspective and I saw some little differences but I cannot say if they were better or worse without a good test image.

Can you please shed some light and your thoughts on this?

sgotti:
@Lawrence37 I think this is now ready for review.

I removed my previous question since I now have a clearer vision (I forgot that the transformations are applied in reverse order since they map from the output/transformed image to the input image). Also updated the PR descriptions and the commits that move CA correction near/with distortion correction.

roth-m:
For Panasonic, I don't believe the reference/formula you mentioned will work for most lenses. It's even specified in on syscall.eu page that there are different formulas depending on the lens.
I believe Pana Exif coefficients are for Brown Conrady model and not Barel (I tried a few years ago to fit the exif coefficients to Barel model without success, but that is just my tries).
(See: https://github.com/darktable-org/darktable/pull/7092#issuecomment-2180741190 )

sgotti:
> For Panasonic, I don't believe the reference/formula you mentioned will work for most lenses. It's even specified in on syscall.eu page that there are different formulas depending on the lens.
> I believe Pana Exif coefficients are for Brown Conrady model and not Barel (I tried a few years ago to fit the exif coefficients to Barel model without success, but that is just my tries).
> (See: [darktable-org/darktable#7092 (comment)](https://github.com/darktable-org/darktable/pull/7092#issuecomment-2180741190) )

@roth-m definitely. It just works well (better than lensfun correction on short focus distances) for my Lumix S and Sigma DG DN (L-mount FF) lenses but it should be removed/commented out before merging. If you have more details (in the DT post I can't better understand your findings) to share or need some exif data from my lenses let me know. A dedicated place/issue could be better to discuss this.

Lawrence37:
I did some brief testing with images from the RPU. It works well for Sony and Fujifilm images in general. I noticed these issues:
- Olympus SP565UZ: Metadata option is selectable, but none of the corrections are available.
- Panasonic DC-FZ10002: Distortion correction looks wrong.
- Panasonic DMC-G80: Distortion correction looks insufficient.
- YI Technology M1 (dng): CA correction looks insufficient.

My findings support the concern regarding Panasonic images. The problem with the dng may just be bad correction parameters in the metadata.

sgotti:
> Olympus SP565UZ: Metadata option is selectable, but none of the corrections are available.

Good catch. Fixed.

> My findings support the concern regarding Panasonic images. 

For Panasonic I'm talking oob with @roth-m trying to find a common correction algorithm (with the current implemented one I have good corrections for my L-Mount lenses but it doesn't fit other mount/lenses).

> * YI Technology M1 (dng): CA correction looks insufficient.

Tested the unique image on raw.pixls.us. The algo looks right, it's just that the reported corrections aren't good (other programs does the same corrections with the same uncorrected CA).

rtengine/iptransform.cc
```diff
@@ -1213,52 +1205,56 @@ void ImProcFunctions::transformGeneral(bool highQuality, Imagefloat *original, I
...
-                    break;
-                case PerspType::CAMERA_BASED:
-                    const double w = p_matrix[3][0] * x_d + p_matrix[3][1] * y_d + p_matrix[3][3];
-                    const double xw = p_matrix[0][0] * x_d + p_matrix[0][1] * y_d + p_matrix[0][3];
-                    const double yw = p_matrix[1][0] * x_d + p_matrix[1][1] * y_d + p_matrix[1][3];
-                    x_d = xw / w;
-                    y_d = yw / w;
-                    break;
-            }
+            for (int c = 0; c < (doCACorrection ? 3 : 1); ++c) {
```
Lawrence37: Why iterate through `c` here? It seems it is not needed until after rotation, possibly resulting in unnecessary re-computation of scale, perspective, and rotate if the compiler doesn't know how to optimize.

rtengine/lcp.cc
```diff
@@ -985,11 +985,27 @@ rtengine::LCPMapper::LCPMapper(
     isFisheye = pProf->isFisheye;
 }
 
-bool rtengine::LCPMapper::isCACorrectionAvailable() const
+bool rtengine::LCPMapper::hasDistortionCorrection() const
+{
+    return true;
+}
```
Lawrence37: Add a comment explaining why this always returns true.

rtengine/lcp.cc
```diff
@@ -985,11 +985,27 @@ rtengine::LCPMapper::LCPMapper(
...
+
+bool rtengine::LCPMapper::hasCACorrection() const
 {
     return enableCA;
 }
 
+bool rtengine::LCPMapper::hasVignettingCorrection() const
+{
+    return true;
+}
```
Lawrence37: Same here

rtengine/lensmetadata.h
```diff
@@ -0,0 +1,88 @@
...
+ *  You should have received a copy of the GNU General Public License
+ *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
+ */
+#pragma once
+
+#include "lcp.h"
+#include "procparams.h"
+#include "rtengine.h"
+#include "metadata.h"
+#include <memory>
```
Lawrence37: Sort these includes alphabetically and place the standard library include at the top in its own section.

rtgui/lensprofile.cc
```diff
@@ -144,17 +146,18 @@ LensProfilePanel::LensProfilePanel() :
...
-    modesGrid->attach(*lensfunLenses, 1, 4, 1, 1);
+    modesGrid->attach(*corrMetadata, 0, 1, 3, 1);
+    modesGrid->attach(*corrLensfunAutoRB, 0, 2, 3, 1);
+    modesGrid->attach(*corrLensfunManualRB, 0, 3, 3, 1);
+
+    modesGrid->attach(*lensfunCamerasLbl, 0, 4, 1, 1);
+    modesGrid->attach(*lensfunCameras, 1, 4, 1, 1);
+    modesGrid->attach(*lensfunLensesLbl, 0, 5, 1, 1);
+    modesGrid->attach(*lensfunLenses, 1, 5, 1, 1);
     modesGrid->attach(*warning, 2, 3, 1, 2);
```
Lawrence37: The warning should be bumped down too.

rtengine/lensmetadata.h
```diff
@@ -0,0 +1,88 @@
...
+    void correctDistortionAndCA(double &x, double &y, int cx, int cy, int channel) const override;
+    void correctDistortion(double &x, double &y, int cx, int cy) const override;
+    void correctCA(double &x, double &y, int cx, int cy, int channel) const override;
+    void processVignette(int width, int height, float **rawData) const override;
+    void processVignette3Channels(int width, int height, float **rawData) const override;
+
+    void processVignetteNChannels(int width, int height, float **rawData, int channels) const;
+    void initCorrections(int width, int height, const procparams::CoarseTransformParams &coarse, int rawRotationDeg);
+
+    /* Implementors should implement the blow methods */
```
Lawrence37:
```suggestion
    /* Implementers should implement the below methods */
```

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+    {
+        knots.resize(nc);
+        dist.resize(nc);
+        vig.resize(nc);
+        for (int i = 0; i < 3; ++i) {
+            ca[i].resize(nc);
+        }
+
+        for (int i = 0; i < this->nc; i++) {
+            knots[i] = (double)(i + 0.5) / (nc - 1);
```
Lawrence37: It's already a `double`

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+    int nc;
+    std::array<short, 16> distortion;
+    std::array<short, 16> ca_r;
+    std::array<short, 16> ca_b;
+    std::array<short, 16> vignetting;
+
+    std::vector<double> knots;
+    std::vector<double> dist;
+    std::array<std::vector<double>, 3> ca;
+    std::vector<double> vig;
```
Lawrence37, line 202: These can be private. Same for the other classes.

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+*/
+
+class SonyMetadataLensCorrection : public CenterRadiusMetadataLensCorrection
+{
+public:
+    int nc;
+    std::array<short, 16> distortion;
+    std::array<short, 16> ca_r;
+    std::array<short, 16> ca_b;
+    std::array<short, 16> vignetting;
```
Lawrence37, line 197: `MAXKNOTS` instead of 16?

rtengine/lensmetadata.cc
```diff
@@ -278,10 +280,212 @@ class SonyMetadataLensCorrection : public CenterRadiusMetadataLensCorrection
...
+        knots_dist.resize(MAXKNOTS);
+        dist.resize(MAXKNOTS);
+        for (int i = 0; i < 3; ++i) {
+            ca[i].resize(MAXKNOTS);
+        }
+
+        // convert from spline related to source image (input is source image
+        // radius) to spline related to dest image (input is dest image radius)
+        for (int i = 0; i < MAXKNOTS; i++) {
+            const double rin = (double)i / (double)(nc - 1);
```
Lawrence37: `static_cast`, or use `1.0`

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+        auto &exif = metadata.exifData();
+
+        std::array<double, 4> distortion;
+        std::array<double, 6> cacorr;
+
+        auto it = exif.findKey(Exiv2::ExifKey("Exif.OlympusIp.0x150a"));
+        if (it != exif.end() && it->count() == 4) {
+            for (int i = 0; i < 4; ++i) {
+                const float kd = it->toFloat(i);
+                distortion[i] = kd;
```
Lawrence37: Why use a variable?

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+                has_dist = true;
```
Lawrence37: Move `has_dist = true;` outside the loop

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+            }
+            it = exif.findKey(Exiv2::ExifKey("Exif.OlympusIp.0x150c"));
+            if (it != exif.end() && it->count() == 6) {
+                for (int i = 0; i < 6; ++i) {
+                    const float kc = it->toFloat(i);
+                    cacorr[i] = kc;
```
Lawrence37: Same question with the variable

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+                }
+                has_ca = true;
+            }
+
+            if (has_dist) {
```
Lawrence37: `has_dist` is always true here

rtengine/lensmetadata.cc
```diff
@@ -481,11 +481,140 @@ class FujiMetadataLensCorrection : public CenterRadiusMetadataLensCorrection
...
+        }
+    }
+
+    double distortionCorrectionFactor(double rout) const override
+    {
+        // The distortion polynomial maps a radius Rout in the output
+        // (undistorted) image, where the corner is defined as Rout=1, to a
+        // radius in the input (distorted) image, where the corner is defined
+        // as Rin=1.
+        // Rin = Rout*dk0 * (1 + dk2 * (Rout*dk0)^2 + dk4 * (Rout*dk0)^4 + dk6 * (Rout*dk0)^6)
```
Lawrence37: Is `dk0` supposed to be `drs`?

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+
+        auto &exif = metadata.exifData();
+
+        auto it = exif.findKey(Exiv2::ExifKey("Exif.PanasonicRaw.0x0119"));
+        if (it != exif.end()) {
+            std::vector<Exiv2::byte> buf;
+            buf.resize(it->value().size());
+            it->value().copy(&buf[0], Exiv2::littleEndian);
+
+            const Exiv2::byte *data = &buf[0];
```
Lawrence37: `buf.data()` instead of `&buf[0]`?

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+            b = Exiv2::getShort(data + 8, Exiv2::littleEndian) / 32768.0f;
+            c = Exiv2::getShort(data + 22, Exiv2::littleEndian) / 32768.0f;
+
+            has_dist = true;
+        }
+    }
+
+    double distortionCorrectionFactor(double rout) const override
+    {
+        const double rin = scale * (rout + a * std::pow(rout, 3) + b * std::pow(rout, 5) + c * std::pow(rout, 7));
```
Lawrence37: You can use [Horner's method](https://en.wikipedia.org/wiki/Horner%27s_method) to avoid all the `std::pow()`s.

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+        const double cf = rout / rin;
```
Lawrence37, line 673: Interesting that the formula calculates the undistorted radius instead of the other way around. This could lead to bad results if the distortion is strong.

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+// be done.
+// Currently we only handle a subset of all the available opcodes and only one
+// WarpRectilinar for distortion and FixVignetteRadial for vignetting (that's
+// usually the case with Leica DNGs and ADC generated DNGs).
+// This should be extended to support more exotic operations lists (i.e.
+// multiple WarpRectilinear)
+class DNGMetadataLensCorrection : public MetadataLensCorrection
+{
+public:
+    const FramesMetaData *meta;
```
Lawrence37: Unused field

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+        auto &exif = metadata.exifData();
+
+        auto it = exif.findKey(Exiv2::ExifKey("Exif.SubImage1.OpcodeList3"));
+
+        if (it != exif.end()) {
+            std::vector<Exiv2::byte> buf;
+            buf.resize(it->value().size());
+            it->value().copy(&buf[0], Exiv2::invalidByteOrder);
+
+            const Exiv2::byte *data = &buf[0];
```
Lawrence37: `buf.data()` instead of `&buf[0]`?

rtengine/lensmetadata.cc
```diff
@@ -689,6 +689,308 @@ class PanasonicMetadataLensCorrection : public CenterRadiusMetadataLensCorrectio
...
+    double cx_v;
+    double cy_v;
+    double m_d;
+
+    int planes;
+
+    bool has_dist, has_ca, has_vign;
+    std::array<std::array<double, 6>, 3> warp_rectilinear;
+    std::array<double, 5> vignette_radial;
+    std::set<int> processed_opcodes;
```
Lawrence37: `processed_opcodes` can be a local variable.

rtengine/lensmetadata.cc
```diff
@@ -0,0 +1,1035 @@
...
+                    has_dist = true;
+                    if (planes == 3) {
+                        has_ca = true;
+                    }
+
+                    // we currently handle only one vignetting correction
+                } else if (opcodeID == 3 && !has_vign) { // FixVignetteRadial
+                    size_t start = idx;
+                    size_t end = idx + 7 * 8;
+                    if (end <= buf.size()) {
```
Lawrence37: Should this be `>`?

Lawrence37 (commented):
I reviewed all the code. Let me know when the Panasonic correction has been sorted out or disabled. I'm hoping to include this in 5.11.

FYI GitHub makes it hard to review changes after rebasing, or maybe there's a trick that I don't know about?

rtengine/iptransform.cc
sgotti: I was waiting for a review on the logic before optimizing. I assume that the logic of the first 3 commits looks good to you so I updated them.

rtengine/lensmetadata.cc
sgotti: These are the max sony knots in metadata. While MAXKNOTS is the number of knots to use to compute the inverted fuji spline. So it's a concidence that they are 16. I'll move MAXKNOTS inside the fuji class since it's used only there to avoid such confusion.

rtengine/lensmetadata.cc
sgotti: indeed there was a wrong nesting. Fixed.

rtengine/lensmetadata.cc
sgotti: indeed there was a wrong nesting. Fixed.

rtengine/lensmetadata.cc
sgotti: Yes, copied from darktable comment which has the same wrong naming.

rtengine/lensmetadata.cc
sgotti: Right. It checks that the expected end is inside the buffer size.

sgotti:
> I reviewed all the code. Let me know when the Panasonic correction has been sorted out or disabled. I'm hoping to include this in 5.11.

To include this PR in 5.11 I'd prefer to disable the Panasonic correction and then we could easily add the right implementation if it'll be found before or after 5.11.

Let me know how you prefer to disable it:
* disable it in MetadataLensCorrectionFinder
* disable it in the PanasonicMetadataLensCorrection constructo
* Just remove the whole panasonic commit
* something else

> 
> FYI GitHub makes it hard to review changes after rebasing, or maybe there's a trick that I don't know about?

I prefer to do a development process like the linux kernel one with a patchset composed of one or multiple commits with self contained in logical changes instead of piling up tons of commits for every change, fix, review, improvements etc...

This makes review difficult with the current github web interface. I usually use `git range-diff` to compare changes of a patch series/pull request.

Lawrence37:
Unfortunately, `git range-diff` only works if you have all the commits. I don't see any way to fetch the commits from before the rebase.

rtengine/lensmetadata.h
Lawrence37: "Implementers" is still spelled wrong

Lawrence37:
Thanks for addressing the review comments. There are two remaining comments (excluding the one about the Panasonic model).

As long as the rebase is not changing the base commit, GitHub can show the changes just fine. The problem comes when rebasing onto a new base, like when adding new commits from `dev`. The GitHub comparison gets cluttered by the new `dev` commits.

Lawrence37:
> Let me know how you prefer to disable it

Any of these would be fine. I would do it in the constructor since that's closest to where the to-do is.

rtengine/lensmetadata.cc
sgotti: Right. At the end the current formula is just a not perfect guess.
Note that something similar happens also for the fuji one. The provided spline in metadata calculates the undistored radius from the distorted radius and we have to reverse it.

Lawrence37: If this turns out to be the correct formula for some lenses, then we can improve accuracy with a lookup table of the inverse function. Best to leave this until after figuring out the correction model(s) though.

rtengine/lensmetadata.cc
Lawrence37: One more comment here

Lawrence37:
Hi @sgotti, there's one more comment in the Panasonic correction regarding the use of `&buf[0]`. I'd like to get this merged and resolve any merge conflicts in the LibRaw PR, then enter a feature freeze. If you are busy, let me know if you want me to address the final comment.

sgotti:
@Lawrence37 sorry I'm on vacation. Feel free to address it. Thanks!

PR #7104, opened 2024-06-16
Change order of image/raw loading to generate thumbnails

### LINKED ISSUE #7103: Some jpeg files are treated as raw files
**Steps to reproduce**
1. Apply the Neutral profile to https://raw.pixls.us/getfile.php/3648/nice/Nikon%20-%20Z%2050%20-%2012bit%20compressed%20(3:2).NEF and save a jpeg.
2. View the jpeg in the file browser. The thumbnail does not render correctly.
3. Open the jpeg in the editor. Notice the file is treated as a raw file.

**Additional information**
Possibly introduced in #6367 because it changed how files are decoded. It attempts to load images as raw files first, then tries other formats upon failure.

xiota:
Fixes #7103

### DIFF
--- a/rtgui/thumbnail.cc
+++ b/rtgui/thumbnail.cc
@@ -248,44 +248,44 @@ void Thumbnail::_generateThumbnailImage()
     cfs.exifValid = false;
     cfs.timeValid = false;
 
-    // RAW works like this:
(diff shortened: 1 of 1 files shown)


Lawrence37 (approved):
Looks good to me. Thanks for the quick fix @xiota!

PR #7105, opened 2024-06-16
Remove Itanium architecture support for Windows

Lawrence37:


Lawrence37:
This pull request removes the ability to create a Windows installer for the Intel Itanium architecture. As of version 6.3.0, Inno Setup does not support Itanium (see https://jrsoftware.org/files/is6-whatsnew.htm).

PR #7107, opened 2024-06-19
Keep values scope from 5.10 Selective Editing - for color & light - shadow/highlight - vibrance #7102

### LINKED ISSUE #7102: Default Local Adjustments Scope from 5.10 not used in dev
**Short description**
When applying a pp3 from 5.10 with the Local Adjustments color tools Scope set to the default value of 30, the scopes of the individual tools are not set to 30.

**Steps to reproduce**
1. In 5.10, start with the Neutral profile.
2. Create a Local Adjustments spot and add the Color & Light tool.
3. Save the pp3.
4. In dev, apply the pp3. Observe the Color & Light tool's Scope is set to 15.

**Additional information**
The scope is properly set if it was saved with a value other than 30. Applying pp3s from dev works as expected. The issue only happens when applying a pp3 with the "Scope (color tools)" set to 30.

Desmis:
I replace a complex manner (in improccoordinator) to manage this problem, by using ppversion (<=350)  in procparams

In 5.10 and before, les 3 scopes for Color &  Light, Shadows/Highlight, Vibrance were grouped and used the value of scope in settings

Now, with the new manner for manage spot, each tool has its own scope.

Issue #7102

### DIFF
--- a/rtgui/ppversion.h
+++ b/rtgui/ppversion.h
@@ -1,11 +1,13 @@
 #pragma once
 
 // This number has to be incremented whenever the PP3 file format is modified or the behaviour of a tool changes
-#define PPVERSION 350
+#define PPVERSION 351
 #define PPVERSION_AEXP 301 //value of PPVERSION when auto exposure algorithm was modified
 
 /*
   Log of version changes
+   351  2024-06-19
+        take into account Global in selectibe editing
    350  2023-03-05
         introduced white balance standard observer
    349  2020-10-29
--- a/rtgui/ppversion.h
+++ b/rtgui/ppversion.h
@@ -7,7 +7,7 @@
 /*
   Log of version changes
    351  2024-06-19
-        take into account Global in selectibe editing
+        take into account Global in selective editing
    350  2023-03-05
         introduced white balance standard observer
    349  2020-10-29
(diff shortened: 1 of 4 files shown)


rtgui/ppversion.h
```diff
@@ -1,11 +1,13 @@
...
 
 // This number has to be incremented whenever the PP3 file format is modified or the behaviour of a tool changes
-#define PPVERSION 350
+#define PPVERSION 351
 #define PPVERSION_AEXP 301 //value of PPVERSION when auto exposure algorithm was modified
 
 /*
   Log of version changes
+   351  2024-06-19
+        take into account Global in selectibe editing
```
Lawrence37:
```suggestion
        take into account Global in selective editing
```

Lawrence37:
Now all the tools that used the common color scope always have the value 30. Here's a pp3 with the color scope set to 37.
[scope.pp3.txt](https://github.com/user-attachments/files/15948340/scope.pp3.txt)

rtgui/ppversion.h
Desmis: I just change this text

Thank you

Desmis:
@Lawrence37 
Then i Will merge this change to morrow, if I solve the problem with value 30... what is not won 

Jacques

Desmis:
@Lawrence37 

I think maybe I solved the problem? For me these are things that are not obvious. Can you check and if necessary make changes.
Then if it works, it's the same thing I should do for "feather" #7113

Thank you

Jacques

Lawrence37:
Works fine now! Feel free to merge.

Desmis:
@Lawrence37 

Thank you. I will merge soon.
Now, with a same change I will fixe issue #7113 and #7114

Jacques

PR #7122, opened 2024-07-03
Selective editing - modifies the activation of the Laplacian according to the spots with Dehaze

### LINKED ISSUE #7121: Dynmamic range and exposure spot crashes a particular edit
Using RawTherapee Dev 5.10-298 and editing the RAF image in this thread https://discuss.pixls.us/t/between-you-me-and-the-lamp-post/44334 
This pp3 in this post https://discuss.pixls.us/t/between-you-me-and-the-lamp-post/44334/28?u=chaimav crashes RT. Spot is named appropriately

crash

Desmis:
#7121


rtengine/iplocallab.cc
```diff
@@ -20521,6 +20536,37 @@ void ImProcFunctions::Lab_Local(
...
+    float epsi = 0.000001f;
+    if((lp.laplacexp > 1.f && lp.exposena) || (lp.strng > 2.f && lp.sfena)) {//clip value above 65535.f and > epsilon when Contrast attenuator with high values Laplacian or Original Retinex 
+#ifdef _OPENMP
+        #pragma omp parallel for
+#endif
+            for (int i = 0; i < bh; ++i)
+                for (int j = 0; j < bw; ++j) {
+                    prov1->r(i, j) = clipR((float) rtengine::max(prov1->r(i, j), epsi));
+                    prov1->g(i, j) = clipR((float) rtengine::max(prov1->g(i, j), epsi));
+                    prov1->b(i, j) = clipR((float) rtengine::max(prov1->b(i, j), epsi));
```
Lawrence37:
```suggestion
                    prov1->r(i, j) = clipR(rtengine::max(prov1->r(i, j), epsi));
                    prov1->g(i, j) = clipR(rtengine::max(prov1->g(i, j), epsi));
                    prov1->b(i, j) = clipR(rtengine::max(prov1->b(i, j), epsi));
```

rtengine/iplocallab.cc
```diff
@@ -20521,6 +20536,37 @@ void ImProcFunctions::Lab_Local(
...
+                }
+    } else {//standard case only with small values Laplace no clip
+#ifdef _OPENMP
+        #pragma omp parallel for
+#endif
+            for (int i = 0; i < bh; ++i)
+                for (int j = 0; j < bw; ++j) {
+                    prov1->r(i, j) = (float) rtengine::max(prov1->r(i, j), epsi);
+                    prov1->g(i, j) = (float) rtengine::max(prov1->g(i, j), epsi);
+                    prov1->b(i, j) = (float) rtengine::max(prov1->b(i, j), epsi);
```
Lawrence37:
```suggestion
                    prov1->r(i, j) = rtengine::max(prov1->r(i, j), epsi);
                    prov1->g(i, j) = rtengine::max(prov1->g(i, j), epsi);
                    prov1->b(i, j) = rtengine::max(prov1->b(i, j), epsi);
```

rtengine/procparams.cc
```diff
@@ -9229,7 +9229,22 @@ int ProcParams::load(const Glib::ustring& fname, ParamsEdited* pedited)
...
+                                    spot.complexexpose = 0;//set to Advanced
+                                    spotEdited.complexexpose = true;
+                                }
+                            }
+                        }
+                } else {
+                    assignFromKeyfile(keyFile, "Locallab", "Laplacexp_" + index_str, spot.laplacexp, spotEdited.laplacexp);
+                    assignFromKeyfile(keyFile, "Locallab", "Complexexpose_" + index_str, spot.complexexpose, spotEdited.complexexpose);
+                }
+
```
Lawrence37: Is the only goal to change the tool complexity to "advanced" if contrast attenuator is enabled and the pp3 is from 5.10 or before? If so, wouldn't this simple code work?

```suggestion
                assignFromKeyfile(keyFile, "Locallab", "Laplacexp_" + index_str, spot.laplacexp, spotEdited.laplacexp);
                assignFromKeyfile(keyFile, "Locallab", "Complexexpose_" + index_str, spot.complexexpose, spotEdited.complexexpose);
                if (ppVersion <= 350 && spot.laplacexp > 0.f) { // Contrast attenuator moved to "advanced" after 5.10. Set complexity to "advanced" if Contrast attenuator is in use.
                    spot.complexexpose = 0;
                    spotEdited.complexexpose = true;
                }

```

rtengine/iplocallab.cc
```diff
@@ -20521,6 +20536,37 @@ void ImProcFunctions::Lab_Local(
             }
         }
     }
+    //verify that RGB values are > 0.f issue 7121 to avoid crash
```
Lawrence37: Which part of the code is crashing if values are not positive? Shouldn't that part of the code clip the RGB values? That would prevent crashes caused by other tools that we do not know about yet.

rtengine/iplocallab.cc
```diff
@@ -18000,8 +18006,15 @@ void ImProcFunctions::Lab_Local(
...
-
+                    bool execlaplace = true;//verify if others spot use dehaze
+                    /*
+                    const int sizespot = (int)params->locallab.spots.size();
+                    for (int sp = 0; sp < sizespot; sp++) {
+                        if(params->locallab.spots.at(sp).dehaz){//if true desable Laplace
+                            execlaplace = false;
+                        }
+                    }
+                    */
```
Lawrence37: It's ok to delete this code completely and remove the `execlaplace` variable.

rtengine/iplocallab.cc

rtengine/iplocallab.cc

rtengine/procparams.cc
Desmis: ok, it is better now
Thank you

rtengine/iplocallab.cc

rtengine/iplocallab.cc
Desmis: I checked possible crash cases with negative values. 

There is of course Laplacians, but also RGB curves (Tab Color main).
I found other cases where negative values ​​cause a crash : HSV curve, Channel mixer, Color Toning, Tone curves 1 and 2...

There were also, but this is handled calls to workingtrc  (Abstract Profile and Selective editings).

I think it is "better" not to clip the negative RGB values, knowing that with Prophoto as a "working profile" we are in imaginary colors. But since the user can change the working profile, it's better this way.

If new crashes appear (with negative values) then these cases will have to be added gradually: which function causes the crash and therefore activate the variable "notzero".

Additionally, to save time and memory, I only call functions if necessary.

Thank you

Jacques

Lawrence37: Ok. Where was the crash happening? I'm not able to reproduce the crash, but I'm curious.

Desmis:
@Lawrence37 
The chosen image DSCF9835.RAF is a treasure for achieving slightly unusual RGB values. With other images I had trouble getting crashes. But with this and the stacking of processing before the RGB module, we end up with negative values, or zero. These are amplified by the use of Laplacians.
I think but I'm not 100% sure that many uses with dehaze overlap zones result in either zero or negative values. The Laplacian increasing this phenomenon.

The majority (all?) of crashes are in the RGB module (rgbproc in improcfun.cc) and in particular all modules using curves: RGB curves, HSV curves, Tone curves (in exposure), but also in Channel Mixer and ColorToning.  
These modules should crash as soon as RGB <=0 in all cases. Probably this image is a reveal of an old bug but also of having moved Local Adjustment (Selective Editing) in the process before RGB.

With this image and the planned settings I deactivated each of the activated functions one by one and reactivated them until I got to the crash and I fond RGB curves (If I disable Laplacian no crash). I tried the other RGB or Lab functions and thus found the functions which caused a crash(HSV, Tone curve, etc.) .I didn't encounter any other crashes. which doesn't mean there won't be others.

Thank you for testing

Jacques

Lawrence37:
Which line of code was crashing?

Desmis:
@Lawrence37 
I didn't look for which line(s) of code caused the crash, just the function. All in rgbproc whether for negative values ​​or for those (with the Laplacian) greater than 65535. Crashes occur on TIF output (segmentation fault). I didn't have any crashes in preview mode
I also wonder why in 2024, we are still keeping this complex code (rgbproc) with tiles which complicates things and does not bring many advantages. In ART this function is removed.
jacques


Before making this comment, I tried working with debug and gdb?
I'm not a specialist in this kind of thing, but we see that some LUTs don't have an index (nan). This probably comes from the RGB2Lab conversion, but it must contain incorrect data coming from the functions used in rgbproc (rgb curves, tonecurves, etc.).
Hence my proposal to say it is function "X" which causes these problems, so let's find a remedy.
But maybe we can do otherwise?

Jacques

Lawrence37:
> I also wonder why in 2024, we are still keeping this complex code (rgbproc) with tiles which complicates things and does not bring many advantages. In ART this function is removed.

Tiles improve performance when the processing is not CPU intensive. rgbproc has a lot of simple operations, so it benefits greatly from tiles. ART also has a restructured pipeline. I don't know how much work it will require to adopt something similar.

> I'm not a specialist in this kind of thing, but we see that some LUTs don't have an index (nan). This probably comes from the RGB2Lab conversion, but it must contain incorrect data coming from the functions used in rgbproc

Thanks for the hint. I wouldn't say some functions are creating incorrect data. Negative values and values above 100% (especially) can be valid. The problem is that some tools only work if the input is in a certain range. For such tools, I believe it is the responsibility of the tool to preprocess the input or handle out-of-range values gracefully. It is especially important for tools that have a range smaller than 0-100%. This issue is a perfect example. Values of 0 are problematic even though 0 is typically considered to be valid.

Lawrence37:
Looks good to me. I still prefer addressing the issue at where the crash occurs, but I'm ok with merging this.

PR #7128, opened 2024-07-14
Language file updates for 5.11

Lawrence37:
Reviewed `default` and ran `generateTranslationDiffs` in preparation for the release.


Desmis:
@Lawrence37 

Thank you for these changes, syntax and conventions. Each language has its rules, French is obviously different.

PR #7139, opened 2024-07-20
Mac: clarify installation instructions

### LINKED ISSUE #6580: OSX: Fails to load when not in "/Application"
**Short description**

RawTherapee.app fails to load when it is not installed directly in the system directory "/Applications", e.g. if user has it installed in "~/Applications" or on a separate partition.

On initial installation and the subsequent launch OS X's _CoreServiceUIAgent_ reports `File /Volumes/Storage/Applications/RawTherapee.app/Contents/MacOS/bin/rawtherapee-bin failed on rPathCmd /Applications/RawTherapee.app/Contents/Frameworks/libomp.dylib` and `Fails dylib check`, afterwards RawTherapee.app via com.apple.xpc.launchd: `Service exited with abnormal code: 126`.

Moving the application bundle to `/Applications` fixes things. But moving it out again and it will continue to  `Service exited with abnormal code: 126`, albeit the inital report by _CoreServiceUIAgent_ no longer seems to happen.

Setting up an ugly softlink fixes the issue too:
```
$ cd "/Applications"
$ ln -s /Volumes/Storage/Applications/RawTherapee.app RawTherapee.app
```

**Steps to reproduce**
* Install the application in a path that is NOT "/Applications".
* Launch it.

**Expected behavior**
Installing the application in any user controlled path should not block it from launching, e.g. the user may not have write permissions in the system "/Applications" directory.

**Additional information**
- Version of RawTherapee: 5.8.0
- Version of your operating system: latest OS X 10.10.5 (14F2511)

Other useful information:
n/a

Benitoite:
Resolves #6580
Per @databu https://github.com/Beep6581/RawTherapee/issues/6580#issuecomment-2220407772
Clarifies that app build must be installed to `/Applications`


databu (approved):

Benitoite:
Sorry @databu I accidentally deleted your review.  Thanks for your input.

databu (approved):

databu:
No problem 🙂

tools/osx/INSTALL.readme.rtf
```diff
@@ -1,7 +1,7 @@
 To install the RawTherapee application:
 
 1. Open the RawTherapee .dmg Disk Image included alongside this document.
-2. Drag the RawTherapee icon onto the /Applications folder icon
+2. To install RT, drag icon onto /Applications (it will not start if in another location).
```
Lawrence37: "To install RT" is redundant due to the first line.

"RT" should be expanded to the full name. (I'm going by the RawPedia guidelines at https://rawpedia.rawtherapee.com/Contributing: "_Write "RawTherapee" instead of "RT" or "Rawtherapee". Never abbreviate "RawTherapee"._")

I think it can be more clear to the user that the installation location is a requirement, not a limitation. Instead of "it will not start if in another location", something to the effect of "RawTherapee MUST be installed in /Applications" would work. You can elaborate if you want, e.g. "For security reasons, RawTherapee MUST be installed in /Applications. RawTherapee will refuse to start if installed elsewhere.".

Benitoite: Thanks @Lawrence37 for the tip. Updated PR per your suggestions.

Lawrence37 (approved):

PR #7161, opened 2024-07-29
FramesData: add missing isDNG initialization.

sgotti:
FramesData isDNG wasn't initialized.

This will randomly cause images to be considered as dng files when the underlying value is different than 0 since the metadata parsing could exit before the code that checks for the Exif.Image.DNGVersion tag.


Lawrence37 (approved):
Nice catch

sgotti:
> Nice catch

It was my fault 😅

PR #7168, opened 2024-08-04
Selective Editing icon

### LINKED ISSUE #7129: New icon for Selective Editing
The current icon for the Selective Editing tab is the open hand: hand-open
To avoid confusion, the toolbar icon for the hand tool was changed to the high contrast version: hand-open-hicontrast

The problem is that the high contrast icon doesn't fit in with the other icons. I propose restoring the hand tool icon and giving the Selective Editing tab a new icon. The RT spot is an important aspect of Selective Editing, so my idea for an icon is an RT spot:
rt-spot

Lawrence37:
Adds a new icon for Selective Editing and reverts the hand tool icon back to the normal version instead of the high-contrast version. The new icon is a rectangular RT spot, similar to the one shown in https://github.com/Beep6581/RawTherapee/issues/7129#issuecomment-2241265977.
The actual icon: 
Here's how it looks in the GUI with the RawTherapee theme:
|Scale|Screenshot|
|-|-|
|Standard DPI|image|
|HiDPI (2x)|image|

I made the lines thicker and reduced the opacity to 70% to better match the style of the other tab icons. This is a compromise icon for 5.11 until better icons can be made for the Selective Editing and possibly Advanced tabs.

@Benitoite @Desmis @marter001 @rebio

### DIFF
--- a/rtdata/icons/rawtherapee/scalable/apps/rt-spot.svg
+++ b/rtdata/icons/rawtherapee/scalable/apps/rt-spot.svg
@@ -1,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="24px"
+   height="24px"
+   viewBox="0 0 24 24"
+   version="1.1"
+   id="SVGRoot"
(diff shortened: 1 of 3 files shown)


Desmis:
@Lawrence37 
For me it's good :)

Thank you

PR #7182, opened 2024-08-16
Mac: update copyright year

Benitoite:
Fix fancy dmg graphics
use `@version@` for versioning.


tools/osx/rtdmg-bkgd.png
Lawrence37, line 1: This image has the extra `.` after `/Applications`.

Benitoite: Good catch.

PR #7205, opened 2024-09-22
Gamut compression  - reduce artifacts resulting from out of gamut

Desmis:


rtdata/languages/default
```diff
@@ -1404,7 +1404,17 @@ HISTORY_MSG_COLORTONING_LABREGION_SHOWMASK;CT - region show mask
 HISTORY_MSG_COLORTONING_LABREGION_SLOPE;CT - region slope
 HISTORY_MSG_COMPLEX;Wavelet complexity
 HISTORY_MSG_COMPLEXRETI;Retinex complexity
-HISTORY_MSG_DEHABLACK;Dehaze Bayer Black
+HISTORY_MSG_CG_COLORSPACE;Gamut Compression - Target
+HISTORY_MSG_DEHABLACK;Gamut Compression - Target Gamut
```
Lawrence37: `HISTORY_MSG_DEHABLACK` was changed by mistake

rtengine/color.cc
```diff
@@ -2105,6 +2106,162 @@ float Color::eval_ACEScct_curve(float x, bool forward)
 
 // end code take in ART thanks to Alberto Griggio
 
+//functions needs to use ACES 
+
+//transpose Matrix
+void Color::transpose(Matrix ma, Matrix &R)
+{
+    float r[3][3];
+    double a[3][3] = {{ma[0][0],ma[0][1],ma[0][2]},{ma[1][0], ma[1][1], ma[1][2]},{ma[2][0], ma[2][1], ma[2][2]}};
```
Lawrence37: Is `r` supposed to be a `double` array instead of `float`?

What's the reason for using `r` and `a` instead of writing the values to `R` directly? Seems like it can be simplified to just
```c++
R[0][0] = ma[0][0];
R[0][1] = ma[1][0];
R[0][2] = ma[2][0];
R[1][0] = ma[0][1];
.
.
.
```

rtengine/color.cc
```diff
@@ -2105,6 +2106,162 @@ float Color::eval_ACEScct_curve(float x, bool forward)
...
+    double b[3][3] = {{mb[0][0],mb[0][1],mb[0][2]},{mb[1][0], mb[1][1], mb[1][2]},{mb[2][0], mb[2][1], mb[2][2]}};
+    for( int i = 0; i < 3; ++i){
+        for( int j = 0; j < 3; ++j){
+            r[i][j] = 0.0f;
+            for( int k = 0; k < 3; ++k){
+                r[i][j] = r[i][j] + a[i][k] * b[k][j];
+            }
+        }
+    }
```
Lawrence37: Same comment about `r` using `float` instead of `double` and why `r`, `a`, and `b` are used instead of using the `Matrix`s directly.

rtengine/color.cc
```diff
@@ -2105,6 +2106,162 @@ float Color::eval_ACEScct_curve(float x, bool forward)
...
+{
+    float r[3];
+    float x[3] = {in[0], in[1], in[2]};
+    double a[3][3] = {{ma[0][0],ma[0][1],ma[0][2]},{ma[1][0], ma[1][1], ma[1][2]},{ma[2][0], ma[2][1], ma[2][2]}};
+    for( int i = 0; i < 3; ++i){
+        r[i] = 0.0f;
+        for( int j = 0; j < 3; ++j){
+            r[i] = r[i] + x[j] * a[j][i];
+        }
+    }
```
Lawrence37: Same comment about using `r`, `x`, and `a` instead of using the matrix and arrays directly.

rtengine/color.cc
```diff
@@ -2105,6 +2106,162 @@ float Color::eval_ACEScct_curve(float x, bool forward)
...
+// ISO 17321-1 and Ohta (1997)
+//const float THR_CYAN = 0.815;
+//const float THR_MAGENTA = 0.803;
+//const float THR_YELLOW = 0.880;
+
+// Aggressiveness of the compression curve
+//const float PWR = 1.2;
+//https://www.gujinwei.org/research/camspec/
+
+void Color::gamut_compress(float rgb_in[3], float threshold[3], float distance_limit[3], Matrix to_out, Matrix from_out, float pwr, bool rolloff, float &R, float &G, float &B)
```
Lawrence37: This function should be renamed to something more specific, like `aces_reference_gamut_compression`

rtengine/procparams.cc
```diff
@@ -1884,6 +1884,46 @@ bool SHParams::operator !=(const SHParams& other) const
...
+CGParams::CGParams() :
+    enabled(false),
+    th_c(0.815),
+    th_m(0.803),
+    th_y(0.880),
+    d_c(1.147),
+    d_m(1.264),
+    d_y(1.312),
+    pwr(1.2),
+    colorspace("srgb"),
```
Lawrence37: ACES RGC was developed for the ACEScg color space (ACES AP1 primaries) and the threshold and limits were calculated for ACEScg. Should the default color space be `acesp1` instead?

rtengine/rtthumbnail.cc
```diff
@@ -1327,6 +1327,11 @@ IImage8* Thumbnail::processImage (const procparams::ProcParams& params, eSensorT
         ipf.filmNegativeProcess(baseImg, baseImg, params.filmNegative);
     }
 
+  //  if (params.cg.enabled) {//gamut compression
+   //     ipf.gamutcompr(baseImg, baseImg);
+  //  }
```
Lawrence37: Is this needed?

rtgui/paramsedited.cc
```diff
@@ -3359,6 +3382,46 @@ void ParamsEdited::combine(rtengine::procparams::ProcParams& toEdit, const rteng
...
+    }
+
+    if (cg.d_m) {
+        toEdit.cg.d_m = mods.cg.d_m;
+    }
+
+    if (cg.d_c) {
+        toEdit.cg.d_y = mods.cg.d_y;
+    }
+
```
Lawrence37:
```suggestion
    if (cg.d_y) {
        toEdit.cg.d_y = mods.cg.d_y;
    }

```

rtgui/compressgamut.h
```diff
@@ -0,0 +1,74 @@
+/*
+ *  This file is part of RawTherapee.
+ *
+ *  Copyright (c) 2004-2010 Gabor Horvath <hgabor@rawtherapee.com>
```
Lawrence37: Update the year and name :)

rtgui/compressgamut.cc
```diff
@@ -0,0 +1,366 @@
+/*
+ *  This file is part of RawTherapee.
+ *
+ *  Copyright (c) 2004-2010 Gabor Horvath <hgabor@rawtherapee.com>
```
Lawrence37: Update the year and name

rtgui/compressgamut.cc
```diff
@@ -0,0 +1,366 @@
...
+Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPRESSGAMUT_LABEL"), false, true)
+{
+    auto m = ProcEventMapper::getInstance();
+    EvcgColorspace = m->newEvent(COMPR, "HISTORY_MSG_CG_COLORSPACE");//FIRST to reinit histogram
+    Evcgthc = m->newEvent(COMPR, "HISTORY_MSG_CG_CYANTH");
+    Evcgthm = m->newEvent(COMPR, "HISTORY_MSG_CG_MAGENTATH");
+    Evcgthy = m->newEvent(COMPR, "HISTORY_MSG_CG_YELLOWTH");
+    Evcgdc = m->newEvent(COMPR, "HISTORY_MSG_CG_CYANLIM");
+    Evcgdc = m->newEvent(COMPR, "HISTORY_MSG_CG_CYANLIM");
+    Evcgdc = m->newEvent(COMPR, "HISTORY_MSG_CG_CYANLIM");
```
Lawrence37: Duplicate lines
```suggestion
    Evcgdc = m->newEvent(COMPR, "HISTORY_MSG_CG_CYANLIM");
```

rtgui/compressgamut.cc
```diff
@@ -0,0 +1,366 @@
...
+    pack_start(*iFrame);
+    colorspaceconn = colorspace->signal_changed().connect(sigc::mem_fun(*this, &Compressgamut::colorspaceChanged));
+
+    // Percentage of the core gamut to protect Limits
+    // Values calculated to protect all the colors of the ColorChecker Classic 24 as given by
+    // ISO 17321-1 and Ohta (1997)
+
+    th_c = Gtk::manage (new Adjuster (M("TP_COMPRESSGAMUT_CYANTH"), 0., 0.999, 0.001, 0.815));//0.999 to avoid: 1 - th = 0
+    th_m = Gtk::manage (new Adjuster (M("TP_COMPRESSGAMUT_MAGENTATH"), 0., 0.999, 0.001, 0.803));
+    th_y = Gtk::manage (new Adjuster (M("TP_COMPRESSGAMUT_YELLOWTH"), 0., 0.999, 0.001, 0.880));
```
Lawrence37: Is there really a problem with a threshold of 1? The equations look like they are fine with it. The only problem I see is when threshold is equal to the limit, but the limit is always greater than 1.

rtgui/compressgamut.cc
```diff
@@ -0,0 +1,366 @@
...
+    Gtk::Box *limVBox = Gtk::manage(new Gtk::Box(Gtk::ORIENTATION_VERTICAL));
+    limVBox->pack_start (*d_c);
+    limVBox->pack_start (*d_m);
+    limVBox->pack_start (*d_y);
+    limFrame->add(*limVBox);
+    pack_start(*limFrame, Gtk::PACK_SHRINK);
+
+    // Aggressiveness of the compression curve Pwr
+    rolloff = Gtk::manage(new Gtk::CheckButton(M("TP_COMPRESSGAMUT_ROLLOFF")));
+    pwr = Gtk::manage (new Adjuster (M("TP_COMPRESSGAMUT_PWR"), 0.5, 2.0, 0.01, 1.2));
```
Lawrence37, line 109: According to [this](https://paper.dropbox.com/doc/ACES-Gamut-Mapping-Architecture-VWG-Technical-Documentation-Deliverable-tZHiuOCj0RdYw8PPkrTam), the power "needs to remain above one, but cannot go much higher than 1.5". I'm not sure why the lower limit is 1 because the curve is well behaved down to 0.

They do mention that 1.1 or 1.2 are good for SDR output and 1.4 is good for HDR. It is a very small range, so maybe we should keep the range small here too?

rtgui/compressgamut.cc
```diff
@@ -0,0 +1,366 @@
...
+        } else {
+            listener->panelChanged(Evcgroll, M("GENERAL_DISABLED"));
+        }
+    }
+}
+
+void Compressgamut::colorspaceChanged()
+{
+    if (listener && getEnabled()) {
+        listener->panelChanged(EvcgColorspace, M("GENERAL_ENABLED"));
```
Lawrence37: The message should be the color space name

rtgui/compressgamut.cc
```diff
@@ -0,0 +1,366 @@
...
+{
+    ToolPanel::setBatchMode (batchMode);
+    th_c->showEditedCB ();
+    th_m->showEditedCB ();
+    th_y->showEditedCB ();
+    d_c->showEditedCB ();
+    d_m->showEditedCB ();
+    d_y->showEditedCB ();
+    pwr->showEditedCB ();
+}
```
Lawrence37: I think the color space needs and "Unchanged" option for batch mode.

rtgui/paramsedited.cc
```diff
@@ -3359,6 +3382,46 @@ void ParamsEdited::combine(rtengine::procparams::ProcParams& toEdit, const rteng
         toEdit.sh.lab = mods.sh.lab;
     }
 
+    if (cg.th_c) {
+        toEdit.cg.th_c = mods.cg.th_c;
+    }
```
Lawrence37, line 3387: No batch editor add/set behavior?

Lawrence37:
@Desmis I will do some code cleanup before merging.

rtdata/languages/default
Desmis: OK

rtengine/color.cc
Desmis: OK, I'll let you make this change as you mentioned in your last message.

rtengine/color.cc
Desmis: I'll let you make this change as you mentioned in your last message.

rtengine/color.cc
Desmis: OK, I'll let you make this change as you mentioned in your last message.

rtengine/color.cc
Desmis: OK, it's better

rtengine/procparams.cc
Desmis: Why not ? But in most cases most users' screens (like mine) are sRGB. but I don't see any drawbacks to that.

rtgui/paramsedited.cc
Desmis: Yes a mistake

rtgui/compressgamut.cc
Desmis: I noticed - especially on high gamut reds - that 1 brought artifacts
Here 
    for (int i = 0; i < 3; i = i+1) {
        s[i] = (1.0f  - threshold[i]) / sqrt(fmax(1.001f, distance_limit[i]) - 1.0f);

1.0f - - threshold[i]  becomes null

rtgui/compressgamut.cc
Desmis: From the tests I was able to do, on values ​​lower than 1, I took 0.5 and it works correctly

rtgui/paramsedited.cc
Desmis: is it useful, because this algorithm is really dedicated to a type of image (Red flowers, Blue LED,...)

rtengine/procparams.cc
Lawrence37: The thresholds and limits are optimized for ACES AP1. If we use sRGB for the default, then different thresholds and limits will give better results. At least, we should have different thresholds because they are computed to protect the colors of the 24 patch color checker. I am not too concerned about the limits because they are based on the color spaces of cinema cameras.

Desmis:
@Lawrence37 
To avoid mishandling (or misunderstandings on my part), I suggest that you make the modifications you suggest before merging.
Jacques

rtgui/compressgamut.cc
Lawrence37: 1 - threshold is 0, so s becomes zero. I think that works ok for the compression curve.

rtgui/paramsedited.cc
Lawrence37: Just checking if it was intentional or if it was left out by mistake.

rtgui/compressgamut.cc
Desmis: No problem on principle... In some images (red flowers) artefacts.

Lawrence37:
@Desmis I have made the changes as indicated in my comments and did some cleanup.

Desmis:
@Lawrence37 

I just compiled, all works fine.
Good optimizations. Thank you very much.

Ready to merge ?

Jacques

Lawrence37:
Not yet. They're are some unresolved comments.

Lawrence37:
I want to discuss the default values. As I mentioned in one of my review comments, the gamut compression algorithm was designed to operate using the ACES AP1 primaries. The thresholds were chosen to preserve the colors in the 24 patch color checker (see [Compression Threshold](https://paper.dropbox.com/doc/ACES-Gamut-Mapping-Architecture-VWG-Technical-Documentation-Deliverable-tZHiuOCj0RdYw8PPkrTam#:h2=Compression-Threshold)). The limits were chosen to completely compress gamuts of cinema cameras into the ACEScg color space (see [Distance Limit](https://paper.dropbox.com/doc/ACES-Gamut-Mapping-Architecture-VWG-Technical-Documentation-Deliverable-tZHiuOCj0RdYw8PPkrTam#:h2=Distance-Limit)). Mixing the thresholds and limits designed for ACES AP1 with the sRGB color space is suboptimal.

I see three better alternatives:
1. Change the default color space to ACES AP1.
    - Pros: Follows the ACES Reference Gamut Compression exactly. No need to recalculate the thresholds and distances.
    - Cons: RawTherapee does not use ACES AP1 anywhere by default.
2. Change the default thresholds and distances so they work better for sRGB.
    - Pros: Strong gamut compression reduces risk of color artifacts. Matches the default output profile.
    - Cons: Strong gamut compression may be too limiting since it happens early in the processing pipeline. Need to find the optimal thresholds and distances. sRGB is very different from ACES AP1, so the resulting colors will be different and possible inferior.
3. Change the default color space to ProPhoto and change the default thresholds and distances so they work well for ProPhoto.
    - Pros: Matches the default working profile.
    - Cons: Need to find the optimal thresholds and distances.

What do you think @Desmis?

Desmis:
@Lawrence37 
My English is bad, so I have to use a translator... 

Your question is a good one... should I or shouldn't I change the Threshold and Distance parameters, or let (by creating documentation) the user modify them himself? 
Of course I had read the documentation, but ACES explanations seem succinct and not very explicit.

The system is complex and I had already asked myself the question of how to fit it into the gamut.
I did some prospecting on the subject:
* in the “apseimprov” branch, where it's the most up-to-date (it's also partly found in the other branches), with the use of Gamutmap (Emil Martinec's excellent algorithm), plus the possibility of modifying the position of the white point and the dominant color, and also the possibility (settings) of choosing (Avoid color shift) “XYZ relative”.
* in the “ghs” Generalized Hyperboloc Stretch branch, where
** I've changed the “gamut” button, which modifies the action according to the choice of “Working profile” or “Output profile” - removing the gamma for Working profile. This really shows the differences in behavior
https://discuss.pixls.us/t/generalized-hyperbolic-stretch/45827/54
** in “advanced” mode (more complete than Basic) I've introduced a calculation with current White Point and Black Point data. Action on BP, for example, yields much the same results as Dehaze's last modification (with Raw black point). Action on WP is also important, as it also cuts out data >1 (or 65535) - without 'Clipping' that ACES seems to have “forgotten” ?. Switching to Stretch settings then toggles the gamut (histogram) knob to view data spread and compression.

When I saw this “Gamut compression” option, I was seduced by its apparent simplicity. However, it's extremely complex:
* we don't know the gamut of every camera
* we don't know the camera's Observer
* we don't know the spectral data of the illuminants, especially in problem images (LED, multiple illuminants...)
If all this were known, we could use an approach somewhat similar to that of WB auto, but this is not the case. 
* The reference Working Profile (WP) is variable - the default is Prophoto, which contains imaginary data.

Coming back to the 2 settings (Threshold, Distance), their values seem (I think) correlated by the distance of the white point (CIExy) from each side of the triangle, the general gamut overflow characteristics, which are most often towards Reds / Magentas, and the Pointer's gamut (reflected light)
The “smaller” the destination Output profile, the higher the Threshold values should be, and the higher the Distances values should be. I don't think there's an infallible way of doing this calculation in all cases, depending on the Working Profile and Output profile Target compression. I can make an approximation for these values - which won't be perfect - but better than at present. It's not easy to do (especially the GUI, which is not my strong point).

In the meantime, we can start with ACESP1 as the default value, merge the PR and work on this improvement. What do you think?

After some estimates and tests, here are the approximate orders of magnitude for sRGB
Threshold 
* cyan = 0.25  (look at CIExy diagram)
* Magenta = 0.925
* Yellow = 0.934

Max distance limits
* cyan = 1.050
* Magenta = 1.080
* Yellow = 1.1

These values are reasonably good for AdobeRGB...but we can refine

The default values for ACESP1
* 0.815 , 0.803, 0.88 (threshold)
* 1.147, 1.264, 1.312 (Limits) 
are reasonably good for Rec2020, Prophoto, ACESP1...but we can refine 

Jacques

Lawrence37:
Changing the default color space to ACES AP1 is ok for me.

There is some code in rtthumbnail.cc that is commented out. Is the code necessary?

Desmis:
@Lawrence37 
OK for me.
The code in rtthumbnail is on necessary and can be removed.

Who makes these (minor) changes (delete code, change default to ACESP1). You said you still had some modifications (comments) to make.

Lawrence37:
I don't have any code changes in mind yet. There are 4 remaining comments I made. Two of them are resolved, minus the code change (default values and commented code). The last two I will experiment with (maximum threshold and compression power).

Desmis:
@Lawrence37 

I'll let you complete if necessary. You can merge, no worries for me. I'm "absent" from this morning.

I would make the changes to adapt the values ​​of "limits" and "threshold" afterwards after merge.

Desmis:
I made a change that allows to take into account the "Target compression gamut".

A default choice is possible in Preferences

Of course it is to be refined by tests in real values.

Do i push this commit, or do i wait?

Thank you

Jacques

Lawrence37:
I think if a user wants different default values, then they should set it in their default processing profile. Changing the default values in preferences could cause problems with processing profile operations.

I experimented with the threshold limit and the compression power. A threshold of 1 requires a different equation when using the power curve. Since 0.999 produces almost identical results as 1, I don't have a preference for setting the maximum threshold at 0.999 or 1. The compression power has a very subtle effect. The allowed range in this pull request is fine.

Desmis:
I think there is some kind of misunderstanding between us.

What I did was to adapt the Threshold and Limits coefficients to each workspace. They are really different for example between sRGB and Propohoto.

When the user changes in the "Target compression Gamut" menu the values ​​of 6 coefficients are changed.

The "reset" of each slider automatically sets itself to the new default value specific to the workspace ( I used "setlimits", and change a little adjuster.cc).
I added a checkbutton to allow (this is generally not good) to keep the same values ​​when going for example from ACESP1 to Rec2020, but by default it is the new values ​​that are taken into account.

The choice in Preferences (is of the same type as that for example for the level of complexity of SE), it allows to have for a user a default profile that corresponds to his installation (default ACESP1) - which he can of course change in "Target Compression Gamut".

If the maximum value of the Threshold is set to 1.0 instead of 0.999, we obtain on some images strong artifacts or even worse, segmentation fault. We can have fun and leave the maximum at 1 (instead of 0.999) and limit it in the "aces_reference_gamut_compression" code in color.cc. That's what I just did.

I obtained these coefficients (threshold , limits)  in a somewhat empirical way, because I do not have, neither the Observer of each camera, neither the gamut of each camera,  nor the specific illuminants (LED, etc.)  and no way to simulate it (I do not have a laboratory).
I used the 468 color chart that I developed in 2006 to calibrate the cameras (it is a kind of Colochecker ++) with a higher gamut. I also checked on specific images (red, yellow, purple flowers; underwater, etc.). The starting base being the distances of the white point to each edge of the CIExy triangle.
Of course these coefficients need to be tested and modified by experience, but overall I think they are generally quite satisfactory.

I will create a branch during the day so that this change can be tried without disturbing the PR.

I just (today 9h30 in France) create a new branch "compgamutimprov"

Jacques

Lawrence37:
What I mean is that users already have a way to set the default compression color space by setting it in their default processing profile. Adding a new setting could cause problems because it changes the way processing profiles work. For these two reasons, I think it is better to not add anything to preferences.

I understand the issue with a threshold of 1 when using the power curve. That's why a different formula is needed (equal to the limit of the function as threshold approaches 1). The formula is simply { x<1: x, x>=1: 1}. It is almost identical to a threshold of 0.999. As I wrote, I don't care if the maximum is 1 or 0.999 because there is virtually no visual difference.

There is a way to calculate the "correct" thresholds for each color space by finding the RGB values of each color in the ColorChecker Classic and computing the maximum distance for R, G, and B. It is possible to calculate the distance limits too, if we have the primaries for the color spaces shown in the chromaticity diagram.

Desmis:
I think we've misunderstood what I've done and the purpose of this module.

The choice in Preferences corresponds to the user's habits/installations: does he have an sRGB or Rec2020 monitor, for example. This choice only changes the default value in “Target gamut compression” and doesn't change the rest of the process. But it's not essential - it's comfort. I think we can all agree that ACESP1 is not the usual default choice for users.

The modification I made to avoid crashes when Threshold=1 is very similar to the one you suggest - and results in a crash in some images (underwater...), if th=1.

Perhaps the determination of Threshold and Limit values can be done with a ColorChecker24 (I have my doubts) ?. But as I said, the Colorchecker24 is essentially (if not a deep blue) in the sRGB gamut, and therefore can't properly "guide" the compress values above the gamut, as found in flowers (red, yellow, purple...) for example. It's not just negative RGB values that are at issue, but also those above 1 (or 65535).

The (rather empirical) values I propose are the combination of:
* global evaluation of threshold coefficients based on the relative variation in distance from the white point to the 3 sides of the primary triangle (CIExy). Global evaluation of limits as a function of gamut size and previous values.
* changes in high gamut values on my "Colorchecker 468", while changing workspace (target gamut)
* changes with images (flowers) with red, yellow,.. out of sRGB (out of Colorchecker24), while changing workspace (target gamut)
* changes in underwater images, while changing workspace (target gamut)
* change in saturated blue Led image, while changing workspace (target gamut)

I have tested these values with about 5 to 10 difficult images.
Of course we can refine this settings by user feedback

Another thing, the XYZ values of the DCI-p3 matrix (which I had copied into ART) are not correct. They don't take into account the chromatic adaptation between the White point of DCI-P3 (about 6300K) and D50. I've made this change.

I'll let you do what you want, merge this PR with the minimal modification to avoid crashes or do more at your convenience. Ignore what I've done if you don't want to.

Jacques

Lawrence37:
Maybe the message is lost through translation. What you wrote is consistent with my understanding of what you have done. I also reviewed the code in the compgamutimprov branch. I understand you made these changes with respect to the color spaces, thresholds, and limits.
- Added a color space option in Preferences. This changes the default value of the compression color space to whatever the user wants.
- Changed the maximum threshold to 1 in the GUI. When compressing, the threshold is limited to 0.999.
- Added an option in the tool for automatically changing the thresholds and limits. A user can decide if the values should be kept or if they should be changed to the empirically-derived values for the current color space.
- and other changes

I will restate my comments and provide examples to try to communicate my opinions.

- The color space option in Preferences is **not necessary**. This is because it is already possible to set the default color space. It can be done using the Default Processing Profile in Preferences. Another reason why the default color space should not be set directly in Preferences is because dynamic default values can cause problems. There may be some parts of the code that assume the defaults are static. It will be very difficult to ensure there are no bugs.
  - image
- I am fine with the maximum threshold of 0.999. I am fine with the maximum threshold of 1. Either one is fine for me. You do not need to make any change to the maximum threshold. If you choose to use a maximum of 1 in the GUI, then I want the following changes too.
  - Remove the limit of 0.999 when doing the compression. In other words, remove this line: https://github.com/Beep6581/RawTherapee/blob/7843a8d7248616246a00ae458d2e6dc4d436be35/rtengine/color.cc#L2225 The reason is to avoid misleading users. If they set a threshold of 1, they expect it to be 1, not 0.999.
  - If `rolloff` is `true` and the threshold is 1, use the equation `y = { x < t: x, x ≥ t: t }` to prevent a crash. If threshold is not 1, use the current equation. For example,
  ```c++
            if (d[i] >= threshold[i]) {
                if (threshold[i] == 1.f) {
                    // Special case for threshold of 1. The limit as threshold goes to 1 of the compression equation is
                    // y = { x < t:  x,
                    //       x >= t: t
                    cd[i] = 1.f;
                } else {
                     // Calculate scale factor for y = 1 intersect
                    const float limit = distance_limit[i];
                    const float thres = threshold[i];
                    //                     l - t
                    // Scale s = --------------------------
                    //           ( ( 1 - t )-p     )(1 / p)
                    //           ( ( ----- )   - 1 )
                    //           ( ( l - t )       )
                    const float scale = (limit - thres) / pow(pow((1.0f - thres) / (limit - thres), - pwr) - 1.0f, 1.0f / pwr);
                    // Normalize distance outside threshold by scale factor
                    // x' = (x - t) / s
                    const float nd = (d[i] - thres) / scale;
                    //                  x'
                    // y = t + s ----------------
                    //           (1 + x'^p)^(1/p)
                    const float po = pow(nd, pwr);
                    cd[i] = thres + scale * nd / (pow(1.0f + po, 1.0f / pwr));
                }
            }
  ```

Regarding the derivation of the thresholds, they **can** be calculated using the ColorChecker Classic. That's what ACES did. To try it, I used the [L\*a\*b\* values from the ColorChecker Classic](https://www.xrite.com/service-support/new_color_specifications_for_colorchecker_sg_and_classic_charts) and converted to ACES AP1. The dark blue color is on the lower left of the chromaticity diagram and defines the cyan and magenta thresholds. The RGB values for the dark blue are (0.04246, 0.04885, 0.2528). The distances are (0.832, 0.807, 0). The first two distances corresponding to cyan and magenta are very close to the thresholds found by ACES. The small error may be due to differences between where ACES and I got the ColorChecker values from. I used the new ColorChecker and used GIMP to convert to RGB using the RTv4_ACES_AP1 profile.

Desmis:
As I said before, I'm not going to go over why I did it like this and the limitations of using a Colorchecker24 (even though that's what ACES used  - of course I tried with the Colorchecker24 data as you).

I have expressed my point of view, extended the initial concept, but I am no longer physically (and morally) fit to continue.

You do what you think is right, it will be right. So continue on what seems good to you.

My general participation in RT's development (my age, my health, my morale) will be significantly reduced. This is to be taken into account for everything that is in progress (PR, issues, etc.). 
Nevertheless, I'm always on the lookout and reserve my comments for subjects where I think I can make a contribution.

Jacques

Lawrence37:
I am ok with this PR as it is.

Desmis:
No problem :)

Desmis:
@Lawrence37 

I think it is still necessary to change:
* the values ​​of the XYZ matrix for DCI-P3 which are wrong. They do not take into account the chromatic adaptation between 6300K (Dci-P3 Theater) and D50.
* modify the code to take into account the values ​​of threshold=1 which can lead to a crash on some (rare) images. The solution you propose is good.
* put in the tooltips that the values ​​of threshold and limits are for AcesP1 and that they must be adapted for the other Target Compression gamut

If you agree I can make this change tomorrow.

Possible adaptations of thresholds and limits can be reported (if there is demand) and/or mentioned with indicative values ​​in the documentation when Rawpedia works again.

Jacques

Lawrence37:
Ok @Desmis

Desmis:
@Lawrence37 

I made the changes agreed between us. Threshold in GUI and Color.cc, Matrix DCI-P3, tooltip (simple)
 
Can you check and what did we do to merge?

Thank you

Jacques

rtdata/languages/default
```diff
@@ -2501,6 +2511,29 @@ TP_COLORTONING_TWOALL;Special chroma
...
+TP_COMPRESSGAMUT_THRESHOLD_TOOLTIP;Threshold: controls the percentage of the outer gamut to affect. A value of 0.8 will compress out of gamut values into the outer 20% of the gamut. The inner 80% of the gamut core will not be affected. You can adjust the colors separately: for example if you wanted to protect magenta and yellow a little more than cyan, you could set the threshold a bit higher for cyan.\n\nThese thresholds must be changed depending on the Target Compression Gamut.
+TP_COMPRESSGAMUT_DCIP3;DCI-P3
```
Lawrence37, line 2518: `TP_COMPRESSGAMUT_DCIP3` is a duplicate.

rtdata/languages/default
```diff
@@ -2501,6 +2511,29 @@ TP_COLORTONING_TWOALL;Special chroma
...
+TP_COMPRESSGAMUT_THRESHOLD_TOOLTIP;Threshold: controls the percentage of the outer gamut to affect. A value of 0.8 will compress out of gamut values into the outer 20% of the gamut. The inner 80% of the gamut core will not be affected. You can adjust the colors separately: for example if you wanted to protect magenta and yellow a little more than cyan, you could set the threshold a bit higher for cyan.\n\nThese thresholds must be changed depending on the Target Compression Gamut.
```
Lawrence37: I think this could be more clear about why they should be changed (and technically, a user isn't required to change them to get good results). How about this?

> The thresholds are relative to the Target Compression Gamut. They may need to be adjusted after changing the target gamut if you want the protected color range to remain unchanged.

rtdata/languages/default
```diff
@@ -2501,6 +2511,29 @@ TP_COLORTONING_TWOALL;Special chroma
...
+TP_COMPRESSGAMUT_ACESP1;ACES AP1
+TP_COMPRESSGAMUT_POWER_TOOLTIP;Power: Control the aggressiveness of the compression curve.\n Higher power values result in more compressed values distributed closer to the gamut boundary. This will increase the "color purity" or "saturation" once the image has been rendered through a display transform.\n Higher values have a smoother transition.
+TP_COMPRESSGAMUT_LIMIT_TOOLTIP;Maximum Distance Limit: is the distance outside of the gamut boundary to compress to the gamut boundary. For example, a value of 1.2 will compress distance values of 1.2 to a value of 1.0. \nIndividual controls are given for each color component. They are named cyan magenta and yellow because they actually control the max distance from the edge of the triangle (primaries related to the CIExy diagram), not the corner. For example, boosting the value of the cyan distance limit will increase the distance from which values are mapped to the edge between the blue and green corners of the triangle.\nThe distance limits also control the RGB ratios of the compressed values, which affects the apparent hue.\n\nThese distances must be changed depending on the Target Compression Gamut.
```
Lawrence37: and this one can be

> The distances are relative to the Target Compression Gamut. They may need to be adjusted after changing the target gamut if you want the amount of compression to remain unchanged.

rtdata/languages/default
Desmis: ok

rtdata/languages/default
Desmis: OK, it is more clear

rtdata/languages/default
Desmis: Ok it is more clear
I will make change soon

Jacques

Lawrence37 (approved):

PR #7207, opened 2024-09-24
Remove GUI mask in Global mode

Desmis:
Masks can't work well in Global mode, due to the lack of deltaE.
The solution in place in "dev" didn't work in Global mode, but let the GUI show.

This (easy) modification to the GUI solves the problem of disabling and hiding
* Mask and modifications
* Recovery based on luminance mask
I only change all void  as : void Locallabxxx::updateguixxx(int spottype) 

Jacques


Lawrence37 (approved):

Desmis:
@Lawrence37 

Thank you
I have an internet connection... sufficient. For 3 weeks.

Jacques

PR #7209, opened 2024-09-28
LibRaw 0.21.3

### LINKED ISSUE #7206: [Dependency Update] Update LibRaw to 0.21.3
LibRaw 0.21.3 has been released, bringing support and fixes to  over 40 cameras. 
https://github.com/LibRaw/LibRaw/releases
https://github.com/LibRaw/LibRaw/tree/master/src/tables

Lawrence37:
Applies 0.21.3 commits since 0.21.2 on top of the current internal Libraw (snapshot 202403).

### DIFF
--- a/rtengine/libraw/Changelog.txt
+++ b/rtengine/libraw/Changelog.txt
@@ -1,19 +1,36 @@
+2024-09-18  Alex Tutubalin <lexa@lexa.ru>
+
+ LibRaw 0.21.3-Release
+  * new compile-time define LIBRAW_CALLOC_RAWSTORE
+    If defined: calloc() will be used for all big buffer
+    allocations to prevent uninitialized heap data leak
+
+  * Small allocations: calloc() instead of malloc() to prevent
+    uninitialized heap data leak
+
+  * Support for 4-component JPEG-compressed DNG files
+
+  * CR3-Qstep table: avoid wrong 64-bit code generation
+
+  * Fixed several bugs related to specially-crafted files
+    processing (thanks to OSS-Fuzz project)
+
 2024-03-30  Alex Tutubalin <lexa@lexa.ru>
 
  LibRaw-snapshot-202403
 
- * Raw format support 
+ * Raw format support
     1) Panasonic encoding 8
 
-    2) Sony YCC (Medium/Small compression) pseudo-RAW support: 
+    2) Sony YCC (Medium/Small compression) pseudo-RAW support:
       -- normal processing (3-channel RGB output)
       -- or YCbCr output if LIBRAW_RAWSPECIAL_SRAW_NO_RGB bit is set
-         in imgdata.rawparams.specials (one may specify the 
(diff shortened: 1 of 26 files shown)


kmilos:
@Lawrence37 I'm a bit confused - are you using 0.21.x stable or the 202403 snapshot in RT?

These branches are not necessarily sharing the same code base, and the API/ABI can generally differ...

If you're using the 202403 snapshot, your local `libraw_version.h` is incorrect/misleading, and you should just update the patches from and track the `master` branch (w/ references to commit hashes and then your local mods), and forget about referencing `0.21-stable`...?

Lawrence37:
`dev` is currently based on the snapshot. ABI changes do not matter, and any API changes will have to be taken into account anyway whenever there is a new snapshot or major release.

The goal of this pull request is to combine the latest snapshot and release. When LibRaw lengthened the release cycle, I was concerned there would be more frequent situations like this when a minor release is published after a snapshot. master does have the release commits, but it also has other commits. LibRaw does not say if those commits are tested as extensively as the release and snapshots, so without review from RT developers, I do not feel it is right to synchronize with master.

kmilos:
> dev is currently based on the snapshot

So why is your `libraw_version.h` saying 0.21.3 API then? (Previously even 0.21.0 for what is supposedly snapshot 202403 code base!?)

> The goal of this pull request is to combine the latest snapshot and release.

That's my point. All the important patches done to the stable (release) branch are (or should be, going by LibRaw devs prior behavior) on the master branch as well. There's nothing to "combine" manually if you simply track master (= last snapshot + security/bugfix patches already there), and then just put your local mods on top.

Cf.

stable: https://github.com/LibRaw/LibRaw/compare/0.21.2...0.21.3

and 

master: https://github.com/LibRaw/LibRaw/compare/12b0e5d...9bcb8a1 (w/ commits prior to March already in there, plus other goodness)

> LibRaw does not say if those commits are tested as extensively as the release and snapshots

They don't say much about how extensively snapshots are tested either, yet we (are somehow obliged to) use them. (From my point of view, any commit on master is a "snapshot", it is just a commit number they declare; they don't even tag a "snapshot". The last declared one [forgot to bump `libraw_version.h` API](https://github.com/LibRaw/LibRaw/commit/0d011198b2f75ffa3ede51566616b9b2797104e9), so you shouldn't have used that declared commit anyway...)

I understand any of this might not matter as you're in full control of the fork, but it just makes maintenance more confusing and burdensome IMHO. But it's your prerogative... 🤷

Lawrence37:
This branch is saying 0.21.3 because it is not `dev`. `dev` is saying 0.21.0 because, as you wrote, LibRaw forgot to change it for the snapshot.

The main page on the LibRaw website reads
> Public snapshots are always tested on a fairly large user base

Lawrence37:
Closing in favor of the newer snapshot (202502).

PR #7210, opened 2024-09-28
Generalized Hyperbolic Stretch in Selective Editing

Desmis:


Desmis:
**Introduction**
The Generalized Hyperbolic Stretch (GHS) process allows you to transform the values of pixels in your image to improve the representation of the underlying data for human visualisation.
The generalised hyperbolic equations used in the GHS process have five defining parameters. This allows significant flexibility in designing the "shape" of the transformation.
Typical uses of pixel intensity transformations include:
* Initial stretch of pixel data from linear state.
* Addition of contrast to key areas of the image.
* Overall brightening or darkening of the image.
* Adjustment of the dynamic range of the image.
* Adjustment of pixel data in one colour channel to balance better with other channels.

The control provided by the GHS process is designed to perform well in all these situations.

**Origin**
I believe that the origin of this algorithm, initially dedicated to astrophotography, is to be attributed to PixlnSight ©, the 2 authors are David Payne in 2021 and Mike Cranfield.
A description of the process and possible settings :
https://www.ghsastro.co.uk/doc/tools/GeneralizedHyperbolicStretch/GeneralizedHyperbolicStretch.html


This algorithm is found in Siril ©  Copyright (C) 2005-2011 Francois Meyer (dulle at free.fr)
Copyright (C) 2012-2023 team free-astro (see more in AUTHORS file)
Reference site is https://free-astro.org/index.php/Siril

You will also find this algorithm adapted to CTL by Alberto in ART @agriggio - thanks to him

**In Rawtherapee**
I ported this algorithm to Selective Edition (SE) in C++. The choice of SE is important, as it makes it easy to open several RT-spots (not compulsory, of course) and thus to process difficult images of all kinds (landscape, flowers, astro, etc.). 
For a better treatment, you can proceed in several steps, creating 2 or more RT-spots  each superimposed with the GHS tool. For each of the RT-Spots, the Symmetry point (SP) base value will be different (and of course other settings). Use the peak marker on the histogram to estimate the position of the SP value. For example, you can start the system in RGB mode (1st Spot), then the 2nd Spot in RGB mode with a different SP, and the 3rd in Saturation or Hue mode.
Full Image allows you to differentiate the action according to the color based on the deltaE.

But of course in the majority of cases only one RT-spot is needed.

You will find this algo in: Shadows/Highlights & Equalizer & GHS

**Six settings are available :**
* Stretch factor (D)
This parameter controls the amount of stretch. If the Stretch factor is set to zero there is no stretch, ie the transformation is the identity transformation.
 
* Local intensity (b)
This parameter controls how tightly focused the stretch is around the Symmetry point (SP) by changing the form of the transform itself. For concentrated stretches (such as initial stretches on linear images) a large 'b' factor should be employed to focus a stretch within a histogram peak while de-focusing the stretch away from the histogram peak (such as bright stars). For adjustment of non-linear images, lower 'b' parameters should be employed to distribute contrast and brightness more evenly. Large positive values of 'b' can be thought of as a histogram widener, ie spreading the histogram wider about the focus point, SP. By contrast, lower values of 'b' tend to shift the histogram to a brighter (or dimmer) position without affecting its width too greatly. As a general rule, the level of 'b' employed will decrease as a stretch sequence nears completion, although larger 'b' values can still be employed for precise placement of additional contrast.

* Symmetry point (SP)
Sets the focus point around which the stretch is applied - contrast will be distributed symmetrically about SP. While 'b' provides the degree of focus of the stretch, SP determines where that focus is applied. SP should generally be placed within a histogram peak so that the stretch will widen and lower the peak by adding the most contrast in the stretch at that point. Pixel values will move away from the SP location

* Protect shadows (LP)
Sets a value below which the stretch is modified to preserve contrast in the shadows/lowlights. This is done by performing a linear stretch of the data below the 'LP' level by reserving contrast from the rest of the image. Moving the LP level towards the current setting of SP changes both the scope (range) and the amount of this contrast reservation, the net effect is to push the overal stretch to higher brightness levels while keeping the contrast and definition in the background. The amount of contrast reserved for the lowlights is such that the continuity of the stretch is preserved. This parameter must be greater than or equal to 0 and not greater than the Symmetry point

* Protect highlights (HP)
Sets a value above which the stretch is modified to preserve contrast in the highlights/stars. This is done by performing a linear stretch of the data above the 'HP' level by reserving contrast from the rest of the image. Moving the HP level towards the current setting of SP increases both the scope (range) and the amount of this contrast reservation, the net effect is to push the overal stretch to lower brightness levels while keeping the contrast and definition in the highlights. The amount of contrast reserved for the highlights is such that the continuity of the stretch is preserved. This parameter must be less than or equal to 1 and not less than the Symmetry point.

* Highlight Attenuation
In some cases, heavy stretching can lead to an excessive increase in Highlights, despite the use of HP.
Smooth and soften highlights areas. Amplifies 'Protects Highlights (HP)' action.

**I will be away from home for 3 weeks** (since Monday) but I will remain in contact with possible less material means and therefore the possibility of making changes (at least I hope so)

Jacques

Desmis:
@Lawrence37 

I just merge this PR... But It is a mistake 
Can we reverse easily

Jacques

Issue #5433, opened 2019-09-01
RT crashes on loading Hasselblad H6D-100cMS pixelshift files


heckflosse:
Fix follows later.

PR #7212, opened 2024-10-07
Support for images with more than 6 sub-frames

### LINKED ISSUE #7143: Add support for Canon CR3 Raw Burst picture selection and extraction
Some newer Canon EOS cameras with DIGIC X can use a feature named "RAW Burst Mode".
It basically means that the cam already starts taking pictures while the trigger is half pushed.
They are stored in the CR3 file like some sort of RAW video sequence (picture roll).
Right now, only Canons DPP software and the cam itself can change the "starting" picture and extract the other pictures from the sequence.
Known Cams to support this feature are: R1, R3, R5m2, R6m2, R7, R8, R10...

Currently, RAW Therapee doesnt show these files in the file browser.

My feature request is therefore:
- please add support for these files (i can supply demo files from a Canon R6 Mark II)
- add support for changing the 1st picture
- add support for extracting other RAW pictures from the sequence

Lawrence37:
This lifts the restriction on images with 7 or more sub-frames. The restriction is there to prevent a crash with Hasselblad pixelshift files (#5433) by only allowing sub-frames 2 to 7 to be selected. With this pull request, the restriction applies only to images with the make "Hasselblad".

I also refactored a bit to use `unique_ptr`.

Closes #7143.

### DIFF
--- a/rtengine/rawimagesource.cc
+++ b/rtengine/rawimagesource.cc
@@ -505,14 +505,6 @@ RawImageSource::~RawImageSource()
     delete greenCache;
     delete blueCache;
 
-    for (size_t i = 0; i < numFrames; ++i) {
-        delete riFrames[i];
-    }
-
-    for (size_t i = 0; i + 1 < numFrames; ++i) {
-        delete rawDataBuffer[i];
-    }
-
     if (camProfile) {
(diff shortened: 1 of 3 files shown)


Lawrence37:
I will merge this soon.

PR #7220, opened 2024-10-15
Fix duplicate tag in Deutsch locale

pano9000:
Hello,

The "Deutsch" locale has one message, with an unmatched additional `<b>` tag, which causes a Gtk-WARNING message:

```(rawtherapee:617420): Gtk-WARNING **: 08:29:35.684: Failed to set text 'Bewertung entfernen.
Taste: <b><b>0</b>' from markup due to error parsing markup: Fehler in Zeile 2, Zeichen 28: Element »markup« wurde geschlossen, aber das derzeit offene Element ist »b«
```
The German error message essentially just mentions that the b tag is not closed.

This PR removes the unmatched tag.

Kind regards,

pano


Lawrence37:
Nice catch!

Lawrence37 (approved):

PR #7224, opened 2024-10-20
Fix for rank, color label, and trash status reset

### LINKED ISSUE #7218: Pasting copied profile on photo removes photo rating
**Short description**
Not entirely sure whether it's bug or intended behavior, but I noticed that for some strange reason my photo ratings began disappearing between editing sessions as I edited the images. I was able to troubleshoot it little bit and in the end it seems that when user copies profile from one photo, if processing profile fill mode is on, the rating is not preserved. Same happens for color coding and I presume also  for "in trash" property (but that one does not really matter)

**Steps to reproduce**
1. Turn on processing profile fill mode
2. Load at least two photos
3. Rate both of them
4. Check pp3 profile of second image (it contains general.rank value)
5. Edit randomly first image
6. Copy profile from first image (via the "copy current processing profile to clipboard" button)
7. Paste profile to second image (via the "paste profile from clipboard" button)
8. Check second image rating in UI (it does not change)
9. Check pp3 profile of second image (it is missing general.rank value)
10. Restart RawTherapee
11. Check second image rating in UI (now the change propagated and rating is also missing)

**Expected behavior**
I would expect that ranking, color coding and other similar properties are exempt from the replacement with hardcoded values and are preserved no matter what.

**Additional information**
- Version of RawTherapee: 5.11

Lawrence37:
Prevents the rank, color label, and trash status from getting reset after editing an image or clearing its processing profile.

Whenever the processing parameters get reset or saved, the rank, color label, and trash status start with the default values and the original values must be transferred. This was the case with 5.10 and older, but after some refactoring for the XMP synchronization feature (#6988), the values are only transferred if changed during the current session. There are flags for each parameter that indicates if the value has changed. If the flag is not set, the value does not get transferred.

@sgotti I introduced a way to force-update the values in the processing parameters. I'm thinking it is better to just remove the check for the edited flag, but I am hesitant to use that approach. I get the feeling that there is a reason for the check due to the fact that there is the check *in addition to* comparing the old and new values.

Closes #7218.

### DIFF
--- a/rtgui/thumbnail.cc
+++ b/rtgui/thumbnail.cc
@@ -485,7 +485,7 @@ void Thumbnail::clearProcParams (int whoClearedIt)
         pparams->setDefaults();
 
         // preserve rank, colorlabel and inTrash across clear
-        updateProcParamsProperties();
+        updateProcParamsProperties(true);
 
         // params could get validated by updateProcParamsProperties
         if (pparamsValid) {
@@ -594,7 +594,7 @@ void Thumbnail::setProcParams (const ProcParams& pp, ParamsEdited* pe, int whoCh
         pparamsValid = true;
 
         // do not update rank, colorlabel and inTrash
-        updateProcParamsProperties();
+        updateProcParamsProperties(true);
 
         if (updateCacheNow) {
             updateCache();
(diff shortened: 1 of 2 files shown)


Lawrence37:
Merging this.

@sgotti Let me know if you have any concerns.

Issue #7226, opened 2024-10-21
Blur & Noise in Blur/Grain & Denoise not applied globally with the global spot method


Lawrence37:
**Short description**
The blur, noise, and film grain in Blur/Grain & Denoise do not get applied to the whole image when the spot method is set to global.

**Steps to reproduce**
1. Enable Selective Editing and add a spot.
2. Change the Spot method to Global.
3. Add the Blur/Grain & Denoise tool.
4. Play around with the radius, noise, and film grain settings.

**Expected behavior**
The blur, noise, and film grain should be applied uniformly to the entire image. Instead, it is applied to regions similar to the color below the spot, as if the scope is still active.

**Additional information**
Using the other spot methods, the scope in the tool becomes available. Changing the scope, then changing the spot method back to global shows that the scope affects the regions affected by the tool.

Desmis:
@Lawrence37 

Well spotted, it's just an oversight. I'll add that to the iplocallab.cc code in : ImProcFunctions::BlurNoise_Local
                const float reducdE = calcreducdE(dE, maxdE, mindE, maxdElim, mindElim, lp.iterat, limscope, lp.sensbn);

                Replace by:
                float reducdE = calcreducdE(dE, maxdE, mindE, maxdElim, mindElim, lp.iterat, limscope, lp.sensbn);
                if(lp.fullim == 3 ) {//disabled scope
                    reducdE = 1.f;
                }
Jacques

PR #7235, opened 2024-10-28
FFTW blur radius fix

### LINKED ISSUE #7225: Selective Editing blur radius different in preview
**Short description**
When using Selective Editing > Blur/Grain & Denoise > Blur & Noise > Radius, the preview gives an inaccurate preview of the blur amount at zoom <100%.

**Steps to reproduce**
1. Enable Selective Editing, add a new spot, and add the Blur/Grain & Denoise tool.
2. Increase the blur radius.
3. Observe the blur amount at different zoom levels.

**Expected behavior**
The blur in the preview is close to the amount in the saved image.

Preview at 20% zoom:
image

Exported image, scaled down to 20%:
image

Quick tests show that the blur amount is inversely proportional to the zoom, e.g. double the radius at 50%, triple the radius at 33%, quadruple the radius at 25%, etc.. It seems the radius is not scaled to compensate for the zoom factor.

Lawrence37:
Closes #7225.

### DIFF
--- a/rtengine/iplocalcontrast.cc
+++ b/rtengine/iplocalcontrast.cc
@@ -52,21 +52,8 @@ void ImProcFunctions::localContrast(LabImage *lab, float **destination, const rt
 #endif
         gaussianBlur(lab->L, buf, width, height, sigma);
     } else {
-        float kr = 1.f;
-        //emprical adjustment between FFTW radius and Gaussainblur
-        //under 50 ==> 10.f
-        //above 400 ==> 1.f
-        if(settings->fftwsigma == false) {//empirical formula
-            float ak = -9.f / 350.f;
-            float bk = 10.f - 50.f * ak;
-            kr = ak * sigma + bk;
-            if(sigma < 50.f) kr = 10.f;
-            if(sigma > 400.f) kr = 1.f;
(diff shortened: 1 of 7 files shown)


Desmis:
@Lawrence37 

This PR also fix #7226.
Change in iplocallab.cc  = > ImProcFunctions::BlurNoise_Local

Desmis:
@Lawrence37 

It's good for me.

I will be away from home for 15 days or 3 weeks. But I think I can be reached.

Lawrence37:
Thanks for reviewing @Desmis.

Desmis:
It is ok for me for this part (fftw).

In my proposition (fixbug-blur), there was also the solution to issue 7226, as well as a notable improvement for the user when he wants to blur a background. I changed the way the "Recovery based on luminance mask" function works. Now this function works with "Gaussian blur (noise and grain) , Median, and Guided Filter instead of only Guide Filter.

I assume that these proposals will be dealt with later, separately from this PR.

Lawrence37:
Since this pull request also handles the pp3 compatibility, I will merge it and see if the other fixes and improvements can be incorporated directly from your branch @Desmis.

PR #7250, opened 2024-11-14
Revert "Generalized Hyperbolic Stretch in Selective Editing"

Desmis:
Reverts Beep6581/RawTherapee#7210


Desmis:
@Lawrence37 
Yesterday I did something really stupid. Then I was away from my temporary place of stay (family in Picardy). Today I am in Paris at my daughter's and I realize when I reopen the PR that there is a "revert" function to remove the effect of the previous PR.
That's what I'm going to do.
I'm going back home to Fréjus (Côte d'Azur) on Monday

Jacques

Issue #4299, opened 2018-01-12
Better source tree


Hombre57:
RT's source tree for the code part is quite flat actually : only 3 folders (rtengine, rtexif - that might be dropped, and rtgui).

Would it be useful for newcomers to have a more self-explanatory source tree ? I don't think we can put each tool's GUI and engine in their own directory, because the engine part is mostly made os shared code (between tools)

I'm thinking about :
```
- [rtgui]
   - [custom-widgets]
      - adjuster.cc
      - adjuster.h
      - checkbox.cc
      - checkbox.h
      - coloredbar.cc
      - coloredbar.h
      - coordinateadjuster.cc
      - coordinateadjuster.h
      - curves
      - inspector.cc
      - inspector.h
      - lockablecolorpicker.cc
      - lockablecolorpicker.h
      - lwbutton.cc
      - lwbutton.h
      - lwbuttonset.cc
      - lwbuttonset.h
      - on-preview-editing
      - popupbutton.cc
      - popupbutton.h
      - popupcommon.cc
      - popupcommon.h
      - popuptogglebutton.cc
      - popuptogglebutton.h
      - previewwindow.cc
      - previewwindow.h
      - profilestorecombobox.cc
      - profilestorecombobox.h
      - rtimage.cc
      - rtimage.h
      - shcselector.cc
      - shcselector.h
      - splash.cc
      - splash.h
      - thresholdadjuster.cc
      - thresholdadjuster.h
      - thresholdselector.cc
      - thresholdselector.h

      - [curves]
         - curveeditor.cc
         - curveeditor.h
         - curveeditorgroup.cc
         - curveeditorgroup.h
         - curvelistener.h
         - diagonalcurveeditorsubgroup.cc
         - diagonalcurveeditorsubgroup.h
         - flatcurveeditorsubgroup.cc
         - flatcurveeditorsubgroup.h
         - mycurve.cc
         - mycurve.h
         - mydiagonalcurve.cc
         - mydiagonalcurve.h
         - myflatcurve.cc
         - myflatcurve.h

      - [on-preview-editing]
         - edit.cc
         - edit.h
         - editid.h

   - [tools]
      - bayerpreprocess.cc
      - bayerpreprocess.h
      - bayerprocess.cc
      - bayerprocess.h
      - bayerrawexposure.cc
      - bayerrawexposure.h
      - blackwhite.cc
      - blackwhite.h
      - chmixer.cc
      - chmixer.h
      - coarsepanel.cc
[...]

PR #7255, opened 2024-11-16
Get rid of relative include paths

digitalcarp:
* Use target_include_directories to specify include paths
* Specify project root (parent of rtgui and rtengine) as base include path
* Replace relative includes with normal includes

One refactoring step of #4299 that should reduce the need for `#include "../../../rtengine/header.h"`


Lawrence37:
Looks good to me

PR #7259, opened 2024-11-24
Raw dehaze negative values fix

Lawrence37:
Fixes negative black values for raw dehaze by rejecting pixels lower than the black level (noise, dead pixels, etc.).


Lawrence37:
Fixed a typo.

Lawrence37:

PR #7264, opened 2024-12-08
Demosaiced floating point DNGs

### LINKED ISSUE #7263: Issue loading float DNG file from Photomatix7
Trying to open a 32 bit float image generated by Photomatix7, rawtherapee doesn't seem to understand how to interpret that file. The preview turns black when trying to develop it and no options for demosaicing are listed.
rt
When clearing the processing profile/deleting the .pp3 the preview is displayed correctly again - so apparently rt can load the thumbnail from the raw file
rt2

A 32 bit float dng generated with hdrmerge can be developed just fine however, both files can be opened/developed by AffinityPhoto

to make sure it's not a DNG 1.7 issue, checked with exiftool and both .dng files have DNG version 1.4 with 32bit float samples. comparing the exif tag values (via python+exifread) they also seem identical - image make and model are present, so rt should be able to identify it's a camera with an x-trans sensor?
...so perhaps it's an issue with the exif tag locations? but copying the exif from the hdrmerge file over to the photomatix file also doesn't change anything

i uploaded the two .dng files in question here:
https://nuki.m4studios.at/_DSF3289And8more_photomatix.dng
https://nuki.m4studios.at/_DSF3289And8more_hdrmerge.dng

and here's the exif info of both files (read via python/exifread):
[_DSF3289And8more.txt](https://github.com/user-attachments/files/18014088/_DSF3289And8more.txt)

Lawrence37:
Allows demosaiced floating point DNGs read by LibRaw to be used.

### DIFF
--- a/rtengine/rawimage.cc
+++ b/rtengine/rawimage.cc
@@ -465,6 +465,7 @@ int RawImage::loadRaw(bool loadData, unsigned int imageNum, bool closeFile, Prog
 {
     ifname = filename.c_str();
     image = nullptr;
+    image_from_float.reset();
     verbose = settings->verbose;
     oprof = nullptr;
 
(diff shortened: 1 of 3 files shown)


Lawrence37:

PR #7265, opened 2024-12-09
Quick info font size

Lawrence37:
The quick info font size is always 10pt. This makes it use the main font size set in Preferences.


Lawrence37:
I will merge soon if there are no objections.

Issue #5291, opened 2019-04-19
Real HiDPI support


Hombre57:
RT actually support theme scaling and icon scaling, but it do so manually based on the initial condition (at startup time), and with forcing `GDK_SCALE` to 1. It works fine in Windows and in Linux (hat has been tested), but not on MacOS, at least not with HiDPI resolutions, as reported in issue #5282.

This issue will list the list of change in RT so have real HiDPI support, compliant with the Gtk API, that should enable on-the-fly resolution change handling.

Here is an initial TODO list :

- [ ] `Gtk::Pixbuf` doesn't handle HiDPI mechanism (or someone will have to show me a documentation that prove me wrong), so all `Gtk::Pixbuf` will have to be replaced by `Cairo::ImageSurface`.
- [x] create a `Gtk::CellRenderer` subclass to display an _expander_ arrow closed or opened using two `Cairo::ImageSurface`, not `Gtk::Pixbuf` trough the `Gtk::CellRendererPixbuf`.
- [ ] since HiDPI mode consist in using a logical resolution that is half the physical resolution, the mouse event coordinate are given in this virtual resolution. All mouse location event handler has to be updated to handle sub-pixel accuracy (e.g. X=120.5 in _virtual space_ will be converted to X=241 in _physical space_)
- [ ] all custom widgets should be updated to handle HiDPI mode. They've been updated for the actual implementation, so there might not be much work here.
- [ ] all widget that displays `Cairo::ImageSurface` must handle an invalidation mechanism for on-the-fly HiDPI resolution support, e.g. when moving from one screen to another.
- [ ] for dual monitor mode, some icons (e.g. from the ToolPanel) may have to be created for standard DPI and for HiDPI mode. The `RTWindow` will have to evolve to BE A `RTScalable`, and that class should remove the `static` attribute for `dpi` and `scale`.
- [ ] for the reason explained above, `RTImage` caches will have to handle multiple resolution of the same image in `surfaceCache`

...to be completed.

moomin:
Hi team, sorry if it is not the right place to describe my issue.

I've read #3547 and the discussion there is focused mostly on UI elements and fonts rendering. As I understand it was indeed the goal of the enhancement. But I couldn't find any clear statement or description of how RT is supposed to handle preview in HiDPI environments, especially in combination with different OS settings.

In my case, I have a 4K monitor and Ubuntu 19.04, and RT 5.6 from ppa:dhor/myway.
[...]

PR #7285, opened 2025-01-09
Fix blurry GUI with HiDPI scaling

### LINKED ISSUE #5591: 100% vs 200% image preview in macOS
RawTherapee 5.7 and dev show photos zoomed to 200% instead of 100% under macOS Catalina 10.15.2. This is the same for the built-in Retina display of a MacBook Pro (late 2013) and for an external 4K monitor.

This issue is related to #5291, but hopefully there is a simpler solution for macOS. This is because all menus and icons just look right without additional scaling on HiDPI monitors. MacOS seems to use abstract pixels -- my 4K monitor has 1920x1080 abstract pixels per default in the system settings, whereas the physical signal resolves to 3840x2160. Web pages, pdf previews, fonts and also Nikon Capture NX-D exploit the full 4K resolution to show content. This means that using e.g. NX-D, I can view about 8 MP on the 4K monitor at once. With RawTherapee I can only view about 2 MP on the same monitor with the same system settings in macOS.

RawTherapee might be able to leave all menus and icons as they are and "just" scale  image previews to physical pixels?

GTK might support this. See Alex Larsson's talk at GUADEC 2013. He presents abstract pixels at 8:25: http://videos.guadec.org/2013/High%20res%20display%20support%20GNOME/

### LINKED ISSUE #6712: Photo preview 1:1 is 200% under Wayland at 150% OS display scale
**Short description**
With a 31,5" 4K display running at 4K and 150% global display scale running Fedora KDE with Wayland the photo being viewed/edited in RT is slightly blurry at all zoom levels and zooming to 1:1  reveals that that is twice the scale of the same raw image at 1:1 in Digikam or Darktable (where the photo being edited remains sharp)

**Steps to reproduce**
Employ the Wayland display protocol for a session shown on a high DPI display like 4K 32" with the OS set to a (fractional) global scale like 150%, then start RT and observe your photos in the editor (most of the UI incl the overlay UI is slightly blurry but that isn't as problematic to me as the actual photo being edited being blurry)

**Expected behavior**
The same as under X11/other display protocols where this works i.e 100% zoom is actually that and the UI and photo being edited looks good

**Additional information**

develop - ca08a279d
-Fedora 37 Workstation KDE
- I used the auto build script. Here's the contents of AboutThisBuild.txt:
```
Version: 5.9-126-gca08a279d
Branch: dev
Commit: ca08a279d
Commit date: 2023-03-10
Compiler: cc 12.2.1
Processor: x86_64
System: Linux
Bit depth: 64 bits
Gtkmm: V3.24.7
Lensfun: V0.3.3.0
Build type: release
Build flags:  -std=c++11 -ffp-contract=off -march=native -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas -O3 -DNDEBUG -ftree-vectorize
Link flags:  -march=native
OpenMP support: ON
MMAP support: ON
Build OS: Linux 6.1.14-200.fc37.x86_64 x86_64
Build date: Sun, 12 Mar 2023 14:01:31 +0000 UTC
Build epoch: 1678629691
Build UUID: 7561cc43-14ac-4ccc-9881-ae4d7faa01bf
```

Other useful information:
- The pseudo highDPI mode makes the UI elements a bit too large and does not affect the size or quality of the photo preview.
- RT doesn't react to changing the "Legacy Applications" setting in KDE with Wayland (let them scale themselves or let the OS scale them). I guess it isn't expected to.  Running with `GDK_SCALE=2 rawtherapee` and the pseudo high DPI mode on the UI is very much too large but the Photo preview 1:1 is still 200%.
- Relates to https://github.com/Beep6581/RawTherapee/issues/5291 and https://github.com/Beep6581/RawTherapee/issues/5591 and potentially several older closed issues. There's a very real chance that this is a duplicate of an older issue what with quite a number having to do with scaling;  please point me in that direction if so, I don't mean to make noise : )
-  Since this isn't an issue under X11 I wanted to see what the intention is for Wayland and if it simply is a case of 5291 not being implemented in full and that will take care of this issue when it is.   (I don't know what the state is under other display protocols)
- I'm up for testing things if asked
- Thanks for a great app.

### LINKED ISSUE #7189: Smalll icons on HiDPI display with large fractional scaling
A lot of graphical elements of the interface are not scaled in version 5.11, and because of that they are inconvenient to use on 4K UHD monitor. I don't have such problem with Adobe Camera Raw.

**Screenshot of RawTherapee 5.11:**

RawTherapee 5 11 - blank octagons

**Screenshot of RawTherapee 5.10:**

RawTherapee - blank octagons

OS: Windows 11 (23H2)

digitalcarp:
This PR resolves blurry GUI HiDPI issues I've been running into when using RawTherapee on Wayland at 125% scaling.

With HiDPI, cairo drawing is done in logical pixels. A logical pixel represents some integer number of device pixels.

To handle HiDPI, back buffers and preview data needs to be sized for device pixels. Then `cairo_set_device_scale(.., device_scale, device_scale)` should be used on the cairo surface to indicate that it does not need to be upscaled.

For example, if `device_scale=2`, a preview that needs to cover a (10x20) logical pixel area should be sized (20, 40) and then `hidpi::setDeviceScale(surface, 2)` is called.

I refactored several `int` variables into more descriptive structs (e.g. `hidpi::LogicalCoord` and `hidpi::DeviceCoord`). This was partly to help me reason about the code and partly as documentation. Unfortunately, this refactor makes the PR a bit noisy.

This PR fixes #5591, fixes #6712, fixes #7189. It should help issue #5291 along.

# Features Fixed

* Blurry text/icons
* Blurry file browser/batch queue thumbnails
* Blurry navigator preview
* Blurry editor panel image

# Limitations

The inspector is still blurry. I couldn't untangle the math. The code implies it is already HiDPI aware, but doesn't seem to work in my testing.

If your compositor scales to a fractional value (e.g. 125%) by rendering at 200% and downscaling, the 1:1 zoom option will not be a 1:1 pixel match. The image will be smaller than expected. This effect is worse when the fractional value needs to be rounded up more. We could potentially fix it with the following, but I have not made the change here. The change makes fractional scaling behavior more like what I observed in Windows.

```cpp
#ifdef GDK_WINDOWING_WAYLAND
    gdk_set_allowed_backends("x11,*");
#endif
```

# Testing

* Arch Linux - KDE Plasma 6 - Wayland (for some reason DPI doesn't change?)
  * 100% => DPI=96 device_scale=1
  * 125% => DPI=96 device_scale=2
    * Effectively downscaled from 200% which means 1:1 zoom is too small, but at least the image is not blurry
  * 200% => DPI=96 device_scale=2
 * Windows 10
  * 100% => DPI=96 device_scale=1
  * 125% => DPI=120 device_scale=1
  * 200% => DPI=96 device_scale=2


Lawrence37:
I am using Plasma Wayland too, but the inspector looks sharp to me. The way Wayland works and what GTK is capable of leads me to believe that it is not possible to get true 1:1 rendering with fractional scaling. I have done some research into this topic. I did not manage to find a way to ask Wayland to not scale the graphics using GTK. There were comments indicating that it is not possible. The closest I found was a way to simulate 1:1 rendering by getting the fractional scale using [gdk_surface_get_scale](ttps://docs.gtk.org/gdk4/method.Surface.get_scale.html) or similar, but it is only possible in GTK4.

digitalcarp:
I guess the build failed because whatever version of cairomm is being used doesn't expose a certain API. I'll use a custom wrapper for the C version instead.

digitalcarp:
> I am using Plasma Wayland too, but the inspector looks sharp to me. The way Wayland works and what GTK is capable of leads me to believe that it is not possible to get true 1:1 rendering with fractional scaling. I have done some research into this topic. I did not manage to find a way to ask Wayland to not scale the graphics using GTK. There were comments indicating that it is not possible. The closest I found was a way to simulate 1:1 rendering by getting the fractional scale using gdk_surface_get_scale or similar, but it is only possible in GTK4.

I noticed that darktable doesn't have scaling issues when I'm on 125%. I debugged it a little bit. The reason is that when they call the following code in their setup code, the device scale remains at 1 until the user goes to 200% where it changes to 2. Mysterious GTK behaviour that I don't understand.

```cpp
#ifdef GDK_WINDOWING_WAYLAND
    gdk_set_allowed_backends("x11,*");
#endif
```

Lawrence37:
The documentation for `gdk_set_allowed_backends` says it uses the first available backend in the list. Putting x11 first means GDK will use X11 if it's available. X11 does not have true fractional scaling. I'm guessing you will see the same scaling issues if you don't have xwayland installed.

Lawrence37:
I just did some testing. It's a great improvement. I realized there's a difference between the inspector tab and the fullscreen inspector. The fullscreen version already works fine with HiDPI. Maybe you can see what the difference is between the two?

digitalcarp:
Do you mind if I look at fixing the inspector issues in another PR?

Lawrence37:
That's fine

rtgui/crophandler.cc
```diff
@@ -237,66 +234,71 @@ void CropHandler::setWSize (int w, int h)
...
-{
-    cax = x;
-    cay = y;
+    compDim();
 
-    compDim ();
-
-    if (enabled && update_) {
-        update ();
+    if (update_) {
```
Lawrence37: Why is it no longer checking `enabled`?

rtgui/filebrowserentry.cc
```diff
@@ -226,56 +242,68 @@ void FileBrowserEntry::updateImage(rtengine::IImage8* img, double scale, const r
...
+    int imh = update.img->getHeight();
+
+    bool newLandscape = imw > imh;
     bool rotated = false;
 
-    if (preh == img->getHeight()) {
-        prew = img->getWidth ();
+    if (update.size == previewSize && update.device_scale == pendingDeviceScale) {
+        activeDeviceScale = pendingDeviceScale;
+        previewSize = update.size;
```
Lawrence37: `previewSize` and `update.size` are already equal.

rtgui/rtimage.cc
```diff
@@ -70,6 +70,8 @@ RTImage::RTImage (const Glib::ustring& iconName, const Gtk::IconSize iconSize) :
     if (surface) {
         set(surface->get());
     }
+
+    conn = RTScalable::connectToChanged(sigc::mem_fun(*this, &RTImage::onUpdate));
```
Lawrence37, line 74: Is `conn` used anywhere?

rtgui/rtsigc.h
```diff
@@ -0,0 +1,52 @@
...
+public:
+    RtScopedConnection() = default;
+    RtScopedConnection(sigc::connection&& conn) noexcept : m_connection(std::move(conn)) {}
+
+    RtScopedConnection(const RtScopedConnection&) = delete;
+    RtScopedConnection(RtScopedConnection&& other)
+    {
+        m_connection.disconnect();
+        m_connection = other.m_connection;
+        m_connection = {};
```
Lawrence37: Should this be
```suggestion
        other.m_connection = {};
```
?

Lawrence37 (commented):
When adjusting crops in the file browser, there's a mismatch between the crop and mouse pointer location.
image

rtgui/crophandler.cc
```diff
@@ -643,21 +630,40 @@ void CropHandler::compDim ()
...
     if (cropH > fullH) {
         cropH = fullH;
     }
 
-    displayHandler->setDisplayPosition(imgX, imgY);
+    // Update display position
+    ImageCoord scaledAnchor;
+    if (zoom >= 1000) {
+        scaledAnchor.x = cax * (zoom/1000);
+        scaledAnchor.y = cay * (zoom/1000);
```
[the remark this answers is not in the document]
Lawrence37: Ignore (existing code was like this)

rtgui/crophandler.cc
```diff
@@ -216,18 +214,17 @@ float CropHandler::getZoomFactor ()
...
     if (zoom >= 1000) {
-        cw = ww * 1000 / zoom;
-        ch = wh * 1000 / zoom;
+        cw = output.width * 1000 / zoom;
+        ch = output.height * 1000 / zoom;
     } else {
-        cw = ww * (zoom / 10);
-        ch = wh * (zoom / 10);
+        cw = output.width * (zoom / 10);
+        ch = output.height * (zoom / 10);
```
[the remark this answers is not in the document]
Lawrence37: Ignore (existing code was like this)

rtgui/rtimage.cc
digitalcarp: `conn` is an `RtScopedConnection` so it needs to be assigned for proper lifetime/scope. Even if it is not used elsewhere

rtgui/rtsigc.h
digitalcarp: Fixed

rtgui/filebrowserentry.cc
digitalcarp: Fixed

rtgui/crophandler.cc
digitalcarp: I don't remember. It's probably a bug. Reverted.

digitalcarp:
Still looking at the file browser crop mouse position issue

Resolved: 991895c

Lawrence37:
Hovering over the crop boundary and grabbing it works correctly now, but there are remaining issues with the crop. The crop boundary does not keep up with the cursor. Creating a new crop (after selecting the crop tool or pressing the "c" key) also doesn't work well.

digitalcarp:
Could you elaborate on the differences you are seeing? I'm on a branch without these hidpi changes and the file browser crop behaves the same. The crop boundary really only tracks in one direction and starting a new crop sometimes doesn't have a visual update.

Lawrence37:
Here's a comparison between 5.12 and this branch.
| 5.12 | max-hidpi |
| - | - |
| crop-resize-5 12 | crop-resize-hidpi |
| new-crop-5 12 | new-crop-hidpi|

digitalcarp:
I think I've fixed the issue. The crop in file browser is still kind of janky, but seemingly nothing that isn't already in the dev branch.

Lawrence37 (approved):
Looks good now!

kaesa:
Something in this change is causing RawTherapee to segfault when directory has images but some are filtered out from the view and the user re-reads the directory.

Steps to reproduce:
1. Enter in a directory with images.
2. Write arbitrary characters to the "Find:" query box (top right) until some or none of the  thumbnails are no longer shown.
3. Click on the refresh button left from the query box. Or use any other method to re-read the directory.

Relevant part of trace produced with release build:

```
#0  0x0000555555a86873 in hidpi::setDeviceScale(Cairo::RefPtr<Cairo::ImageSurface> const&, int, int) ()
#1  0x0000555555c275bf in ThumbBrowserEntryBase::updateBackBuffer() ()
#2  0x0000555555a38709 in FileBrowserEntry::_updateImage(ThumbImageUpdateListener::ImageUpdate&&) ()
#3  0x0000555555a38820 in std::_Function_handler<bool (), std::_Bind<FileBrowserEntry::updateImage(ThumbImageUpdateListener::ImageUpdate&&)::{lambda(ThumbImageUpdateListener::ImageUpdate&)#1} (ThumbImageUpdateListener::ImageUpdate)> >::_M_invoke(std::_Any_data const&) ()
#4  0x0000555555a70201 in IdleRegister::add(std::function<bool ()>, int)::{lambda(void*)#1}::_FUN(void*) ()
```

This code path is most likely executed after all the previews have been generated and `FileCatalog::previewsFinishedUI` interface function is called (through a middle step) by the `PreviewLoader`. The `FileCatalog::previewsFinishedUI` calls `fileBrowser->refreshQuickThumbImages();`, which, through middle steps, launches  a pool of threads from `ThumbImageUpdater` to recreate all thumbnails in low priority background threads applying processing profiles to them.

PR #7291, opened 2025-01-19
Delete pp3 when clearing processing parameters

### LINKED ISSUE #7289: 'delete profile' doesn't delete the pp3-file
Hi there,
deleting a profile should also delete the sidecar file, isn't it? 
What am i overlooking? I was quite sure this was the function in former versions.

How to reproduce: 
Open RT filemanager, choose any former editet picture, rightclick, menu profile operations - delete profile.
In filmstrip you will see, that the applied changes for this picture will be reset, but he checkmark still remains. As a result, a pp3-file sidecar will stay there and is not deleted.

expected behavior:
Sidecar should be deleted also. When a sidecar (even empty) is available, no dynamic profile will be applied, when opened next time.

btw: I am on windows 10 with RT 5.11-146 dev
marter

edit:  i tried the same with 5.11-rc and here it behaved as expected: checkmark has gone and pp3-file was deleted

Lawrence37:
See the commit message.

Fixes #7289.


Lawrence37:

PR #7292, opened 2025-01-19
Format in output template

### LINKED ISSUE #7196: Add file format specifier to batch queue output location template
For the batch queue output location template, it would be helpful to have a specifier for the file format.

Usually when processing my files with the batch queue, I output in more than one file format.  For example, I might process once, writing .jpg files and then a second time writing out .png files.

For the output location, I use something like "%p1/converted_jpg/%f" for the jpg files to output in a folder called converted_jpg in the source photo's folder.  Then, when I output to the .png format, I (hopefully remember to) change the output location to "%p1/converted_png/%f".

I propose adding a "%e" specifier that is replaced with the output format.  That way, I can use an output location of "%p1/converted_%e/%f" that works in both cases. A different specifier than "%e" could be used.  I picked "e" for extension.

I think this capability would be helpful for anyone who wants to output in folders with the format included in the name.

I've added this capability to my own compiled version of RawTherapee. I've attached a diff.

There were only a few minor changes required to make this update.  I only updated rtgui/batchqueue.h and rtgui/batchqueue/cc.  I added a third argument (the format) to the calcAutoFileNameBase function, which is where the template is expanded.  This function appears to be called from a couple other place in the code with only a single argument.  My limited testing did not specifically ensure that I didn't break anything when the function is called this way.  I also didn't update the full instructions that come up when you press the "?" button.

digitalteg:
I added in a "%e" format specifier for the batch queue output location template.

I also updated the help panel for the batch queue output location to document the added specifier.

Now, for example, an output location of "%p1/converted_%e/%f" will include the file extension (output format) in the folder path for the output image.  So, the folder path will contain "converted_jpg" or "converted_png", etc., depending on the output file extension.

Resolves #7196

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -2222,8 +2222,10 @@ QUEUE_FORMAT_TITLE;File Format
 QUEUE_LOCATION_FOLDER;Save to folder
 QUEUE_LOCATION_TEMPLATE;Use template
 QUEUE_LOCATION_TEMPLATE_HELP_BUTTON_TOOLTIP;Show or hide a help panel with instructions for creating location templates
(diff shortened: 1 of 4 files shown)


Lawrence37:
Thank you!

Lawrence37 (approved):

PR #7299, opened 2025-02-02
Fujifilm electronic shutter raw crop for dcraw

Lawrence37:
Fixes the raw crop for certain Fujifilm cameras when using the electronic shutter. See https://discuss.pixls.us/t/issue-with-raw-files-from-fuji-x-e4/44083.


Lawrence37:
I will merge soon.

PR #7319, opened 2025-02-25
Selective Editing - Avoid negatives values  - see on Pixls.us

Desmis:
I am trying to solve this problem knowing that on my machine both on Windows 11 (24H2) and under Ubuntu 24.04 in all cases that I tried with the Raw and the pp3 provided, in "preview" mode, in TIF output or in Queue (fast export, etc.) no crash occurs.

On the other hand I tried "Contrast by detail levels" then SE with Color & Light contrast, there I noticed a crash.
From tests done on another branch in development, it seems that upstream of SE, CBDL or Dehaze can generate RGB values ​​with very low negative values ​​(of the order of -0.01 on a scale of [0 65535]) which are poorly accepted by SE in some cases only.

I added at the beginning of the SE code, a check that ensures that there are no negative RGB values, in 3 cases:
* Contrast by detail levels enable 
* Haze removal enable
* I added a checkbox, enable by default in SA > Settings > Specific cases > **Avoid negatives values**. I don't know if this will be effective for what is discussed in the thread on Pixls.us, but it's worth a try (because I don't have any crashes in this case). Maybe it is due to a specific processor (AMD, Intel or a Bios configuration...) which could be more (or less) precise for certain calculations where the results are close to zero or to a Raw Type with particular black point, or ??.

https://discuss.pixls.us/t/crash-when-exporting-or-moving-detail-window/48453/1

### DIFF
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -1488,7 +1488,7 @@ HISTORY_MSG_ICM_WORKING_PRIM_METHOD;Primaries method
 HISTORY_MSG_ICM_WORKING_SLOPE;TRC - Slope
 HISTORY_MSG_ICM_WORKING_TRC_METHOD;TRC method
 HISTORY_MSG_ILLUM;CAL - SC - Illuminant
-HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Avoid zero and negative values
+HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Pre-filter zero and negative values
 HISTORY_MSG_LOCAL_CIE_CONTSIG;Local CIECAM - Sigmoid contrast
 HISTORY_MSG_LOCAL_CIE_SKEWSIG;Local CIECAM - Sigmoid skew
 HISTORY_MSG_LOCAL_CIE_SMOOTHTH;Local CIECAM - Attenuation threshold
@@ -2998,7 +2998,7 @@ TP_LOCALLAB_AVOID;Avoid color shift
 TP_LOCALLAB_AVOIDCOLORSHIFT_TOOLTIP;Fit colors into gamut of the working color space and apply Munsell correction (Uniform Perceptual Lab). Default: Munsell only.\n\nMunsell only: Fixes Lab mode hue drifts due to non-linearity when chromaticity is changed (Uniform Perceptual Lab).\nLab: Applies a gamut control in relative colorimetric. Munsell is then applied.\nXYZ Absolute: Applies gamut control in absolute colorimetric. Munsell is then applied.\nXYZ Relative: Applies gamut control in relative colorimetric. Munsell is then applied. The result is not the same as Lab.
 TP_LOCALLAB_AVOIDMUN;Munsell correction only
 TP_LOCALLAB_AVOIDMUN_TOOLTIP;Munsell correction always disabled when Jz or CAM16 is used.
-TP_LOCALLAB_AVOIDNEG;Avoid zero and negative values
+TP_LOCALLAB_AVOIDNEG;Pre-filter zero and negative values
 TP_LOCALLAB_AVOIDRAD;Soft radius
 TP_LOCALLAB_BALAN;ab-L balance (ΔE)
 TP_LOCALLAB_BALANEXP;Laplacian balance
--- a/rtdata/languages/default
+++ b/rtdata/languages/default
@@ -1488,7 +1488,7 @@ HISTORY_MSG_ICM_WORKING_PRIM_METHOD;Primaries method
 HISTORY_MSG_ICM_WORKING_SLOPE;TRC - Slope
 HISTORY_MSG_ICM_WORKING_TRC_METHOD;TRC method
 HISTORY_MSG_ILLUM;CAL - SC - Illuminant
-HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Avoid negatives values
+HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Avoid zero and negative values
 HISTORY_MSG_LOCAL_CIE_CONTSIG;Local CIECAM - Sigmoid contrast
 HISTORY_MSG_LOCAL_CIE_SKEWSIG;Local CIECAM - Sigmoid skew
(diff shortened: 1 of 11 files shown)


Desmis:
@Lawrence37 

I tried to create executables with the yml files.

With,  publish_pre_dev_labels: '["Beep6581:avoidneg"]' the system does not give errors but does not create executables, or not in https://github.com/RawTherapee/RawTherapee/releases/tag/pre-dev-github-actions

With,  publish_pre_dev_labels: '["Rawtherapee:avoidneg"]' the system give errors : PR run failed

I see that in the yml files there are references (towards the end) to Beep6581, but I don't know anything about it, and don't know what should or shouldn't be changed

But you can find them in "actions"

Lawrence37:
"Beep6581" should be changed to "RawTherapee". This should have been done after moving the repository to the organization, but I did not remember we were using Beep6581 in the workflow files.

rtdata/languages/default
```diff
@@ -1488,6 +1488,7 @@ HISTORY_MSG_ICM_WORKING_PRIM_METHOD;Primaries method
 HISTORY_MSG_ICM_WORKING_SLOPE;TRC - Slope
 HISTORY_MSG_ICM_WORKING_TRC_METHOD;TRC method
 HISTORY_MSG_ILLUM;CAL - SC - Illuminant
+HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Avoid negatives values
```
Lawrence37:
```suggestion
HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Avoid negative values
```

rtdata/languages/default
```diff
@@ -2997,6 +2998,7 @@ TP_LOCALLAB_AVOID;Avoid color shift
 TP_LOCALLAB_AVOIDCOLORSHIFT_TOOLTIP;Fit colors into gamut of the working color space and apply Munsell correction (Uniform Perceptual Lab). Default: Munsell only.\n\nMunsell only: Fixes Lab mode hue drifts due to non-linearity when chromaticity is changed (Uniform Perceptual Lab).\nLab: Applies a gamut control in relative colorimetric. Munsell is then applied.\nXYZ Absolute: Applies gamut control in absolute colorimetric. Munsell is then applied.\nXYZ Relative: Applies gamut control in relative colorimetric. Munsell is then applied. The result is not the same as Lab.
 TP_LOCALLAB_AVOIDMUN;Munsell correction only
 TP_LOCALLAB_AVOIDMUN_TOOLTIP;Munsell correction always disabled when Jz or CAM16 is used.
+TP_LOCALLAB_AVOIDNEG;Avoid negatives values
```
Lawrence37:
```suggestion
TP_LOCALLAB_AVOIDNEG;Avoid negative values
```

There could be confusion about what this does. Some people can read this and think it will prevent the spot from creating negative values. Maybe something like "Pre-filter negative values" would work?

rtengine/iplocallab.cc
```diff
@@ -14276,13 +14278,50 @@ void ImProcFunctions::Lab_Local(
...
+    int bw0 = transformed->W;
+    int bh0 = transformed->H;
+
+    float epsi0 = 0.000001f;
+    bool nocrash = false;
+    bool cbdl = false;
+    if(params->dirpyrequalizer.cbdlMethod == "bef") {//If user choose "after black and white" this function which removes negative values is not used, hence CBDL is best performed, after Selective Editing in Lab mode
+        cbdl = true;
+    }
+    bool dirpyrdeha = (params->dirpyrequalizer.enabled && cbdl)  ||  params->dehaze.enabled  || lp.avoidneg;//lp.avoidneg in setting
```
Lawrence37: `dirpyrdeha` is just a duplicate of `nocrash`.

rtengine/iplocallab.cc
```diff
@@ -14276,13 +14278,50 @@ void ImProcFunctions::Lab_Local(
...
+
+
+
+
+    //avoid negative values RGB then Lab when using before SE CBDL or Dehaze, or processor type...
+
+    int bw0 = transformed->W;
+    int bh0 = transformed->H;
+
+    float epsi0 = 0.000001f;
```
Lawrence37, line 14298: This means it avoids negative and zero values. Should the setting be renamed to "Avoid zero and negative values" (or "Avoid non-positive values")?

rtengine/iplocallab.cc

rtdata/languages/default
Desmis: I will push "Avoid zero and negative values"

rtdata/languages/default
Desmis: I will push "Avoid zero and negative values", I think it is understantable..

rtengine/iplocallab.cc
Desmis: I will push "Avoid zero and negative values"
Just a note to show the specificity of languages ​​in French, it is a mistake not to put an "s" in "negative"

Lawrence37: Is it because adjectives in French must have an extra -s when they refer to plural nouns? In English, adjectives don't have plural forms. Spell check might not catch "negatives" because "negative" can also be a noun.

rtdata/languages/default
Lawrence37: You don't think there will be confusion about what it does?

Desmis: After re-reading, I think you are true. I change for
"Pre-filter zero and negative values"

rtengine/iplocallab.cc
Desmis: I am not disputing the grammar of another language, just whether these "subtleties" are well perceived by automatic translators...

Lawrence37: I know. I'm just asking out of curiosity. Besides English, I have only ever learned (some) Spanish. Spanish has plural forms of adjective and I was wondering if it's the same in French.

Lawrence37 (approved):

rtengine/iplocallab.cc
Desmis: French is a complex language. There are several types of adjectives. The agreement of the conjunctions depends on the auxiliary verb and the position of the object complement.
It is a very old language taking its sources from Latin, Greek, a (little) bit of Germanic languages, Arabic... and now recently English (through social networks, students). But did you know that about 50% of English words come from French (and not the other way around).

PR #7322, opened 2025-02-28
Fixes issue #7282

### LINKED ISSUE #7282: Noise Reduction Luminance Curve changed when using Batch Edit
**Short description**
The Noise Reduction Luminance Curve ([Directional Pyramid Denoising], LCurve parameter) is overwritten when using Batch Edit

**Steps to reproduce**
You need two (partial) profiles that apply different luminance noise curves, e.g. those attached.
Then:
- start RawTherapee, FileBrowser in a folder with at least two photos
- right-click on first photo / Processing profile operations / Apply --> profile 1
- right-click on second photo / Processing profile operations / Apply --> profile 2
- open the two .pp3 files that belong to your photos and verify that "LCurve" parameters are indeed different (as they should be)
- select the two photos
- batch edit / exposure / adjust exposure compensation (or any other batch edit operation)
- deselect the two photos so .pp3 files get updated
- open the two .pp3 files that belong to your photos and verify that "LCurve" parameters are now the same (which should not be)

**Expected behavior**
Batch editing something else should not affect the Noise Reduction Luminance Curve.
(For chrominance curve, i.e. CCCurve parameter, it works as it should.)

**Additional information**
Sample partial profiles that can be used to reproduce (rename to .pp3):

[Profile2.txt](https://github.com/user-attachments/files/18321752/Profile2.txt)

[Profile1.txt](https://github.com/user-attachments/files/18321751/Profile1.txt)

Bezierr:
This is supposed to fix issue [#7282] "Noise Reduction Luminance Curve changed when using Batch Edit".

Please note: I have not tried to understand the underlying logic but was only looking for places where "lcurve" (affected by the bug) and "cccurve" (not affected by the bug) were treated differently. Please verify that this fix does not introduce unintended side effects.

### DIFF
--- a/rtgui/dirpyrdenoise.cc
+++ b/rtgui/dirpyrdenoise.cc
@@ -574,6 +574,7 @@ void DirPyrDenoise::read (const ProcParams* pp, const ParamsEdited* pedited)
         passes->setEditedState     (pedited->dirpyrDenoise.passes ? Edited : UnEdited);
         set_inconsistent           (multiImage && !pedited->dirpyrDenoise.enabled);
         median->set_inconsistent   (!pedited->dirpyrDenoise.median);
+        lshape->setUnChanged       (!pedited->dirpyrDenoise.lcurve);
         ccshape->setUnChanged      (!pedited->dirpyrDenoise.cccurve);
 
         //      perform->set_inconsistent (!pedited->dirpyrDenoise.perform);


Lawrence37 (approved):
This makes sense to me. The batch editor relies on the paramsedited to know if a value has been changed. Since the curve was getting overwritten, either the paramsedited for the curve was getting set or it was always assumed to be true. By not setting the unchanged state of the curve, the paramsedited would have been set by the GUI regardless of if the curve was changed.

Issue #6467, opened 2022-04-28
DCP profile size inflated with redundant data


Entropy512:
In the process of making a DCP profile for my Sony A7M4 (will be uploading the shots and creating a separate issue for that later this week), I realized that every single dual-illuminant DCP profile likely has identical tone curve and LookTable settings (since the LookTable is tied to the tone curve, and all of the profiles were generated from the same tone curve).  I plan on confirming this assumption later this week.  DCP profiles without tone curve and NTRO are only 65k, vs 1MB per file.

This is really redundant, especially since RT defaults to not using the tone curve in the file, and using its own tone curve.

The LookTables are used by the "Standard Film Curve" profiles, however I think that this approach is not a good idea:  If the user changes the tone curve settings, the LookTable will no longer match the tone curve, but will still be turned on.

It appears that the idea is to get the behavior of the Perceptual tone curve mode without the performance hit, as (from a rough initial check of the source), Perceptual appears to be similar to if not identical to dcamprof's default NTRO configuration.

@heckflosse what are your thoughts on trying to speed up Perceptual in the same way as the current "precalculated LookTable" setup, by recalculating an HSV LUT when the tone curve is changed?

This would allow skipping the tone curve and LookTable in any future DCP profiles.

Entropy512:
It looks like all profiles copied from ART are ones with no tone curve and no LookTable, so @agriggio appears to already be going this route.

Thanatomanic:
For reference, you need the `-L` flag: `dcamprof.exe make-dcp -n "camera id string" -d "camera id string" -c "RawTherapee CC0" -L tungsten.json daylight.json "camera id string.dcp"`

I think this is a pretty reasonable idea btw, especially considering that the Auto-Matched Tone Curve usually does a pretty good job to get a decent tone curve going by default.

Entropy512:
-L (matrix-only) is not what we want - there are two HueSatMap tables in the profiles that contain a tone curve:
A 2.5D HueSatMap LUT that is exposure-independent and only performs color corrections, applied after the matrix but before the tone curve.  -L disables this, we do not want to disable this

A 3D HueSatMap LUT that is exposure-dependent and is tied to the tone curve.  It's my opinion that we do not want this, since it will no longer be properly applicable the moment someone changes the tone curve.  It appears to be intended to perform the same function as the "Perceptual" tone curve option, but faster since it is LUT-based.  Passing "-t linear" will disable this.

[...]

PR #7326, opened 2025-03-03
Canon EOS R8 DCP

### LINKED ISSUE #7300: Default color profile for the Canon EOS R8
The Canon EOS R8 does not have a DCP profile in RT, and the standard profile produces colors so off it's impossible for an amateur like me to fix it. I followed the directions [here](https://rawpedia.rawtherapee.com/How_to_create_DCP_color_profiles#Shooting_the_Color_Target) to shoot a color target. You may find my photos [here](https://filebin.net/vrp6axczgphp2xmy).

For reference, the daylight pictures were taken at noon in winter in the northeast US, with some snow on the ground. I used the less exposed daylight photo when making my own DCP profile. That profile yields significantly better results than the default, so I wanted to share.

Lawrence37:
This DCP was created using the procedure described in RawPedia. I'm open to the idea of removing the tone curve and look table as described in #6467 to reduce the file size.

Closes #7300.


Lawrence37:
I replaced the DCP with the smaller version without the tone curve and look table.

PR #7340, opened 2025-03-26
Reorg toolpanels

digitalcarp:
Moves tool panel source code to `rtgui/tools` as part of #4299.

Issue #5238, opened 2019-03-21
Moving towards modern c++


okoncevoy:
Looking through RT code, I often see mixins from modern c++ (closures, constexpr, auto, etc), but also there are many parts, that are written in old style. Can I start moving some parts to more modern approach? Will this be acceptable?

heckflosse:
Of course that will be acceptable after review. :+1:

Benitoite:
Are you talking c++14?  I'm using AppleClang 10, which is mostly compatible with '14.

okoncevoy:
I would be nice, if we can start using c++14. I have clan 7.0.1 which also supports 14 standard. But I where talking not only about compiler support, or just enabling flag for compilation. But about refactoring of already written code. This can decrease amount of code, increase maintainability and might be event performance at some places.

Benitoite:
That is great to hear.  One issue I ran into on mac with '14 code in Filmulator was that Appleclang10 did not have `std::make_unique` but there was an equivalent workaround found that compiled. 
https://github.com/CarVac/filmulator-gui/commit/ee5505c99cfbe24ffd0d85f7ec54f07a491288a1

PR #7341, opened 2025-03-27
Migrate to C++20 and CMake 3.25

digitalcarp:
gtk4mm requires C++17. Notable C++17 features:
* `std::optional`
* `std::variant`
* `std::string_view`
* Useful template programming helpers
* NRVO
* CTAD

New versions of gcc [12](https://gcc.gnu.org/gcc-12/), clang, MSVC, and Apple clang all have good support for C++20 ([link](https://en.cppreference.com/w/cpp/compiler_support/20)). Let's jump to C++20 directly to additionally support:
* Concepts
* `std::jthread`
* `<numbers>`
* `<bits>`
* `std::span`

Our previous minimum version of CMake is also very old and newer CMake versions mark it as deprecated. Migrating to [3.25](https://www.kitware.com/cmake-3-25-3-available-for-download/) gives:
* FetchContent
  * Add [fmt](https://github.com/fmtlib/fmt) as an example
* CMake presets

This PR is dependent on #7340, so please ignore the moved files and curve code changes. This PR helps with #5238.


digitalcarp:
Builds are failing as an old version of GCC is used: `-- The CXX compiler identification is GNU 9.4.0`.
The Actions config will need to be updated to install gcc 13 or 14.

Lawrence37:
We'll be upgrading the AppImage runner to Ubuntu 22.04 which [officially has up to GCC 11](https://packages.ubuntu.com/search?keywords=gcc&searchon=names&suite=jammy&section=all).

Lawrence37:
According to https://gcc.gnu.org/projects/cxx-status.html, GCC has full support of C++14, almost full support of C++17, and **experimental** support for C++20.

https://clang.llvm.org/cxx_status.html says Clang fully implemented C++17 and partially implemented C++20.

I'm hesitant to adopt C++20 due to the experimental status of GCC.

digitalcarp:
That's fair. For submission, I'll split this PR into two PRs. One for CMake 3.25. One for bumping up to C++17.

P.S. It looks like Ubuntu 24.04 is supported now for Actions. https://github.com/actions/runner-images/issues/10636

PR #7353, opened 2025-04-06
GHS tool panel width fix

### LINKED ISSUE #7352: Tool tab width becomes wider when GHS BP >= 0.0009
**Short description**
When setting the black point value the a bigger value, the side panel automatically becomes wider. Specifically for my setup, the panel expands itself when black point is equal or bigger than 0.0009, and gets wider with every 0.0001 or 0.0002 increment until it reaches 0.0019. Note that I set the panel to its narrowest, and if the user uses a wider panel (wider than the glitch would cause the panel to be), the problem would not occur.

**Steps to reproduce**
1. Enable Local adjustment, add a spot with GHS tool, set stretch factor to 0.001 or 0.002.
2. Black Point (BP linear) is now adjustable. Dragging the slide to the right and make the value near positive.
3. Add 0.0001 to the value by hand. The side panel should suddenly become wider when a certain value is reached.

**Expected behavior**
Step 3 does not happen and GUI keeps itself consistent.

**Additional information**
- Self-built Rawtherapee 5.10-384-gf99132b48 on 31 Mar 2025.
- For reference, I'm using a 2K, 27-inch screen for what's matter.

Lawrence37:
Allow the black and white point information to wrap.

Closes #7352.

### DIFF
--- a/rtgui/locallabtools.cc
+++ b/rtgui/locallabtools.cc
@@ -4448,7 +4448,11 @@ LocallabShadow::LocallabShadow():
     ghs_MID->setAdjusterListener(this);
     ghs_BLP->setAdjusterListener(this);
     ghs_HLP->setAdjusterListener(this);
+    ghsbpwpLabels->set_line_wrap();
+    ghsbpwpLabels->set_justify(Gtk::Justification::JUSTIFY_CENTER);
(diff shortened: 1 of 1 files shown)


Desmis:
@Lawrence37 

Thank you.
I didn't know this command existed....which makes things easier, especially when the values ​​displayed in the Label can have a lot of numbers.
Jacques

PR #7359, opened 2025-04-16
parse_qt: possible integer overflow

npt-1707:
Description
This PR fixes a possible integer overflow in parse_qt() that was cloned from LibRaw but did not receive the security patch. The original issue was reported and fixed under https://github.com/LibRaw/LibRaw/commit/1334647862b0c90b2e8cb2f668e66627d9517b17. This PR applies the same patch to eliminate the vulnerability.

References
https://nvd.nist.gov/vuln/detail/CVE-2018-5815
https://github.com/LibRaw/LibRaw/commit/1334647862b0c90b2e8cb2f668e66627d9517b17

### DIFF
--- a/rtengine/dcraw.c
+++ b/rtengine/dcraw.c
@@ -6817,8 +6817,6 @@ void CLASS parse_qt (int end)
   while (ftell(ifp)+7 < end) {
     save = ftell(ifp);
     if ((size = get4()) < 8) return;
-    if ((int)size < 0) return; // 2+GB is too much
-    if (save + size < save) return; // 32bit overflow
     fread (tag, 4, 1, ifp);
     if (!memcmp(tag,"moov",4) ||
 	!memcmp(tag,"udta",4) ||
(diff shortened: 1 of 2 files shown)


Lawrence37:
Thanks for bringing this to our attention, but all modifications should be made to `dcraw.cc` instead. `dcraw.c` does not get compiled into RawTherapee. It only serves as a base on top of which all the RawTherapee-specific modifications are applied to. (see https://github.com/RawTherapee/RawTherapee/blob/dev/rtengine/dcraw.patch.README)

npt-1707:
Thank you for your feedback, @Lawrence37! I have updated the change in the draw.cc file. Please review it at your convenience.

Lawrence37:
Could you revert `dcraw.c`? We want to keep it identical to the original.

npt-1707:
I reverted `draw.c`. Please review it at your convenience.

Lawrence37 (approved):

npt-1707:
Thanks for merging my PR, @Lawrence37! 

Just wanted to let you know that I plan to report this as a CVE. Please let me know if you have any concern. Thanks!

Issue #7361, opened 2025-04-17
New Chinese (Simplified) translation for Rawtherapee 5.12


syyrmb:
Incremental translation for local adjustment's GHS tool. 

Corrected one error introduced by last update (I replaced all Chinese translation for "Local Adjustments" with "Selective Editing" (in Chinese), forgetting that LA tool itself still exist without any name change).

Corrected very few legacy errors that was there since before 2020 (when I was not involved so they have been overlooked).

I found one copy-paste mistake in Default file: TP_LOCALLAB_GHS_HP_TOOLTIP: This parameter must be at most 1 and no less than the Symmetry point.\n * The adjuster will automatically limits the **_maximum_** value to the Symmetry Point (SP).

This one is probably copied from Protect Shadows' tooltip and I think "maximum" should read "minimum". Also I believe there are some other typo in other GHS tooltips but I didn't document them. Would be great if someone can do further proof reading. I'll also provide update in this issue if I rediscover them.

[Chinese (Simplified).txt](https://github.com/user-attachments/files/19794066/Chinese.Simplified.txt)

Desmis:
@syyrmb 

For me (with my bad english) the tooltip is correct, it is the "maximum" the good values. I just added a small extra value of 0.001 to make sure there won't be a crash.
For example if SP=0.00996, HP will have a minimum value of 0.0096 + 0.001 = 0.01096

But I admit it's quite subtle. The goal of having automatic adjustment of the 3 sliders linked SP, LP, HP is to avoid crashes...

In fact it's SP  who leads the movement, after it's only GUI

But as you are the first to make a remark if it seems clearer to you, it is easy to change and put "minimum" for HP

Desmis:
@syyrmb @Lawrence37 

I suggest changing the two tooltips for clarity. I think @syyrmb is right about the "minimum," and I'll clarify by adding
"less (or more) a small value to avoid crashes."

The 2 tooltips:
TP_LOCALLAB_GHS_LP_TOOLTIP ; ....., The adjuster will automatically limit the maximum value to the Symmetry Point (SP), minus a small value to avoid crashes.....
and
TP_LOCALLAB_ GHS_HP_TOOTIP;...., The adjuster will automatically limits the minimum value to the Symmetry Point (SP), plus a small value to avoid crashes.....

Of course, it is not essential, but desirable to understand how the GUI and also GHS work... especially due to the lack of current documentation

I Have create a pull request 
#7363

[...]

PR #7363, opened 2025-04-18
Change tooltips TP_LOCALLAB_GHS_LP_TOOLTIP  TP_LOCALLAB_ GHS_HP_TOOTIP

Desmis:
#7361 

I suggest changing the two tooltips for clarity. I think @Syyrmb is right about the "minimum," and I'll clarify by adding 
"less (or more) a small value to avoid crashes."

the 2 tooltips
TP_LOCALLAB_GHS_LP_TOOLTIP
and
TP_LOCALLAB_ GHS_HP_TOOTIP

Of course, it is not essential, but desirable to understand how the GUI and also GHS work... especially due to the lack of current documentation


kaesa:
Respectfully proposing without really understanding the tool.

How about,
in TP_LOCALLAB_GHS_HP_TOOLTIP from:
`* This parameter must be at most 1 and no less than the Symmetry point.\n * The adjuster will automatically limits the minimum value to the Symmetry Point (SP), plus a small value to avoid crashes.\n\n`
to:
` * The adjuster will automatically limit the minimum value based on the Symmetry Point (SP).\n\n`

and in TP_LOCALLAB_GHS_LP_TOOLTIP from:
`* This parameter must be at least 0 and no greater than the Symmetry point.\n * The adjuster will automatically limit the maximum value to the Symmetry Point (SP), minus a small value to avoid crashes.\n\n`
to:
`* The adjuster will automatically limit the maximum value based on the Symmetry Point (SP).\n\n`

Because, the adjustable range is not explained in other tooltips for the GHS tool either. The adjuster minimum/maximum values change when SP is changed, so that is probably good to explain. Crashes mean nothing to people who are not coding the algorithm (or reporting bugs in tracker) and probably is not helpful information in a tooltip.

Desmis:
@kaesa 

No problem at all... That's the purpose of these PRs: to enable constructive dialogue.
I'll make the change you're proposing (or thereabouts, because there have already been 'checks')

Desmis:
@Lawrence37 
I don't know, given the progress of 5.12, whether or not we should validate this PR for 5.12?

Nevertheless, I'm in favor.

Lawrence37 (approved):

PR #7376, opened 2025-04-27
Selective Editing CAM16 - change Tone Mapping Operators to Highlight Tone Mapping - order Tone mapping Q

Desmis:


Desmis:
@waynesutton50 @nosle @Lawrence37 

What are we doing about this PR? From my point of view, as the designer at the base of the system, I am in a bad position, because I am both judge and arbitrator.

To add a little water to the mill, if:
* Slope based, RGB channel slope, Sigmoid based, are clearly (provided we agree on the definition) tone-mappers that act on both dynamics range(DR), shadows, and highlights
* This is only partially true for:
* * EV based , which compresses the data roughly from the middle-grey to the maximum. While this mainly affects the highlights and midtones, the algorithm is neither a curve nor linear, but an exponential.
* * Gamma based, which is actually a '1/2 tone-mapper' (Slope based) and which only acts above middle-grey with an asymptotic hyperbolic type function. This also affects midtones and highlights, but with a different algorithm.

From my point of view, both proposals, if we look at it from the user's point of view, each have their advantages and disadvantages.
May be the initial wording we agreed with Wayne some months ago , will create less  confusion (Tone Mapping Operators).

As for whether or not the default method should be the first one that appears in the title, there is again a vast debate with advantages and disadvantages... There are several examples in Selective Editing where the method proposed by default, notably due to the progress of algorithms, is not the first one in the title (ex: Wavelet, GHS)

I propose for the sake of efficiency and to facilitate 5.12 to leave everything as it was before this PR, but temporarily not to close it, but to put it in 'draft', then to reopen it to start a debate on these difficult and contradictory points (labels, default method, tooltip, etc.)... vast and delicate debate.

Thank you all for your contribution to these less than obvious and subjective issues.

jacques

Lawrence37:
"Tone mapping" is a broad term that fits all the methods. "Highlight tone mapping" is too specific because the slope and sigmoid methods affect more than the highlights. For correctness, we should avoid using "highlight tone mapping".

waynesutton50:
> May be the initial wording we agreed with Wayne some months ago , will create less confusion (Tone Mapping Operators).

@jdc as I mentioned in the pixls.us forum post, more explicit tooltips are probably the best answer. As things stand, the tooltip for Tone Mapping Operators begins with: 

> TP_LOCALLAB_SMOOTHCIE_TOOLTIP;Completes the processing carried out by gamma, slope and midtones **by causing a slight lowering of lights.**

This is misleading in light of your explanation above and is probably a carryover from the original versions of this module. If you care to revise the tooltip I can translate it for you.

> As for whether or not the default method should be the first one that appears in the title, there is again a vast debate with advantages and disadvantages.

This doesn't make sense to me. The tool title is "Sigmoid Q & Slope based" and the tooltip starts  with 

> Allows you to simulate a tone mapping appearance using both 'CIECAM' and 'Sigmoid Q' **or** 'Slope based Q'

To me this implies that "Slope based" is an alternative to Sigmoid, not the default choice.

The explanation that follows first describes the Sigmoid function with "Slope based" being mentioned only towards the end. If you think that the best default option is "Slope based", wouldn't it be better to change the tool title to Slope based and Sigmoid Q" and reword the tooltip accordingly?

Desmis:
@waynesutton50 @Lawrence37 

First of all, thank you for your participation. :)

There are undeniably inconsistencies (not just in Selective Editing CAM16), whether it's the label names, the associated tooltips, or the default methods), and I see no problem (quite the contrary) in reviewing them. Another example in 'Exposure tab' is Tone Mapping, which, although it is indeed a tool of this type, is confusing (ART made another choice).

Along the same lines, and this is a conversation we've had before, but it would be nice to share more broadly, what about the order of 'Tabs' and tools in Rawtherapee? From what I've heard, it's due to the 'usual' order of using the tools...But that was in 2006. For example, one of the major tools is Capture Sharpening, which is completely isolated, relegated to the back of the pack...

What I propose, as soon as 5.12 is released (so as not to disrupt the release, nor the translations), this should only take a few days - to close this Pull Request and open a new one based on version 5.12. And of course take up your current proposals

Labels are not my strong point but one could (suggestion which can be changed) choose
"Review of labels, tooltips and order of methods and tools in Rawtherapee"

Of course, these changes should not affect compatibility (with 5.12...), and of course, this could have an impact on the documentation. So, proceed with caution.
To be more effective, work in advance by 2 or 3 people could precede a more collective exchange.

Thanks again

Jacques

PR #7385, opened 2025-05-07
SE sharpening - avoid amount = 0 and show slider in all cases - hide slider gamma in inverse

Desmis:
I think I have solved this issue... very, very old

In all cases if amount < 1 no action with sharpening (but wrong display), in this sense, the GUI slider will start at 1

I change  the GUI and now "amount" is shown in Basic, standard and Advanced mode 

For slider gamma in inverse
This is also an old bug... which must have existed since the creation of Local Adjustments (10 years ago already...) even if it only appeared recently.
After examination, the gamma function malfunctions in inverse mode. Since it is only present in "advanced" mode, I removed it from the "inverse" mode.
Of course, it remains in "normal" mode... You can always simulate the "inverse" mode by "Excluding spot," which appeared later in Local Adjustments.


Lawrence37:
This is ok for 5.12, but I would like to see the bugs be fixed in the future so that a zero amount and gamma in inverse mode are usable.

Lawrence37 (approved):

Desmis:
@Lawrence37 

In all case 0 cannot work....
But I will look in the future for inverse (if possible)

I merged into release-5.12, I'll let you finalize with release

Jacques

PR #7405, opened 2025-05-25
Adds new functionality to framing: annotation.

stolk:
This adds an annotation text that can be printed into the border. ASCII only, with a single fixed-width font.

To use it, Framing+Resizing needs to be enabled, and Framing needs to be enabled.

Frame size must be at least 30 pixels.

[This is what it looks like.](https://github.com/RawTherapee/RawTherapee/discussions/7404)


Benitoite:
The History entry for the annotation change is titled `HISTORY_MSG_FRAMING_ANNOTATION`, but should be in a readable format like the others for instance "Framing Annotation Message".
IMG_2952

Benitoite:
</img>
Resizing (Changed Size from .10 to .07) the frame makes the annotation oblique.
</img>

stolk:
Thanks for finding that.

Yeah, I can't really explain why that happens.

My best guess is that scan-lines are padded?

So if the framed image is 5000 pixels wide, it is rounded to nearest 32 multiple, 5024 perhaps?

digitalcarp:
I like the feature :)

Have u considered drawing fonts using a cairo surface and pango layout? Cairo is the 2d drawing lib and Pango does the font rendering. I think that would let you draw any system font.

stolk:
Thank you, fixed in new commit.

stolk:
@Benitoite Both the oblique-bug and the history message have been fixed.

stolk:
> Have u considered drawing fonts using a cairo surface and pango layout? Cairo is the 2d drawing lib and Pango does the font rendering. I think that would let you draw any system font.

I went for simplicity over bells n whistles. A bare-bones annotation, not unlike film manufacturers do, like Ilford does here:

image

I spaced out the letters horizontally a bit more, so it looks less ugly.

digitalcarp:
FYI framing gui code got moved see #7340. You should be able to cleanly merge the changes. If not a rebase will definitely be clean.

Benitoite:
We could have annotations with user-selectable fonts. 

Screenshot 2026-09-10 at 00 29 31-2
Screenshot 2026-09-10 at 00 37

Annotation Sans [digitized Liberation Sans]: (proportional w/kerning)
IMG_3357-8
Film Plotter: (original annotation font)
IMG_3357-7
Cochin Bold: (user-selected font)
IMG_3357-6

PR #7409, opened 2025-05-27
Adding some code documentation to IdleRegister helper class

kaesa:
Little bit of code documentation for the IdleRegister helper class. To speed up code reading.


kaesa:
I took the note " It's best to call `IdleRegister::destroy()` in the destructor of the class owning the `IdleRegister` instance. Otherwise make sure, it is the last member which will be deleted first." directly from the authors commit message 824ecae. Is that ok, or should it be worded otherwise. I didn't want to make a mistake of misinterpretation.

Lawrence37 (approved):

PR #7419, opened 2025-06-02
Review of Labels, Tooltips and Order of methods and tools in Rawtherapee

Desmis:
@waynesutton50 @Lawrence37

As I suggested, as soon as 5.12 is published in 'dev' I propose to open a discussion on Labels, tooltips, order of methods, and the location of tools in Rawtherapee. We could choose as a title, or any other proposal :
‘Review of labels, tooltips and order of methods and tools in Rawtherapee’ – of course we can change... and I close the #7376

There are undeniably inconsistencies (not just in Selective Editing CAM16), whether it's the label names, the associated tooltips, or the default methods), and I see no problem (quite the contrary) in reviewing them. Another example in 'Exposure tab' is Tone Mapping, which, although it is indeed a tool of this type, is confusing (ART made another choice).

Along the same lines, and this is a conversation we've had before, but it would be nice to share more broadly, what about the order of 'Tabs' and tools in Rawtherapee? From what I've heard, it's due to the 'usual' order of using the tools...But that was in 2006. For example, one of the major tools is Capture Sharpening, which is completely isolated, relegated to the back of the pack…

Everyone has their own logic for assigning names and classifying them (I'm not sure I'm an authority). But some choices that date back to 2006 should, I think, at least be examined, even if nothing is changed. 

Compared to 2006, new methods have emerged, such as tone-mappers, wavelets, and white balance. 

Limitations: don't question the pipeline order to avoid compatibility issues, nor do you make any major changes to the algorithm code…

This new Pull Request has only one change to the one proposed by Wayne
"Sigmoid Q & Slope based", replace by "Slope based & Sigmoid Q".

Before opening this debate on Pixls.us, I would like to have your opinion in principle.

Thank you

Jacques


Lawrence37:
Are you thinking of reviewing all the strings, or do you have a list?

I'm open to rearranging the tools. We did have a discussion on which tools to remove in 6.0. If we agree that some tools are no longer that useful, we can move them to the bottom of their respective tabs. You make a good point about capture sharpening. One problem is that it only works on raw images, so it must stay in the raw tab, and the raw tab is full of stuff that normally should be left alone. That's why the tab is placed near the end. If I'm not mistaken, capture sharpening can in theory work on all linear images. We could enable it for all images, just like what was done to the film negative tool.
I also have this idea to add some tools to the favorites tab by default for new users. If we place tools like exposure, white balance, sharpening, etc. there, that will give new users a good place to start learning RawTherapee.

Desmis:
@Lawrence37    @waynesutton50
No, I don't have any list, it's more out of principle

There's what Wayne spotted, but the same thing in SE with Local contrast & Wavelet, and now with GHS... Is it annoying, I don't know.
There's the term 'Tone mapping'...
There's Capture-Sharpening, of course, but with the version of the 'captur-noise' branch in Selective Editing, you can use all files.

As you point out, there are old algorithms (obsoletes) , where should we put them (at the end of the list, group them together, etc.).

We also need to consider the 'new' processes introduced by new tools (Tone Mapper, Selective Editing, etc.).
The current configuration is linked to history and habits. Changing habits will be difficult, but it's often one of the few solutions for introducing a tool.

For me the interest is not to mess everything up, but to think about the Labels, the order of the methods in each method and more generally the order and place of the tools, review the tooltips, and of course the obsolete algorithms.

Thank you for your principled response.

The other fundamental question is "do we resolve this among three or four people?" or do we open the debate on Pixls.us (we risk being overwhelmed by the proposals).

Jacques

kaesa:
What if the obsoleted algorithms were hidden by default with the possibility to enable them with a single option in the settings? I don't think moving obsolete algorithms to the bottom of the tabs makes much difference to the casual users.

Desmis:
@kaesa 

Thank you for this intervention... 

I know there's a 'favorite' tab in Preferences, but it's a little-known feature and doesn't really simplify the interface.

Can't we 'take advantage' of this PR to allow (in Preferences / options) to select - as in Selective Editing - a level of "complexity". Maybe not in the SE sense, but to offer several version interfaces. A simplified one, a bit like ART with the most relevant algorithms (which ones?) and another more complete one with all the algorithms including the obsolete ones (which ones?). If we do something like this, we will have to be very educational, otherwise the user 'accustomed' to a command will see it disappear as soon as we have implemented this functionality.

From a personal point of view, I am satisfied with your intervention, which clearly shows that something must be done.

But I remind you, we must not - except in exceptional cases - question the pipeline, or the basic code of the algorithms. This should only be an appearance, a possible simplification / optimization.

Lawrence37:
Keep in mind that hiding tools isn't as simple as just hiding them. If one of the hidden tools is active, either it must be revealed, or it must be disabled. In Selective Editing, for example, hidden adjustments are disabled/defaulted. I'm sure Jacques knows the extra complexity it involves.

Desmis:
I completely agree with @Lawrence37 , but just because it's a complex subject doesn't mean it shouldn't be done if overall there's an interest in it.

To clear/make visible a tool, a module is not enough (currently it is managed in Toolpanelcoords). It is also necessary to take into account the settings that may have been made previously in the pp3 with the previous versions.

At a minimum, if the tool is "hidden" and the PP3s do not use it, the tool must be deactivated or neutralized, without disrupting the pipeline.
If the PP3 for this tool is enabled, you must consider this tool as a choice (not an option, but 'inline', a bit like in Wavelet Levels...) to validate this inclusion.
And probably other problems to solve.

This doesn't mean it's impossible, but it's 'less simple' than it seems and requires adaptations of the GUI or Engine code

Desmis:
I made various changes:
* Move "Capture Sharpening" (Raw tab) to the top of the list. As a reminder, there is a version that works in non-raw mode in the Pull Request for captur-noise in Selective Editing.
* Replace in Selective Editing "Local Contrast & Wavelets" to "Wavelets & Local Contrast" to be consistent with the default method.
* Added a choice in Preferences > General > Define general Complexity with 2 choices :"Essential tools" and "All tools"

To test this module I arbitrarily chose the "Edge" module in Detail Tab. 
* If the user chooses "All Tools," "Edge" is still visible.
* If the user chooses "Essential Tools" and "Edge" was previously disable - so it was disabled in PP3 - "Edge" is no longer visible.
* If the user chooses "Essential Tools" and "Edge" was previously enabled - so it was enabled in PP3 - "Edge" is accessible and take into account.

Of course, I can add other tools to hide in "Essential" mode

Desmis:
I just added several tools to the list of tools to hide in "Essential tools" mode (arbitrary!).

These are little-used tools or obsolete functions or replaced by more efficient functions... (arbitrary).

If you used one of these tools (pp3) in a previous version of Rawtherapee, they will be present in the "Essential tools".

The tools concerned:
* Shadows/Highlighs (Exposure Tab) - Can be replaced by Tone Equalizer.
* Dynamic Range Compression (Exposure Tab) - Can be replaced by the Tone-mappers present in Selective Editing : Log Encoding, Sigmoid, Generalized Hyperbolic Stretch, TRC, etc..
*  Edges (Detail Tab).
* Microcontrast (Detail Tab).
* Contrast by Detail Levels (Detail Tab) - Can be replaced by Abstract profile - Contrast enhancement.
* Retinex (Advanced Tab).

PR #7423, opened 2025-06-05
Fix various build issues. Drop -l (login) from shell calls except on WIN32.

mandree:
Let me revise the initial comment.

This originally did set out to just remove the -l from $(SHELL) invocations, but it turns out we've had to fix a lot of issues along the way. I believe I fixed the MacOs, AppImage and Windows Github workflows, and made builds on Linux with Unix Makefiles generator, and FreeBSD, more robust. (I tested on x86_64 machines running Fedora Linux 42 and FreeBSD 14.3.)


mandree:
Some more context on this, I've been packaging graphics/rawtherapee on FreeBSD for a long time now. This PR strives to improve portability of the RT source code; I have been sitting on this patch for too long (almost a year) and had forgotten I'd made it.

This stems from FreeBSD 14.2 amd64, where the default build shell chokes on -l:

```text
$ /bin/sh -l  -c 'echo hello'
Illegal option -l
```

The recent v5.12 release has prompted me to look through what I have, and I've found it again and thought to submit it.

Lawrence37:
Converting this to a draft because Windows builds need to be taken care of.

mandree:
> Converting this to a draft because Windows builds need to be taken care of.

Is this something that has to with the content of this change, or is it that the Windows CI itself is broken so this change just isn't testable?

Where else do we start to get this moving?

mandree:
Let's see if adding libtool fixes the Windows/MSYS2 build. 

If that's insufficient, we can also add this patch to rtengine/libraw/configure.ac to fix autotools warnings:
[0001-libraw-update-libtool-macros-in-configure.ac.patch.txt](https://github.com/user-attachments/files/20834654/0001-libraw-update-libtool-macros-in-configure.ac.patch.txt)

mandree:
Euh, that's a tough one with the MSYS2 Windows build.  It complains about AC_SUBST missing now, but that might be some other macro actually be the culprit, maybe AX_OPENMP -- and I've wasted a few hours trying this and that.

mandree:
Whoah, I can only conclude that MSYS2/MINGW64 autoconf 2.71 and 2.72 or the MinGW64 m4/autom4te contents are messed up and either the latter are in a SNAFU or autoconf can't find its generic.m4 macro.  The symptom is that autoconf-2.72 complains that AC_SUBST and AC_MSG_WARN (which is an alias to some `AS_*` macro) are potentially undefined, aborts, and forgets to copy auxiliary scripts such as `compile` to `.`

I've wasted a whole day on this. 

The only reliable recourse seems to be to configure libraw on a unixoid computer (MacOS/Linux/FreeBSD) with `libtoolize --copy --install --force && autoconf --install --force` and check that stuff in.  I did so Fedora 42 x86_64, and pushed to my fork, and the Windows workflow gets libraw configured properly.  I know it's not ideal and I am not at all familiar with MSYS2.  For the WIN32 build, we also need to add AM_MAINTAINER_MODE for libraw's configure.ac and configure with --disable-maintainer-mode to avoid autoconf from being run.

I know some Cygwin and I do cmake/MS Visual Studio based development (mostly C++14/17) at work, but MSYS2 isn't part of the equation.  Is there someone who can look into the MSYS autoconf issues and file a report with the MSYS2 folks?

mandree:
That has been quite an odyssee and not exactly efficient.  MSYS2/MINGW64 can't run autoconf properly (or wouldn't find some default macro definitions that should come from generic.m4) -- maybe that's due to MSYS2 path issues or maybe packages are broken.  We still need help on that, see this comment above https://github.com/RawTherapee/RawTherapee/pull/7423#issuecomment-2992616412

Having said that, https://github.com/RawTherapee/RawTherapee/commit/93362739fdd36294f1d31c2f40828ef9792179a4 passes the AppImage, macOS, Windows builds via Github work flow, and pass local builds on Fedora Linux 42 x86_64 and FreeBSD 14.3 x86_64 for me.  Please review and merge.

Lawrence37:
Whoa, that's a lot. I would have stopped you from spending so much time on this if I wasn't occupied with other things. The problem with MSYS2 is that it sets up a bunch of things in the environment and all of that is gone if you invoke a new shell. The simple solution is to use the `-l` flag on MSYS2 to tell it to do all the setup. If you want a more "proper" solution, you'll need to capture the environment state and restore it for new shells. As for the current solution, 50,000+ new lines to make LibRaw build on MSYS2 makes it feel way over-engineered.

mandree:
> Whoa, that's a lot. I would have stopped you from spending so much time on this if I wasn't occupied with other things. The problem with MSYS2 is that it sets up a bunch of things in the environment and all of that is gone if you invoke a new shell. The simple solution is to use the `-l` flag on MSYS2 to tell it to do all the setup. If you want a more "proper" solution, you'll need to capture the environment state and restore it for new shells. As for the current solution, 50,000+ new lines to make LibRaw build on MSYS2 makes it feel way over-engineered.

Most of that is generated code as you can see if you read through the commits one by one. If the -l "fixes" MSYS2, then MSYS2 is really broken and needs a bug report and fixing.  -l isn't supported by POSIX so can only be used if you are absolutely sure it will work. As such, it has its host of Unix portability issues, and it imports the user environment which you explicitly do not want for normal scripts.

So what do we do instead?  Nuke the -l everywhere and add it back in for MSYS2 builds?

Lawrence37:
Generated or not, 50,000 lines is still a lot. Just because `-l` makes it work on MSYS2 doesn't mean MSYS2 is broken. `-l` is just one convenient way to set up the environment.

`-l` for MSYS2 will be necessary unless there is a reasonably simple alternative.

mandree:
Why would a system that can't run a plain script (such as the sub-processes of autoreconf) through a plain /bin/sh without bash's --login option _not_ be broken? But let's not quarrel about this assessment.

So the options seem to be
- fix MSYS2 (which would by my preferred alternative -- but I never interacted with those folks, not sure how accessible they are) such that /bin/sh scripting works without "login shell", or
- find a lighter weight compromise in Rawtherapee that improves the build for Unixoid non-bash shells such as FreeBSD's /bin/sh, without breaking MSYS2 builds.  

I might try to poke at the latter by cherry-picking some changes and splitting the "-l" addition to a separate part that's OS dependent, but no promise as to the timing of my next attempt.

mandree:
I think we're good now without adding a truckload of generated code. Github CI builds here -> https://github.com/mandree/RawTherapee/actions

The change from 0b5114e to 989ae22 is best viewed with "git diff -bw" because I only strip one `if(TRUE)` and one `endif()` line and reformat the block in between, which Github cannot show easily because it can't ignore whitespace diff.  

When reviewing, the "Files changed" view offers the option to hide whitespace changes, which is probably better to use.

With 0b5114e and again with 989ae22, all builds have passed and only Windows uses `-l` to work around MSYS2 shortcomings.  macOS/Linux (AppImage) would no longer get the -l for $(SHELL).

For good measure, let's try to minimize excess UpdateInfo.cmake induced rebuilds.  I haven't checked if the whole package is reproducible, yet.  But in a clean Git environment, UpdateInfo.cmake only depends on Git contents now, not on the host time or ID through UUIDGEN.

rtengine/libraw/configure.ac
Lawrence37, line 1: FYI this file is part of LibRaw and they are (appear to be) unaware of issues with libtool 2.

rtengine/LibRaw.cmake
```diff
@@ -67,47 +85,44 @@ endif()
 
 # Build flags.
 set(LIBRAW_MAKE_FLAGS "")
-if(CMAKE_GENERATOR MATCHES ".*Makefiles.*")
-    set(LIBRAW_MAKE_COMMAND "$(MAKE)")
-else()
```
Lawrence37, line 72: Please make it possible to use the jobserver. It is very inconvenient for developers otherwise.

mandree: I didn't take away parallel builds, only jobserver.  Meaning that libraw has its own job count defaulting to the CPU count.  So it will continue to be fast, assuming there is enough RAM.

mandree: I checked things a little bit, so the cause for the Linux build issues I'd observed in the past with Makefiles is a missing dependency between rtengine and "LibRaw", the target that gets defined to "ALL" and to depend on a non-existing phantom file.

That's one part.

Still we need to cater for incompatible "nmake" and other garbage in $(MAKE) and just use gmake or make in that case.  I've pused new material to my branch.

mandree:
OK, so the merge request has been reworked and rebased onto current dev.

Note that as with the previous state of a few weeks ago, libraw get built in parallel mode unless the user explicitly prevents that.  For compatible makes, we keep the jobserver mode, but we avoid nmake or other garbage for the libraw build.  Tested in the Github workflows and on FreeBSD 14.3 RELEASE amd64 with clang 19.1 for c++.

rtengine/LibRaw.cmake
Lawrence37: My point about the jobserver is that it allows better control over the parallelism. That not only includes increasing the parallelism for faster builds, but also reducing it to keep the rest of the system reasonably responsive.

mandree: We have that in what I pushed yesterday.  I have added a reformatting push today.

I am aware of what the jobserver does, but is that still a topic these days? 
In cmake contexts, I prefer Ninja over GNU make any time, and Ninja can't act as a jobserver itself as of 1.13, only as client to GNU make, or you need to `cmake -G 'Unix Makefiles'`, and with ninja, if you're under memory pressure relative to the number of CPU cores (say, 16 GB or less on a HTT/SMT enabled 8+ core CPU that does 16+ threads) and/or on an OS that responds badly to load/pressure, you'll probably want `ninja LibRaw && ninja` or similar to avoid ninja spawning another dozen of compilers in parallel with the libraw build.

Lawrence37: I'm on a laptop with 2/4 physical/logical cores. `-j3` keeps the system responsive while I work on other tasks, at least on Linux. I need to reduce the parallelism even more on Windows. Otherwise it becomes sluggish.

rtengine/LibRaw-make.sh
```diff
@@ -0,0 +1,25 @@
+# this deliberately has no shebang (#!) line.
+# See LibRaw.cmake's add_custom_command(), which should
+# execute it with a POSIX-compliant shell.
+# It expects exactly two arguments, which will be mapped to
+# MAKE and LOGICAL_PROCESSORS
+
+if [ 2 -gt $# ] ; then
```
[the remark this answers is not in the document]
mandree: It's safer to put constants first in comparisons. While `$#` is safe, other `$...` are not.

rtengine/LibRaw-make.sh
```diff
@@ -0,0 +1,25 @@
...
+
+_gmake="$(set +e; type >/dev/null gmake && echo gmake)"  # gmake if on PATH, else empty
+case "${_make##*/}" in
+    gmake|make) printf '%s\n' >&2 "-- ${0##*/} is using inherited $_make"
+        ;;
+    *)  _make="${_gmake:-make} -j${LOGICAL_PROCESSORS}"
+        printf '%s\n' >&2 "-- ${0##*/} is using generated $_make"
+        ;;
+esac
+exec ${_make}  "$@" # $_make unquoted, need to split away the -jN arg
```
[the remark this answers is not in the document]
mandree: The proposal is garbage, it does not prevent _make_j from environment to break us. Solved by writing proper code.

UpdateInfo.cmake
```diff
@@ -68,10 +86,25 @@ if(REL_INFO_FILE STREQUAL REL_INFO_FILE-NOTFOUND)
...
+
+    if (SOURCE_DATE_EPOCH)
+        # reproducible build path
+        execute_process(COMMAND uname -ms OUTPUT_VARIABLE BUILDINFO_OS OUTPUT_STRIP_TRAILING_WHITESPACE)
+        execute_process(COMMAND "${SHELL}" -c "LC_ALL=C ; LANG= ; export LC_ALL LANG ; DATE_FMT=\"+%a, %d %b %Y %T %z\"; date -u -d \"@${SOURCE_DATE_EPOCH}\" \"\$DATE_FMT\" 2>/dev/null || date -u -r \"${SOURCE_DATE_EPOCH}\" \"\$DATE_FMT\"" OUTPUT_VARIABLE BUILDINFO_DATE OUTPUT_STRIP_TRAILING_WHITESPACE)
+        set(BUILDINFO_EPOCH "${SOURCE_DATE_EPOCH}")
+        set(BUILDINFO_UUID "git-${GIT_COMMIT_FULL}")  # use GIT_COMMIT instead of a UUID
+    else()
+        execute_process(COMMAND uname -mrs OUTPUT_VARIABLE BUILDINFO_OS OUTPUT_STRIP_TRAILING_WHITESPACE)
+        execute_process(COMMAND "${SHELL}" -c "LC_ALL=C ; LANG= ; export LC_ALL LANG ; date -u \"+%a, %d %b %Y %T %z\"" OUTPUT_VARIABLE BUILDINFO_DATE OUTPUT_STRIP_TRAILING_WHITESPACE)
```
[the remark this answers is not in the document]
mandree: OK, rewritten

UpdateInfo.cmake
```diff
@@ -68,10 +86,25 @@ if(REL_INFO_FILE STREQUAL REL_INFO_FILE-NOTFOUND)
...
+    if (DEFINED ENV{SOURCE_DATE_EPOCH} AND NOT ENV{SOURCE_DATE_EPOCH} STREQUAL "")
+        # override from environment
+        set(SOURCE_DATE_EPOCH "$ENV{SOURCE_DATE_EPOCH}")
+        message(STATUS "SOURCE_DATE_EPOCH overriden from environment to ${SOURCE_DATE_EPOCH}")
+    endif()
+
+    if (SOURCE_DATE_EPOCH)
+        # reproducible build path
+        execute_process(COMMAND uname -ms OUTPUT_VARIABLE BUILDINFO_OS OUTPUT_STRIP_TRAILING_WHITESPACE)
+        execute_process(COMMAND "${SHELL}" -c "LC_ALL=C ; LANG= ; export LC_ALL LANG ; DATE_FMT=\"+%a, %d %b %Y %T %z\"; date -u -d \"@${SOURCE_DATE_EPOCH}\" \"\$DATE_FMT\" 2>/dev/null || date -u -r \"${SOURCE_DATE_EPOCH}\" \"\$DATE_FMT\"" OUTPUT_VARIABLE BUILDINFO_DATE OUTPUT_STRIP_TRAILING_WHITESPACE)
```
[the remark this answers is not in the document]
mandree: This has been tested. Rewriting and extracting to a script is prone to bugs, and this does not need maintenance.

mandree:
Bump - this is now almost three months old and I would like to avoid this getting stale when the main^W dev branch advances.  Please let me know what's acceptable and what isn't.

rtengine/LibRaw-make.sh
```diff
@@ -0,0 +1,34 @@
...
+case "${_make##*/}" in
+    gmake|make) printf '%s\n' >&2 "-- ${0##*/} is using inherited $_make"
+        _make_arg=
+        ;;
+    *)  _make="${_gmake:-make}"
+	case "$MAKEFLAGS" in
+            *--jobserver-auth*) ;;
+            *)                  _make_arg="-j${LOGICAL_PROCESSORS}" ;;
+        esac
+        printf '%s\n' >&2 "-- ${0##*/} is using generated $_make $_make_arg"
```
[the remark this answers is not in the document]
mandree: Resolved in later push to my branch.

mandree:
@Lawrence37 I have also rebased to minimize diffs to upstream dev branch.

Lawrence37 (approved):

Lawrence37:
FYI rebasing can make re-reviewing difficult because GitHub has no way of showing new changes while excluding those from the base branch. Luckily, there was no overlap between your changes and those in `dev`.

mandree:
> FYI rebasing can make re-reviewing difficult because GitHub has no way of showing new changes while excluding those from the base branch. Luckily, there was no overlap between your changes and those in `dev`.

Uh, sorry. I am so used to Gitlab (use it at daily work), where reloading the merge request page will suffice to it gets the merge base sorted aside.  

I'll try to remember not to rebase on Github pull requests unless I have to; I got myself confused when fixing the 2nd robot review finding because I needed to do that locally.

Thanks for merging the changes.

PR #7429, opened 2025-06-09
Wavelet levels - crash with Highlight threshold in residual image

### LINKED ISSUE #7427: Crashes when apply a .pp3 and change residule image Highlight Threshold to 0
**Short description**
RawTherapee crashes if loading [this .pp3 sidecar](https://github.com/user-attachments/files/20639447/IMG.pp3.txt) for a raw and changing Wavelet -> Residual Image -> Highlight Threshold to 0.

**Steps to reproduce**
1. Download the provided .txt, rename it to .pp3, apply it to a raw file. 
2. Enable Wavelet -> Residual Image, change Highlight Threshold to 0.
3. RawTherapee should crash.

**Expected behavior**
Rawtherapee would not crash.

**Additional information**
Rawtherapee 5.12 release version. 
I thought there was some conflict with "Sharp-mask and Clarity" and "Residual Image" modules since disabling one another (or set "Sharp-mask and Clarity" method to sharp mask) would avoid the crash, however, I tried to recreate this with another raw by manually enabling modules and adjusting values to the same but failed to replicate the crash (but pasting .pp3 and change threshold still crashes), so there must be another cause hidden in my .pp3 file (probably not GHS since disabling it has no effect, and other enabled tools don't have their parameters changed, so not sure what did I miss).

Desmis:


Desmis:
#7427

Lawrence37 (approved):

PR #7430, opened 2025-06-10
macOS: update app icon

Benitoite:
</img>
Update to macOS app icon; designed in latest [Apple Icon Composer](https://developer.apple.com/icon-composer/) with 45° specular highlight glass look on an unobtrusive default dark chicklet.


Lawrence37:
The outer ring doesn't stand out as much and the proportions are different. Here's a comparison with the border between the the inner and outer rings scaled to match.
logo

Benitoite:
I didn't notice Icon Composer ate the highlights for some reason! (I'm using the svg logo from data/images as the source.) What I had to do was separate the highlight ring in a SVG editor to import into its own layer in Icon Composer. 
</img>

Lawrence37:
That's interesting. I checked all the SVGs and they all have the same proportions, but different from the Icon Composer output. Does it have a problem with SVGs?

Benitoite:
My guess is Icon Composer seems to want to do layering itself, so the glass effect can be applied.  Layering in icons is also part of the system glass look.  Both layers get 888% scaled.  The highlight ring is in Screen mode over the fruit loop.

Benitoite:
</img>
Here's how it looks with no glass effect on the rings.

Lawrence37:
If both layers are scaled the same amount, there shouldn't be any proportion issues. The hole is too large in comparison with the rings. If the outer ring's layer had a smaller scale, that would have explained the discrepancy.

Benitoite:
I just noticed the `.ico` when overlaid has different dimensions but couldn't find an svg that corresponds.
IMG_2975

Lawrence37:
I haven't tried the .ico, but all the SVGs are the same compared to each other (and different from the logo on the new icon) when viewed on my computer.

Benitoite:
FYI the present mac icon vs WIP:
</img><-- present mac icon

</img></img></img></img>

Benitoite:
Is there a source document for the logo icon?

Lawrence37:
I don't think there is, but Roel recreated the logo in #6446 and now it lives at [rtdata/icons/hicolor/scalable/apps/rawtherapee.svg](https://github.com/RawTherapee/RawTherapee/blob/dev/rtdata/icons/hicolor/scalable/apps/rawtherapee.svg)

Benitoite:
Thanks for pointing that out to me.  I could only find an outdated link previously. Apple IconComposer rendered the svg's highlight ring in a distorted fashion, so what Ive done is export the svg as a giant png with Inkscape, and load that into IconComposer.



</img>--> present mac icon VS </img>--> currently proposed
___
Previous work:
</img></img></img></img>

Benitoite:
Another way to obtain a similar result is to Edit>Select All, then Object>Transform>Scale up the original svg in Inkscape to 650%x650%, and select File>Document Properties>Display>Resize to content before resaving. IconComposer then interprets the file as much larger than what is originally designed into the file.

</img>

Benitoite:
Here's a look at a *Tahoe* mockup Application folder with the new icon:

Screenshot 2025-07-23 at 07 24 17

Lawrence37 (approved):
Nice! It's an exact match now.

PR #7435, opened 2025-06-15
Lean windows.h and fix compiler warnings

digitalcarp:
* Replace all `windows.h` with `rtengine/leanwindows.h`
* Unify windows header version
* Resolve include `winsock2.h` before `windows.h` warning
* Remove unneeded `#undef`  with leaner `windows.h`
* Fix `-Wstringop-truncate`


rtengine/improccoordinator.cc
```diff
@@ -3208,9 +3208,6 @@ void ImProcCoordinator::stopProcessing()
 
 void ImProcCoordinator::startProcessing()
 {
-
-#undef THREAD_PRIORITY_NORMAL
```
Lawrence37, line 3212: Do you know where `THREAD_PRIORITY_NORMAL` comes from? Why is it removed here but not in a couple other places?

digitalcarp: It comes from windows threading headers and glib thread headers. It gets undef previously because the windows macro ends up messing with using the glib enum

rtengine/rawimage.cc
```diff
@@ -538,21 +538,23 @@ int RawImage::loadRaw(bool loadData, unsigned int imageNum, bool closeFile, Prog
...
-        strncpy(model, d.normalized_model, sizeof(model)-1);
-        model[sizeof(model)-1] = 0;
         RT_software = d.software;
         dng_version = d.dng_version;
         filters = d.filters;
         is_foveon = d.is_foveon;
         colors = d.colors;
         tiff_bps = 0;
 
+        auto rt_strncpy = [](char* dest, char* src, size_t n) {
```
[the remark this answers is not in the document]
digitalcarp: The passed in arrays are struct members. It is safe as they cannot be null

rtengine/improccoordinator.cc
digitalcarp: I see your point. I removed the undefs that I missed previously.

rtengine/rawimage.cc
digitalcarp, line 1: Merged with dev, this was a merge conflict and should get a closer review.

rtengine/rawimage.cc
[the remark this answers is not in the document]
digitalcarp: Added checks to the lambda

Lawrence37 (approved):

Issue #7443, opened 2025-06-27
Segfault since gtkmm stack upgrade


casta:
**Short description**
Since gtkmm stack upgrade, rawtherapee segfault at startup. Not sure about the package that triggered the upgrade. But please see the stack trace below: I thing that there is an issue with default parameters. Patch proposal will follow.

**Steps to reproduce**
1. Update to gtkmm-3.24.10/pangomm-2.46.4/cairomm-1.14.5/atkmm-2.28.4/glibmm-2.66.8/libsigc++-2.12.1
2. Start rawtherapee, it’s segfault

**Expected behavior**
No segfault.

**Additional information**
Version: 5.12-8-g9f9b6ede8
Branch: dev
Commit: 9f9b6ede8
Commit date: 2025-06-15
Compiler: x86_64-pc-linux-gnu-gcc 15.1.0
Processor: 12th\ Gen\ Intel(R)\ Core(TM)\ i5-12500
System: Linux
Bit depth: 64 bits
Gtkmm: 3.24.10
Lensfun: 0.3.4.0
libjxl: Disabled
Build type: RelWithDebInfo
Build flags: -march=native -O3 -mtune=native -pipe -std=c++11 -ffp-contract=off  -Werror=unused-label -Werror=delete-incomplete -fno-math-errno -Wno-attributes -flto -Wall -Wuninitialized -Wcast-qual -Wno-deprecated-declarations -Wno-unused-result -Wunused-macros -fopenmp -Werror=unknown-pragmas  -ftree-vectorize
Link flags: -Wl,-O1 -Wl,--hash-style=gnu -Wl,--as-needed -Wl,--sort-common  -flto
OpenMP support: yes
MMAP support: ON
Build OS: Linux 6.15.4-gentoo x86_64
Build date: Fri, 27 Jun 2025 19:03:13 +0000 UTC
Build epoch: 1751050993
Build UUID: 3284c1d5-6afc-46bb-ad79-90bbf656ccbd

Other useful information:
```

Program received signal SIGSEGV, Segmentation fault.
0x00007ffff6fd6c20 in Glib::ustring::ustring(Glib::ustring const&) () from /usr/lib64/libglibmm-2.4.so.1
(gdb) bt
#0  0x00007ffff6fd6c20 in Glib::ustring::ustring(Glib::ustring const&) () at /usr/lib64/libglibmm-2.4.so.1
[...]

PR #7446, opened 2025-06-29
Fix read before initialize in options init

digitalcarp:
Fast export opts are reading from an empty string since `rtSettings.srgb` is set later in the function. Reorder code to fix dependency issue.

Not sure if this helps #7443


digitalcarp:
Have a better fix in a separate branch.

Issue #5314, opened 2019-05-06
Change format of translation files


glixx:
To be easy translatable I suggest to change format of translation files on:
```key=value```
Not too many changes in code, but using Mozilla PROPERTIES file (.properties) is better for translators.
Transifex supports Mozilla PROPERTIES file (.properties).

So just to change the first ```;``` in all strings on ```=```.

glixx:
``` sed 's/;/=/1' default > rawtherapee_en.properties```
```sed 's/;/=/1' Russian > rawtherapee_ru.properties```
etc.

Thanatomanic:
> Mozilla PROPERTIES file (.properties) is better for translators

Can you elaborate on that? You are suggesting only a minor change, but why we should change this at all? Only to be able to use Transifex?

glixx:
To use any tools, appications, transifex that support properties files.

Beep6581:
@glixx please link to a specification of "Mozilla PROPERTIES" and to a list of software which supports it.

glixx:
This is almost Java properties utf8 format.

AleksiKl:
Could the [Poeditor](https://poeditor.com) be used in this project?

mkyral:
> Could the [Poeditor](https://poeditor.com) be used in this project?

Not directly. I'm using custom scripts to convert RT language file to .po and back. 
https://github.com/mkyral/rawtherapee_l10n

mkyral:
I don't use that page, but maybe you can use simple conversion to csv.

vomikan:
I have set up synchronization of my project with [crowdin.com](https://crowdin.com) .
Now all transfers are made by a team of volunteers online. Synchronization with the repository is automatic. It's convenient.

https://crowdin.com/project/moneymanagerex

Issue #2209, opened 2015-08-11
Interface language files with modified text support


Beep6581:
Originally reported on Google Code with ID 2225

```
I planned to implement a way to have language files which would support letting the
translator know when an already translated text is outdated because the original English
one in "default" changed. I wanted to do this in Bash, but of course that limits us
to Bash users. Better to do it in something cross-platform like Python, but I don't
know that yet.

It would work like this:
All language files would look like this:
KEY:checksum;Human text
After adding text or making changes to "default", the script would be run, which will
update the checksum for each "Human text".
The script would also copy these checksums into the translation files.
e.g. the checksum stored for each key in Espanol would correspond to the one in "default"
at the time of translation.
Now if I want to translate Espanol a month from now, I run the script again, and the
script generates new checksums for each key in "default" and matches them against those
already stored in Espanol. If they match, we know the "default" text did not change.
If they do not match, we know the "default" text changed, and even if there already
is a Spanish translation of that key, we know it's outdated and must be revised.
Even though the text changed, it's likely the old translation is still better than
none, so instead of prepending an exclamation mark to the changed line, the script
could figure out some other way of letting the user know of the change while at the
same time letting RawTherapee still use that line, e.g. instead of prepending "!" it
could prepend "?", or maybe just moving it into the front of the translation file between
some "# CHANGED KEYS FOLLOW" section.
```

Reported by `entertheyoni` on 2014-01-28 07:01:57

PR #7449, opened 2025-07-02
Add utilities for future gettext migration

digitalcarp:
## Motivation

- Translators have difficulty keep track of changes to key-based translation strings
- In-house localization method missing features like fallback chaining and plural forms

## Proposal

Move to a gettext-based i18n API.

- Standardized translation workflow with other open source projects
- Leverage existing tools for editing translation/PO files (e.g. Poedit)
- Use `LANGUAGE` env var to implement language fallback chains
- Better API for handling plural forms and context specific text
- Allow using pre-distributed (in Linux) ISO639-3 gettext MO files to translate locale to language names

## Changes

- Add Python script for simplify working with gettext with handling legacy in-house translations
  - Automates detecting new or changes translation reference text (mark as fuzzy)
- Add `i18n.h` but prevent including it until fully implemented in future PR

## Changes in Future PR

- Mark as many possible translatable strings in code with `i18n.h` macros as possible
- Setup locale and gettext domains
- Possibly map PO file back to in-house translation files (so translators have easier time, but dev is still the same)

### DIFF
--- a/rtengine/i18n.h
+++ b/rtengine/i18n.h
@@ -1,0 +1,108 @@
+/*
+ *  This file is part of RawTherapee.
+ *
+ *  Copyright (c) 2025 Daniel Gao <daniel.gao.work@gmail.com>
+ *
+ *  RawTherapee is free software: you can redistribute it and/or modify
+ *  it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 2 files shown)


digitalcarp:
For future reference, the following issues could be related/improved.
#2209 #4319 #5314

PR #7480, opened 2025-07-27
rawtherapee-cli: overhaul command-line parsing logic

apoelstra:
I spent a few hours this morning refactoring the command-line parsing logic. There are a few high-level issues with the current CLI interface:

* Any --long-flags are silently ignored.
* There are several options (`-o` and `-O`, `-s` and `-S`, `-j` and `-t` and `-n`) which are incompatible and clobber each other and can be specified multiple times; the current code silently takes the most recent one.
* If the user provides an input file without `-c` this is a silent error.
* If the user provides no input files this is a silent error.
* In general, most forms of malformed input result in silent errors.

I have preserved the CLI interface (with the one change that I allowed `-c` to appear anywhere in the list of arguments, not just the end, which I'm happy to revert if people feel strongly about it) in the "happy case" that it's invoked correctly so there should be no breaking changes. However, in the unhappy case I provided much more detailed error information as well as rejecting lots of things that are currently accepted (e.g. extra `--help` args, or multiple `-j` args, and so on). I also added a `-v` verbose flag.

A few samples of changes:

Before:
```
$ rawtherapee-cli
RawTherapee, version 5.12, command line.
Terminating without anything to do.

$ rawtherapee-cli --help
RawTherapee, version 5.12, command line.

$ rawtherapee-cli -h
RawTherapee, version 5.12, command line.
<full help text snipped>

$ rawtherapee-cli ~/media/photos/original/baby-karina.jpg
RawTherapee, version 5.12, command line.

$ rawtherapee-cli -c ~/media/photos/original/baby-karina.jpg
RawTherapee, version 5.12, command line.

"/home/apoelstra/media/photos/original/baby-karina.jpg"  doesn't exist!
$ rawtherapee-cli -c ~/media/photos/original/baby-karina.jpeg
RawTherapee, version 5.12, command line.
Output is 8-bit integer.
Processing: /home/apoelstra/media/photos/original/baby-karina.jpeg

$ rawtherapee-cli -c ~/media/photos/original/baby-karina.jpeg -O .
RawTherapee, version 5.12, command line.
"-O"  doesn't exist!
"./.git"  is a folder. Folder skipped
"./.clang-format"  is not one of the selected parsed extensions. Image skipped.
<snip several lines of it interpreting `-O` and `.` as input files>

$ rawtherapee-cli -j 50 -o . -c ~/media/photos/original/baby-karina.jpeg
RawTherapee, version 5.12-dev, command line.
```

After:
```
$ rawtherapee-cli
RawTherapee, version 5.12, command line.
No input files provided. Use -c to add input files or -h for help.
(Hint: if you intended to do nothing, provide -c by itself with no filenames.)

$ ./rawtherapee-cli --help
RawTherapee, version 5.12-dev, command line.
Usage:
  rawtherapee-cli -c <dir>|<files>   Convert files in batch with default parameters.
  rawtherapee-cli <other options> -c <dir>|<files>   Convert files in batch with your own settings.
Options:
  rawtherapee-cli[-o <output>|-O <output>] [-q] [-a] [-s|-S] [-p <one.pp3> [-p <two.pp3> ...] ] [-d] [ -j[1-100] -js<1-3> | -t[z] -b<8|16|16f|32> | -n -b<8|16> ] [-Y] [-f] -c <input>


Unrecognized flag --help.
(Hint: rawtherapee-cli has no --long-form arguments. Use -h or -? to see help.)

$ rawtherapee-cli ~/media/photos/original/baby-karina.jpg
<version, short help snipped>

Invalid CLI argument /home/apoelstra/media/photos/original/baby-karina.jpg.
(Hint: if you meant to provide an input file, use -c. See -h for full help.)

$ rawtherapee-cli -c ~/media/photos/original/baby-karina.jpeg -O .
<version, then processes the file same as if -O . had been provided earlier>

$ rawtherapee-cli -j 50 -o . -c ~/media/photos/original/baby-karina.jpeg
<version, short help snipped>

Invalid CLI argument 50.
(Hint: the -j flag does not allow a space. For example, use -j99 rather than -j 99.)
```


apoelstra:
While writing the PR description I noticed that if you pass `-O.`, which you might expect to be interpreted as `-O .`, the `.` is ignored and the argument simply does nothing. (Well, it flips the "copy params file" bit on and leaves the destination the same as whatever the most recent `-O` or `-o` invocation set it to.) I'll update the PR to address that at some point.

But first, what are the next steps here?

Presumably nobody wants to review a 24-commit unsolicited PR. I'm happy to split this out into multiple PRs, to refactor or revert changes, to rebase the whole thing on #7478 (which would simplify some stuff since I'd really like to have `std::variant` and `std::optional`..), etc. It's already roughly organized as a series of small refactors with the "real changes" coming in at the end. So it should be easy to pull apart or to remove/move individual parts.

If there is a total concept NACK I'll probably maintain this as a fork, but with a few exceptions, I think most of the changes here are clearly a benefit to the CLI interface.

digitalcarp:
I guess with the cmake bump, did u consider using a library like CLI11?

apoelstra:
I did not -- I am not very familiar with cmake or with any specific cli libraries.

I can say that if we wanted to preserve the current CLI interface more-or-less as-is, there wouldn't be much benefit to it. Right now every single flag is a single-character flag, we don't support accumulating them (like `-avj95`), we don't have long forms of flags, we don't have subcommands, etc. So the work of parsing is pretty small. (Although we still aren't doing it properly -- e.g. before and after this PR, `-awhatever` is parsed the same as `-a` :)).

Maybe we want to have those things! But that would be a big discussion and probably result in a bigger PR than this. (The bulk of this PR is basic code refactorings that would be necessary regardless.)

rtengine/procparams.cc
```diff
@@ -12143,8 +12151,12 @@ void PartialProfile::applyTo(ProcParams* destParams, bool fromLastSave) const
...
-AutoPartialProfile::AutoPartialProfile() :
-    PartialProfile(true)
+AutoPartialProfile::AutoPartialProfile(bool createInstance, bool paramsEditedValue) :
+    PartialProfile(createInstance, paramsEditedValue)
+{
+}
+
+AutoPartialProfile::AutoPartialProfile(AutoPartialProfile&& other)
 {
 }
```
Lawrence37, line 12161: Why is the move constructor empty?

rtengine/procparams.cc
```diff
@@ -12143,8 +12151,12 @@ void PartialProfile::applyTo(ProcParams* destParams, bool fromLastSave) const
     }
 }
 
-AutoPartialProfile::AutoPartialProfile() :
-    PartialProfile(true)
+AutoPartialProfile::AutoPartialProfile(bool createInstance, bool paramsEditedValue) :
+    PartialProfile(createInstance, paramsEditedValue)
+{
+}
```
Lawrence37, line 12157: `createInstance` is `false` by default, so `PartialProfile` will be constructed with `createInstance = false` instead of `createInstance = true` by default. Is this an intentional change? If so, why?

rtgui/main-cli.cc
```diff
@@ -56,32 +56,254 @@
...
+        case OutputType::JPG: return ".jpg";
+        case OutputType::PNG: return ".png";
+        case OutputType::TIF: return ".tif";
+    }
+    std::abort();
 }
 
+class OutputPath {
+public:
+    // TODO remove default constructor once we move parsing into constructor for CliArgs
```
Lawrence37, line 93: Is it still planned to have the parsing done in the CliArgs constructor?

rtgui/main-cli.cc
```diff
@@ -193,517 +412,449 @@ int main (int argc, char **argv)
...
-#endif
-        if ( currParam.length() > 1 && currParam.at(0) == '-' && currParam.at(1) == 'q' ) {
-            return true;
-        }
-    }
-
-    return false;
+static void longUsage (const std::string& cmd_name) {
+    Glib::ustring pparamsExt = paramFileExtension.substr (1);
+    std::cerr << "  An advanced, cross-platform program for developing raw photos." << std::endl;
```
Lawrence37, line 430: Is it standard practice to send help text to stderr? Testing a few commands like grep, git, and awk, they all use stdout unless the help text is due to invalid parameters.

rtgui/main-cli.cc
```diff
@@ -193,517 +412,449 @@ int main (int argc, char **argv)
...
 
-                        if (outputPath.substr (0, 9) == "/dev/null") {
-                            outputPath.assign ("/dev/null"); // removing any useless chars or filename
-                            outputDirectory = false;
-                            leaveUntouched = true;
-                        } else if (Glib::file_test (outputPath, Glib::FILE_TEST_IS_DIR)) {
-                            outputDirectory = true;
-                        }
-                    }
+        switch ( currParam.at (1) ) {
```
Lawrence37, line 533: Is `currParam` guaranteed to be at least 2 characters long?

rtgui/main-cli.cc
```diff
@@ -193,517 +412,449 @@ int main (int argc, char **argv)
...
+                    std::cerr << "(Hint: if you meant to provide a filename, prefix it with ./.)" << std::endl;
+                }
+                return -1;
+
+            default:
+                shortUsage (Glib::path_get_basename (argv[0]));
+                std::cerr << "Unrecognized flag " << currParam << "." << std::endl;
+                std::cerr << "(Hint: use -h or -? to see help.)";
+                if (last_ch == 'c' || last_ch == 'p') {
+                    std::cerr << "(Hint: if you meant to provide a filename, prefix it with ./.)" << std::endl;
```
Lawrence37, line 697: Might want to quote the prefix to make it more clear.
```suggestion
                    std::cerr << "(Hint: if you meant to provide a filename, prefix it with \"./\".)" << std::endl;
```

rtgui/main-cli.cc
```diff
@@ -193,517 +412,449 @@ int main (int argc, char **argv)
...
 
-                        if (Glib::file_test (argument, Glib::FILE_TEST_IS_REGULAR)) {
-                            bool notAll = allExtensions && !options.is_parse_extention (argument);
-                            bool notRetained = !allExtensions && !options.has_retained_extention (argument);
+            case '-':
+                shortUsage (Glib::path_get_basename (argv[0]));
+                std::cerr << "Unrecognized flag " << currParam << "." << std::endl;
+                std::cerr << "(Hint: rawtherapee-cli has no --long-form arguments. Use -h or -? to see help.)" << std::endl;
+                if (last_ch == 'c' || last_ch == 'p') {
+                    std::cerr << "(Hint: if you meant to provide a filename, prefix it with ./.)" << std::endl;
```
Lawrence37, line 688: Might want to quote the prefix to make it more clear.

rtgui/main-cli.cc
```diff
@@ -193,517 +412,449 @@ int main (int argc, char **argv)
...
-                                deleteProcParams (processingParams);
-                                return -3;
-                            }
-                        }
-                    }
+            case 'j':
+                if ( alreadySetJ ) {
+                    shortUsage (Glib::path_get_basename (argv[0]));
+                    std::cerr << "At most one of -j, -t or -n may appear, at most once." << std::endl;
+                    return -1;
```
Lawrence37, line 607: Reduce code duplication.

rtgui/main-cli.cc
```diff
@@ -193,517 +412,449 @@ int main (int argc, char **argv)
...
-    int compression = 92;
-    int subsampling = 3;
-    int bits = -1;
-    bool isFloat = false;
-    std::string outputType;
-    unsigned errors = 0;
+    bool alreadySetB = false;
+    bool alreadySetO = false;
+    bool alreadySetJ = false;
+    bool alreadySetS = false;
```
Lawrence37, line 502: I'm not convinced it's a good idea to avoid clobbering. Some programs (like gcc) leverage clobbering to make it easier to use in scripts. It allows default values to be set and optionally overridden.

Lawrence37 (commented):
Hi @apoelstra, thanks for this pull request. Sorry for taking so long to complete a review. I've added a few comments. The way you organized the commits was very helpful.

apoelstra:
I'm sorry, I don't think I'll be able to revisit this PR anytime in the next several months. These are great review comments but I've moved away from trying to do photo processing via CLI, and some of these imply fairly large refactors. (And I don't know C++ well enough to LLM-generate this, and as a fellow open-source maintainer I am certainly not going to submit slop that I haven't reviewed myself.) I've also transitioned from being on parental leave with a helpless baby to having both a job and a very mobile baby :)

I don't mind if somebody else takes over the branch and/or rewrites my commits (if somebody does do this, I'd kinda like a `Coauthored-by:`, but don't stress about it).

If I get back to this I'll rebase and reopen.

PR #7489, opened 2025-08-04
Automated builds for Windows ARM

### LINKED ISSUE #7484: ARM Windows build
As ARM processors become more common in today's computers, it is beneficial to test builds in ARM environments and provide official ARM-compatible builds. #7477 adds ARM builds for Linux. There is already a workflow for ARM on macOS. Windows has not been done yet.

Lawrence37:
This adds Windows ARM builds using the new `windows-11-arm` runner, which is currently in public beta. Build artifacts now have the architecture in the name. For example,
x86: `RawTherapee_dev_win64_x86_64_release.exe`
ARM: `RawTherapee_dev_win64_arm64_release.exe`

I do not have an ARM device with Windows, so I cannot test.

Closes #7484.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -335,3 +335,58 @@ jobs:
           tag_name: pre-dev-github-actions
           repository: RawTherapee/RawTherapee
           files: build/${{env.PUBLISH_NAME}}.exe
+
+  test-launch:
+    needs: build
+    runs-on: ${{ matrix.arch == 'arm64' && 'windows-11-arm' || 'windows-latest' }}
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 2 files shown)


Lawrence37:
I've split the publishing steps into a separate job and added a test launch job that checks if RawTherapee runs.

ifq:
Hi, where can I get the win arm64 build, thanks.

Lawrence37:
@digitalcarp FYI I upgraded the fmt library because I was getting a compile error which has since been fixed in v12.0.0 (see the linked PR in the commit message). There's probably nothing to be concerned about, but I'm letting you know just in case. It is a major version upgrade after all.

Lawrence37:
@ifq Click on the "Checks" tab, then click on the Windows entry. It should take you to another page. The downloads are in the Artifacts section at the bottom of the page.

PR #7451, opened 2025-07-04
Fix static initialization order fiasco crashes

digitalcarp:
Moves several extern variables behind a `App::get()` singleton. The singleton is implemented as a Meyer's singleton for thread safety and defined initialization order.

The most "noisy" change is changing `options` extern global to `App::get().options()`. For places where `options` is changed, `mut_options()` is called instead.

Fixes #7443

Edit: Also fixes #7052 hopefully

PR #7491, opened 2025-08-04
Fix static initialization order fiasco crashes for release-5.12 branch

digitalcarp:
Backports #7451 to release-5.12 branch. The changes are squashed into a single commit and a patch file was generated by running `git format-patch -1`.


Benitoite:
To me it looks like this would somehow introduce a regression affecting local and (current) automated arm64 macos builds fixed two weeks ago in https://github.com/RawTherapee/RawTherapee/commit/89534913f849a0d423b0ef28aeb9e4d4f3947a0f
Why?  One of the macos libraries has a baked-in relative install_name path that references a non-flat Frameworks directory.  Changes were made against the file `tools/osx/macosx_bundle.sh` to normalize the builds.
What?  It affects the generation of the gtk gdk pixbufloader database on systems with default system integrity protection, specifically in relation to the rsvg loader.

digitalcarp:
This change is to the release-5.12 branch. I don't know if that's allowed. Your fix is in dev and I assume not in the release branch, so it wasn't captured for this automated build.

Benitoite:
Indeed since the release there were updates to the Homebrew packages causing the need to fix the building since then.  The macos CI always gets the latest available things from Homebrew, so causing a rebuild on a stale branch will inevitably show more and more errors as time goes on.

digitalcarp:
Addressed comments from #7451.

digitalcarp:
I'll leave my own branch open in case future people wanted this patch. But for now closing PR since there's not enough time or will to merge.

PR #7493, opened 2025-08-09
Improve PP3 loading robustness.

gibbotronic:
I note that the PP3 file is an INI file where various string parameters have non-uniform capitalization, and the code within RawTherapee's processing looks for specific formatting to set up various variables, eg. `improcfun.cc` `ImProcFunctions::ciecam_02float` expects the ColorAppearance.SurrSrc parameter to be "Average" when setting up how it performs and "AVERAGE" is not an applicable value, thereby leading to various parameters not being set up correctly, so you can get a divbyzero and therefore a segfault.

Ideally all parameters would be converted to lowercase and all logic would work on lowercase parameters, but in any case this pull request adds the ability to cope with variations in case of various string parameters such that internal logic for processing still succeeds.

This is to help performance with relation to issues I was having in AI-PP3 at https://github.com/tychenjiajun/art/issues/5

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -25,6 +25,7 @@
 #include <glibmm/fileutils.h>
 #include <glibmm/miscutils.h>
 #include <glibmm/keyfile.h>
+#include <glibmm/ustring.h>
 
 #include "color.h"
 #include "colortemp.h"
@@ -42,6 +43,73 @@ using namespace std;
 
 namespace
 {
+
+Glib::ustring camelCapped(const Glib::ustring& str)
+{
+    Glib::ustring result(str);
+    result.replace(0, 1, 1, Glib::Unicode::toupper(result[0]));
+    bool nextShouldBeCapital = false;
+    for (size_t i = 1; i < result.size(); ++i) {
+        if (result[i] == ' ') {
+            nextShouldBeCapital = true;
+        } else {
+            if (nextShouldBeCapital) {
+                result.replace(i, 1, 1, Glib::Unicode::toupper(result[i]));
+                nextShouldBeCapital = false;
+            } else {
+                result.replace(i, 1, 1, Glib::Unicode::tolower(result[i]));
+            }
+        }
+    }
+    return result;
+}
+
+enum class KeyFormat {
+    Original,
+    CapitalizeWords,
+    CapitalizeFirstLetterOnly,
+    AllUppercase,
+    AllLowercase
+};
+
+Glib::ustring formatKey(const Glib::ustring& key, KeyFormat format = KeyFormat::Original)
+{
+    if (key.empty()) {
+        return key;
+    }
+    switch (format)
+    {
+        case KeyFormat::Original:
+            return key;
+        case KeyFormat::CapitalizeWords:
+            return camelCapped(key);
+        case KeyFormat::CapitalizeFirstLetterOnly: {
+            Glib::ustring result(key);
+            result.replace(0, 1, 1, Glib::Unicode::toupper(result[0]));
+            for (size_t i = 1; i < result.size(); ++i) {
+                result.replace(i, 1, 1, Glib::Unicode::tolower(result[i]));
+            }
+            return result;
+        }
+        case KeyFormat::AllUppercase: {
+            Glib::ustring result;
+            for (auto c : key) {
+                result += Glib::Unicode::toupper(c);
+            }
+            return result;
+        }
+        case KeyFormat::AllLowercase: {
+            Glib::ustring result;
+            for (auto c : key) {
+                result += Glib::Unicode::tolower(c);
+            }
+            return result;
+        }
+        default:
+            break;
+    }
+    return key;
+}
 
 Glib::ustring expandRelativePath(const Glib::ustring &procparams_fname, const Glib::ustring &prefix, Glib::ustring embedded_fname)
 {
@@ -92,7 +160,7 @@ Glib::ustring expandRelativePath2(const Glib::ustring &procparams_fname, const G
 		&& !procparams_fname2.empty()
 		&& Glib::path_is_absolute(procparams_fname2)) {
 		// embedded_fname is not a valid path;
-		// try with procparams_fname2 (the path defined in Preferences) as a prefix 
+		// try with procparams_fname2 (the path defined in Preferences) as a prefix
 		rPath = expandRelativePath(procparams_fname2 + G_DIR_SEPARATOR_S, prefix, embedded_fname);
 	}
 	return(rPath);
@@ -142,7 +210,7 @@ Glib::ustring relativePathIfInside2(const Glib::ustring &procparams_fname, const
 		// (the path defined in Preferences)
 		rPath = relativePathIfInside(procparams_fname2 + G_DIR_SEPARATOR_S, false, embedded_fname);
 	}
-	return(rPath);		
+	return(rPath);
 }
 
 
@@ -2072,7 +2140,7 @@ CGParams::CGParams() :
     pwr(1.2),
     colorspace("acesp1"),
     rolloff(true)
-    
+
 {
 }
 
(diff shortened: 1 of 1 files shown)


Lawrence37:
In my opinion, it is better to have a separate tool to align externally-generated pp3s to the format expected by RawTherapee. The pp3 format belongs to RawTherapee. It should be able to enforce a strict format to keep the pp3 reader simple. It should not have to cater to 3rd party applications that make mistakes in the pp3 generation.

gibbotronic:
I understand that the pp3 format is indeed RawTherapee's. But the issue is that there is no standardisation on the naming of text items due to the disparate and diverse contributors.

Eg. in some instances, text values are interpreted as enums (which is sensible) like TWO_DEGREES. Then sometimes it is capitalised (Coloropp), then other times lowercase (normal / high), then sometimes all upper (SLI / MAN / AUTO) then sometimes with spaces (As Image), sometimes without (AsImage), then sometimes camel-capped (ProPhoto).
So the format of each text-based parameter in the INI file is strangely non-uniform.

I did this as a method to make the usabilty of RawTherapee more helpful for external tools such as AI-PP3 without them having to track the idiosyncratic (random?) capitalisation of text parameters.
If the loading of the INI file can cope with a range of capitalisation of text parameters, surely the better. And if this handling of methods is within the source code of RawTherapee's loading function itself, it would save the maintenance of another external tool which could drift from the format in RawTherapee and stagnate.

Since submitting this, my machine has been working through photos from a European trip whilst I have been out and about for a few days (and at work), with rawtherapee-cli coping with AI-PP3's output happily.

I did the fix in here because AI-PP3's source is in JS and I much prefer C++, plus adding it to RawTherapee means other PP3 generators (which I know nothing about) can write their PP3s and RawTherapee will happily load them without fuss, or (as is the case here) won't segfault because a variable like coefQ has bypassed correct initialisation all because the word "Average" was written as "AVERAGE".

This pull request won't affect the handling of how RawTherapee saves its own pp3 files, of course. It just makes it more accepting of text parameters in the motley-assorted INI file (of which there is no written spec, as far as can tell). How can we expect external tools to write a standard format when the format is not documented?

Lawrence37:
I understand that the non-uniformity of the values and names is strange, but I still don't think that RawTherapee should support non-compliant values and names from external tools. Doing so would mean added complexity that must be maintained as long as RawTherapee keeps using pp3. The bigger reason why I think this way is because if the external tool knows which values and names are available, then surely it should know their proper spelling, capitalization, etc. If a external tool puts "average" instead of "Average", or "Mean" instead of "Average", I think that is 100% the responsibility of the external tool to correct. Non-AI powered pp3 generators don't have this issue. I just don't see how we can justify the added effort and complexity of having RawTherapee be more flexible.

gibbotronic:
I understand the worry in some respects, but where is the definition of "non-compliant values" for all to see?
If it isn't published anywhere then how can anything be considered non-compliant?

If the file format was XML we'd likely have a XSD which would be trivial to maintain, and we could point people to the XSD when they send us a malformed XML to load.
Instead we have an INI file with no standardisation or published spec and then fail to operate correctly (crash) when they don't meet the unpublished/imaginary standard.

Whether the external tool is AI or not is not overly relevant. I could load a PP3 in an editor and type with caps lock on and the file would load but the program would segfault when processing the pp3. Make it a binary format if we don't want text edits.

It's no different to accepting bad user input. You can't type out of range values into edit boxes and we validate the input.

As I say, I thought ensuring the text values met the expected internal formats on load was the simplest route instead of changing every function to work on lowercase values.
You can reject the pull request with the small changes to a single function which is no problem but then it would appear that there's no interest in validating the format of the loaded pp3, like we do for other user input, which to my mind is absolutely baffling.
You then end up with bugs in other places (as demonstrated by incorrect variable initialisation during the colour correction phase) and produce more work down the line all because you haven't validated user input from file input.

Lawrence37:
The pp3 specification is implicitly defined, but is nevertheless unambiguous. Deviations from the specification is what I mean by "non-compliant". Sorry if my wording is confusing.

It seems I'm having difficulty communicating my rationale as to why this is better off as an external tool. I will phrase it in a different way. Coaxing the text in the pp3 from within RawTherapee is essentially masking the problems.
Problem 1 - Crash with unknown values: If I understand you correctly, *any* unknown value leads to a crash for certain parameters. Adjusting a bad input's capitalization, whitespace, etc. only avoids crashes for a small set of bad inputs. A proper fix is using the default value if the parameter cannot be mapped to a recognized value.
Problem 2 - "Non-compliant" values from pp3 generators: Invalid values point to a bigger underlying problem. The pp3 generator does not understand the pp3 specification well enough. This could partially be due to the lack of a specification document. It could also be due in part to the pp3 generators' inability to follow instructions exactly (no matter how clear the specification is). This part is unique to LLM-based generators, which is why I referenced non-AI generators in my previous comment.

Problem 1 is to be solved inside RawTherapee and does not require coaxing inputs into recognized values. Problem 2 can be solved by writing a pp3 specification document, and in the case of LLM-based pp3 generators, creating an output correction tool that knows the shortcomings of the LLM(s).

gibbotronic:
Thank you for your patience and efforts to communicate clearly. It is much appreciated.

_"A proper fix is using the default value if the parameter cannot be mapped to a recognized value."_
Yes, this is probably the ideal. In my case I found that `improcfun.cc` `ImProcFunctions::ciecam_02float` does checks like `if (params->colorappearance.surrsrc == "Average")` .... AVERAGE obviously doesn't match.
The initialization of various floats to correct values (eg. `c = 0.f` is initialised) therefore doesn't happen, and subsequent function calls which expect sensible values (`Ciecam02::initcam1float`) now outputs a `wh = 0` value, so then the final calls to generate the computation value `coefQ = 32767.f / wh` means `coefQ` is `0`, and finally the call `Qpro = CAMBrightCurveQ[(float)(Qpro * coefQ)] / coefQ;` means you are getting a divide by zero, and the entire program segfaults.

_"The pp3 generator does not understand the pp3 specification well enough."_
This is most certainly true. But as the standard is only implicit, you can't blame it that much...
As you state a boring spec document would probably be required. But as I thought this was a headache (and definitely more maintenance than just keeping a single C++ function up-to-date, which is trackable in terms of effort in a single place - Git) and not enforcable by programmatic means like an XSD is to an XML (very simple), this is why I chose the route to make RawTherapee forgiving of the "non-compliant" values.
As the non-compliance was not the actual parameter values but how they were presented (shouted instead of whispered / uppercase instead of lower etc.) I took the fix I did. A shouted word has the same meaning as a whispered word; merely the volume is the difference. But it's the same meaning.

_"The pp3 specification is implicitly defined, but is nevertheless unambiguous."_
I would say that an implicit definition is ambiguous by default. An explicit file definition removes all ambiguity. It just currently means the program happens to work as expected when the PP3 file meets the varied idiosyncratic values for each parameter, which are only "defined" in the program logic at their point of use.
Eg. the C++ standard is very explicit how many things work, as is the C standard. The bits they leave implicit are typically the "undefined behaviour" areas, or more accurately, the areas where programmers can write things that fall outside the exceedingly explicit standard.

As it turns out (and a wonderful positive outcome so far), I've used this fix and tool to run 4500 photos through Anthropic Claude, and then have RawTherapee apply the suggested modifications, then made a LLM evaluate what is a "good" picture from all of them to whittle down to a few hundred photos, then programmatically arranged these in a Scribus XML file so that I have the start of a coffee-table book. It's saved me a lot of time (thanks!).
Also the edits are far less clumsier than my overblown efforts. I intend on working through the other 80,000 photos from my life for an annual for each year, since there is a horrible habit of never printing any digital photos.

Lawrence37:
I disagree with equating implicitness with ambiguity. It is possible to logically derive a complete and explicit pp3 specification document from the code. If such a document is considered unambiguous, then there was never any ambiguity because the before and after are logically equivalent.

I forgot to add that if the pp3 generator gets the capitalization wrong, it could indicate that it doesn't actually understand what it's doing and is just making stuff up that happen to be close to valid values. This brings us back to masking the problems.

PR #7541, opened 2025-11-11
innosetup 6 doesn't support Icelandic

Benitoite:
Fixes windows CI build. Icelandic is not supported by InnoSetup 6.


Desmis:
@Benitoite 

OK, I will merge soon :)

Jacques

PR #7551, opened 2025-12-07
Manage Gamut Compression constant in one place

kaesa:
This PR make changes to the Gamut Compression tool so that some constants that were formerly spread around the source file are now all handled in one place. Also, fixes an issue where the threshold defaults were being applied to images that had different working profile than the Rec2020, which is the requirement for these defaults.

This PR is a continuation to the PR #7550 previously created by @desmis.

Additional notes:
* Requesting review and comments from @digitalcarp.

### DIFF
--- a/rtgui/tools/compressgamut.cc
+++ b/rtgui/tools/compressgamut.cc
@@ -34,7 +34,6 @@ using namespace rtengine::procparams;
 
 namespace
 {
-// clang-format off
 
 struct ListItem
 {
@@ -44,6 +43,7 @@ struct ListItem
     const std::array<const double, 3> thresholdDefault; ///< { th_c, th_m, th_y }
 };
 
+// clang-format off
 const std::array<ListItem, 7> COLORSPACE_LIST_ITEMS = {{
     {0,  "rec2020",    "TP_COMPRESSGAMUT_REC2020",   { 0.815, 0.803, 0.880} },
     {1,  "prophoto",   "TP_COMPRESSGAMUT_PROPHOTO",  { 0.815, 0.803, 0.880} },
@@ -53,8 +53,8 @@ const std::array<ListItem, 7> COLORSPACE_LIST_ITEMS = {{
     {5,  "acesp1",     "TP_COMPRESSGAMUT_ACESP1",    { 0.815, 0.803, 0.880} },
     {6,  "beta",       "TP_COMPRESSGAMUT_BETA",      { 0.90 , 0.95 , 0.95} }
 }};
-
 // clang-format on
+
 }
 
 const Glib::ustring Compressgamut::TOOL_NAME = "compressgamut";
--- a/rtgui/tools/compressgamut.cc
+++ b/rtgui/tools/compressgamut.cc
@@ -59,7 +59,7 @@ const std::array<ListItem, 7> COLORSPACE_LIST_ITEMS = {{
 
 const Glib::ustring Compressgamut::TOOL_NAME = "compressgamut";
 
-Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPRESSGAMUT_LABEL"), false, true)
+Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPRESSGAMUT_LABEL"), true, true)
 {
     auto m = ProcEventMapper::getInstance();
     EvcgColorspace = m->newEvent(COMPR, "HISTORY_MSG_CG_COLORSPACE");//FIRST to reinit histogram
--- a/rtgui/tools/compressgamut.cc
+++ b/rtgui/tools/compressgamut.cc
@@ -32,6 +32,29 @@
 using namespace rtengine;
 using namespace rtengine::procparams;
 
+namespace
+{
+    struct ListItem
+    {
+        const int       ID;                 ///< The row number in the combobox
+        const char*     pp3Name;
+        const char*     translationName;
+        const double    limits_th_c[4];
+        const double    limits_th_m[4];
+        const double    limits_th_y[4];
+    };
+
+    ListItem ColorspaceListItems[] = {              
+        {   0,  "rec2020",    "TP_COMPRESSGAMUT_REC2020",   {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   1,  "prophoto",   "TP_COMPRESSGAMUT_PROPHOTO",  {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   2,  "adobe",      "TP_COMPRESSGAMUT_ADOBE",     {0., 1., 0.001, 0.79}, {0., 1., 0.001, 0.82}, {0., 1., 0.001, 0.85} },
+        {   3,  "srgb",       "TP_COMPRESSGAMUT_SRGB",      {0., 1., 0.001, 0.51}, {0., 1., 0.001, 0.82}, {0., 1., 0.001, 0.82} },
+        {   4,  "dcip3",      "TP_COMPRESSGAMUT_DCIP3",     {0., 1., 0.001, 0.68}, {0., 1., 0.001, 0.89}, {0., 1., 0.001, 0.95} },
+        {   5,  "acesp1",     "TP_COMPRESSGAMUT_ACESP1",    {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   6,  "beta",       "TP_COMPRESSGAMUT_BETA",      {0., 1., 0.001, 0.90}, {0., 1., 0.001, 0.95}, {0., 1., 0.001, 0.95} },
+    };
+}
+
 const Glib::ustring Compressgamut::TOOL_NAME = "compressgamut";
 
 Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPRESSGAMUT_LABEL"), false, true)
@@ -58,13 +81,9 @@ Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPR
     Gtk::Box *iVBox = Gtk::manage(new Gtk::Box(Gtk::ORIENTATION_VERTICAL));
    
     colorspace = Gtk::manage(new MyComboBoxText());
-    colorspace->append(M("TP_COMPRESSGAMUT_REC2020"));
-    colorspace->append(M("TP_COMPRESSGAMUT_PROPHOTO"));
-    colorspace->append(M("TP_COMPRESSGAMUT_ADOBE"));
-    colorspace->append(M("TP_COMPRESSGAMUT_SRGB"));
-    colorspace->append(M("TP_COMPRESSGAMUT_DCIP3"));
-    colorspace->append(M("TP_COMPRESSGAMUT_ACESP1"));
-    colorspace->append(M("TP_COMPRESSGAMUT_BETA"));
+    for (auto item : ColorspaceListItems) {
+        colorspace->append(M(item.translationName));
+    }
     colorspace->set_active(3);
     iVBox->pack_start(*colorspace);
     iFrame->add(*iVBox);
@@ -241,20 +260,10 @@ void Compressgamut::read (const ProcParams* pp, const ParamsEdited* pedited)
 
     colorspaceconn.block (true);
 
-    if (pp->cg.colorspace == "rec2020") {
-        colorspace->set_active(0);
-    } else if (pp->cg.colorspace == "prophoto") {
-        colorspace->set_active(1);
-    } else if (pp->cg.colorspace == "adobe") {
-        colorspace->set_active(2);
-     } else if (pp->cg.colorspace == "srgb") {
-        colorspace->set_active(3);
-    } else if (pp->cg.colorspace == "dcip3") {
-        colorspace->set_active(4);
-    } else if (pp->cg.colorspace == "acesp1") {
-        colorspace->set_active(5);
-    } else if (pp->cg.colorspace == "beta") {
-        colorspace->set_active(6);    
+    for (auto item : ColorspaceListItems) {
+        if (pp->cg.colorspace == item.pp3Name) {
+            colorspace->set_active(item.ID);
+        }
     }
     colorspaceconn.block (false);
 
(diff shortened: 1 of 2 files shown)


rtgui/tools/compressgamut.cc
```diff
@@ -32,6 +32,29 @@
...
+namespace
+{
+    struct ListItem
+    {
+        const int       ID;                 ///< The row number in the combobox
+        const char*     pp3Name;
+        const char*     translationName;
+        const double    limits_th_c[4];
+        const double    limits_th_m[4];
+        const double    limits_th_y[4];
```
digitalcarp: Suggestion: Use `std::array<double, 4>`. Consider making `setLimits()` accept `const std::array<double, 4>&` as the argument directly.

rtgui/tools/compressgamut.cc
```diff
@@ -32,6 +32,29 @@
...
+
+    ListItem ColorspaceListItems[] = {              
+        {   0,  "rec2020",    "TP_COMPRESSGAMUT_REC2020",   {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   1,  "prophoto",   "TP_COMPRESSGAMUT_PROPHOTO",  {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   2,  "adobe",      "TP_COMPRESSGAMUT_ADOBE",     {0., 1., 0.001, 0.79}, {0., 1., 0.001, 0.82}, {0., 1., 0.001, 0.85} },
+        {   3,  "srgb",       "TP_COMPRESSGAMUT_SRGB",      {0., 1., 0.001, 0.51}, {0., 1., 0.001, 0.82}, {0., 1., 0.001, 0.82} },
+        {   4,  "dcip3",      "TP_COMPRESSGAMUT_DCIP3",     {0., 1., 0.001, 0.68}, {0., 1., 0.001, 0.89}, {0., 1., 0.001, 0.95} },
+        {   5,  "acesp1",     "TP_COMPRESSGAMUT_ACESP1",    {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   6,  "beta",       "TP_COMPRESSGAMUT_BETA",      {0., 1., 0.001, 0.90}, {0., 1., 0.001, 0.95}, {0., 1., 0.001, 0.95} },
+    };
```
digitalcarp: Suggestion: Unindent the namespace contents. Add a `// clang-format off` and `// clang-format on` around the `ColorspaceListItems` array. This is to help with auto-formatting in the future.

rtgui/tools/compressgamut.cc
```diff
@@ -32,6 +32,29 @@
...
+    {
+        const int       ID;                 ///< The row number in the combobox
+        const char*     pp3Name;
+        const char*     translationName;
+        const double    limits_th_c[4];
+        const double    limits_th_m[4];
+        const double    limits_th_y[4];
+    };
+
+    ListItem ColorspaceListItems[] = {
```
digitalcarp: Nit: Use `lower_snake_case` for variable names (as suggested in coding conventions).

Suggestion: Prefer `std::array<ListItem, N>`. More idiomatic C++, but might require more typing due to poor type inferencing in C++11.

Important: Probably should make this a `const` or `constexpr`, in which case name as `UPPER_SNAKE_CASE`.

rtgui/tools/compressgamut.cc
```diff
@@ -58,13 +81,9 @@ Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPR
...
    
     colorspace = Gtk::manage(new MyComboBoxText());
-    colorspace->append(M("TP_COMPRESSGAMUT_REC2020"));
-    colorspace->append(M("TP_COMPRESSGAMUT_PROPHOTO"));
-    colorspace->append(M("TP_COMPRESSGAMUT_ADOBE"));
-    colorspace->append(M("TP_COMPRESSGAMUT_SRGB"));
-    colorspace->append(M("TP_COMPRESSGAMUT_DCIP3"));
-    colorspace->append(M("TP_COMPRESSGAMUT_ACESP1"));
-    colorspace->append(M("TP_COMPRESSGAMUT_BETA"));
+    for (auto item : ColorspaceListItems) {
```
digitalcarp: Performance: Use `const auto&` to prevent unnecessary copies.

rtgui/tools/compressgamut.cc
```diff
@@ -241,20 +260,11 @@ void Compressgamut::read (const ProcParams* pp, const ParamsEdited* pedited)
...
-        colorspace->set_active(2);
-     } else if (pp->cg.colorspace == "srgb") {
-        colorspace->set_active(3);
-    } else if (pp->cg.colorspace == "dcip3") {
-        colorspace->set_active(4);
-    } else if (pp->cg.colorspace == "acesp1") {
-        colorspace->set_active(5);
-    } else if (pp->cg.colorspace == "beta") {
-        colorspace->set_active(6);    
+    for (auto item : ColorspaceListItems) {
```
digitalcarp: Performance: Use `const auto&` to prevent unnecessary copies.

rtgui/tools/compressgamut.cc
```diff
@@ -307,23 +317,13 @@ void Compressgamut::write (ProcParams* pp, ParamsEdited* pedited)
...
-        pp->cg.colorspace = "adobe";
-    } else if (colorspace->get_active_row_number() == 3){
-        pp->cg.colorspace = "srgb";
-    } else if (colorspace->get_active_row_number() == 4){
-        pp->cg.colorspace = "dcip3";
-    } else if (colorspace->get_active_row_number() == 5){
-        pp->cg.colorspace = "acesp1";
-    } else if (colorspace->get_active_row_number() == 6){
-        pp->cg.colorspace = "beta";
+    for (auto item : ColorspaceListItems) {
```
digitalcarp: Performance: Use `const auto&` to prevent unnecessary copies.

rtgui/tools/compressgamut.cc
```diff
@@ -347,45 +347,27 @@ void Compressgamut::updategamutGUI()
...
+
+    * Furthermore, what about the necessary color adaptation, both at this stage of compression and at the stage implied by the output temperature (see CIECAM)?
+
+    Therefore, these parameters correspond roughly to the ratio of the distances between the white point of the Working Profile, 
+    * The white point of the "Target compression gamut" corrected by chromatic adaptation and the Yellow, Magenta, and Cyan values ​​of the primary color triangle.
+    * These calculated values ​​are for 'normal' cases where the illuminants are not exhaustive, and the colors to be recovered are within reasonable limits.
+    * In other cases, the 3 Threshold sliders need to be adjusted (often lowered). Pay attention to the artifacts.
+    */
+
+    for (auto item : ColorspaceListItems) {
```
digitalcarp: Performance: Use `const auto&` to prevent unnecessary copies.

digitalcarp (changes requested):
Thanks for doing the refactor cleanup!

rtgui/tools/compressgamut.cc
```diff
@@ -32,6 +32,29 @@
...
+        const double    limits_th_m[4];
+        const double    limits_th_y[4];
+    };
+
+    ListItem ColorspaceListItems[] = {              
+        {   0,  "rec2020",    "TP_COMPRESSGAMUT_REC2020",   {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   1,  "prophoto",   "TP_COMPRESSGAMUT_PROPHOTO",  {0., 1., 0.001, 0.815}, {0., 1., 0.001, 0.803}, {0., 1., 0.001, 0.880} },   // default limits
+        {   2,  "adobe",      "TP_COMPRESSGAMUT_ADOBE",     {0., 1., 0.001, 0.79}, {0., 1., 0.001, 0.82}, {0., 1., 0.001, 0.85} },
+        {   3,  "srgb",       "TP_COMPRESSGAMUT_SRGB",      {0., 1., 0.001, 0.51}, {0., 1., 0.001, 0.82}, {0., 1., 0.001, 0.82} },
+        {   4,  "dcip3",      "TP_COMPRESSGAMUT_DCIP3",     {0., 1., 0.001, 0.68}, {0., 1., 0.001, 0.89}, {0., 1., 0.001, 0.95} },
```
digitalcarp: Critical: Typo during refactor. Last element should be `0.9` not `0.95`.

Desmis:
@kaesa @digitalcarp 

Thank you for this work... To me it's all Greek, but I think it's better.

I tried it on the current version of @kaesa , and it seems to work. 

I only found one (minor) difference, but it doesn't concern 'Threshold', only 'Maximum Distance Limits', and only on the image with the LED (AT00129.DNG)... I used the exact same pp3. Perhaps there are other reasons? But I don't see how this change would affect these calculations.

There are no differences on the other images.

Thank you

Jacques

kaesa:
@Desmis 
I've now noticed the comment at `updategamutGUI()` states `// Only for working profile Rec2020` is this comment true? I don't see anything that checks the working profile either in here or at `colorspaceChanged()`, where `updategamutGUI()` is called.

My second question is that since the adjuster properties are not really modified, would it be more appropriate to use setValue instead of setLimits. It would help keeping the amount of numbers more manageable and to the point.

Desmis:
@kaesa 

Yes, the calculations are only valid if and only if you start with Rec2020. 

Doing this for a more detailed "Working Profile" is a real chore - and what's the point? I did this to make things easier, knowing that what's displayed is only indicative. Anyone who tells me how a sensor 'XXX' behaves under an illuminant 'YYY' without knowing either one is either lying or ignorant. And even if we knew them, what understandable translation could we provide to the user? Simply put, in these specific extreme conditions (LEDs in the image...cameras 0bserver unknown, Illuminant unknown...), we don't really know what's happening. Fortunately, these conditions are rare.

The only dot I put in the GUI is '*' next to 'Target compression' and the Tooltips... I don't think we should bother controlling the Working Profile... we'll just create a complicated mess. But if you know how to do it (simply), why not...

The second question it's part of the GUI. If you think it's more appropriate, no problem. I found 'setLimits' elsewhere in the code. Just a reminder that I'm absolutely terrible at coding, especially GUIs.

Thank you

Jacques

kaesa:
@digitalcarp
Thank you for the review and spotting the typo. Please let me know if these changes need more work.

@Desmis 
I think you miss understood what I asked. I meant if the Rec2020 needs to be checked. Based on the comment @lawrence37 made to the original PR and your comment, I now tested and found that the defaults were being applied to pictures that had ProPhoto as their working profile. I added a test to make sure the defaults are applied only when using the correct working profile. Please confirm if this is ok.

I also removed one call to the `updategamutGUI()` in the read function because right after that, the threshold values were being applied from the procparams. Please confirm if this is ok.

Desmis:
@kaesa 

In French, we say "je suis dur à la comprenoir"... I have problems with both English and understanding code.

I notice that the changes you made are working correctly.

Thank you

Jacques

rtgui/tools/compressgamut.cc
```diff
@@ -32,6 +32,31 @@
 using namespace rtengine;
 using namespace rtengine::procparams;
 
+namespace
+{
+// clang-format off
```
digitalcarp: Could you move the clang-format stuff to just wrap `COLORSPACE_LIST_ITEMS`? Sorry if I wasn't clear before.

Other than that, looks good to me.

kaesa:
Thank you.

Desmis:
@kaesa 
I'm taking advantage of this open pull request to avoid opening another one.

I think adding 1:1 to the 'Gamut Compression' label would be a good idea. It indicates that the behavior isn't perfect.

Indeed, sometimes small colored dots (similar to impulse noise) appear if the user hasn't selected 1:1

These artifacts disappear as soon as the zoom is at 100% or when outputting TIFF/JPG.


Could you change in 'compressgamut.cc':

Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPRESSGAMUT_LABEL"), false, true)

Replace by
Compressgamut::Compressgamut () : FoldableToolPanel(this, TOOL_NAME, M("TP_COMPRESSGAMUT_LABEL"), **true**, true)

Thank you

Jacques

Desmis:
@kaesa 

OK, thank you

kaesa:
@desmis Did I cause problems for merging with rebase and force-push?

Desmis:
@kaesa 
I hadn't tried the last(s) change(s) you made: 8685693 and force-pushed (For your information, I have absolutely no idea what it's for 'force-pushed' - Computer code, compilation directives, Git, etc., are as unclear to me as ancient Egyptian).
 
So I went to the repository created for your pull request.

git reset --hard HEAD
git pull

And here's what I get... a 'rebase' request

image

I created a new clone of your fork. I went into the 'combobox', did:
* cmake...
* make --jobs=16 install

And everything is fine.

I hope this helps you.

But the system suggests 'Squash and merge'

image


Jacques

kaesa:
In the future if you can do `git pull --rebase` in this kind of situation and that way you can avoid creating a new clone. I'll be more careful with the rebase in the future. I'm also not that not experienced with git. Squash merging keeps the commit history cleaner.

Desmis:
@kaesa 

Two questions: 

* Can this PR be merged without risk? Because currently I'm the only one (apart from Lawrence37) with the rights, . If things go wrong, what will I do? This is completely abnormal. From the outside, it looks like they want to kill Rawtherapee.
* Can I merge this PR or do I have to wait for something? For me, all works fine (in my repo).

Thank you

Jacques

kaesa:
I thought it is ok. But like I said, I'm not experienced with git and it does have a tendency to surprise me. So, lets ask @digitalcarp. It's no problem for me to create a new branch and PR for merging.

Desmis:
ok :)

digitalcarp:
Should be fine to merge. GitHub will auto-detect merge conflicts and it currently says "no conflicts with base branch".

Squash and merge is fine. I assume "squash and merge" means git will take all 5 commits on the branch and "squash" it into one commit before merging. Some projects like doing that for cleaner git history.

When you rebase interactively you can do the squash operation interactively.

```bash
git checkout combobox
git rebase -i dev

# In the text editor change "pick" to "squash" to combine a commit with the commit above.
pick commit-A
squash commit-B
squash commit-C
```

Your branch will change from "commit-A -> commit-B -> commit-C" to "commit-A" where all 3 commits are combined.

Desmis:
@kaesa @digitalcarp 

I will merge this PR soon. As soon as the 'Refactor locallab params' is validated in Dev

I will be away from home from Wednesday, December 17th until January 4th. At least in theory, I should have internet access.

Thank you

Jacques

PR #7554, opened 2025-12-11
Fixes use after free bug in the CacheManager::deleteEntry

kaesa:
There is a use after free bug in the CacheManager::deleteEntry:

When deleteEntry is called, the execution is directed to the Thumbnail::decreaseRef ();, which calls back to the CacheManager::closeThumbnail. In CacheManager::closeThumbnail the thumbnail is deleted if the reference count goes to zero. Then execution returns back to the CacheManager::deleteEntry, where the thumbnail->getMD5 is called using a pointer to a now deleted Thumbnail object.

### DIFF
--- a/rtgui/cachemanager.cc
+++ b/rtgui/cachemanager.cc
@@ -170,20 +170,11 @@ void CacheManager::deleteEntry (const Glib::ustring& fname)
 
     auto thumbnail = iterator->second;
 
-    // decrease reference count;
-    // this will call back into CacheManager,
-    // so we release the lock for it
-    {
-        lock.release ();
-        thumbnail->decreaseRef ();
-        lock.acquire ();
-    }
-
-    // check again if in the editor,
-    // the thumbnail still exists,
-    // if not, delete it
-    if (openEntries.count (fname) == 0) {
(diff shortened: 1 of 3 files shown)


Lawrence37:
Do you know a way to trigger the bug?

kaesa:
`deleteEntry` is executed when the user uses "File Operations -> Delete permanently" from mouse menu at the File Browser. Doing the same thing in the film strip when editing the picture, in which case the Thumbnail can't be deleted or it triggers bad behavior. And from `FileCatalog::reparseDirectory` when a file system event is handled on that same directory and the image is found missing. This happens, for example, when the trash can is emptied.

I have used RawTherapee for managing my pictures for a long time and I have never experienced bad behavior I can attribute to this. Of course it can be just luck. Sometimes emptying the trash is slow and sometimes even seems to freeze for a really long time, but I have primarily attribute these to the delete events triggering an excessive amount of reparseDirectory and sleeping hard drives. Now I'm not so sure anymore. I found this while I was going through mutexes for another thing.

Lawrence37 (approved):

PR #7576, opened 2025-12-29
Ciecam clean and improvments - GHS improvments

Desmis:
As I mentioned on Pixls.us, I (partially) cleaned up the CIECAM02/16 code, parts of which were over 10 years old. There are no changes to GHS in this Pull Request today.

I added a feature, simple at the Ciecam code level, but affecting many files for the GUI... I know how to do this, even if it's not at all my field (IT specialist). However, this needs to be verified, as I'm a bit absent-minded and make mistakes.

* **Change the hue** for the 3 mains colors known to the user: Red, Green, Blue, by rotating the data of a few degrees.
* **Modify the saturation** (in the CIECAM sense), particularly by lowering it, for these 3 colors. I chose the saturation component (s). I could have chosen Chroma (C) or Colorfulness (M), but (s) is simpler to implement, and I think it's the appropriate answer in this case (small variations).

I'm attaching the tooltip (in bad english)

...
 Corrects any color deviations, due to either: gamut overshoot following various adjustments in the process, poor white balance, or desires for more natural colors.
The system operates after the first process (scene) which converts the 'Working profile' data into that of 'CAM16'. At this stage, the J, Q, C, s, M, h data take into account the physiological aspects considered by CIECAM, and a chromatic adaptation is performed.
For each main color of CIECAM ac-bc [chroma based] hue plane - Red, Green, Blue - you can perform a slight hue(h) rotation in degrees, and reduce the saturation(s) as a percentage.
These adjustments are made just before the other CIECAM adjustments: Lightness (J), Brightness (Q), Chroma (C), Saturation (s), Colorfullness (M), hue (h), contrast(J), contrast(Q) and curves.
Next, the third process converts the data from the 'CAM16' color space to the 'Working profile', taking into account the Viewing conditions. Color adaptation is then performed.
This hue and saturation adjustment does not involve primaries. It is based on the principles of CAM (Color Appearance Models) in 3 processes.
...

image

I'm attaching a link that dates back to Ciecam02 (not Ciecam16), but which briefly explains how Ciecam works.
https://www.handprint.com/HP/WCL/color7.html#CIECAM


Desmis:
@kaesa @Lawrence37 @waynesutton50

For your information, this is another way to retouch colors... There's more to life than just primaries.

The actual (main) change is in improcfun.cc, starting at line 1183. It might seem like I made a mistake with 'Cp', but that's intentional, because if I need to change something, for example, replace 's' with 'M'...or others...

Of course you can check what I've done; in fact, there are a lot of files involved. I've also cleaned it up (by removing parts of the code that dated back to its creation in 2012).  Of course, I see 'typos' in my code, the Labels and Tooltips need to be corrected... But the main thing is to understand this approach... which doesn't quite go in the direction others are going.
You could also say that, given my age, I'm rambling and out of touch. For the record, CIECAM16 dates back to 2020... an eternity ago on such subjective topics...

Jacques

.github/workflows/appimage.yml
```diff
@@ -15,7 +15,7 @@ on:
   workflow_dispatch:
 
 env:
-  publish_pre_dev_labels: '[]'
+  publish_pre_dev_labels: '["Rawtherapee:ciecam-ghs"]'
```
digitalcarp, line 18: Should the labels be removed before merging?

rtgui/tools/locallabtools2.cc
```diff
@@ -13576,7 +13576,7 @@ void Locallabcie::convertParamToNormal()
     colorflcie->setValue(defSpot.colorflcie);
     lightqcie->setValue(defSpot.lightqcie);
     chromlcie->setValue(defSpot.chromlcie);
-    huecie->setValue(defSpot.huecie);
+    //huecie->setValue(defSpot.huecie);
```
digitalcarp: Is this comment intentional?

rtengine/iplab2rgb.cc
```diff
@@ -1062,7 +1133,7 @@ void ImProcFunctions::workingtrc(int sp, Imagefloat* src, Imagefloat* dst, int c
         D60 = 6005  // for ACES AP0 and AP1
     };
     double tempv4 = 5003.;
-    double p[6]; //primaries
+ //   double p[6]; //primaries
```
digitalcarp: Delete commented code.

rtengine/iplab2rgb.cc
```diff
@@ -737,13 +737,84 @@ void ImProcFunctions::apsatur(int sp, Imagefloat* tmpImage, Imagefloat* tmpImage
...
+    const float dx_new = scaling * distance_to_edge * cos_angle;
+    const float dy_new = scaling * distance_to_edge * sin_angle;
+    newprimxy[0] = dx_new + whitepoint[0];
+    newprimxy[1] = dy_new + whitepoint[1];
+}
+//end code Darktable
+
 void ImProcFunctions::workingtrc(int sp, Imagefloat* src, Imagefloat* dst, int cw, int ch, int mul, Glib::ustring &profile, double gampos, double slpos, int cat, int &illum, int prim, int locprim,
-                                 float &rdx, float &rdy, float &grx, float &gry, float &blx, float &bly, float &meanx, float &meany, float &meanxe, float &meanye,
+                                 float &rdx, float &rdy, float &grx, float &gry, float &blx, float &bly, float &meanx, float &meany, float &meanxe, float &meanye, double *p,
```
digitalcarp, line 823: Suggestion: Instead of `double *p`, reuse or define a new struct that has variable names. `p.red.x` is much easier to understand than `p[0]`.

The specifics are up to you, but I've provided an example.

```cpp
struct PrimaryCoord {
    double x = 0;
    double y = 0;
};

struct RgbCoord {
    PrimaryCoord red;
    PrimaryCoord green;
    PrimaryCoord blue;
};
```

rtengine/iplab2rgb.cc
```diff
@@ -1707,6 +1781,49 @@ void ImProcFunctions::workingtrc(int sp, Imagefloat* src, Imagefloat* dst, int c
...
+            float newprimxy[2];
+            for (int i = 0; i < 3; i++) { 
+                newprimxy[0] = 0.f;
+                newprimxy[1] = 0.f;
+                rotate_and_scale_primary(primaries, 1.f - inset[i], rotation[i], i , newprimxy, xyD);//xyD takes into account the default illuminant or the one chosen by the user
+                if(i == 0) {
+                    p[0] = newprimxy[0];
+                    p[1] = newprimxy[1];
+                    if(rtengine::settings->verbose) {
+                        printf("newRx_prim=%f newRy_prim=%f \n", (double)newprimxy[0], (double)newprimxy[1]);
```
digitalcarp: Optional: With the `fmt` library, you can simplify printing and formatting. Types are also automatically detected for you so casting is not needed.

```cpp
#include <fmt/format.h>

fmt::println("newRx_prim={} newRy_prim={}", newprimxy[0], newprimxy[1]);
```

.github/workflows/appimage.yml
Desmis: of course....but i am not ready..not at all to merge

Thank you

rtengine/iplab2rgb.cc
Desmis: I will have a look

rtgui/tools/locallabtools2.cc
Desmis: This is a work in progress. I pushed the commit to have a backup (before traveling...and returning home), but I'm far from finishing this PR.

rtengine/iplab2rgb.cc
Desmis: Ok

rtengine/iplab2rgb.cc
Desmis: It's probably 'better' from a code perspective, but the changes needed are significant, and my skills are limited. So I'm leaving it as is (I keep p*), but I don't see any problem with changing it.

Desmis:
@kaesa @Lawrence37 @waynesutton50 @digitalcarp 

What am I trying to do in this Pull-Request?

First, it means prioritizing the work of researchers in the field of colorimetry (for example, CIECAM, which officially began in 1997) over the work of individuals (or companies linked to the film industry). I'm not saying these initiatives are bad, but for me, we're reinventing the wheel a bit.

The problem of taking into account the physiological aspects (and there are many), some discovered centuries ago, others more recently around 1960, is no small matter. Believing that the problem will be solved by simply addressing one or two effects is a delusion. A CAM (Color Appearance Model) is a whole system, and to my knowledge, the only one that works (almost perfectly) in the majority of cases is CIECAM.

As a reminder, when I set out in 2012 to create user-friendly software that addressed the enormous complexity of CIECAM, some people thought I was crazy. I believe the RT version is quite advanced... It lacked, among other things, the ability to fine-tune the Hue, which is what I'm proposing in the first commits. For me (and for renowned researchers... who are getting old... like me), CIECAM is the Rolls Royce of colorimetry. To take an example, we don't convert data to Rec2020 or BT709 (sRGB), but to CAM16... It took researchers 20 years to find this compromise and all the "formulas" to account for the physiological effects, which aren't limited to just one or two effects, but at least five or six, distributed across 3 processes, all of which are taken into account by CIECAM.

However, in some cases, particularly with unusual illuminants (LEDs, halogens, illuminant mixtures, etc.), CIECAM cannot solve seemingly impossible equations. Hence the two modules "primaries & illuminants" in SE CAM16 and Abstract Profile.
Retouching primaries is a complex operation; it fundamentally changes the data, altering the RGB=>XYZ conversion matrix... It's not trivial. This also assumes, but I repeat myself (probably due to age), that the white balance, before starting any algorithm, is as good as possible (hence Temperature correlation WB auto) in maths.

If we deviate from the CIExy diagram, which, let's remember, is only a projection of the visible gamut... we end up with imaginary colors (the ones we want to eliminate)... what happens then? Short of being a magician or a fortune teller, we enter uncharted territory.
If we further concentrate the data, we will end up with out-of-gamut values... How do we control them? To my knowledge, there is no satisfactory answer.

The "primaries & illuminants" version of Abstract Profile had two ways to edit: linear primary data and the CIExy diagram. Since the latter is too small to view, I'm going to add the ability to modify the primaries in polar coordinates (I took the code from Darktable... by adapting it slightly and limiting it to geometry. Why reinvent the wheel? ). The first step is done; now I need to improve the GUI, keeping in mind that not everything is possible (at least not for me, given my limited coding skills). The interactions between the three possible modes - linear, graphical, and polar - are complex, both in terms of the code (the rtengine code) and the GUI. I don't think it's necessary to solve all the GUI issues for a tool that's rarely used. I've attached the tooltip. Of course, it's written in poor English, and I'm not sure it will be understandable.

--
You have three options for modifying the primaries by selecting them from the 'Destination primaries' combo box:
1) By selecting 'Custom (Linear sliders)' and changing the linear coordinates of Red, Green, and Blue (Rx,  Ry,  Gx,  Gy,  Bx,  By). In this case: a) the graphical representation of the CIExy diagram is updated; b) the polar coordinates (R.rot,  R.sat,  G.rot,  G.sat,  B.rot, B.sat) remain at zero. The system calculates the new X, Y, and Z values, taking into account the actual White Point.

2) By directly changing the values ​​using the CIExy diagram. When 'Custom CIExy diagram' is selected in 'Destination primaries', you can modify the values ​​of the three primaries directly on the graph. The system calculates the new X, Y, and Z values. Note that in this case, the White Point position on the graph will not be updated. The polar coordinates (R.rot,  R.sat,  G.rot,  G.sat,  B.rot,  B.sat) remain at zero.

3) With the 'Custom' (Polar sliders) option: a) the graphical representation of the CIExy diagram is updated; b) the linear coordinates (Rx, Ry, Gx, Gy, Bx, By) are not updated. The system calculates the new X, Y, and Z values. After selecting polar coordinates: a) if you change the combo box to 'Custom (CIE xy Diagram)', the linear coordinate values ​​will be updated. The polar coordinates (R.rot, R.sat, G.rot, G.sat, B.rot, B.sat) will remain at zero; b) if you change the combo box to 'Custom (Linear sliders)', the system will revert to the default values.

Be very careful when using this type of tool - whether it's polar sliders, linear sliders, or CIExy graphs. If you go outside the CIExy diagram, you'll generate imaginary colors, how will they be handled? If you reduce the primary values ​​too much, you risk out-of-gamut data. Use with caution, and with small values, while monitoring the primaries' deviation from the CIExy diagram.

--

There's one last part to do, and it's a significant one. In GHS, I added (by copying the code into a CTL) the ability to use a conversion matrix to AgX. I also added the ability to use a matrix to JzAzBz and Cat16. These calculations, even if they produce good results, are subject to significant cognitive biases... and the results can be inconsistent with certain images. Applying a conversion matrix to RGB data is, at the very least, very approximate. I'm going to try, at least for JzAzBz and Cat16, to do this in a more methodical way, that is, starting from the XYZ values ​​and the chosen white point (at least I hope!).

In conclusion, colorimetry is complex; the very concepts of primaries and white point are not easy to grasp. This system, conceived in 1931 by the ‘Commission Internationale de l’Eclairage (CIE)’, is showing its limitations with the increasing capabilities of cameras and software. I prefer to rely on research and a certain scientific rigor. Nevertheless, I am not criticizing what has been implemented elsewhere, even if it seems complex from the user's perspective (I'm not saying that CIECAM is simple...).

Your comments on the algorithms (but also the code, knowing that I am not at all an expert in this field), the labels and the tooltips are welcome.

Thank you

Jacques

rtgui/tools/locallabtools2.cc
Desmis: Yes, this is necessary to avoid resetting the GUI when changing complexity levels and to maintain the capabilities of the Hue slider.

Desmis:
@kaesa @Lawrence37 @waynesutton50 @digitalcarp 

I made the change to avoid relying solely on cognitive biases in GHS "matrix". I added two options: "JzAzBz (XYZ High dynamic)" and "Cat16 (XYZ harmonious). The modification is simple (RGB<=>XYZ conversion taking into account the Working Profile, plus, of course, the JzAzBz or Cat16 matrix). I'm referring to matrices, not CIECAM or the JzCzHZ algorithm. 

This approach seems less trivial to me, and more respectful of the principles of colorimetry. Nevertheless, I have retained the previous choices.

I tried to comment the code so that a developer could better understand the system's logic. I also tried my best (in broken English) to write tooltips.

I don't see why I would keep the default AgX Matrix setting, which implies a preference. So I changed it to 'none', meaning GHS alone.

Similarly, even though in 'RGB Luminance' mode the automatic Symmetry Point (SP) predetermination works quite well in most cases, it doesn't work in other situations.

Therefore, I prefer to put the system in 'neutral' mode, allowing the user to make the choices:
* 'Matrix': one choice from the 6 available
* 'Symmetry point' automatic or not,  as 'Auto Black point & White point' which is disabled by default

This is more neutral. But this should be written in the documentation, which is still very difficult to access (or perhaps I've missed a step).

@Lawrence37 
It is being produced, but that must be normal, many more outputs (for executables)  than before... probably due to the ARM versions... It takes a little longer. But no problem.

Jacques

rtengine/iplab2rgb.cc
```diff
@@ -741,13 +742,84 @@ void ImProcFunctions::apsatur(int sp, Imagefloat* tmpImage, Imagefloat* tmpImage
...
                     tmpImage->g(y, x) = G0 * 65535.f;
-                    tmpImage->b(y, x) = B0 * 65535.f;                                     
+                    tmpImage->b(y, x) = B0 * 65535.f;
                 }               
 }
 
 
+// code taken from Darktable - and adapted to Rawtherapee
+// It was only used to calculate polar coordinates. Why reinvent the wheel when it's already been done elsewhere?
+float determinant(float a, float b, float c, float d)
```
kaesa: More constness to improve compiler chances for optimizatons:
`constexpr float determinant(const float a, const float b, const float c, const float d)`

Desmis: OK

rtengine/iplab2rgb.cc
```diff
@@ -741,13 +742,84 @@ void ImProcFunctions::apsatur(int sp, Imagefloat* tmpImage, Imagefloat* tmpImage
...
+{
+    return a * d - b * c;
+}
+
+
+float intersect_line_segments(float x1, float y1,
```
Desmis, line 759: OK

Desmis:
@kaesa   Thank you very much :)

I will probably start a thread in Pixls.us, today or tomorrow morning to discuss this Pull Request and the changes it entails. I will try to be clear and explain: why CIECAM is used, why the use of primaries should be limited to the minimum necessary, and why there are two additional XYZ matrices in GHS. I will also reiterate the steps I recommend in 'Game changer'. Of course, this is just my opinion, and leaves everyone free to think and develop what they think is best.

Jacques

Desmis:
@kaesa 

I just change your last comment ... which is now deleted. For my part, I see no difference in how it works.

What should I do now? No problem for me, to change again

Thank you

kaesa:
I should note that apart from the `constexpr` copilot did not approve my suggestions. It suggests that passing by value for the small types can be more efficient. In this case it probably doesn't matter as the compiler should be able to inline the functions anyway. Making compiler more eager to inline was my goal. I may have been wrong or useless.

That is why I deleted the comment.

Desmis:
@kaesa 
So, I leave it as it is now?

kaesa:
Keep `constexpr` and remove other stuff I suggested. That is what copilot suggests. I agree based on the argument  that potential optimizations with `const float&` references are unreliable.

I.e.:
`constexpr float determinant(float a,  float b,  float c,  float d)`

and
````
float intersect_line_segments(float x1,  float y1,
                              float x2,  float y2,
                              float x3,  float y3,
                              float x4,  float y4)
````
is enough.

Desmis:
OK, I will change soon . No problem at all

Jacques

Desmis:
I've made a few improvements. Specifically, I've moved some tooltips that were 'intrusive' because they were assigned to a frame, for the primaries in icmpanel.cc and for the hue modifications (red, green, blue) in ColorAppearance.cc.

I've cleaned up the code and removed the white space and old comments (parts of the code were 5 to 12 years old!).

I added 3 small fixed dots in Labgrid (only for the Primaries and Illuminants section), which represent the Rec2020 primary values ​​(default)). This allows the user to see what they are doing.

Desmis:
@kaesa @Lawrence37 @waynesutton50 @digitalcarp 

I made some "improvements" (which I think you would have asked me for, or which Copilot would have suggested)... I first made several modifications concerning Abstract profile (Primaries & Illuminant part) without worrying too much about the GUI code for icmpanel.cc (a lot of duplicated code), nor the 'magic numbers' which had already been present for 5 years.

Adding the Rec2020 primary markers to the CIExy diagram led me to do things (that I don't always know how to do very well... it's code!)

* So I optimized the code duplication in icmpanel.cc with a function "upgateGUI_lin_pol_graph()"
* And for the transfer of variables coming from either iplocallab.cc, or iplab2rgb.cc or improccoordinator.cc, icmpanel.cc, localabtools2.cc, concerning the 3 primaries Red Green Blue, the White point of the illuminant and the Dominant color, I have completely revised the Coefficients of 1.81818, 0.55 and 0.1 to no longer use these 'magic numbers'.
* I also cleaned up the code again by removing comments and whitespace (I hope I didn't mess anything up...). I also tweaked a few tooltips a bit.

From my perspective, aside from some changes to labels and tooltips, I think I've finished what I set out to do. Of course, that doesn't mean the code is "perfect"... Your feedback is welcome.

I'll briefly recap what I had planned:
* clean CIECAM  (Colorappearance & lighting - Advance Tab) GUI and engine code (and a few minor adjustments to SE CAM16 for consistency) . This process handles possible hue variations (Red, Green, Blue) through rotation and saturation, without using primaries. It operates within the second CIECAM process. Physiological effects are addressed by CIECAM processes 1 (scene) and 3 (viewing). I still think (but maybe I'm just a silly old man) that CIECAM (CAM16) is a safe, proven solution, developed by leading researchers over the last 30 years or so. Perhaps I'm being a bit provocative, but the physiological aspect - what our eye/brain system perceives - are not related to new technologies (nor to AI - unless we're assuming brain implants...).
* Bringing rigor to the conversion matrices before and after GHS. Using a matrix on RGB data  is, to say the least, "curious"... So I added something more conventional (but it's not CIECAM, nor SE JzCzHz), using XYZ.
* Add a third option for modifying primaries in Abstract Profile, offering, as other software does (ART, Darktable, etc.), rotation and modification of the distance between the primary and the white point. Adding Rec2020 markers and viewing the changes on the CIExy diagram allows you to see what you're doing. You have to be very careful not to go outside the CIExy diagram...or conversely (unless it's intentional) not to force the gamut. This generates strong imaginary colors that you can't control. From my point of view, modifying the primaries must be perfectly understood by the user.
Thank you

Jacques

Desmis:
Even though I said I was finished: 
* I found a malfunction in the calculation of the equivalent luminance (a problem related to multiple RT spots). I took the opportunity to implement a conventional calculation (diagonale XYZ matrix) when the White Point value is reasonable (less than 1...it sometimes reaches around 0.5).

* I also added, for informational purposes only, an estimate of the 'Middle Grey' at the end of the GHS algorithm itself. So, before 'Stretch Regularization & Midtones' and 'Highlight Attenuation'. This demonstrates (also) that there is little to no connection with Symmetry Point (SP).

Desmis:
If no one objects, or suggests improvements for the Labels and tooltips, I will merge this PR sometime next week.

Jacques

PR #7583, opened 2026-01-05
Automatic language detection for Korean

Lawrence37:
Associates Korean with the "ko" locales and adds the untranslated keys to the language file.


digitalcarp:
LGTM

PR #7590, opened 2026-01-18
Workaround for Windows workflow Lensfun error

Lawrence37:
The latest version of MSYS2's Lensfun package installs the `lensfun` Python module in a new location (\<prefix>/lib/site-packages instead of \<prefix>/lib/python\<version>/site-packages) outside of the default paths used by Python (., \<prefix>/lib/python\<version>.zip, \<prefix>/lib/python\<version>, \<prefix>/lib/python\<version>/lib-dynload and \<prefix>/lib/python\<version>/site-packages). This provides a temporary workaround by adding the new directory to the Python path when invoking `lensfun-update-data`.

I see no issues with merging this immediately, but please inform me if there is a concern.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -96,7 +96,7 @@ jobs:
       - name: Include Lensfun
         run: |
           echo "Updating Lensfun database."
-          PYTHONPATH=/mingw64/lib/site-packages lensfun-update-data  # Workaround for lensfun path mismatch
+          PYTHONPATH=/${{env.MSYS2_ENV_PREFIX}}/lib/site-packages lensfun-update-data  # Workaround for lensfun path mismatch
           echo "Creating Lensfun directory in the build directory."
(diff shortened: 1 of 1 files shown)


kmilos:
Is this something that should be addressed upstream? Either Lensfun itself, or MSYS2 packaging?

Desmis:
All works fine... but why MSYS2's Lensfun package installs the lensfun Python module in a new location ? Curious..

Thank you

Jacques

Lawrence37:
I think the new directory is a mistake. I don't see any justification for it.

Lawrence37:
I've opened an issue with MSYS2.
https://github.com/msys2/MINGW-packages/issues/27455

Desmis:
@Lawrence37 

OK thank you

kmilos:
The corrected MSYS2 package is now available, so you could revert this workaround if you wish.

PR #7591, opened 2026-01-19
Foveon raw image thumbnail fix

### LINKED ISSUE #7495: Thumbnail is garbled after processing is done in X3F (foveon) files despite opening correctly
Currently using RawTherapee 5.12 on Windows 11 Pro 24H2

I was processing some X3F (foveon) files from my Sigma SD14 and Sigma SD1 Merril and the images come up just fine, especially when I thinker with colour noise reduction and use a dark frame they look better than using Sigma’s own PhotoPro App, but the thumbnails are completely garbled after the file is processed. Before it’s processed the thumbnail is fine because presumably it’s generated from the embedded JPEG.

By the way, some other bugs I found, but are not so critical:

The EXIF information also doesn’t show up in XF3 files, despite it being supported in dcraw, for example, but that’s not really a problem for me.

Trying to use multiple dark frames (from a directory) instead of a single one crashes the app immediately once I select Dark-Frame-Auto-Selection.

Tell me if those other bugs should have their own issues? I understand support for X3F is really low priority… few people have these cameras...

You can get some samples, including a black frame, here: https://rnbdc.ddns.net/pub/sigma_sd14_raws_OA36yH2Kc18Hyg.zip

Lawrence37:
Fixes the width and height used to generate the thumbnails for Foveon raw images. The source image contains the full raw data.

Closes #7495.

### DIFF
--- a/rtengine/rtthumbnail.cc
+++ b/rtengine/rtthumbnail.cc
@@ -238,7 +238,11 @@ void scale_colors (rtengine::RawImage *ri, float scale_mul[4], float cblack[4], 
             }
         }
     } else {
-        const int size = ri->get_iheight() * ri->get_iwidth();
+        // Foveon has full raw image.
+        const int size =
+            ri->getSensorType() == rtengine::eSensorType::ST_FOVEON
+                ? (ri->get_topmargin() + ri->get_iheight()) * ri->get_rawwidth()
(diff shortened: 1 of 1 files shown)


digitalcarp:
LGTM

PR #7594, opened 2026-01-25
Multiple small changes related to _refreshProgressBar - function

kaesa:
* This PR changes the way the loaded images are tracked. In the dev, the additions are tracked manually. This PR changes the code so that the value is acquired from the container.
* Changes the progress bar visually:
Current progressbar:
orig_all orig_filtered orig_loading
From left to right; All files loaded, some files filtered, loading new files.
Proposed:
new_all new_filtered new_loading_countdown
This change adds consistency to the way the brackets are used. Changes the way the `previewsToLoad` is displayed. The goal is to make it easier to register what is happening. In the dev, it is little inconvenient to calculate how many images are still to be loaded if the folder has more than a small amount of files.
Proposed look with horizontal bar:
new_all_horizontal
new_filtered_horizontal
new_loading_horizontal_countdown
Counting the "to be loaded images" down makes sense also, if my future PR, which changes the way events are handled gets accepted. However, this is also a matter of taste, and it is trivial to change the count down to `(previewsLoaded+previewsToLoad)` and is perfectly fine for me to do if this change gets opposition.
* The if in the `_refreshProgressBar` was causing the progress bar not updating when in the editor. This was buggy behavior for example when resetting the file catalog in the editor with X,Y or SHIFT+F3/F4 for example. I haven't found reasons to have it, but I think if there is a reason, the check could probably be more specific.
* Calling _refreshProgressBar inside a loop is not right. The progressbar is not getting redrawn until the execution returns back to the framework.

Additional note:
There is a bug when File Browser is processing a folder and user switches to another folder. The blue progress bar is left to a state all images minus one image. The bug happens because `if ( dir_id != selectedDirectoryId ) {` in `void FileCatalog::previewsFinishedUI(int dir_id)` is evaluated true for some reason. This bug is also present in dev and 5.12.

PS.
I closed this PR draft (and page) and opened it again by copy pasting old text. To my surprise, the images appeared correctly without reuploading. Please let me know if the images are missing and I'm seeing them just because of some caching reasons.

### DIFF
--- a/rtgui/filebrowser.h
+++ b/rtgui/filebrowser.h
@@ -130,7 +130,7 @@ protected:
     BatchPParamsChangeListener* bppcl;
     FileBrowserListener* tbl;
     BrowserFilter filter;
-    int numFiltered;
+    std::size_t numFiltered;
 
     void toTrashRequested   (std::vector<FileBrowserEntry*> tbe);
     void fromTrashRequested (std::vector<FileBrowserEntry*> tbe);
@@ -172,7 +172,7 @@ public:
     void applyPartialMenuItemActivated (ProfileStoreLabel *label);
 
     void applyFilter (const BrowserFilter& filter);
-    int getNumFiltered()
+    std::size_t getNumFiltered()
     {
         return numFiltered;
     }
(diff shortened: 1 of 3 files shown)


kaesa:
Regarding the bug in the additional note, I figured it happens because `PreviewLoader::removeAllJobs` leaves behind jobs that are already running and those have the old `dir_id`. Surprisingly they get executed in the end and thus the `previewsFinished` is executed with wrong `dir_id`. I managed to clear them by copying the relevant logic from `ThumbImageUpdater::removeAllJobs()`, which waits for the running jobs to finish before returning. Now I'm wondering what would be the right way to address it. What ever the fix would be, in my opinion, it does not belong to this PR.

kaesa:
This PR is now part of #7677.

PR #7599, opened 2026-02-04
User-Customizable Aspect Ratios in Crop and Framing Tools

Figuera:
## Summary

This PR introduces user-customizable aspect ratio management for the Crop and Framing tools. Users can now:
- Enable/disable built-in aspect ratios to reduce clutter in the ratio selection dropdown
- Create custom aspect ratios with user-defined labels and proportions
- Manage all aspect ratios through a new preferences panel

## Motivation

The previous implementation provided a fixed set of 28 aspect ratios with no user control. This caused two issues:
1. The aspect ratio dropdown became cluttered with ratios many users never needed
2. Users couldn't add custom aspect ratios for specialized workflows without  a convoluted process that needed to be repeated every time they wanted to use a non-built-in ratio.

This PR addresses both issues by giving users complete control over which ratios appear in their workflow.

### User Interface

A new "Aspect Ratios" section has been added to the Image Processing tab in Preferences:

2026-02-03_23-02-1770179221_partial

The interface provides:

**Control Buttons**
- "Select All" - Enables all aspect ratios (both built-in and custom)
- "Deselect All" - Disables all aspect ratios

**Aspect Ratio Table**
A scrollable table displaying all available aspect ratios with three columns:
- **Enabled** - Checkbox to enable/disable each ratio in the Crop/Framing tools
- **Label** - Display name (e.g., "3:2", "16:9", "21:9 Ultrawide")
- **Ratio Value** - Calculated decimal value of the aspect ratio (e.g., 1.500, 1.778, 2.333)

Built-in ratios are shown with checkboxes that can be toggled but cannot be deleted. Custom ratios can be both toggled and deleted.

**Custom Ratio Input**
Below the table, users can define new aspect ratios:
- **Custom Ratio** - Text field for the ratio label (e.g., "21:9 Ultrawide")
- **Value** - Two numeric fields separated by a colon for entering width:height (e.g., 21:9)
- **Calculated Result** - Live preview showing the decimal value (e.g., "= 2.333")
- **Add Button** (+) - Adds the custom ratio to the list
- **Delete Button** (−) - Removes the selected custom ratio from the list

**Input Validation**
The add button is only enabled when:
- Both width and height values are provided and greater than 0
- A label is provided
- The ratio value doesn't already exist (within 0.001 tolerance)
- The label doesn't duplicate an existing ratio name

I am open to suggestions for improving the UI design. The add and delete buttons, in particular, may be slightly confusing at first, but I haven't found a better solution.

A restart is required for changes to take effect. I don't see a way around this limitation, but more experienced developers may have suggestions.

### Technical Implementation

**Files Modified:**
- `rtgui/windows/preferences.h` (+33 lines) - Added UI components for aspect ratio management
- `rtgui/windows/preferences.cc` (+256 lines) - Implemented aspect ratio preferences panel
- `rtgui/options.h` (+7 lines) - Added storage vectors for user-defined ratios
- `rtgui/options.cc` (+40 lines) - Added serialization/deserialization for custom ratios
- `rtgui/aspectratios.cc` (-39 lines) - Built-in values moved to options.h, fill function modified to read from it.
- `rtdata/languages/default` (+14 lines) - Added translation strings

**Data Model:**

The following variables are added to the Option Object.
```cpp
std::vector<Glib::ustring> sysAspectRatiosLabels;   // 28 predefined labels
std::vector<double> sysAspectRatiosValues;          // Calculated values
std::vector<int> sysAspectRatiosEnabled;            // Enable/disable state
std::vector<Glib::ustring> userAspectRatiosLabels;  // Custom labels
std::vector<double> userAspectRatiosValues;         // Custom values
std::vector<int> userAspectRatiosEnabled;           // Enable/disable state
```

Two of these (`sysAspectRatiosLabels` and `sysAspectRatiosValues`) cannot be changed by any means and always use the default values defined at options.h (This is something I may have to work on). All the others can be modified on the options file under the "Crop Settings" section, preferably through the preferences settings UI.

Here is an example of the options set:

```yaml
sysAspectRatiosEnabled=1;1;1;0;1;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;0;
userAspectRatiosLabels=21:9;
userAspectRatiosValues=2.3333333333333335;
userAspectRatiosEnabled=1;
```

This means that only a couple of main built-in ratios are enabled, and the custom 21:9 ratio has been added and enabled, the user will see 21:9 (it could have been 21:9 Ultrawide) and rawtherapee will use the 2.3333333333333335 internally.

If these options are not set, as they would be for everyone not using this branch, all built-in aspect ratios would be enabled and all the user aspects would be empty.

## Testing Performed

- Verified enabling/disabling built-in ratios updates Crop/Framing dropdowns
- Tested adding custom ratios with various width:height combinations
- Confirmed duplicate detection prevents identical ratios or labels
- Tested deletion of custom ratios (delete button disabled for built-in)
- Verified settings persistence across application restarts
- Tested Select All/Deselect All buttons
- Confirmed restart requirement notice is displayed

## Usage Example - summary

To add a custom 21:9 ultrawide aspect ratio:
1. Open Edit → Preferences → Image Processing
2. In the "Aspect Ratios" section, enter:
   - Custom Ratio: "21:9 Ultrawide"
   - Value: 21 : 9 (the result "= 2.333" appears automatically)
3. Click the + button to add
4. Restart RawTherapee
5. The new ratio appears in Crop and Framing tool dropdowns

To hide unused built-in ratios:
1. Uncheck unwanted ratios in the Enabled column
2. Restart RawTherapee
3. Disabled ratios no longer appear in tool dropdowns

## Compatibility

The feature is fully backward compatible:
- Existing configurations continue to work with all 28 built-in ratios enabled
- New options are added to the settings file without affecting existing keys
- I checked compatibility with the PR RawTherapee/RawTherapee#7503, as suggested in a previous discussion. I confirmed that the PR doesn't address the same problem and are mostly compatible with one another, however since that PR moves the location of aspectratios.(h,cc) a manual intervention would be required for the merge to happen, which should be easy enough.

## Related Issues

This PR addresses issue RawTherapee/RawTherapee#5693


kaesa:
Looks pretty good. I noticed couple things.

* The program crashes to an unhandled exception if you try to insert anything but a number to the number field.
* The resulting ratio has 4 digits of accuracy, but the input fields can take only 3 digits. Maybe the input fields could be expanded by one digit.

Figuera:
@kaesa

Thank you for your time.

I pushed a commit to address the crash you mentioned in (1).

I had previously added logic to block non-numeric characters, but it seems that it wasn't working as expected. The crash was ultimately caused by the Result label attempting to compute a division when the input wasn’t numeric. This commit does two things: Actually Ignores non-numeric inputs and, updates the Result label to fail silently if something goes wrong.

Regarding (2): I don’t feel strongly about allowing 3 or 4 digits. However, I’m not aware of any real-world aspect ratios that need that level of precision, in fact all aspect ratios I know use 1 or 2 digits, more than 3 feels unnecessary. The resulting ratio uses 4 digits because it is trying to show the ration as a decimal rather than a fraction.

kaesa:
Someone might want to define golden ratio (1.618). Users may come up with various ways to achieve their preferred ratios. Including raw pixel counts. Ultimately, it is just about allowing users to do easily whatever they want to do. As long as it is safe. However, it is not that important to me.

Figuera:
I see your point.

I considered increasing the precision even higher, however since the results ratio and the aspect table ratio only display 4 digits it would create a source of confusion, so I left it at 4.

kaesa:
It works well now. Thank you. 

There is a use case for a shared workflow or, if the user re-installs RawTherapee and loses old configurations. The crop on the opened image displays the crop correctly, until the users tries to adjust the crop in anyway, at which point the old data is permanently lost. I think it would satisfy the POLA principle, if the ratio would be automatically switched to "Current" when the ratio is unrecognized. Even better would be, if the custom ratio name in the pp3 could be left unchanged. I don't know if that would be feasible with a reasonable amount of work for the benefit. The most complete solution would be, if the program would ask the user if they want to add custom ratio found from the pp3. However, adding the functionality might make this feature unnecessarily big.

I don't think the above use case is really important. But if there is something what could be done with a reasonable amount of effort, it would be a great detail. There is also some amount of previously existing buggy behavior with the Crop module. Similar issue happens with the orientation, but right now, I just don't remember the steps to reproduce.

PS. I noticed, that if I run old and new version of the program on the same machine, the previously added custom ratios disappear from the config. I personally wouldn't consider that a big deal with this kind of feature.

Figuera:
I added a commit for the crop tool to default to "Current" when the aspect ratio of the processing parameters file is unknown. There is no equivalent behavior in the framing tool, but it already default to "As Image" which is not bad.

I did put some work on the idea of having a dialog for "Add a new custom aspect ratio" however it as adding to much complexity and points of failure to the editor code and I decided to drop it. If someone thinks it is really important I can come back to it.

kaesa:
I've tested and it works good now. I have no complaints.

digitalcarp:
Looks very useful! One minor UI improvement may be to split the custom aspect ratio entry UI into two rows instead of just one. In certain languages, it may not fit as nicely as in English.

Figuera:
> Looks very useful! One minor UI improvement may be to split the custom aspect ratio entry UI into two rows instead of just one. In certain languages, it may not fit as nicely as in English.

To test this, I asked an AI to generate translations for all the languages in rtdata/languages and checked whether the UI broke in any of them. I manually verified the Portuguese and Spanish translations (languages I know); they seemed correct overall, though somewhat awkward in places. I also kept the dialog window close to its minimum possible width. Under these conditions, some languages did exhibit UI issues—in particular, in some cases the “–” button was hidden from view.

That said, when this happens, many other entries in the Preferences dialog also show layout problems. For example, in the Image Processing tab, the directory combo boxes are often too narrow and the label to their right to be clipped or not visible at all. However, the most severe issues appear in the Layout section of the General tab (the first section shown): the entire second column of checkboxes are often be hidden, as well as all the “restart required” labels.

For these reasons, I tend to conclude that I should not change the UI. I like the current layout, and the Preferences dialog as a whole appears to assume a window width large enough for all elements to fit cleanly.

kaesa:
@Figuera you're still here? 5.14 roadmap has started so maybe this could get accepted for it?

Figuera:
@kaesa

I am still here but have been busy, I may have time for the upcoming week.

What would need to be done for this to get to 5.14?

rtdata/languages/default
```diff
@@ -2199,6 +2190,20 @@ PREFERENCES_PARSEDEXTADDHINT;Add entered extension to the list.
 PREFERENCES_PARSEDEXTDELHINT;Delete selected extension from the list.\nPredefined extensions cannot be deleted.
 PREFERENCES_PARSEDEXTDOWNHINT;Move selected extension down in the list.
 PREFERENCES_PARSEDEXTUPHINT;Move selected extension up in the list.
+PREFERENCES_ASPECTRATIOS;Aspect Ratios
+PREFERENCES_ASPECTRATIO_ENABLED;Enabled
+PREFERENCES_ASPECTRATIO_LABEL;Label
+PREFERENCES_ASPECTRATIO_VALUE;Ratio Value
+PREFERENCES_ASPECTRATIO_CUSTOMLABEL;Custom Ratio
```
Lawrence37, line 2197: Column labels and individual settings are in sentence case.
```suggestion
PREFERENCES_ASPECTRATIO_VALUE;Ratio value
PREFERENCES_ASPECTRATIO_CUSTOMLABEL;Custom ratio
```

rtdata/languages/default
```diff
@@ -2199,6 +2190,20 @@ PREFERENCES_PARSEDEXTADDHINT;Add entered extension to the list.
...
+PREFERENCES_ASPECTRATIO_CUSTOMVALUE;Value
+PREFERENCES_ASPECTRATIO_LABELPLACEHOLDER;e.g., 21:9 Ultrawide
+PREFERENCES_ASPECTRATIO_SELECTALL;Select All
+PREFERENCES_ASPECTRATIO_DESELECTALL;Deselect All
+PREFERENCES_ASPECTRATIO_ADDHINT;Add custom aspect ratio
+PREFERENCES_ASPECTRATIO_DELHINT;Delete selected custom aspect ratio.\nBuilt-in aspect ratios cannot be deleted.
+PREFERENCES_ASPECTRATIO_SELECTALL_HINT;Enable all default aspect ratios
+PREFERENCES_ASPECTRATIO_DESELECTALL_HINT;Disable all aspect ratios
+PREFERENCES_ASPECTRATIO_INFO;Aspect ratios are used in Crop and Framing tools. Built-in ratios cannot be deleted but can be disabled. Restart Required.
```
Lawrence37, line 2206:
```suggestion
PREFERENCES_ASPECTRATIO_INFO;Aspect ratios are used in Crop and Framing tools. Built-in ratios cannot be deleted but can be disabled. Restart required.
```

rtgui/windows/preferences.cc
```diff
@@ -777,6 +788,118 @@ Gtk::Widget* Preferences::getImageProcessingPanel ()
     cropFrame->add(*cropGrid);
     vbImageProcessing->pack_start(*cropFrame, Gtk::PACK_SHRINK, 4);
 
+    // Aspect Ratios
+    Gtk::Frame* frar = Gtk::manage(new Gtk::Frame(M("PREFERENCES_ASPECTRATIOS")));
+    Gtk::Box* vbar = Gtk::manage(new Gtk::Box(Gtk::ORIENTATION_VERTICAL));
+    vbar->set_spacing(4);
+
+    Gtk::Box* hbButtons = Gtk::manage(new Gtk::Box());
```
Lawrence37, line 796: A couple of the variables (`hbButtons`, `valueColumn`, `valueRenderer`, etc.) could be renamed to make them clear that they are used only in the aspect ratio section.

rtgui/windows/preferences.cc
```diff
@@ -777,6 +788,118 @@ Gtk::Widget* Preferences::getImageProcessingPanel ()
...
+    aspectRatios->append_column_editable(M("PREFERENCES_ASPECTRATIO_ENABLED"), aspectRatioColumns.enabled);
+    aspectRatios->append_column(M("PREFERENCES_ASPECTRATIO_LABEL"), aspectRatioColumns.label);
+
+    auto* valueColumn = Gtk::manage(new Gtk::TreeViewColumn(M("PREFERENCES_ASPECTRATIO_VALUE")));
+    auto* valueRenderer = Gtk::manage(new Gtk::CellRendererText());
+    valueColumn->pack_start(*valueRenderer, true);
+    valueColumn->set_cell_data_func(*valueRenderer, [this](Gtk::CellRenderer* cell, const Gtk::TreeModel::iterator& iter) {
+        auto* textCell = dynamic_cast<Gtk::CellRendererText*>(cell);
+        if (textCell) {
+            char buf[32];
```
Lawrence37, line 824: I would prefer to use `std::array` instead

rtgui/windows/preferences.cc
```diff
@@ -777,6 +788,118 @@ Gtk::Widget* Preferences::getImageProcessingPanel ()
...
+    vbar->pack_start(*scrollAR, Gtk::PACK_EXPAND_WIDGET, 4);
+
+    Gtk::Box* hbAdd = Gtk::manage(new Gtk::Box());
+    hbAdd->set_spacing(4);
+    Gtk::Label* labelLabel = Gtk::manage(new Gtk::Label(M("PREFERENCES_ASPECTRATIO_CUSTOMLABEL") + ":", Gtk::ALIGN_START));
+    aspectRatioLabel = Gtk::manage(new Gtk::Entry());
+    aspectRatioLabel->set_placeholder_text(M("PREFERENCES_ASPECTRATIO_LABELPLACEHOLDER"));
+    aspectRatioLabel->set_width_chars(15);
+    Gtk::Label* valueLabel = Gtk::manage(new Gtk::Label(M("PREFERENCES_ASPECTRATIO_CUSTOMVALUE") + ":", Gtk::ALIGN_START));
+    aspectRatioWidth = Gtk::manage(new Gtk::Entry());
```
Lawrence37, line 841: Why not a `Gtk::SpinButton` for the width and height entry?

rtgui/windows/preferences.cc
```diff
@@ -2009,6 +2141,23 @@ void Preferences::storePreferences()
         moptions.parseExtensionsEnabled.push_back(c[i][extensionColumns.enabled]);
     }
 
+    moptions.sysAspectRatiosEnabled.clear();
+    moptions.userAspectRatiosLabels.clear();
+    moptions.userAspectRatiosValues.clear();
+    moptions.userAspectRatiosEnabled.clear();
+    Gtk::TreeNodeChildren ar = aspectRatioModel->children();
+
+    for (size_t i = 0; i < ar.size(); i++) {
```
Lawrence37, line 2150: This can be converted into a ranged-based `for` loop

rtgui/windows/preferences.cc
```diff
@@ -2818,6 +2985,110 @@ void Preferences::moveExtDownPressed()
...
+    }
+    int width = std::stoi(widthText);
+    int height = std::stoi(heightText);
+    if (width <= 0 || height <= 0) {
+        addAspectRatio->set_sensitive(false);
+        aspectRatioResult->set_text("= ");
+        return;
+    }
+    double newValue = static_cast<double>(width) / static_cast<double>(height);
+    char buf[32];
```
Lawrence37, line 3021: I would prefer to use `std::array` instead

rtgui/windows/preferences.cc
```diff
@@ -2818,6 +2985,110 @@ void Preferences::moveExtDownPressed()
...
+    aspectRatioResult->set_text(buf);
+    if (aspectRatioLabel->get_text().empty()) {
+        addAspectRatio->set_sensitive(false);
+        return;
+    }
+    Gtk::TreeNodeChildren children = aspectRatioModel->children();
+    Glib::ustring newLabel = aspectRatioLabel->get_text();
+    for (const auto& row : children) {
+        if (row[aspectRatioColumns.label] == newLabel ||
+            std::abs(row[aspectRatioColumns.value] - newValue) < 0.001) {
```
Lawrence37, line 3032: Aspect ratios that are inverses of each other (e.g. 3:2 and 2:3) are not treated as duplicates, but they behave the same in the crop and framing tools.

PR #7601, opened 2026-02-08
Fixes Thumbnail reference counting issues when deleting files

kaesa:
### This PR changes two things
The calls to `CacheManager::deleteEntry` messes `Thumbnail` manual reference counting. This PR changes those two calls to `CacheManager::clearFromCache`. So that the cached files will be deleted, but the `Thumbnails`s are left alone.

Also, notice that previously cached files were not deleted, if the file was open in the editor or it was opened in the editor. After this change, the cached files are always deleted if the image file is deleted.

### Steps to reproduce
1. Have an image in the File Browser
2. Choose to add the image to the processing queue using the mouse right button menu.
3. Delete the image
4. Switch to Processing queue and click the x on the corner to remove the thumbnail.

On my tests the program seems to go in deadlock. Or maybe not. It is essentially a use after free bug so anything should be possible, but the `Thumbnail`s are heavily mutexed, so maybe it indeed just locks up.

### Longer explanation

A short while ago I noticed the batch processing has Thumbnail reference counting issue. I started looking into it and obviously I was able to fix the problems by adding an extra `Thumbnail::incrementRef` call. The trouble began when I couldn't find the right place to decrement. I eventually proceeded to surround debug printing with all thumbnail reference manipulations to map out who was doing what. Then I found out, the calls to  `CacheManager::deleteEntry` are wrong. 

Here are three scenarios with reference counting and file deletion:
1. Image is open only in File Browser. The image is deleted. The `Thumbnail` references go to zero already on the call to `fileBrowser->delEntry (fname);` after that, the code calls the  `CacheManager::deleteEntry`  which will merely clear cache files if existing and otherwise return as a no-op. Because it can no longer find the `Thumbnail` from the internal container. -- There are no ill effects.
5. Image is open in file browser and then opened in editor. Opening the image to the editor increments the reference count twice, because for some reason, registering as a listener to the `Thumbnail`, also increments the reference count. When the image is deleted, there are still two references before the call to  `CacheManager::deleteEntry`. So the `Thumbnail` survives with the lifeline of having the listener interface increment the references too. -- The program continues with no ill effects. [The editor never ever decrements the references, so once an image has been opened in the editor, its `Thumbnail` will exist the entire execution of the program.]
6. Image is open in File Browser, then added to the Processing queue, then deleted. Unfortunately, in this case the call to `CacheManager::deleteEntry` will decrement the reference count to zero and the `Thumbnail` will get deleted while the processing queue still needs a valid pointer.

In my previous PR #7554, I erroneously claimed the bad path is executed always, when files are deleted. This is not true. I was basing my analysis on reading the code. In reality, it requires very specific conditions for the bad path to be actually executed.

### DIFF
--- a/rtgui/filecatalog.cc
+++ b/rtgui/filecatalog.cc
@@ -1042,10 +1042,11 @@ void FileCatalog::deleteRequested(const std::vector<FileBrowserEntry*>& tbe, boo
     if (msd.run() == Gtk::RESPONSE_YES) {
         for (unsigned int i = 0; i < tbe.size(); i++) {
             const auto fname = tbe[i]->filename;
+            const auto md5 = tbe[i]->thumbnail->getMD5();
             // remove from browser
             delete fileBrowser->delEntry (fname);
             // remove from cache
-            cacheMgr->clearFromCache (fname, true);
+            cacheMgr->clearFromCache (fname, md5, true);
             // delete from file system
             ::g_remove (fname.c_str ());
             // delete paramfile if found
@@ -1396,8 +1397,9 @@ void FileCatalog::clearFromCacheRequested(const std::vector<FileBrowserEntry*>& 
 
     for (unsigned int i = 0; i < tbe.size(); i++) {
         Glib::ustring fname = tbe[i]->filename;
+        Glib::ustring md5 = tbe[i]->thumbnail->getMD5();
         // remove from cache
-        cacheMgr->clearFromCache (fname, leavenotrace);
+        cacheMgr->clearFromCache (fname, md5, leavenotrace);
     }
 }
 
@@ -1809,12 +1811,11 @@ void FileCatalog::reparseDirectory ()
 
     // check if a thumbnailed file has been deleted or is not in a directory of interest
     const std::vector<ThumbBrowserEntryBase*>& t = fileBrowser->getEntries();
-    std::vector<Glib::ustring> fileNamesToDel;
     std::vector<Glib::ustring> fileNamesToRemove;
 
     for (const auto& entry : t) {
         if (!Glib::file_test(entry->filename, Glib::FILE_TEST_EXISTS)) {
-            fileNamesToDel.push_back(entry->filename);
+            cacheMgr->clearFromCache(entry->filename, entry->thumbnail->getMD5(), true);
             fileNamesToRemove.push_back(entry->filename);
         }
         else if (!App::get().options().browseRecursive && Glib::path_get_dirname(entry->filename) != selectedDirectory) {
@@ -1826,11 +1827,8 @@ void FileCatalog::reparseDirectory ()
         delete fileBrowser->delEntry(toRemove);
         --previewsLoaded;
     }
-    for (const auto& toDelete : fileNamesToDel) {
-        cacheMgr->clearFromCache(toDelete, true);
-    }
 
-    if (!fileNamesToDel.empty()) {
+    if (!fileNamesToRemove.empty()) {
         _refreshProgressBar();
     }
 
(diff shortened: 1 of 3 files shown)


kaesa:
I removed the linked issue. After reading the report again carefully, I realized, the term delete doesn't actually mean deleting files, but just removing images from the queue.

Lawrence37:
In verbose mode, this error message appears after deleting a file.
`Failed to delete all files for cache entry '/path/to/image.jpg': No such file or directory`

It's not really an issue, but it could indicate unnecessary double cache deletion.

kaesa:
This was a small puzzle. There are couple small problems here.
1. In `deleteRequested`
```
    if (md5.empty ()) {
        return;
    }

    auto error = g_remove (getCacheFileName ("images", fname, ".rtti", md5).c_str ());
    error |= g_remove (getCacheFileName ("embprofiles", fname, ".icc", md5).c_str ());

    if (purgeData) {
        error |= g_remove (getCacheFileName ("data", fname, ".txt", md5).c_str ());
    }

    if (purgeProfile) {
        error |= g_remove (getCacheFileName ("profiles", fname, App::PARAM_FILE_EXTENSION, md5).c_str ());
    }

    if (error != 0 && rtengine::settings->verbose) {
        std::cerr << "Failed to delete all files for cache entry '" << fname << "': " << g_strerror(errno) << std::endl;
    }
```
The entries "embprofiles" and "profiles" fail. Others succeeded with the file I tested.

2. When the `reparseDirectory` calls the `clearFromCache`, the image has already been deleted from the filesystem and for that reason, getMD5 fails to produce an MD5 string and the cached files are never deleted.
```
std::string CacheManager::getMD5 (const Glib::ustring& fname)
{

    const auto file = Gio::File::create_for_path(fname);
```
When the files are being deleted through the UI, the image file gets deleted after the cache has been cleared.

rtgui/filecatalog.cc
```diff
@@ -1826,11 +1827,8 @@ void FileCatalog::reparseDirectory ()
...
         --previewsLoaded;
     }
-    for (const auto& toDelete : fileNamesToDel) {
-        cacheMgr->deleteEntry(toDelete);
-    }
 
-    if (!fileNamesToDel.empty()) {
+    if (!fileNamesToRemove.empty()) {
         _refreshProgressBar();
     }
```
Lawrence37, line 1833: There are at least as many files to remove as there are to delete. Is `_refreshProgressBar()` supposed to be called eve if no files were deleted?

kaesa: Oops. Sorry about this.

kaesa: Hold on. The call is made because the `fileNamesToRemove` contains files that are removed from the File Browser. So this is correct behavior. Cached files are not indicated in the UI at all.

Lawrence37:
Is the latest commit supposed to resolve the cache deletion issue? I still get the failure message.

kaesa:
It fixes the issue of reparseDirectory was not deleting cached files. I don't really know about the delete failure. I don't know if there is a real problem with the files in "embprofiles" and "profiles". Are these files supposed to exist for all files? Problem with `deleteFiles` is that it will report if any of the files it tries to delete doesn't exist.

For the original concern, deletion is attempted only once.

Lawrence37:
I know the embprofiles files may not exist (depending on the image file) and the profiles files don't exist unless the image has been edited. It's weird that the failure message did not appear prior to this pull request yet the cache files were still deleted upon deleting the image file.

kaesa:
I tested current dev and release-5.12 and I do get the failure message when deleting through the RawTherapee UI, but not if I delete with an external file browser. The `Thumbnail` reference counting is a mess. It could be that the cached files get deleted very sparsely. The old call, `deleteEntry`, didn't delete the cached files if there was anyone holding a reference to the `Thumbnail`. For that note,  I have tested and deleting the cached files didn't give any trouble when saving jpg from an already deleted image still open in the editor.

Lawrence37:
Maybe I was mistaken. I tried again on dev and I see the message.

Lawrence37 (approved):

PR #7602, opened 2026-02-09
Improve scroll wheel handling in preferences

Figuera:
## Summary

This PR improves scroll wheel behavior in the Preferences dialog by preventing unfocused widgets from capturing scroll events. Key improvements include:

- Scroll wheel events now propagate to the outer preferences window when scrolling over unfocused ComboBoxTexts, SpinButtons, and TreeViews
- Users can now scroll the preferences page naturally without widgets hijacking the scroll events
- Focused widgets continue to respond to scroll events as expected (e.g., changing ComboBoxText values or SpinButton numbers)
- Special handling for nested ScrolledWindows to ensure proper event propagation

## Motivation

The previous implementation had a frustrating UX issue: when scrolling through the Preferences dialog, any unfocused widget under the mouse cursor would capture scroll events, preventing the page from scrolling. Users had to carefully position their mouse to avoid these widgets, making navigation through preferences cumbersome. This presented the user with two problems.

1 - Some Preferences remained hidden from view, since scrolling to them required intent.
2 - It was easy to accidentally modify preferences while attempting to scroll through them.

This PR is tangentially related to PR #7599, which adds a full-width TreeView to the Preferences dialog that makes scrolling challenging. That said, since that PR is not yet merged a slight modification would have to be made to that PR to adjust to the proposed scrolling behavior of this PR.

### Technical Implementation

The main strategy is to replace all instances of ComboBoxText, SpinButton, and TreeView with MyComboBoxText, MySpinButton, and MyTreeView. Both MyComboBoxText and MySpinButton are already defined in guiutils.h and used extensively in the software but somewhat ignored in the preferences, and their definition already implements improved scrolling behavior where values only change when the shift key is pressed or when the widget has focus. 

The transition from Gtk::SpinButton to MySpinButton was smooth, but using MyComboBoxText presented problems since its default width is fixed at 70px and doesn't adapt to entry content (which standard ComboBoxText does and is the expected behavior of a couple of inputs in the preferences). This was fixed by introducing `setPreferredWidthFromEntries()`, which calculates the width of all entries and adjusts the combobox width to fit the longest entry.

MyTreeView had to be implemented from scratch. Its implementation is more complex than the others because each TreeView is wrapped in a ScrolledWindow that continues scrolling even when the TreeView ignores scroll events. To achieve the expected behavior, scroll events must be manually propagated to the second-level parent.

As an addendum, I kept the name MyTreeView to remain consistent with the naming scheme in guiutils.h, though RTTreeView might be more appropriate.

**Files Modified:**
- `rtgui/guiutils.h` (+14 lines) - Added MyTreeView class declaration and setPreferredWidthFromEntries() method
- `rtgui/guiutils.cc` (+54 lines) - Implemented MyTreeView scroll handling and width calculation
- `rtgui/windows/preferences.h` (+28 lines, -28 lines) - Changed widget types from Gtk to custom My* classes
- `rtgui/windows/preferences.cc` (+65 lines, -58 lines) - Replaced all Gtk::ComboBoxText, Gtk::SpinButton, and Gtk::TreeView instances with custom classes

## Testing Performed

- Verified scroll wheel scrolls the preferences page when mouse is over unfocused widgets
- Verified TreeView (file extensions list) scrolls internally when focused, propagates when unfocused
- Confirmed Shift+Scroll continues to work as expected
- Tested across multiple preference tabs and widget types

## User Experience

**Before:**
- Scrolling over any ComboBox/SpinButton/TreeView would capture the scroll event
- Users had to carefully avoid widgets while scrolling the preferences
- Frustrating navigation experience

**After:**
- Natural scrolling behavior throughout the preferences dialog
- Widgets only capture scroll when focused (intentional interaction)
- Smooth, predictable scrolling experience

## Compatibility

Fully backward compatible:
- No changes to data structures or file formats
- No changes to user-visible behavior except improved scroll handling
- Existing configurations continue to work unchanged

### DIFF
--- a/rtgui/guiutils.cc
+++ b/rtgui/guiutils.cc
@@ -1140,6 +1140,26 @@ void MyComboBoxText::get_preferred_width_for_height_vfunc (int height, int &mini
     minimum_width = rtengine::max(minimumWidth, RTScalable::scalePixelSize(10));
 }
 
+void MyComboBoxText::setPreferredWidthFromEntries()
+{
+    Glib::RefPtr<Gtk::TreeModel> model = get_model();
+    if (!model) {
+        return;
+    }
+    Glib::RefPtr<Pango::Layout> layout = create_pango_layout("");
+    layout->set_font_description(get_pango_context()->get_font_description());
+    int max_width = 0;
+    for (const Gtk::TreeModel::Row& row : model->children()) {
+        Glib::ustring text;
+        row.get_value(0, text);
+        layout->set_text(text);
+        int w, h;
+        layout->get_pixel_size(w, h);
+        max_width = std::max(max_width, w);
+    }
+    setPreferredWidth(max_width + 42, max_width + 42);
+}
+
 
 MyComboBox::MyComboBox ()
 {
@@ -1282,6 +1302,40 @@ bool MyHScale::on_key_press_event (GdkEventKey* event)
     } else {
         return Gtk::Widget::on_key_press_event(event);
     }
+}
+
+MyTreeView::MyTreeView ()
+{
+}
+
+bool MyTreeView::on_scroll_event (GdkEventScroll* event)
+{
+    // If widget has focus, propagate the signal to the wrapper.
+    if (event->state & GDK_SHIFT_MASK || has_focus()) {
+        Gtk::TreeView::on_scroll_event(event);
+        return false;
+    }
+
+    // Not focused: find outer ScrolledWindow (skip immediate parent which wraps this TreeView)
+    Gtk::ScrolledWindow* outer_sw = nullptr;
+    int scroll_window_count = 0;
+    for (Gtk::Widget* p = get_parent(); p; p = p->get_parent()) {
+        if (auto* sw = dynamic_cast<Gtk::ScrolledWindow*>(p)) {
+            scroll_window_count++;
+            // Skip the first ScrolledWindow (immediate parent), find the outer one
+            if (scroll_window_count > 1) {
+                outer_sw = sw;
+                break;
+            }
+        }
+    }
+    // Forward scroll to outer scroller (dialog/page)
+    if (outer_sw) {
+        return outer_sw->event(reinterpret_cast<GdkEvent*>(event));
+    }
+
+    // No outer scroller found, just propagate
+    return false;
 }
 
 class MyFileChooserWidget::Impl
--- a/rtgui/guiutils.h
+++ b/rtgui/guiutils.h
@@ -405,6 +405,7 @@ public:
     explicit MyComboBoxText (bool has_entry = false);
 
     void setPreferredWidth (int minimum_width, int natural_width);
+    void setPreferredWidthFromEntries();
     void connect(const sigc::connection &connection) { myConnection = connection; }
     void block(bool blocked) { myConnection.block(blocked); }
 };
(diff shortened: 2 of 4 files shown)


kaesa:
Seemed to work well. I think this is real usability improvement.

rtgui/guiutils.cc
```diff
@@ -1140,6 +1140,26 @@ void MyComboBoxText::get_preferred_width_for_height_vfunc (int height, int &mini
...
+    int max_width = 0;
+    for (const Gtk::TreeModel::Row& row : model->children()) {
+        Glib::ustring text;
+        row.get_value(0, text);
+        layout->set_text(text);
+        int w, h;
+        layout->get_pixel_size(w, h);
+        max_width = std::max(max_width, w);
+    }
+    setPreferredWidth(max_width + 42, max_width + 42);
```
digitalcarp, line 1160: Why is 42 used here?

rtgui/guiutils.cc
```diff
@@ -1284,6 +1304,40 @@ bool MyHScale::on_key_press_event (GdkEventKey* event)
...
+            if (scroll_window_count > 1) {
+                outer_sw = sw;
+                break;
+            }
+        }
+    }
+    // Forward scroll to outer scroller (dialog/page)
+    if (outer_sw) {
+        return outer_sw->event(reinterpret_cast<GdkEvent*>(event));
+    }
```
digitalcarp, line 1335: Hijacking GTK's usual event propagation like this seems like a code smell to me. Have you considered instead implementing a special custom scrolled window widget instead? That way, the scrolled window itself can decide if it wants to propagate the event to the outer window. To check if any children of the scrolled window has focus, ChatGPT say you can do:

```
GtkWidget* toplevel = gtk_widget_get_toplevel(widget);
GtkWidget* focus = gtk_window_get_focus(GTK_WINDOW(toplevel));

bool inner_has_focus =
    focus && gtk_widget_is_ancestor(focus, widget);

if (!inner_has_focus && !(event->state & GDK_SHIFT_MASK)) {
    // forward scroll to outer
}
```

digitalcarp: Implementing a custom scrolled window also has the benefit of being more flexible and reusable.

rtgui/guiutils.cc
Figuera: 42 accounts for the initial padding in the combobox entries.

This function exists because switching from the native GTK ComboBoxText to MyComboBoxText removed automatic width calculation. It iterates over the list entries, finds the longest text, and sets the combobox width accordingly. However, this still left the combobox slightly too narrow, as it didn't account for the initial padding; the + 42 corrects for that. It is not the best solution and if you have a better proposition I am happy to apply it.

rtgui/guiutils.cc
Figuera: That is probably for the best, I will try to get some work done on that later.

rtgui/guiutils.cc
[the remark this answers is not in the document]
Lawrence37: Copilot's suggestion does make a lot of sense. The problem with `MyComboBoxText`, as you stated, is the override that results in a default width of 70 window pixels. The override is done in `get_preferred_width_vfunc`, so calling `Gtk::ComboBoxText`'s `get_preferred_width_vfunc` to get the widths allows us to preserve the width behavior exactly.

Another option is to enhance `MyComboBoxText` to allow the option of using the custom sizing or deferring to `Gtk::ComboBoxText`.

PR #7605, opened 2026-02-10
Fix for #7604 - minor bug in writing history

### LINKED ISSUE #7604: minor bug in history shown
RT 5.12
when modifying the threshold of highlight reconstruction (inpaint) the history only shows 0 or 1. 
The correct value is written in pp3-file, so there is no lack of function.

Image

kaesa:
Fixes history bug with "Highlight reconstruction" -> "Inpaint Opposed" -> "Gain threshold" . Value written to history was truncated to int, while users expect floating point with 2 digit precision. Issue 7604

Fixes #7604


digitalcarp:
LGTM

Lawrence37 (approved):

Issue #4744, opened 2018-08-21
Changes to a module should activate it when it's not


Thanatomanic:
It would be useful for a more rapid editing process when a module gets activated as soon as you start moving a slider or tick a box or anything. The rationale behind this would be that whenever you move a slider you intend to use the module.
For example, right now you might want to add a little vibrance to your image. You then first need to activate the `Vibrance` module and then play with the slider. In the alternative situation you just need to move the slider.

Edit: inspiration for this request was found [here](https://discuss.pixls.us/t/sharpening-not-working/8693/10)

ff2000:
The downside could be that you accidentally activate a feature that you don't need simply by pressing a button/slider.

fatso83:
I am a fresh user that stirred up quite a racket on pixls.us when encountering this and commenting on this. I had tried searching and reading the RT docs for why my changes to the sliders did not take effect for what seemed like way too long, when encountering the linked post in the issue, but I failed to see paragraph four in [Getting Started](https://rawpedia.pixls.us/Getting_Started), which is the only place this seems to be mentioned. 

Having CPU intensive tools disabled by default (the current behavior) seems like a good idea, considering that having a responsive UI is a main priority for a good user experience. What does not strike me as a good idea, is _leaving the tools disabled_ after you start using them. After all, there is a reason (probably several) for why you could not apply this consistently for all tools. The crop tool, for instance, is automatically enabled once you use it. There is no "use step", followed by "enable". That would be super confusing for anyone having used _any image editing software_ and, like in the other cases, would make the program seem like it is not working.

I think having a default setting where tools are enabled once you start fiddling with the sliders is a good default for new users. Existing users could get a toggle to flip in the Preferences panel to appease them. It might be worth considering if this functionality should only be enabled on new installs, as to not annoy existing users too much.

PR #7606, opened 2026-02-10
Add Auto-Enable Feature for Tool Interactions

Figuera:
## Summary

This PR implements automatic tool enabling when users interact with disabled tools, improving workflow by eliminating the need to manually toggle tools on before adjusting their parameters. When users interact with sliders, curves, grids, or other controls in a disabled tool, the tool automatically enables itself.

## Motivation

The current workflow requires users to first enable a tool before they can adjust its parameters. This creates an unnecessary extra step and can create confusion. Auto-enable on user input is the default behavior in other raw editors including Darktable, Lightroom, and Nikon NX Studio, and is standard in editing software, which creates frustrated expectations for new users of RawTherapee.

## Scope

- ✅ **Fully implemented:** Exposure, Detail, and Color tabs
- ⚠️ **Partial:** Advanced tab (most tools work, Wavelet requires additional work)
- ❌ **Not included:** Local Editing and Raw tabs (future work)

### Technical Implementation

The core implementation strategy consists of creating the `tryEnableTool` method on the Adjuster object, which is the main user input across all the different tools. Every time an Adjuster is changed, it will check if the parent tool is disabled and enable if necessary. A method to ignore this behavior is implemented for special cases.

The same method is applied for the LabGrid and Curves objects, which operate similarly.

Per-tool implementation to deal with non-conventional input was also applied for all tools in the Exposure, Detail, and Color tabs. Most tools in the Advanced tab work correctly, but Wavelet will need additional work. Local Editing and Raw tabs were not tested.

In general, I avoided auto-enabling the tool if the input was just a setting that has no practical effect alone, such as the choice of a color space, but I believe this needs more testing.

**Core Infrastructure:**
- Implemented auto-enable logic in `Adjuster` widget (used extensively across tools)
- Added auto-enable support to `CurveEditorGroup` for all curve-based controls
- Implemented auto-enable in `LabGrid` for color/tone grid interactions
- Provided `setAutoEnableTool()` method to allow disabling the behavior per-widget
- Special case handling for many tools

**Tool-Specific Implementations:**
- **FilmSimulation:** Enables when selecting a CLUT profile
- **FilmNegative:** Enables when clicking to pick reference spots
- **Gradient:** Enables when activating edit mode
- **Spot:** Enables when entering edit mode
- **ColorToning:** Enables when activating "Show Mask" mode
- **Dehaze:** Enables when activating "Show Depth Map"
- **DirPyrEqualizer:** Enables when using contrast adjustment buttons
- **ToneEqualizer:** Enables when toggling colormap visualization

**Special Cases:**
- Some widgets explicitly disable auto-enable where the behavior would be undesirable (e.g., `rstprotection` in LCurve)

**Files Modified:**
- `rtgui/toolpanel.h` / `rtgui/toolpanel.cc` - Added enableTool() method
- `rtgui/widgets/basic/adjuster.h` / `rtgui/widgets/basic/adjuster.cc` - Implemented auto-enable logic
- `rtgui/widgets/curves/curveeditorgroup.h` / `rtgui/widgets/curves/curveeditorgroup.cc` - Added curve auto-enable
- `rtgui/labgrid.h` / `rtgui/labgrid.cc` - Implemented grid auto-enable with parent tool reference
- `rtgui/tools/colortoning.cc` - Integrated with LabGrid and show mask button
- `rtgui/tools/dehaze.cc` - Auto-enable on show depth map
- `rtgui/tools/dirpyrequalizer.cc` - Auto-enable on contrast buttons
- `rtgui/tools/filmnegative.cc` - Auto-enable on spot picking
- `rtgui/tools/filmsimulation.h` / `rtgui/tools/filmsimulation.cc` - Auto-enable on CLUT selection
- `rtgui/tools/gradient.cc` - Auto-enable on edit mode activation
- `rtgui/tools/labcurve.cc` - Disabled auto-enable for rstprotection adjuster
- `rtgui/tools/spot.cc` - Auto-enable on edit mode activation
- `rtgui/tools/toneequalizer.cc` - Auto-enable on colormap toggle

## Testing Performed

- Verified tools auto-enable when adjusting sliders/curves/grids while disabled
- Confirmed no auto-enable occurs when explicitly disabled via `setAutoEnableTool(false)`
- Tested all modified tools to ensure proper enabling behavior
- Verified no interference with normal enabled tool operation
- Confirmed edit mode activation properly enables respective tools
- Tested special visualization modes (show mask, show depth map, colormap) trigger enabling

## User Experience

**Before:**
- User must click to enable tool before adjusting any parameters
- Two-step process: enable, then adjust

**After:**
- User can directly interact with any tool control
- Tool automatically enables on first interaction

## Compatibility

Should be backward compatible, extent of tool affected demands testing.

### DIFF
--- a/rtgui/tools/filmsimulation.cc
+++ b/rtgui/tools/filmsimulation.cc
@@ -66,7 +66,7 @@ FilmSimulation::FilmSimulation()
     :   FoldableToolPanel( this, TOOL_NAME, M("TP_FILMSIMULATION_LABEL"), false, true )
 {
     m_clutComboBox = Gtk::manage( new ClutComboBox(App::get().options().clutsDir) );
-
+    m_clutComboBox->setListener(this);
     int foundClutsCount = m_clutComboBox->foundClutsCount();
 
     if ( foundClutsCount == 0 ) {
@@ -86,7 +86,9 @@ FilmSimulation::FilmSimulation()
 void FilmSimulation::onClutSelected()
 {
     Glib::ustring currentClutFilename = m_clutComboBox->getSelectedClut();
-
+    if (!currentClutFilename.empty() && currentClutFilename != "NULL") {
+        m_clutComboBox->tryEnableTool();
+    }
     if ( getEnabled() && !currentClutFilename.empty() && listener && currentClutFilename != m_oldClutFilename ) {
         Glib::ustring clutName, dummy;
         HaldCLUT::splitClutFilename( currentClutFilename, clutName, dummy, dummy );
@@ -210,7 +212,8 @@ std::unique_ptr<ClutComboBox::ClutModel> ClutComboBox::cm2;
 
 ClutComboBox::ClutComboBox(const Glib::ustring &path):
     MyComboBox(),
-    batchMode(false)
+    batchMode(false),
+    listener(nullptr)
 {
     if (!cm) {
         cm.reset(new ClutModel(path));
@@ -464,4 +467,15 @@ Gtk::TreeIter ClutComboBox::findRowByClutFilename( Gtk::TreeModel::Children chil
     }
 
     return result;
+}
+void ClutComboBox::setListener(FoldableToolPanel* l)
+{
+    listener = l;
+}
+void ClutComboBox::tryEnableTool()
+{
+    if (!listener) {
+        return;
+    }
+    listener->enableTool();
 }
--- a/rtgui/guiutils.cc
+++ b/rtgui/guiutils.cc
@@ -15,6 +15,7 @@
  *  You should have received a copy of the GNU General Public License
  *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
  */
+#include <algorithm>
 #include <cairomm/cairomm.h>
 #include "rtengine/rt_math.h"
 
@@ -1089,7 +1090,7 @@ void MyScrolledToolbar::get_preferred_height_vfunc (int &minimumHeight, int &nat
     }
 }
 
-MyComboBoxText::MyComboBoxText (bool has_entry) : Gtk::ComboBoxText(has_entry)
+MyComboBoxText::MyComboBoxText (bool has_entry) : Gtk::ComboBoxText(has_entry), toolPanel(nullptr)
 {
     minimumWidth = naturalWidth = RTScalable::scalePixelSize(70);
     Gtk::CellRendererText* cellRenderer = dynamic_cast<Gtk::CellRendererText*>(get_first_cell());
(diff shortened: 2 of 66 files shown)


rtgui/tools/filmsimulation.cc
```diff
@@ -86,7 +86,9 @@ FilmSimulation::FilmSimulation()
 void FilmSimulation::onClutSelected()
 {
     Glib::ustring currentClutFilename = m_clutComboBox->getSelectedClut();
-
+    if (!currentClutFilename.empty() && currentClutFilename != "NULL") {
```
kaesa, line 89: `if (!currentClutFilename.empty() && currentClutFilename != "NULL") {`
Something off here?

I'm not reviewing!

kaesa: Ok I see. The string content can be "NULL".

kaesa:
I don't notice any difference with Selective Editing or when adding tools to spots. Does this draft include the functionality for it? I'm actually thinking, the Selective Editing could get auto enabled when adding a spot. Were you thinking that when editing a tool in a spot, it would auto enable the Selective Editing, the spot itself and the specific tool that is being adjusted?

Figuera:
I haven't yet implement it for the selective editing or any tool with multi level folding. I am thinking about testing on the standards tools first, consolidate a set of directives of how and when it should work and then implement it on the multi level folding tools.

rtgui/tools/filmsimulation.cc
Figuera: Yeah I mimicked the logic from another part of the file.

kaesa:
The concept feels good and works well on quick testing.

Lawrence37:
Related issue: #4744.
Related discussion: https://discuss.pixls.us/t/sharpening-not-working/8693/39

Kildor:
One advantage of current behaviour is ability to quick adjust tool before you switch it on, especially on slow PC or large raws.

Figuera:
I pushed a commit refactoring the auto-enable logic by introducing an abstract class that centralizes the required auto-enable methods, and by having most user-input objects inherit from it.

The main addition is support for auto-enabling both parent and child tools in multi-level tool groups such as Color Management, Wavelets, and Selective Editing (although the latter hasn’t been tested yet, and the implementation there is likely incomplete).

One remaining issue is the Resize tool, which uses a different multi-level structure. Its sub-tools (Post-resize Sharpening and Framing) are linked directly to the parent tool in toolpanelcoords.cc, rather than being managed inside the tool itself. This makes automatically enabling the parent tool more challenging. I tried a few initial approaches without success, so we may need to refactor this tool to behave more like the others.

Another area of concern is checkboxes. They appear to have special behavior that is only triggered in batch-editing mode, which I don’t normally use and don’t fully understand,  I tried to be extra careful around these, but this part may still need review.

kaesa:
New changes seem to work ok.

Figuera:
I just pushed two, unfortunately large, commits centered on solving most of the remaining  problems. The first replaces checkbox-style inputs from tools with a custom class that inherits  from Gtk::CheckButton and implements the auto-enable capability. This follows the logic that was   already being done with the Adjuster class and the Curve class, for example. I tested the  behavior of the new CheckBoxes, even under batch editing which has special behavior, and the new  inputs seem to be working fine.

The other major step was the introduction of auto-enable capabilities to the Selective Editing tools. Since the Adjusters, Curves, and Buttons were all already using custom classes, most of the work was centered on replacing all 80+ checkboxes in the locallab tools in order for them to have auto-enable capabilities.

Given the extent of the changes, a great deal of testing is recommended.

The only remaining issue now is the framing and post-resize sharpening tools, which have a  different construction logic that breaks the auto-enable capability.

kaesa:
Looks promising on quick testing. I think you need to rebase at some point because there is already a new tool added to the selective editing. Also, @desmis should be testing the tools at selective editing still work as expected when he has time for it.

digitalcarp:
I've only quickly scanned the code, but it looks like you added this feature by injecting the parent tool panel to a bunch of widgets.

Have you considered instead adding an event listener to the refreshmap/history events that every tool emits instead? If you define which events belong to which tool, you can turn the appropriate tool on when an event is received. Might be less intrusive of a change and keeps widgets decoupled.

Figuera:
> I've only quickly scanned the code, but it looks like you added this feature by injecting the parent tool panel to a bunch of widgets.
> 
> Have you considered instead adding an event listener to the refreshmap/history events that every tool emits instead? If you define which events belong to which tool, you can turn the appropriate tool on when an event is received. Might be less intrusive of a change and keeps widgets decoupled.

Hmm, that is an interesting proposition. I haven't thought of that.

My intuition is that it would be bad to hijack other events in order to do this. I also think that the set of custom widgets that I consolidated in these commits might be positive to the project facilitating future changes by consolidating the widget construction in one place.

All that said, I do believe that my method got far too complex and it would demand that every developer do be on board with the new changes, so your solution may be our only feasible option.

I must add that I gotten myself busy and don't know when I will be able to put more work into to this.

PR #7607, opened 2026-02-11
Fixes issues related to thumbnail generation firstpass

kaesa:
### The issues
The thumbnail firstpass generation has two issues:
1. The previewsFinished is not always called
2. Rare condition when the program  segmentation faults when closed while the first pass is still working
3. There is a race condition reading the jobs_empty() in PreviewLoader::processNextJob()

### Steps to reproduce 1
1. Open directory with sufficient amount of files
2. While the File Browser is populating, switch to a directory with small amount of files
4. Observe the status bar does not update correctly.

The small directory needs to be small enough. If there is sufficient amount of files, the bug doesn't trigger. I have made a screen capturing of the bug: https://www.youtube.com/watch?v=b1gy6C6yhZY

### Steps to reproduce 2
1. Delete files from _~/.cache/RawTherapee/images/_ but leave the _data_ directory alone. This will trick the first pass into generating fully processed thumbnails.
2. Open directory with sufficient amount of previously cached images with processing profiles (pp3)
3. Close RawTherapee before the File Browser populates to full

I haven't been able to trigger the segfault when the program creates quick thumbnails. It can be, that there is segfault only when the first pass generates fully processed thumbnails. I don't know the exact segfault mechanism.

### Bug description
Both issues are happening because the `void PreviewLoader::removeAllJobs()` is not waiting for the already started jobs to finish before returning.

### Solution
I copied the logic that waits for the background jobs to finish from the `ThumbImageUpdater::removeAllJobs`.

### Potential adverse effects
The solution shouldn't give much adverse effects. The adverse effects means users waiting for the program. The solution does add wait times when the current directory is closed while the File Browser is populating. This can happen when the user chooses another directory or toggles recursive directories. However, normally the first pass is creating quick thumbnails and the added delay should be minimal. When the delay becomes significant is, when the first pass is generating fully processed thumbnails. As far as I know, that is abnormal behavior from the program.

Also, currently I have not been able to trigger segfaults outside of the sequence to close the program. On the other hand, not waiting the first pass to finish before continuing while the first pass has the potential to segfault may become problematic in the future because of some unrelated changes to thumbnails or otherwise.

### DIFF
--- a/rtgui/previewloader.cc
+++ b/rtgui/previewloader.cc
@@ -141,15 +141,16 @@ public:
         if (--nConcurrentThreads == 0) {
             std::lock_guard<std::mutex> lock(mutex_);
 
-            if (!jobs_removed_ && jobs_.empty()) {
+            if (!jobs_removed_ && !nConcurrentThreads && jobs_.empty()) {
+                // re-check nConcurrentThreads under mutex because it is possible for another
+                // thread to race to take the last job from jobs_ and to avoid double call to
+                // previewsFinished when that happens
                 notifyListener = true;    
             }
             inactive_.notify_all();
         }
 
-        if (notifyListener && !nConcurrentThreads) {
-            // As long as the upper if (--nConcurrentThreads == 0) { is executed outside of mutex,
-            // it will always be a rare possibility that the previewsFinished is called twice
+        if (notifyListener) {
             DEBUG("Previews Finished\n");
             j.listener_->previewsFinished(j.dir_id_);
         }
--- a/rtgui/previewloader.cc
+++ b/rtgui/previewloader.cc
@@ -92,7 +92,7 @@ public:
     std::atomic<int> nConcurrentThreads;
 
     // Need to be a std::mutex because used in a std::condition_variable object...
-    // This is the only exception along with ThumbImageUpdater and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
+    // This is the only exception besides ThumbImageUpdater and GThreadMutex (guiutils.cc). MyMutex is used everywhere else
     std::mutex mutex_;
     bool jobs_removed_;
 
@@ -115,7 +115,7 @@ public:
             j = *jobs_.begin();
             jobs_.erase(jobs_.begin());
             DEBUG("processing %s", j.dir_entry_.c_str());
-            DEBUG("%d job(s) remaining", jobs_.size());
+            DEBUG("%ld job(s) remaining", jobs_.size());
 
             nConcurrentThreads++; // to detect when last thread in pool has run out
         }
@@ -147,7 +147,10 @@ public:
             inactive_.notify_all();
         }
 
-        if (notifyListener) {
+        if (notifyListener && !nConcurrentThreads) {
+            // As long as the upper if (--nConcurrentThreads == 0) { is executed outside of mutex,
+            // it will always be a rare possibility that the previewsFinished is called twice
+            DEBUG("Previews Finished\n");
             j.listener_->previewsFinished(j.dir_id_);
         }
     }
@@ -189,7 +192,7 @@ void PreviewLoader::add(int dir_id, const Glib::ustring& dir_entry, PreviewLoade
 
 void PreviewLoader::removeAllJobs()
 {
-    DEBUG("stop %d", impl_->nConcurrentThreads);
+    DEBUG("stop %d", impl_->nConcurrentThreads.load());
 
     std::unique_lock<std::mutex> lock(impl_->mutex_);
     impl_->jobs_.clear();
--- a/rtgui/previewloader.cc
+++ b/rtgui/previewloader.cc
@@ -17,13 +17,15 @@
  *  along with RawTherapee.  If not, see <https://www.gnu.org/licenses/>.
  */
 
+#include <mutex>
+#include <condition_variable>
+#include <atomic>
 #include <memory>
 #include <set>
 #include "cachemanager.h"
 #include "filebrowserentry.h"
 #include "previewloader.h"
 #include "guiutils.h"
-#include "threadutils.h"
 
 #ifdef _OPENMP
 #include <omp.h>
@@ -52,15 +54,7 @@ public:
         Glib::ustring dir_entry_;
         PreviewLoaderListener* listener_;
     };
-    /* Issue 2406
-        struct OutputJob
-        {
-            bool complete;
-            int dir_id;
-            PreviewLoaderListener* listener;
-            FileBrowserEntry* fdn;
-        };
-    */
+
     struct JobCompare {
         bool operator()(const Job& lhs, const Job& rhs) const
         {
@@ -74,7 +68,9 @@ public:
 
     typedef std::set<Job, JobCompare> JobSet;
 
-    Impl(): nConcurrentThreads(0)
+    Impl(): 
+        nConcurrentThreads(0),
+        jobs_removed_(false)
     {
 #ifdef _OPENMP
         int threadCount = omp_get_num_procs();
@@ -83,20 +79,31 @@ public:
 #endif
 
         threadPool_.reset(new Glib::ThreadPool(threadCount, 0));
+
+        if (App::get().options().rtSettings.verbose) {
+            printf("PreviewLoader::Impl pool thread count is %d\n", threadCount);
+            printf("PreviewLoader::Impl nConcurrentThreads is ");
+            printf(nConcurrentThreads.is_lock_free() ? "lock free\n" : "not lock free\n");
+        }
     }
 
     std::unique_ptr<Glib::ThreadPool> threadPool_;
-    MyMutex mutex_;
     JobSet jobs_;
-    gint nConcurrentThreads;
-// Issue 2406   std::vector<OutputJob *> output_;
+    std::atomic<int> nConcurrentThreads;
+
+    // Need to be a std::mutex because used in a std::condition_variable object...
+    // This is the only exception along with ThumbImageUpdater and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
+    std::mutex mutex_;
+    bool jobs_removed_;
+
+    std::condition_variable inactive_;
 
     void processNextJob()
     {
         Job j;
-// Issue 2406       OutputJob *oj;
+    
         {
-            MyMutex::MyLock lock(mutex_);
+            std::lock_guard<std::mutex> lock(mutex_);
 
             // nothing to do; could be jobs have been removed
             if ( jobs_.empty() ) {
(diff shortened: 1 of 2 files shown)


rtgui/previewloader.cc
```diff
@@ -135,30 +129,22 @@ class PreviewLoader::Impl :
...
-                            oj->listener->previewReady(oj->dir_id,oj->fdn);
-                        }
-                        output_.erase(output_.begin());
-                        delete oj;
-                    }
-                }
-        */
         bool last = g_atomic_int_dec_and_test (&nConcurrentThreads);
 
+        if ( --active_ == 0 ) {
```
Lawrence37: This comparison to a shared value is used in a conditional without a lock. Could there be a race condition?

rtgui/previewloader.cc
```diff
@@ -135,30 +129,22 @@ class PreviewLoader::Impl :
...
+            std::lock_guard<std::mutex> lock(mutex_);
+            if (inactive_waiting_) {
+                inactive_waiting_ = false;
+                inactive_.notify_all();
+                // Jobs are being removed, don't report finished state
+                return;
```
Lawrence37: Can you help me understand the logic here? `active_` is `0` and `inactive_waiting_` is `true`, which mean this happens when there are no more running jobs and `removeAllJobs` has been called. So we don't notify listeners that the previews are finished once the last job is done? When is the listener notified then?

rtgui/previewloader.cc
kaesa: I think there is. It needs addressing. It is a problem only when the jobs are cancelled, but I think the problem sequence exists.

rtgui/previewloader.cc
kaesa: I think it is a logical error to claim the listener the previews finished when in reality they were cancelled. The only reason we are here because we can't kill already started jobs so we need to wait and as a side effect we get an excess `previewsFinished` call.

The `PreviewLoader` is abstracted half way and it is unnecessarily a singleton, but in reality it is tightly coupled with the `FileCatalog`. The potential problem calling the `previewsFinished` is that the `FileCatalog` uses that as a signal to launch `ThumbImageUpdater` for every image in the File Browser. This is not currently an issue as evident by the bug, because the `dir_id_` would be wrong and because the interface functions in the `FileCatalog` give lower priority to the events spawned from the `previewsFinished`. But my thinking is that it is just safer to not do the call at all. Of course, if the interface is thought as something from a truly independent module, I agree the listener would need to be informed so it can do necessary clean up or other logic. I think this could be done by adding a parameter to the `previewsFinished`. However, because the listener is the one who cancels the previews in the first place and also it wouldn't use that information, I think it would be just an unnecessary change.

For `PreviewLoader` to support more than one user would require more work and I think changes to the interface could be done then. For example, the `previewsFinished` is called when the entire job queue is empty to the listener who just happens to be associated to the last job in the queue.

rtgui/previewloader.cc
```diff
@@ -202,8 +188,19 @@ void PreviewLoader::add(int dir_id, const Glib::ustring& dir_entry, PreviewLoade
...
+    }
+
+    while ( impl_->active_ != 0 ) {
+        DEBUG("waiting for running jobs2");
+        {
+            std::unique_lock<std::mutex> lock(impl_->mutex_);
+            impl_->inactive_waiting_ = true;
+            impl_->inactive_.wait(lock);
+        }
+    }
```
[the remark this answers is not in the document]
kaesa: Needs addressing.

rtgui/previewloader.cc
Lawrence37: I see. Renaming `previewsFinished` to prevent accidental misuse should suffice.

kaesa:
I reworked and rebased to keep it cleaner. I think the review needs to start from scratch anyway. Now this also fixes a race condition with `jobs_.empty()` call.

kaesa:
@digitalcarp if you don't mind would you be willing to take a look if you see something wrong.

This is still not thread safe in the absolute sense. If there would be more clients than just FileCatalog, it would be possible to add to the jobs while the jobs are being removed. This is not a real concern because there is only one client who does all the calling. Making `PreviewLoader` support multiple clients would require considerable changes also in other parts and there is no requests to support other clients as far as I know, so I thought this is a good stopping point for now. Also, there is no design how to handle the case when adding to the jobs fails.

rtgui/previewloader.cc
```diff
@@ -86,17 +81,24 @@ class PreviewLoader::Impl :
...
-// Issue 2406   std::vector<OutputJob *> output_;
+
+    // Need to be a std::mutex because used in a std::condition_variable object...
+    // This is the only exceptions along with ThumbImageUpdater and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
+    std::mutex mutex_;
+    std::atomic<unsigned int> active_;
+
+    bool inactive_waiting_;
+
+    std::condition_variable inactive_;
```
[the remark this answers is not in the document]
kaesa: Fixed

rtgui/previewloader.cc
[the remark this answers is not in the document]
kaesa: Fixed

rtgui/previewloader.cc
```diff
@@ -86,17 +81,24 @@ class PreviewLoader::Impl :
     }
 
     std::unique_ptr<Glib::ThreadPool> threadPool_;
-    MyMutex mutex_;
     JobSet jobs_;
     gint nConcurrentThreads;
-// Issue 2406   std::vector<OutputJob *> output_;
+
+    // Need to be a std::mutex because used in a std::condition_variable object...
+    // This is the only exceptions along with ThumbImageUpdater and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
```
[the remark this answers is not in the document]
kaesa: Fixed I think. I think the plural exceptions was the real problem.

rtgui/thumbimageupdater.cc
```diff
@@ -90,7 +90,7 @@ class ThumbImageUpdater::Impl :
     std::unique_ptr<Glib::ThreadPool> threadPool_;
 
     // Need to be a std::mutex because used in a std::condition_variable object...
-    // This is the only exceptions along with GThreadMutex (guiutils.cc), MyMutex is used everywhere else
+    // This is the only exceptions along with PreviewLoader and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
```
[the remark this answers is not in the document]
kaesa: Fixed I think. I think the plural exceptions was the real problem.

rtgui/previewloader.cc
```diff
@@ -202,8 +190,17 @@ void PreviewLoader::add(int dir_id, const Glib::ustring& dir_entry, PreviewLoade
...
-    MyMutex::MyLock lock(impl_->mutex_);
+
+    std::unique_lock<std::mutex> lock(impl_->mutex_);
     impl_->jobs_.clear();
+
+    if (impl_->nConcurrentThreads.load() != 0) {
+        DEBUG("waiting for running jobs2");
+
+        impl_->jobs_removed_ = true;
+        impl_->inactive_.wait(lock, [this] { return impl_->nConcurrentThreads.load() == 0; });
```
digitalcarp, line 205: Important: Doing a quick read of the call stack, waiting on the condition variable gets triggered every time a user selects a different directory. If `removeAllJobs()` is called from the main UI thread, the GUI will be frozen. The previous behaviour would clear the jobs queue/vector and not block the UI thread for too long.

I'm thinking of the case where a user selects a new directory with lots of images. It spawns many image processing jobs. The user quickly/immediately selects a different directory. The UI is frozen until all the previous jobs are complete.

kaesa: > The previous behaviour would clear the jobs queue/vector and not block the UI thread for too long.

It still does that.

But you are right, the freeze time will increase for some cases. But its not because waiting for all the jobs, it's only for the already started jobs. I'm a little bit concerned that the max threads is not set for the ThreadPool and the documentation has no mentioning how smart the ThreadPool is about it, but on my testing the waits were almost non-existent with non-cached directories. The big problem for users is the `ThumbImageUpdater`. Waiting for even just a few processing tasks to finish takes a long time. Luckily, they won't both block. It is always only one or the other.

kaesa: I stand corrected, the max threads is actually set. If OpenMP is enabled, it is the number of cores and otherwise it is 2. It is set in the constructor of the ThreadPool. My mistake for not looking into this. So the wait time will be from 2 to the core count of jobs to finish before returning.

rtgui/thumbimageupdater.cc
```diff
@@ -90,7 +90,7 @@ class ThumbImageUpdater::Impl :
     std::unique_ptr<Glib::ThreadPool> threadPool_;
 
     // Need to be a std::mutex because used in a std::condition_variable object...
-    // This is the only exceptions along with GThreadMutex (guiutils.cc), MyMutex is used everywhere else
+    // This is the only exception along with PreviewLoader and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
```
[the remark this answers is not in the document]
Lawrence37: The Copilot suggestion is wrong. It's better off being
```suggestion
    // This is the only exception besides PreviewLoader and GThreadMutex (guiutils.cc). MyMutex is used everywhere else
```

rtgui/previewloader.cc
```diff
@@ -83,20 +79,31 @@ class PreviewLoader::Impl :
...
 
     std::unique_ptr<Glib::ThreadPool> threadPool_;
-    MyMutex mutex_;
     JobSet jobs_;
-    gint nConcurrentThreads;
-// Issue 2406   std::vector<OutputJob *> output_;
+    std::atomic<int> nConcurrentThreads;
+
+    // Need to be a std::mutex because used in a std::condition_variable object...
+    // This is the only exception along with ThumbImageUpdater and GThreadMutex (guiutils.cc), MyMutex is used everywhere else
```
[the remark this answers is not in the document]
Lawrence37: Same here.
```suggestion
    // This is the only exception besides ThumbImageUpdater and GThreadMutex (guiutils.cc). MyMutex is used everywhere else
```

rtgui/previewloader.cc
Lawrence37: I think with the latest code, the only problem is `previewsFinished` may get called twice during normal completion of jobs.

When removing all jobs, the job queue is empty, so no other thread will increment `nConcurrentThreads`. `nConcurrentThreads` will only ever decrease, making the condition true only once.

kaesa: You're right.

After a thread has decremented to zero while there is still one job left in the queue, there is enough time for another thread to get the last job under mutex before the thread that decremented to zero checks the `jobs_` for empty. Resulting the previewsFinished called twice. I added a second check for the `nConcurrentThreads` that makes the rare situation more rare. But I don't think it is possible to completely eliminate the possibility without decrementing and testing under mutex. Calling the `previewsFinished` multiple times is not dangerous though. The `ThumbImageUpdater` is reasonably efficient at rejecting tasks for images already in queue and tasks on images that have already been processed don't actually do anything.

I added a comment so that future readers realize there is this problem before doing anything where it would become a bigger problem.

kaesa: I take previous back. I was being stupid, sorry. I think it is now safe. The `nConcurrentThreads` is incremented and `jobs_` is being emptied under same mutex. Re-checking the `nConcurrentThreads` with `jobs_` simultaneously under mutex prevents double call.

Lawrence37 (approved):

PR #7609, opened 2026-02-22
Selective Editing - GHS crash when Black point is zero or negative,  user activate "Auto Black Point & White Point"

Desmis:
This is an old (if one can say that for GHS) problem. GHS is complex.

In common and simple cases, Michaelis-Menten can solve problems, and its simplified interface is more intuitive. However, in complex cases, GHS is more efficient.

GHS allows you to adjust the BP (linear) and WP (linear) values. It also allows you to use negative values ​​(which GHS does not support), to provide additional shadow detailing.

If the user, after trying negative BP values ​​(for example -0.05) finds that it does not make a difference, if he clicks on 'Auto Black point & White point', the system crashes.

I spent a lot of time trying to figure out why, and how to solve it.

Why :
* when the Stretch factor (D) values ​​are very low and the Black Point is zero, GHS gives a near-infinite value for rare pixels, which causes the 'rgb2Lab' conversion to crash.

How: 
* I have given up on any GUI solution, because there is already a loop management in the 'events' which is not simple....and I am almost certain that it would not solve anything.
* GHS (in iplocallab.cc) uses 3 loops to calculate linear BP, linear WP, Symmetry point (SP) by passing values ​​to the GUI multiple times (improcoordinator.cc, locallab.cc, rtengine.h).
* Thinking that 'activating this function only when "Auto BP & WP" is activated does not work, certainly we get through the first loop, but not the others.
* Another aspect to consider is that there may then be a 2nd GHS Spot (normal or Inverse), and the data will be close to the interval [0 1]. In fact often near of 1.1 because I left a little margin to prevent GHS from clipping the data.

I completely solved the problem (there may be another solution, but this one works), by adding just before the RGB=>Lab conversion, a limitation of the RGB values ​​to about 1.15 times the maximum (65535).

constexpr float clipRplus(float x)
{
    return rtengine::LIM(x, 0.00001f, 75500.f);//used for GHS if user set BP to 0 or negative - about 1.15 maxi
}

This value of approximately 1.15 times (75500 instead of 65535) ensures that the reading remains slightly above the maximum values ​​of the second GHS spot. Therefore, there are no adverse effects.

There is no difference before/after the modification, or if the user activates 'auto BP & WP' after trying negative BP values:
* in TIFF layers / differences  16 bits.
* for Linear BP, linear WP.
* for Symmetry point (SP).

Jacques


Desmis:
@kaesa 

Can you take a look at what I've done? In terms of code, it's very simple.

Thank you

Jacques

rtengine/iplocallab.cc
```diff
@@ -18482,7 +18486,8 @@ void ImProcFunctions::Lab_Local(
     
     
 }
```
kaesa: This was not part of this PR but the indent with the above closing brace is badly off. It should be at the same level as the bool and it has two lines indented by 4 spaces above it that there shouldn't be.

rtengine/iplocallab.cc
```diff
@@ -18482,7 +18486,8 @@ void ImProcFunctions::Lab_Local(
...
-                        if(params->locallab.spots.at(sp).ghs_autobw == true  && strtype == GHTStrType::NORMAL) { //find probably White point and black point ...Must be adjusted manually in soma cases notably Black point with negatives values...                        
+                        bool autobw = params->locallab.spots.at(sp).ghs_autobw;
```
kaesa: const bool = params->locallab.spots.at(sp).ghs_autobw;

rtengine/iplocallab.cc
```diff
@@ -18482,7 +18486,8 @@ void ImProcFunctions::Lab_Local(
...
+                        if( autobw == true  && strtype == GHTStrType::NORMAL) { //find probably White point and black point ...Must be adjusted manually in soma cases notably Black point with negatives values...
```
kaesa: Missing space after `if` extra space after `(`
Should be:
`if (autobw`
to match coding style elsewhere in this source file.

rtengine/iplocallab.cc

kaesa:
I found couple nitpicks. I didn't find anything actually wrong.

rtengine/iplocallab.cc

rtengine/iplocallab.cc

Desmis:
@kaesa 

OK, thank you very much...
I think this solves a problem that would have happened sooner or later.

Jacques

rtengine/iplocallab.cc
```diff
@@ -109,6 +109,10 @@ constexpr float clipR(float x)
     return rtengine::LIM(x, 0.f, 65535.f);//used when Laplacian Contrast attenuator
 }
 
+constexpr float clipRplus(float x)
```
Lawrence37: Rename for clarity and accuracy, such as `clipGhsRgbOutput`

Lawrence37 (commented):
Do you have a sample image and pp3 that demonstrates the crash? How does GHS result in infinite values? I thought the GHS function doesn't output infinity?

rtengine/iplocallab.cc
```diff
@@ -18481,8 +18485,10 @@ void ImProcFunctions::Lab_Local(
                             }
     
     
-}
-                        if(params->locallab.spots.at(sp).ghs_autobw == true  && strtype == GHTStrType::NORMAL) { //find probably White point and black point ...Must be adjusted manually in soma cases notably Black point with negatives values...                        
+                        }
+
+                        const bool autobw = params->locallab.spots.at(sp).ghs_autobw;
+                        if (autobw == true  && strtype == GHTStrType::NORMAL) { //find probably White point and black point ...Must be adjusted manually in soma cases notably Black point with negatives values...
```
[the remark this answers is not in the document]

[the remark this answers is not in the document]
Desmis: See below

rtengine/iplocallab.cc
```diff
@@ -19012,7 +19018,19 @@ void ImProcFunctions::Lab_Local(
                         if(ghs3sig > lp.maxdataghs / 65535.f) {//if the distribution is not Gaussian, then we take for 3.5 sigmas the real maximum.
                             ghs3sig = lp.maxdataghs / 65535.f;
                         }
-
+// This additional procedure allows 'rgb2lab' to function. In rare cases, when the Stretch function (D) is very small, if the black point is zero or negative, and the user activates "Auto Black Point & White Point", the GHS algorithm generates values ​​close to infinity.
+// This causes a crash during the RGB to Lab conversion.
+// GHS is complex and works in several 'passes'. This solution resolves the problems without disrupting any of the calculations and has no influence on the (visible) image or in 16-bit layer/differences.
+#ifdef _OPENMP
```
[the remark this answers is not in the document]
Desmis: I've done a lot of testing. There's no point in clamping it earlier; it's actually detrimental. The goal is to allow the execution of an additional RT-spot without any issues, which can reach a value of 1.1 instead of 1.

rtengine/iplocallab.cc
```diff
@@ -109,6 +109,10 @@ constexpr float clipR(float x)
     return rtengine::LIM(x, 0.f, 65535.f);//used when Laplacian Contrast attenuator
 }
 
+constexpr float clipRplus(float x)
+{
+    return rtengine::LIM(x, 0.00001f, 75500.f);//used for GHS if user set BP to 0 or negative - about 1.15 maxi
+}
```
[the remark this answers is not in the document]
Desmis: I replace clirPlus by clipGhsRgbOutput.
I don't see the point in adding an extra constant that complicates the code. The value of 75500 (after many attempts) is sufficient. The comment explains why, just in case.

rtengine/iplocallab.cc
```diff
@@ -19012,7 +19018,19 @@ void ImProcFunctions::Lab_Local(
...
+// GHS is complex and works in several 'passes'. This solution resolves the problems without disrupting any of the calculations and has no influence on the (visible) image or in 16-bit layer/differences.
+#ifdef _OPENMP
+            #pragma omp parallel for if (multiThread)
+#endif                       
+                        for (int i = 0; i < bfh; ++i)
+                            for (int j = 0; j < bfw; ++j) {//avoid crash when user set BP to 0 or < 0 , and enable auto BP WP.
+                                tmpImage->r(i, j) = clipRplus(tmpImage->r(i, j));//clipRplus keep data about 1.15 maximum (65535) about 75500. No difference with or without on TIFF layers differences.
+                                tmpImage->g(i, j) = clipRplus(tmpImage->g(i, j));//There are no differences in the calculation of BP (linear) or WP (linear), nor of the Symmetry Point (SP). There is no influence on the subsequent GHS Spot.
+                                tmpImage->b(i, j) = clipRplus(tmpImage->b(i, j));//1.15 : more than enough to be within the limits of a second RT-spot.
+                            }
```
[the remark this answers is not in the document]
Desmis: The problem only appears during the RGB-to-Lab conversion... and it must be done just before to avoid interfering with the other RT-Spots. I don't know exactly why we're getting out-of-range values, but we're starting with a Black point  value close to or at zero or negative, and a very low stretch factor (D) (0.001) leads to this. For example, Reducing the minimum value of Stretch factor (D) below 0.001 results in malfunctions and crashes. When I implemented GHS, I found this value of 0.001, which seems to be a minimum and compatible with the calculation of BP and WP. There is a very small, almost negligible influence. This mode of operation is essential if you want to perform calculations in real time, unlike calculations done 'beforehand' such as Black-Ev and White-Ev.
I've tried many things... I've been trying to solve this bug for a long time. All attempts to limit it to just "auto BP & WP" fail because there are multiple calculation loops.

rtengine/iplocallab.cc

Desmis:
@Lawrence37 
As I explain in the lengthy introduction, which is why I wrote such a long explanation, this applies to any image. Since the user can choose negative values for BP or Stretch factor ​​that GHS doesn't support, a workaround was necessary.

So, if the user wanted to enhance the shadows before GHS, for example by setting BP=-0.05, and then realizes that this doesn't produce the desired result and clicks on "Adjust BP & WP," the system crashes. Of course, we could tell them to manually reset BP to > 0.

I think the solution works correctly. Putting this control (clamp) further upstream in the code not only serves no purpose, but is harmful.

I'm not sure we'll end up with infinite values, but the fact remains that the rgblab function fails and not for negatives values.

Rather than trying (and I did try) to prevent these values... (which must be due to the combination of a very low BP and a very low stretch factor), I started from what I'm trying to achieve.
For a second spot, the white point value needs to be slightly greater than 1.1. Why 1.1? Because I added a value of 0.1 to the WP to give GHS some leeway and prevent it from clipping the data immediately (as GHS does), and data are divided par WP. However, it's ultimately the designer's responsibility to ensure the data is within the range [0, 1] for GHS.
There's no point in using excessively high values, because for 'Inverse GHS' the data is clipped, and for the second spot, it should ideally be greater than 1.1 (just  a little).

I tried with 1.05 around 69000. Everything worked correctly, but in 16-bit layer/differences (Photoshop), I had a few rare (invisible) pixels with tiny values. So 1.1 is indeed an important criterion... (if you can call it that). The value of 75500 is sufficient and has no impact on either the 16-bit layer/difference or any of the BP and WP calculations.

You might say that it's not very 'computer-related', but it now works regardless of the images I've tested. Not everything is rational; for example, CIECAM is anything but rational.
* white point )= 0.9 to 1.1 (standard)
* white point = 1.5 (slightly overexposed)
* white point = 0.7 (under exposed)
* white point = 4 (sunset)
* white point = 11 (LEDs)

This also works for the Black Point:
* negative values and 'auto BP WP' after.
* negative values without 'auto BP WP'.
* BP = 0.001.
* BP = 0.12 - which needs to be reset close to zero to restore contrast. This is something that neither Filmic, nor Log encoding, nor Sigmoid (with Black EV) can do (easily).

This problem doesn't exist with Michaelis-Menten. 
* Firstly, because I wanted two algorithms: a simple MM with few settings, and a more efficient but more complex GHS.
* Secondly, because the hyperbolic function does not strictly need to be between 0 and 1 (although it is highly desirable).

But this bug would have appeared sooner or later... Okay, you have to want it... but it happens.

Jacques

kaesa:
@desmis I haven't been able to reproduce the crash either.

Desmis:
@Lawrence37 @kaesa 

I forgot an important detail.
The crash occurs **in all cases under Windows 11**, and **very rarely under Linux**. I don't know about macOS.

The problem seems to be due to (it's very difficult to say) rounding errors in calculations or a difference in accuracy in extreme cases.

That's why I insisted on having two systems on my machine, to test in different OS configurations. 

It may be due to the type of processor (I5 - 12500H) or SSE calculations, or the BIOS... In any case, under Windows on my machine, in 99% of cases where I trace the situation as indicated, there is a 'segmentation fault'.

This is not the first time I have seen these 'small' differences. In particular, during the development of White Balance auto-temperature correlation, the correlation coefficients and patch composition were (very slightly) different, without any noticeable impact on the result (or the 3rd digit after the decimal point).

Similarly, in GHS, for images with very high WP values, Linux and Windows show small differences. For example:
* Linear WP - Linux : 4.2056
* Linear WP - Windows : 4.2058
This has virtually no impact on the apparent result.

There are other differences, but these are related to the GUI libraries or fonts. The Windows and Ubuntu interfaces look similar, but are different.

As a reminder, I primarily do my development work under Windows, because I use Notepad++ (which is very good and which I've been using for over 15 years) and Photoshop, which I find more intuitive than GIMP, especially for 16-bit Layers/Differences. Then I check under Linux Ubuntu that everything works in a similar (though not completely identical) way.

Jacques

kaesa:
I wasn't able to crash with my images on Windows 10. It would be great if there was a test image with a pp3 just needing to click the "Auto Black point & White point" to trigger the crash.

Desmis:
@Lawrence37 @kaesa 
I think the problem is solved, either:
* by the latest Windows 11 update (which took 2 hours to install) Windows 11 Insider Preview Feature Update (26220.7934)
* either by updating Msys2/mingw64.

Yesterday morning, after I had updated Msys2/Mingw64 last week, there was a **systematic crash** regardless of the image. This happened as soon as the procedure I mentioned was followed (The user want to try negatives values for black point, then changed his mind to "auto BP & WP".

So, yesterday I updated Windows 11 (it was difficult, I think the longest update and one that often fails from what I've seen on the web), I also updated Msys2/Mingw64 again. It all ended late last night.

I tried it this morning, both on an already open branch (redgreenblue) and on a new clone. The crash disappeared... That's an observation, but I don't know what was causing this systematic crash (Windows 11 or Msys2 ?).

I suggest we don't 'merge' this PR but put it in **draft**, in case the problem comes back.

But when I think about the time I spent on it... impressive. The solution provided has only one (minor) drawback: it slightly increases the processing time.

Jacques

kaesa:
Converting to draft sounds reasonable. If the issue comes back then we have this as a starting point and we can try to verify the problem.

Desmis:
@kaesa @Lawrence37 

I don't know exactly what to do? The Windows11 and MsyS2 update has certainly resolved the systematic crash, but it still appears, much rarer and erratic.

My grandson, who has a completely different machine with an AMD Raizen 7-7736U processor instead of an Intel i5-12500H, tried both versions: the 'dev' version and the 'ghscrash' version (which I updated with merge dev to allow for proper comparisons). Furthermore, it has a Linux (Debian) operating system instead of Windows 11 or Ubuntu 24.04.

I'm providing a link with an image and a pp3 file that results in crashes.
These links are in the new RawPedia Game Changer
https://rawpedia.pixls.us/tutorials/#game-changer---how-to-process-a-sunset

* Choose 'dev'.
* Download the Raw image and the two PP3 files with GHS.
* Choose one of the two pp3. Then 'Reset Stretch factor (D)' ( = 0.001), for the **first** Spot
*  Unchecked 'Auto Black point & White point'
* Move the Black point (linear) slider between -0.2 and 1 (sometimes it crashes..) . And reactivate 'Auto Black point & White point'. Repeat these operations several times (via 'neutral' or not).

* Debian + AMD Raizen : The crash occurs 3 to 4 times out of 10 attempts.
* Ubuntu + Intel (my machine): The crash is random
* Windows11 (my machine) : The crash is random.

With the 'ghscrash' patch we never experienced a crash (which does not mean that the problem is completely solved). whether under Windows (Intel), Ubuntu (Intel), or on the other machine : Debian - (AMD).

For your information.

Jacques

kaesa:
I can crash RT by switching between the three spots and clicking the auto BP checkbox in really fast speed. But I think that crash is happening because of the way the GHS is implemented. It has other issues too. Just by selecting the different spots will changes the image in arbitrary ways. This is what I reported before too. It could be that the fix you propose in this PR just slows down the execution in your computer enough to not crash because of an unrelated reason.

Desmis:
@kaesa 
I must admit I don't know exactly what's happening, mainly because the system performs three loops, specifically to calculate the slider positions and incorporate their new values.

Under extreme conditions, the system is also sensitive to the WP.  Crashes (very rare) occur only in exceptional configurations. However, we can't control what a user might do.

It might be a slowdown, but I don't really believe that. The same function with Pragma OMP, just before RGBLAB, is only effective if you reduce the maximum values ​​(CLIP). 

I spent a significant amount of time on this. In the case of Windows 11 (before the last system update), it was "easy" to find "where" it was happening... because every single image was affected.

I remained very pragmatic (the Debug version didn't offer anything extra).

I isolated the code piece by piece, both in iplocallab.cc, improccordinator.cc, and dcrop.cc, as well as in the GUI functions - locallab.cc, locallabtools.cc, and rtengine.h. 

It is an understatement to say that the code is complex (even outside of GHS), but with GHS the fact that I wanted - and it seems essential to me - to do the calculations of BP, WP, SP in real time, has considerably complicated the problem. It's so simple to do what Log encoding, Filmic, and Sigmoid do: either not calculate these values ​​at all, or do it beforehand. We don't run into any problems, but the differences are "enormous"...at least that's how I perceive it, and what my grandson also saw when I told him, "Look at the nuances of the sun" with setting X or with software Y....(for example try "Inpaint opposed' and "Color Propagation"...).

I started with the minimum I wanted to achieve: enabling a second RT-spot under GHS to function. Therefore, it must be within the range [0 + epsilon, 1.1] at a minimum. There's absolutely no point in going any further.

I was ready to abandon this modification, but my grandson, who has a different machine, installed RT (under Debian) with Libraw=OFF  (He's on holiday at his grandpa's for a whole week).  As soon as he tried the modifications I described - which are anything but intuitive - namely, manually modifying the BP and WP and re-enabling 'auto', the crash occurred, not systematically, but many times. Much more frequently than on my machine running Windows 11 or Ubuntu.

He compiled the 'ghscrash' version and there are no more crashes. And I reiterate, the apparent settings are identical, and under Windows 11 the differences in TIFF files (with Photoshop) are invisible, and the discrepancies are zero. So my reasoning of limiting the maximum to around 1.15 seems to work.

My proposal is therefore to merge this solution, even if it might seem strange (unusual) to purists. But if we must delve into the "unusual," look at the code for "WB temperature correlation," or that of CIECAM (adapted to RT), or that of the Selective Editing parameters (for deltaE and transitions...then it's even stranger). My background as a manager, head of very large projects, scientist, and sociologist specializing in human factors, has taught me modesty and humility.

Jacques

kaesa:
It's not about purism.

I think there are cross dependencies that come up when the program gets stressed. I think this PR could very like be just a whack-a-mole bug fixing and the real issue remains unsolved. If you clip the values just before using them, you could be lessening the issues that come from other parts of the code modifying the same values at the same time. But at the same time, you mess the results of the other concurrent execution that is trying to use the values. I think since the fix is about modifying the values before use, this crash should be reproducible without any randomness and without stressing the system.

Can you confirm you can crash the program using only one GHS spot?

kaesa:
I have confirmed I can still crash the program with this fix. I used the standard pp3. First I just click the different spots in quick succession without modifying the values and the image gets completely messed up. Then I reset the stretch of all spots and start selecting the two first spots (not the inverse spot as it doesn't allow it without unchecking the inverse option) and clicking the Auto BP checkbox on and off without waiting for the program to process the image. After a short time the program segfaulted.

Desmis:
I didn't say this PR solved everything...far from it. But the algorithm is complex. I emphasize the real-time calculation aspect. Overcoming this aspect resolves all the problems, because GHS itself clips the data (between 0 and 1). 

The major problem ignored in many algorithms, including gamut compression (I'm only talking about RT), is the assumption that out-of-gamut values, such as sunset or LEDs, are outside the correct CIExy diagram. This problem is approached with a huge cognitive bias. What's most often out of gamut isn't the chroma, but the Y value of xyY. 
Just because you're within the CIExy diagram doesn't mean you're within the gamut. Sure, if you're outside the gamut, you're out of gamut, but for small RGB values, the gamut is very narrow, and for very high RGB values, you're also out of gamut.

For example, a sunset (with Color Propagation) results in values ​​around 4 (which corresponds to approximately 262,000 on the 65535 scale), while for an LED image, the values ​​are close to 11, or 720,000. If you use another highlighting algorithm, such as Inpaint Opposed, the maximum values ​​are halved or more. If you don't use any highlight reconstruction algorithm, the values ​​to consider are always less than 2 (the range 0-1).

The same type of problem arises with black points.

It's clear that I've looked at the minimum and maximum values ​​throughout the algorithm, but it's difficult to draw a clear conclusion. Just before rgb2lab, it's rare to exceed 1.1, but in the 'return' it crashes. The simplest solution, but not the best in terms of results, is to ignore the calculation of BP and WP and let it run. You won't have any bugs or crashes... That's what the GHS algorithm does 'by default'.

When Copilot suggested "upgrading" the clip, that would be a mistake, but it doesn't mean everything is fine.

So in summary, I see no problem either not pushing this PR or the opposite. But I think it's better to do it. What I don't understand is why a (very important) Windows 11 update solved 99% of the problem.

Jacques

kaesa:
> So in summary, I see no problem either not pushing this PR or the opposite. But I think it's better to do it. What I don't understand is why a (very important) Windows 11 update solved 99% of the problem.

I'm pretty sure it didn't solve or influence the problem in any way. Only thing that it changed was to the execution environment and now you are experiencing the effects of the problem less.

I don't think the problem is necessarily the calculations on their own, but it looks like a concurrency issue with data race or something to that line. The mere fact that users are not allowed to even accidentally select the spots that have the GHS tool in selective editing after the result looks fine, tells me that there are logic problems in the code.

I don't believe this PR addresses the problem. I have already confirmed the program keeps crashing with this PR applied. Even if we ignore that the problem remains without a fix and pushing this PR feels harmless. Merging may produce surprising problems in the future when more changes are done.

Desmis:
I agree that this solution included in the PR does not solve everything.

I looked back at my 'ramblings' and realized that I wanted to give GHS some leeway, with the variable WP_LINEAR_FREE which I arbitrarily set to 0.1, meaning that the white point which by calculation was for example at 3.8, becomes 3.9. Unfortunately I think this is poorly reflected in the rest of the code, and it is perhaps in wanting to 'do well' that I have done harm.

I just set this value to zero (as it was before), which isn't a problem for compatibility since this (auto) setting didn't exist in 5.12. And I cancelled the Clip before rgb2lab, by adding comments.

I think it works better this way. It will be up to the user to fine-tune the WP and BP (if 'auto' isn't suitable).

But I'm not saying that this solves everything and that everything will be fine.

Desmis:
I think that, unlike the previous modification (with Clip), this one absolutely must be done (don't add any value to the calculated WP). Of course, I would remove all the Clip code before rgb2lab.

Desmis:
By painstakingly tracking down the (rare) crashes, I made several modifications to avoid negative XYZ or RGB values, or to limit the maximum value - which doesn't seem to have any impact on the results.

In addition, I changed some default values: 'ghs_LC' = 0 or max 'ghs_BLP' = 0.2

I think the number of crashes has been significantly reduced.... After numerous attempts (under Windows 11), I no longer have it. I will try on Linux Ubuntu.

There don't seem to be any differences in the results - except for the one that affects the WP calculation  (remove of WP_LINEAR_FREE), which I think is better now.

///

I also tested it under Linux Ubuntu, and I didn't find any crashes there either... which doesn't mean there aren't any at all.
The sunset file is particularly problematic (when trying to use 3 RT-spot GHS, including 1 inverse), partly because the sun is out of gamut (4 times) and other colors are out of sRGB gamut (see gamut compression).

I also tried with a very difficult image
https://rawpedia.pixls.us/tutorials/#game-changer-using-leds-image
The white point  > 11 . Gamut compression  does a great job, the "maximum achromatic value = 12.2"

But I cannot guarantee that the system cannot crash

Jacques

kaesa:
I'm trying to understand the locallab. I don't. But on my testing I have come to a suspicion that the root of the problems is that the `void ImProcCoordinator::updatePreviewImage(int todo, bool panningRelatedChange)` is updating the UI. It is my understanding that this function is running as a worker and in reality trying to create the preview image.

I don't understand why there is this kind of loop back to the UI. I noticed it is done with some other tools too. However, I think the key to the proper fix would be to first find the right place to set the slider values. There seems to be some loop back to start the preview processing again in `panelChanged`, which is called three times per selected spot. That is also the source of the History spam. The variable names and constant forwarding doesn't really help me making sure I'm understanding the code.

Nevertheless, I think that making calls into that rabbit hole while already working on a preview should be avoided.

kaesa:
I realize my text above wasn't clear reading to the end. What I'm saying, that if we can avoid calls to the `listener->panelChanged`  while calling back to the UI, the issues could lessen.

Desmis:
@kaesa @Lawrence37 

The problem is complex because, depending on the algorithms, calls are sometimes made from 'dcrop.cc'. I'm not the author of the locallab GUI (I'm not shirking responsibility, but I wouldn't have been able to do that).

Yes, I think there are improvements to be made, but I don't think this is the right time,  the time to decide the fate of a Pull Request.

The complexity stems from several things that the user doesn't see:
* a number of Spots which can be very large (beyond normal needs)
* 4 types of spot : with deltaE a) normal; b) excluding; c) full image; or without deltaE - Global.
* 3 levels of complexity : basic, standard, advanced
* at least as many tools, if not more, than RT-main... 
* etc.

The architecture consists of controlspotpanel, locallab, and locallabtools (2) for the GUI. 

The interaction with params (procarams) is complex.
The interaction with iplocallab.cc is complex.

The paradox is that it is generally more efficient, because there is roughly only 1 program with the algorithms (iplocallab) instead of 4 for 'main' (examples: Wavelet, Ciecam... much 'simpler' than 'main')

The first thing to do (from my point of view) is to revise the Selective Editing GUI to make it more intuitive (this was planned... and I think it's already started, though certainly not by me, I'm incapable of doing that). But this would cause a compatibility issue.

But I am open to any proposal, short term (1 or 2 days), medium term (1 or 2 months), long term (more than a year)...If my health and age allow it.

Thank you for all your contributions

Jacques

kaesa:
I'm not talking about making improvements. I'm talking about fixing the problem this PR is trying to address. When you call the `listener->panelChanged`, what exactly you're trying to achieve with it? Maybe it can be done in another way.

kaesa:
Am I correct that when you click on the "Auto BP" checkbox you want the preview generated twice because the first preview is not the final one?

kaesa:
Ok, I see the automatic value calculation is embedded tightly into the preview generation. The only quick and dirty solution I can think of now would be to generate a special preview with a loop back only when the user enables the checkbox and generating the real preview always otherwise. The current implementation will farm a ton of bug reports because the preview keeps changing arbitrarily even though the user is not doing anything to change it.

rtengine/iplocallab.cc
```diff
@@ -18326,6 +18326,7 @@ void ImProcFunctions::Lab_Local(
                         bool ghsautoSP = params->locallab.spots.at(sp).SPAutoRadius;
 
                         const ght_compute_params c = GHT_setup(B, D, LP, SP, HP, strtype);//setup system with entries
+                        const float epsilg = 0.00001f;
```
kaesa: `constexpr`

kaesa:
I fixed some of the problems. I'm not going to make a PR about these unless you want me to. I thought you could copy the changes test and verify.
https://github.com/RawTherapee/RawTherapee/compare/dev...kaesa:RawTherapee:ghslook
These changes fix the arbitrary changes the image gets. These won't fix the full problem.

One of the reasons I'm doubtful that the crash is happening because of the arithmetic is that in my mind the program should crash just the same if the user adjusts the sliders manually. I've started to think you might have inadvertently discovered a way to stress test the preview generation to reveal some bug deep under.

I'll add couple comments on your new changes but else I don't have much else to say about this PR.

rtengine/iplocallab.cc
```diff
@@ -18944,14 +18945,15 @@ void ImProcFunctions::Lab_Local(
...
+
                         //Estimated current Middle grey at the end of GHS (midgrey)
                         //Absolute maximum RGB data at the end of GHS (maxdata)
                         //3 sigma is very representative of the data in use at the end of GHS (stdd)
                         float midgrey = 0.f;
                         float stdd = 0.f;
                         float maxdata = 0.f;
                         const int size = bfh * bfw;
-
+                        const float eps = 0.0001f;
```
kaesa: `constexpr`

rtengine/iplocallab.cc
```diff
@@ -19038,6 +19084,7 @@ void ImProcFunctions::Lab_Local(
                     const bool michwhite = params->locallab.spots.at(sp).mich_white;//Linear White point
                     const float michhigh = params->locallab.spots.at(sp).mich_high;//Highlight reduction
                     const bool midjdx = params->locallab.spots.at(sp).mich_jdx;//Matrix LMS using XYZ transform
+                    const float epsilm = 0.00001f;
```
kaesa: `constexpr`

Desmis:
I fully understood what was at stake. My comment only concerned the timing. My poor English doesn't make things any easier.

These problems with 'passing data to the GUI' using Selective Editing are not new. The solution envisioned by Pierre Cabrera (in 2019) is not simple...and involves some essential elements.

This necessarily involves declaring something like this in Improccordinator or dcrop.
            std::vector<LocallabListener::locallabshGHSbw> locallshgshbw;
This allows us to know which variables are involved, and for which spot (if we forget this, the system works correctly for the first spot... then it's chaos).

I followed the author's instructions, going through Locallab.cc and rtengine.h, then passing the parameters to locallabtools.cc (1 or 2), obviously remembering to retrieve the 'correct' spot in Iplocallab.cc. As I am a computer novice, sometimes it malfunctions and has forced me to do some 'workarounds'.

This problem isn't specific to GHS.

But I believe that improving the GUI, in terms of user interface, would be a huge step forward, especially if we could also draw polygons freehand (with significant constraints to comply with Selective Editing algorithms, which is far from easy)... We would have a 'nearly' perfect system.

Setting up your own 'fork' to improve the GUI is a good option. The question remains how we'll merge it at the end. But I think, again, my IT skills are limited compared to yours, and that you'll know how to handle it, even if it means opening a second pull request that does the same thing as the first, plus your modifications, and canceling mine.  
But if there aren't too many changes, I can make them 'manually'.

///
The other problem concerns the accuracy of the calculations. This same type of calculation is found in JzCzHz and initially in Ciecam02. Both systems were 'double' precision. 
For Ciecam, given the slowness of 'double' we switched to 'float' and similar problems (as GHS) appeared. Solved with CIECAM16.
For Jz taking into account 'PQ' and Absolute luminance results in dividing the RGB or XYZ values ​​by approximately 4. According to the designers (researchers), switching to 'double' is mandatory.... but it's slow.

Roughly speaking, GHS, with the modification I made, is somewhat similar to Jz... That is to say, when the RGB or XYZ values ​​are very low, if you divide them by the WP (3, 4, 11), the values ​​can fall below the processor's accuracy. See, in this regard, the considerable improvement brought about after the Windows 11 update and the significant differences in crash rates between Intel and AMD processors.

I think I've made a lot of changes where we look for negative or out-of-range XYZ or RGB values... But there might still be some left. Combining this with a more efficient GUI and your modifications should make the system more stable and perform better.

Thank you

Jacques

kaesa:
All the changes are highlighted in the link I posted. It really just fixes the conditions I think you tried to do in the first place. The changes fix the images changing arbitrarily when selecting spots and I think most crashes. I was thinking you could verify the changes and incorporate into this PR or tell me if they don't work and I can look at them again. I'm not trying to make your work harder but easier.

Desmis:
@kaesa 

I just added the changes you made to kaesa:RawTherapee:ghslook. I hope I haven't made a mistake.

Thank you

Our messages crossed, I tested it on several images and everything seems to be working correctly.

Jacques

kaesa:
The commit looks correct, I tested this PR and it seems work ok. Thank you for staying open minded in this long conversation. I've learned a lot about GHS capabilities while testing and I think it is a really good tool.

Desmis:
@kaesa @Lawrence37 

Thanks for your work, I think this PR is now ready to 'merge'.

As for GHS (and Michaelis-Menten), with the essential improvements to BP and WP, ​​they are two tools (complex for GHS) but remarkable.

The combination with Gamut Compression, Color Propagation, Abstract Profile, and Ciecam16, etc., makes it, in terms of colorimetry (I'm only mentioning this part here), a must-have... What's a shame is that it's discussed so little. Our communication (RT) isn't very good.

Of course, there are other tools, methods, and concepts in RT... that are also very good and have seen significant improvements... but again, they aren't talked about enough.

Jacques

Desmis:
@kaesa @Lawrence37 

I think, I'm almost 99% sure, that with the change proposed by @kaesa (there is now only one occurrence of GHS instead of 2 or 3 for the Preview), and the modifications made to avoid crashes (negative RGB and XYZ values), it's possible to **remove the restriction with a 'Stretch factor D' between 0.001 and 0.002** to adjust the automatic settings for the BP and WP and Symmetric Point (SP).

It remains to be seen, I think (not sure) that 'Inverse GHS' will likely require an adaptation.

I haven't done it yet; I just started it a week ago... because the deadlines for 5.13 seemed too tight. Nevertheless, I think it's quite easy to do, keeping "Stretch factor D > 0" to enable GHS.

The tooltips also need to be modified.

I open a new Pull request with the changes #7626 

What do you think? 

Jacques

Lawrence37:
@Desmis I see you converted this pull request into a draft. Is it still intended for 5.13?

Desmis:
@Lawrence37 @kaesa 
As I mentioned earlier in the text of this PR, I thought that what interested users - beyond the crashes which were very rare - was greater ease of use.

This PR resolved 2 difficult crash points, which can overlap:

**First type of crash:**
* For certain rare images where the native RGB values ​​are very low (close to or equal to zero) and which simultaneously have a high white point value, crashes occur if the user increases the white point or significantly alters the black point. This drastically reduces the initial RGB values; essentially, they are divided by the white point value... and become less than the system's 'float' precision. 
* This accuracy depends (and this surprised me) on both the operating system (Windows 11, Ubuntu 24.04, Ubuntu 26.04) and the type of processor (Intel or AMD). I was fortunate enough to have my grandson with me for almost a month during this period, using a different machine with an AMD processor. We painstakingly troubleshooted crashes, marking points (with 'printf') at virtually every phase of GHS. 
* We used the challenging images from 'Game changer' (Rawpedia Hugo). And we consistently used the same settings. The surprising thing (which confirms the differences in calculation accuracy) is that it was either my Intel machine (running Windows or Linux) or his (running AMD).
* Ultimately, the errors were of two types: a) negative RGB values, or, more curiously, b) negative XYZ values. This explains the crash I encountered some time ago with "rgb2lab". This led me to split "rgb2lab" into two parts: RgbXYZ, ensuring that XYZ values ​​are greater than zero, and then XYZLab.
* As I mentioned, we could have (like JzCzHz, which also significantly reduces RGB values ​​with the PQ function) used double precision, but the overall calculations would have been much slower.
* In summary, this work on three different systems eliminated crashes caused by unusual WP or BP values. This was achieved without impacting image quality or increasing processing time.

**Second type of crash**
The attempt to perform a 'real-time' calculation of white and black points with GHS enabled resulted in a complex GUI (at least for me). The system as I had designed it required two or three loops to find the values ​​to pass to the GUI. This also sometimes led to crashes, regardless of the BP and WP values. Kaesa solved this problem by allowing only one instance of the BP and WP calculation. Thanks to him.

**Why did I open another PR and put this one in draft?**
Since both types of crashes were resolved, I figured the **most important thing for the user was a simpler interfac**e. Having to set 'Stretch factor (D)' to 0.001 to change the settings was a significant drawback.

So I tried to remove all the constraints related to 'Stretch factor (D)'. I had two options:
*  continue with the same Pull Request, change the title, with the risk that it wouldn't work and I'd have to go back.
* open a 2nd PR #7626 with exactly the same content as this one, but which aims to remove the limitation due to 'Stretch factor D'. 
* This method allows us to see if it works. Otherwise, I would have cancelled this new PR #7626 and used the existing one #7609.
Since the new PR works, I kept the old one (draft) in case you or someone else found a malfunction.

**In conclusion, the Pull Request to be validated is the #7626 which combines this one (the crashes) and an improvement in usability.**

I hope I'm being clear... It's not easy to make choices when there are several bugs in progress... Furthermore, my poor English and the time difference don't make things any easier.

Jacques

Lawrence37:
Ok. In that case, I think we should close this PR to indicate that the plan is to not merge this branch. We can re-open it if plans change.

PR #7623, opened 2026-03-14
Backport gtkmm 4 event controllers

digitalcarp:
Adds the following minimal wrappers for GTK 4 event controllers:

- GestureClick (actually GestureMultiPress in GTK 3 backports)
- EventControllerKey
- EventControllerMotion
- EventControllerScroll

gtkmm already has backports for Gtk::EventControllerZoom and Gtk::GestureDrag. The others aren't backported due to a low number of maintainers for gtkmm.

This is to help facilitate incremental GTK 4 migrations.

### DIFF
--- a/rtgui/CMakeLists.txt
+++ b/rtgui/CMakeLists.txt
@@ -149,6 +149,7 @@ set(NONCLISOURCEFILES
     filepanel.cc
     filethumbnailbuttonset.cc
     filterpanel.cc
+    gtk4.cc
     guiutils.cc
     hidpi.cc
     histogrampanel.cc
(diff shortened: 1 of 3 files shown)


kaesa:
I'm not familiar with gtk4 so I have a dumb question. These emulate gtk4 and when the source is migrated to real gtk4 these backporting files are just removed?

digitalcarp:
Yeah they emulate the gtkmm 4 api, so you'd replace `rt::gtk4::` with `Gtk::` where used. There are still some incompatibilities between gtk3 and gtk4, so you'd need to make other small changes to code using these event controllers.

After the port, this file would be removed.

digitalcarp:
Implemented as part of #7638

PR #7626, opened 2026-03-17
SE - Generalized Hyperbolic Stretch - GHS - remove the Stretch factor D limitation

Desmis:
@kaesa @Lawrence37 

As I mentioned in the Pull Request  #7609, it's easy now that the system no longer performs multiple loops to calculate and account for Black Point (linear) and White Point (linear).

So I removed all controls related to Stretch Factor (D) and modified the tooltips.

I opened another Pull Request to avoid confusion.

But aside from removing the Stretch Factor control and re-enabling LC and Midtones in the GUI and in Iplocallab.cc, there are no other changes.

I've tested it several times on Windows 11 and Ubuntu 24.04 LTS, and it seems to work well. It's definitely 'better' and easier. Thanks to Kaesa for this GUI optimization of locallabtools.cc.

But that doesn't mean everything is perfect (code, tooltips..).

It is also obvious that I will have to revise the documentation on RawPedia (Hugo), but I will only do so if this PR is validated (otherwise a few 'marginal' modifications).

Jacques


kaesa:
It makes sense to remove the restriction.

Desmis:
@kaesa 
Thank you, that seems obvious to me. It's simpler, more efficient.

Thank you again for this modification.

Jacques

Desmis:
@Lawrence37 
See my comments at the end of the PR that I put in draft #7609

Desmis:
From my perspective, this pull request is complete. Of course, there are probably improvements to be made to GHS; what would they be? That will depend on user feedback.

Jacques

digitalcarp:
Looks good to me

Desmis:
@digitalcarp 

Thank you :)

kaesa:
@Desmis I think you could merge this. The PR is pretty clean, it addresses serious issues that need to be fixed, it has been sitting for a long time now without comments and @digitalcarp  has reviewed and ok'd it on his part. If there are more issues they won't be found without more people doing hands on testing.

Desmis:
@kaesa 

OK, Indeed, that doesn't pose any problems. And it allows us to move forward.

Thank you

Jacques

PR #7621, opened 2026-03-09
WIP split of iplocallab.cc

jonathanBieler:
Last time I wanted to contribute to RT I had troubles finding where the tools for local edits where defined and got scared by the 23k lines behemoth file. This is just a test to split iplocallab.cc into several files, which I think would make things a bit more approachable. I did only the first few modules (and it seems to compile & work as before) but in the end it would look something like this :

````
## File Structure After Refactoring

```
rtengine/
├── iplocallab.cc           # Lab_Local + anonymous namespace constants
├── iplocallab.h            # Shared types (local_params, constants)
└── locallab/
    ├── iplocallab_common.cc    # calc_ref, recovm, avoidcolshi
    ├── iplocallab_log.cc       # Log encoding functions
    ├── iplocallab_tonemap.cc   # Tone mapping functions
    ├── iplocallab_ciecam.cc    # CIECAM functions
    ├── iplocallab_softlight.cc # Soft light functions
    ├── iplocallab_exposure.cc  # Exposure functions
    ├── iplocallab_denoise.cc   # Denoise functions 
    ├── iplocallab_retinex.cc   # Retinex functions
    ├── iplocallab_blur.cc      # Blur functions
    ├── iplocallab_mask.cc      # Mask utility functions
    ├── iplocallab_sharp.cc     # Sharpening functions
    ├── iplocallab_exclude.cc   # Exclude functions
    ├── iplocallab_transit.cc   # Transit/shape detection
    └── iplocallab_wavelet.cc   # Wavelet/CBDL functions
````

I've used Claude for help (see rtengine\locallab\REFACTORING_PLAN.md) but this shouldn't introduce any new code if everything goes well, so it's not super risky.

If you guys want this I can try to finish the split, otherwise if you prefer a single large file I'll close this.

PR #7629, opened 2026-03-22
Selective Editing - CAM16 & Wavelet :Update with recent changes in CIECAM, Abstract profile - update thumbnails

### LINKED ISSUE #6029: Local Adjustments doesn't update thumbnail
After successfully applying some Local Adjustments, it appears that the tool doesn't update the miniature in the File Browser, nor in the Filmstrip.

Is that the intended behavior?

This is an example of some Local Adjustments applied to an image (crazy settings so it is clear the difference when no LA tool is applied):

laon

And this is what the Filmstrip and the File Browser shows:

lamini

Compare with the _no LA applied_ version:

laoff

It would be interesting (needed?) that the File Browser shows the current state of the processing.

Desmis:
This Pull Request  is an upgrade of Selective Editing with the recent new features in 'dev'.
* **red green blue** in Color Appearance (CAM16) (as in CIECAM main)
* **attenuation threshold** in Color Appearance (CAM16) * Source Data Adjustments (as in Abstract profile)
* **final gain & Gamut compression** in Color Appearance (CAM16) (as in Abstract profile)
For these first 3 items, it's essentially the GUI; the 'engine' code is almost identical. 

* **Local contrast in Local contrast & Wavelets** (inspired by Contrast Enhancement), with a significative improvment for users  but "small" change in code. It allows you to add a gradient of decomposition levels, plus or minus.

I've also added a feature that was missing since the creation of Selective Editing (in 2014-2018): **thumbnail updates.** Aside from adding numerous lines of code (like in Simpleprocess...), it doesn't present any difficulties (at least I hope not).


Desmis:
@Lawrence37 @hholtm

Regarding the part about "Selective Editing does not update the thumbnails", I think the change made by the PR works.
* I tried with several local Spots,
*  or one Spot or 2 Spot in 'Global' mode
* everything seems to work correctly.

Six years ago, I hadn't considered this solution, which is simple in terms of process but uses a lot of code (over 300 lines). In fact, Selective Editing used three pipeline processes, each with its own specific characteristics. 

I added an indicator to the Selective Editing 'engine' code to identify which part of the pipeline the instruction originated from (variable int call)
* call = 1 - from dcrop.cc - with specific with specific processing for, for example, denoising (a problem with the Preview), but also other functions
* call = 2 - from simpleprocess.cc (output for TIFF/JPG)  with specific processing for, for example, denoising (a problem with the Preview), but also other functions. Here we work on the entire image, and the processing algorithms can be (very) different from those of the preview.
* call = 3 - from improccoordinator.cc general function for calling functions in RT which allows among other things the passing of parameters to the GUI, but which does not call certain specific functions of the code. It is fast because it works with 1 pixel out of 10.

The code that passes parameters for each Spot, along with the curves and parameters passed to the GUI, is complex. The simplest solution is the one from Simpleprocess.cc. That's why I used it to update the thumbnails. But as it stands, it will trigger processes that aren't needed for thumbnails, such as 'denoise'. That's why I'm using the 'simpleprocess' code with `call = 3`, but with one difference: to save processing time, I set `skip` to 16 instead of 10. We only need a preview in the thumbnail, not surgical precision.

On medium-sized images, with 1 global spot, the processing time on my machine is approximately 30 milliseconds.

In **summary**, I think that now the 4 processes in the pipeline are taken into account, without devoting too many resources to updating the thumbnails.

The problem is sharing this PR, for this part (thumbnails) and for others (Cam16, Local contrast Wavelet...). Unfortunately, it works very well in compiled version on my machine (Windows or Linux), but the 'exiv2' library is apparently missing from Github, which means the Windows versions don't work;

Desmis:
From my point of view this Pull Request is finished, but it needs to be merged after the one relating to GHS  #7626 , there may be conflicts.
So, in summary:
1. merge 7626
2. merge 7629

Desmis:
I made a few relatively minor changes, which concern 2 points:
* Add a 'Blur shape detection' slider that does what it says. This function is already present in 'Color & Light', 'Dynamic Range & Exposure', and 'Equalization & pre-tone mapping'...
* I also modified the tooltip in 'Selective Editing > Settings' related to "deltaE Shape detection"

It was during the writing of the ongoing tutorial on 'Game changer' (Pagodas) in Rawpedia that I noticed these needs.
https://github.com/RawTherapee/RawPedia

@kaesa 
Can you take a look at this PR? It's nothing revolutionary. It's essentially the GUI... and a few tweaks. See my comments above and on Pixls.us

kaesa:
You added this block to `void ImProcFunctions::gamutcompr( Imagefloat *src, Imagefloat *dst, int beginend, float powe, int nbsegam,  float &mac, float &mac0, float &mac1, float &mac2) const`
```
    } else if (beginend == 2) {//at the end of the process Selective Editing CAM16. Only 4 cases, which are the cases, at this stage, actually possible 
        if (nbsegam == 1) {
            out = Rec2020;
        } else if (nbsegam == 2) {
            out = adobe;
        } else if (nbsegam == 3) {
            out = srgb;
        } else if (nbsegam == 4) {
            out = dcip3;
        }
    }
```
What is the reason for this change because every time you call the `gamutcompr` you pass 0 as `nbsegam`?

kaesa:
I take the above back. I found the place where you pass something else instead.

Desmis:
@kaesa 

No problem at all. This is the purpose of a PR and the examination by another

kaesa:
I tried looking through it and I didn't spot anything. I have to be honest, this is a very large PR and it is very hard to spot the code lines that can be actually dangerous. I think we may have to rely on copilot for finding typos and inconsistencies. Also, I had a hard time finding the changes from the UI perspective for practical testing. The code overall has some refactoring needs, but I didn't spot this PR adding new targets for that.

Desmis:
@kaesa 

I have already checked it with Copilot, but I can do it again.

Admittedly, there's a lot of code... but as I explained in the issue, I was planning to release this PR for 5.14. The **thumbnail** problem, which is now the root cause, seems to be resolved, since no one is mentioning it.

However, aside from the GUI, there's (almost) nothing new in this PR except for the very simple modification of 'Local contrast'.

rtengine/rtthumbnail.cc
```diff
@@ -1742,8 +2108,10 @@ IImage8* Thumbnail::processImage (const procparams::ProcParams& params, eSensorT
         float mac1 = 0.f;
         float mac2 = 0.f;
         int beginend = 1;
+        int nbsegam = 0;
+        int sp = 0;
         if (params.icm.wgamut != ColorManagementParams::Wwgamut::NONE) {
-            ipf.gamutcompr(provcomp, provcomp, beginend, mac, mac0, mac1, mac2);
+            ipf.gamutcompr(provcomp, provcomp, beginend, sp, nbsegam, mac, mac0, mac1, mac2);
         }
```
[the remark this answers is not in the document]
Desmis: OK

rtengine/rtthumbnail.cc
```diff
@@ -1365,6 +1365,372 @@ IImage8* Thumbnail::processImage (const procparams::ProcParams& params, eSensorT
             ipf.getAutoExp (aeHistogram, aeHistCompression, 0.02, expcomp, bright, contr, black, hlcompr, hlcomprthresh);
         }
     }
+    //Thumbnails update in Selective Editing
+    LabImage* labView2 = new LabImage (fw, fh);
+    {//The code is essentially identical to simpleprocess.cc, but of course with a much smaller scale (sk = 16).
+     //To ensure it is recognized by all transition systems, deltaE, etc., I use call = 3 as the improcoordinator in Lab_local.
+     // call = 3 will avoid activating the resource-intensive 'denoise' functions.
+     //I declare an additional Labimage variable, labview2, to avoid interference with the rest of rtthumbnail.
+        if (params.locallab.enabled && params.locallab.spots.size() > 0) {
```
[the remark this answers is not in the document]
Desmis: OK It's cleaner this way, obviously I had to add some get()

Desmis:
@kaesa @digitalcarp @Lawrence37 @jonathanBieler 

I just merge the PR related to GHS.

I also made a series of changes, not all specific to this PR, but aimed at making the code more robust before  @jonathanBieler implements the PR #7621 

This concerns the 'min' and 'max' functions and also 'std::unique_ptr'...

I think that in a few days, if no one objects, I'll merge this PR. Of course, I'll talk about it on Pixls.us

Jacques

Desmis:
To better showcase the new features, I've updated Rawpedia Game Changer with a challenging "Pagodas" image.

https://rawpedia.pixls.us/tutorials/#pagodas---how-to-process-local-contrast---clarity

Jacques

Lawrence37:
I'm taking this off the 5.13 milestone because it makes significant changes.

Desmis:
I made a change that I think is important. It concerns the ‘red-green-blue’ part in CIECAM (Color Appearance & Lighting) or Selective Editing > Color Appearance - CAM16.

When you modify the Brightness curve (whether for red, green or blue) the system generates artifacts, especially in skies with clouds

Due to the complexity of CIECAM, where 6 variables are active simultaneously (J, Q, C, s, M, h), the solution is not simple and cannot be reduced to a “filter”… I added a “Threshold Brightness Curve” slider which is perhaps a bit excessive, but it works.

rtengine/iplocallab.cc
```diff
@@ -2125,7 +2133,7 @@ float generalized_loglogistic_sigmoid(float value,
                                       float film_power,
                                       float paper_power)
 {
-    const float clamped_value = max(value, 0.0);
+    const float clamped_value = rtengine::max(value, 0.0);
```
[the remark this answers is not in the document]

rtengine/iplocallab.cc
```diff
@@ -2230,7 +2238,7 @@ void  ImProcFunctions::sigmoid_main(float r,
                      paper_exposure, paper_power, middle_grey);
     float rgb[3] = {r, g, b};
     for (int i = 0; i < 3; i = i+1) {
-        rgb[i] = max(rgb[i], 0);
+        rgb[i] = rtengine::max(rgb[i], 0);
     }
```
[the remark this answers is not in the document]
Desmis: OK

Desmis:
I think the result is correct now.

My initial idea of ​​introducing a threshold - which actually takes Chroma (C) into account - was insufficient to reduce artifacts (even if it reduced many of them) when using one or more of the **brightness curves for “Red Green Blue”** (whether in Color Appearance & Lighting or the new Selective Editing > Color Appearance).

Adding a gradual effect avoids the before/after effect and results in smoother transitions.

The problem is far from simple, because here we can’t use the ‘usual’ tools to ensure this gradualness, like “Guided Filiter” or “Gaussian blur”… The concept implemented is, in a way, quite similar to a deltaE.

Lawrence37:
There's a crash when Color Appearance & Lighting > Image Adjustments > Red Green Blue > Threshold Brightness Curves is set to 100.

It might be happening here.
https://github.com/RawTherapee/RawTherapee/blob/ae45453641c751b75df7f650bd5e29520341dcff/rtengine/improcfun.cc#L766-L769
`brighthres` is 100. `limb` becomes 100 which leads to a division by zero when calculating `kam`.

Desmis:
@Lawrence37 
I am still working with limited resources (computer, highly unstable internet connection, etc.).
You are right; a division-by-zero error can occur when using the "Threshold brightness curve" - thanks for spotting that bug -but the issue isn't limited to "Color Appearance & Lighting"; "Selective Editing - Color Appearance (CAM16 & JzCzHz)" is affected as well.

The few users who tried both versions of 'red green blue' (see Pixls.us and the limitations of pull-request-based validation, since it isn't the 'dev' branch) seem to have preferred the Selective Editing version, as it offers more flexibility (it's the same modification).

As for applying the Color Appearance & Lighting modification in PR #7716, It’s a tedious task - if you want everything to go smoothly later on and avoid merge conflicts. It essentially involves copying the code line by line.

Jacques

Desmis:
@Lawrence37 
If no one objects, I'm going to merge this PR on Sunday, August 2nd, or monday

rtdata/languages/default
```diff
@@ -2516,6 +2541,13 @@ TP_COLORAPP_ALGO_TOOLTIP;Lets you choose between parameter subsets or all parame
 TP_COLORAPP_BADPIXSL;Hot/bad pixel filter
 TP_COLORAPP_BADPIXSL_TOOLTIP;Suppression of hot/bad (brightly colored) pixels.\n0 = No effect\n1 = Median\n2 = Gaussian.\nAlternatively, adjust the image to avoid very dark shadows.\n\nThese artifacts are due to limitations of CIECAM02.
 TP_COLORAPP_BRIGHT;Brightness (Q)
+<<<<<<< HEAD
+TP_COLORAPP_BRIGHTHRES;Threshold Brightness Curves
+TP_COLORAPP_BRIGHT_TOOLTIP;Brightness in CIECAM is the amount of perceived light emanating from a stimulus. It differs from L*a*b* and RGB brightness.
+TP_COLORAPP_BRIGHT_CUR_RED;Brightness - red
+TP_COLORAPP_BRIGHT_CUR_GREEN;Brightness - green
```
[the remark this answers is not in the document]
Desmis: OK

rtdata/languages/default
```diff
@@ -1577,6 +1592,11 @@ HISTORY_MSG_LOCAL_CIE_SLOPESMOG;Local - CIECAM Green balance
 HISTORY_MSG_LOCAL_CIE_SLOPESMOQ;Local - CIECAM Gray balance
 HISTORY_MSG_LOCAL_CIE_SLOPESMOR;Local - CIECAM Red balance
 HISTORY_MSG_LOCAL_CIE_SMOOTH;Local - CIECAM Scale Yb scene
+<<<<<<< HEAD
+HISTORY_MSG_LOCAL_CIE_SMOOTHJCIE;Local - CIECAM Smooth
+HISTORY_MSG_LOCAL_CIE_SMOOTHLNK;Local CIECAM - Link R G B
+=======
+>>>>>>> dev
 HISTORY_MSG_LOCAL_CIE_SMOOTHINV;Local CIECAM - Inverse color
 HISTORY_MSG_LOCAL_CIE_SMOOTHLNK;Local CIECAM - Link R G B
```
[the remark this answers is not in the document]
Desmis: OK

rtgui/tools/locallabtools2.cc
```diff
@@ -9814,6 +9983,33 @@ void Locallabcie::updateguicie(int spottype)
...
+{//maximum RGB and Saturation at the end of CAM16
+    idle_register.add(
+    [this, m_rgb, m_sat, gamaut]() -> bool {
+        GThreadLock lock; // All GUI access from idle_add callbacks or separate thread HAVE to be protected
+
+            if (gamaut) {
+                rgbmaxdata->set_text(//RGB maximum
+                    Glib::ustring::compose(M("TP_LOCALLAB_CIE_FINALRGB"),
+                        Glib::ustring::format(std::fixed, std::setprecision(3), m_rgb))
+                );
```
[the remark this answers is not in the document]
Desmis: OK, I chnage also locallabtools.cc

rtdata/languages/default
```diff
@@ -1526,6 +1527,7 @@ HISTORY_MSG_LOCAL_AUTOCAPRADIUS;Local - Sh Auto Capture radius
 HISTORY_MSG_LOCAL_AUTOSHAR;Local - Sh Auto Contrast
 HISTORY_MSG_LOCAL_AUTOSP;Local - GHS - Auto SP
 HISTORY_MSG_LOCAL_AVOIDNEGATIVE;Local - SC - Pre-filter zero and negative values
+HISTORY_MSG_LOCAL_BLURCIEDE;Local - blur - CIECAM
```
Lawrence37: It should go from broad to specific and the first letter of each should be uppercase, i.e. `Local - CIECAM - Blur`. "Blur" is misleading because it is most likely interpreted as blurring the image, not the mask. Be more specific, like "Mask blur" or "Blur shape detection".

rtdata/languages/default
```diff
@@ -1544,7 +1546,20 @@ HISTORY_MSG_LOCAL_CIE_BLUYL;Local - CIECAM Blue Y
 HISTORY_MSG_LOCAL_CIE_BRICOMP;Local - CIECAM Brightness compression
 HISTORY_MSG_LOCAL_CIE_BRICOMPTH;Local - CIECAM Brightness compression threshold
 HISTORY_MSG_LOCAL_CIE_BWCIE;Local - CIECAM Black and white
+HISTORY_MSG_LOCAL_CIE_GAMGAIN;Local - CIECAM Gain
+HISTORY_MSG_LOCAL_CIE_GAMPOWER;Local - CIECAM Power
+HISTORY_MSG_LOCAL_CIE_GAMLIST;Local - CIECAM Target Gamut
```
Lawrence37: There's some inconsistency here, including existing strings. Some have a `-` after `CIECAM` and others don't.

rtdata/languages/default
```diff
@@ -1544,7 +1546,20 @@ HISTORY_MSG_LOCAL_CIE_BLUYL;Local - CIECAM Blue Y
...
 HISTORY_MSG_LOCAL_CIE_CAT;Local - Matrix adaptation
+HISTORY_MSG_LOCAL_CIE_RGB_BRIGHTHRES;Local - CIECAM - Threshold Brightness Curves
```
Lawrence37: "Brightness curves threshold" is more grammatical and follows the established capitalization scheme for adjusters.

rtdata/languages/default
```diff
@@ -1544,7 +1546,20 @@ HISTORY_MSG_LOCAL_CIE_BLUYL;Local - CIECAM Blue Y
...
+HISTORY_MSG_LOCAL_CIE_RGB_COLORHRED;Local - CIECAM - Hue Red
+HISTORY_MSG_LOCAL_CIE_RGB_SCHROMARED;Local - CIECAM - Sat Red
+HISTORY_MSG_LOCAL_CIE_RGB_SHAPERED;Local - CIECAM - Brightness Red
+HISTORY_MSG_LOCAL_CIE_RGB_COLORHGREEN;Local - CIECAM - Hue Green
+HISTORY_MSG_LOCAL_CIE_RGB_SCHROMAGREEN;Local - CIECAM - Sat Green
+HISTORY_MSG_LOCAL_CIE_RGB_SHAPEGREEN;Local - CIECAM - Brightness Green
+HISTORY_MSG_LOCAL_CIE_RGB_COLORHBLUE;Local - CIECAM - Hue Blue
+HISTORY_MSG_LOCAL_CIE_RGB_SCHROMABLUE;Local - CIECAM - Sat Blue
+HISTORY_MSG_LOCAL_CIE_RGB_SHAPEBLUE;Local - CIECAM - Brightness Blue
```
Lawrence37: Capitalization here and in other places also doesn't match the established scheme. Only the first word should be capitalized, e.g. "Brightness blue".

rtdata/languages/default
```diff
@@ -1577,8 +1592,9 @@ HISTORY_MSG_LOCAL_CIE_SLOPESMOG;Local - CIECAM Green balance
 HISTORY_MSG_LOCAL_CIE_SLOPESMOQ;Local - CIECAM Gray balance
 HISTORY_MSG_LOCAL_CIE_SLOPESMOR;Local - CIECAM Red balance
 HISTORY_MSG_LOCAL_CIE_SMOOTH;Local - CIECAM Scale Yb scene
-HISTORY_MSG_LOCAL_CIE_SMOOTHINV;Local CIECAM - Inverse color
+HISTORY_MSG_LOCAL_CIE_SMOOTHJCIE;Local - CIECAM Smooth
 HISTORY_MSG_LOCAL_CIE_SMOOTHLNK;Local CIECAM - Link R G B
+HISTORY_MSG_LOCAL_CIE_SMOOTHINV;Local CIECAM - Inverse color
```
Lawrence37: Add a `-` between `Local` and `CIECAM`.

Desmis:
@Lawrence37 

Apologies for the lack of rigor regarding the merge...with the last Dev change (threshold) but in my defense, this PR has been open for a (very) long time...

rtdata/languages/default
```diff
@@ -1631,6 +1647,8 @@ HISTORY_MSG_LOCAL_GHS_MID;Local - GHS - Midtones
 HISTORY_MSG_LOCAL_GHS_SLOPE;Local - GHS - Lab slope factor
 HISTORY_MSG_LOCAL_GHS_SMOOTH;Local - GHS - Highlight attenuation
 HISTORY_MSG_LOCAL_GHS_SP;Local - GHS - Symmetry point
+HISTORY_MSG_LOCAL_GRADWAV;Local Wav Gradient levels
+HISTORY_MSG_LOCAL_LIMITWAV;Local -Wavelet - Prevent decomposition
```
Lawrence37: Inconsistent use of `-` and missing space after one of the `-`s.

rtdata/languages/default
```diff
@@ -2516,9 +2534,11 @@ TP_COLORAPP_ALGO_TOOLTIP;Lets you choose between parameter subsets or all parame
 TP_COLORAPP_BADPIXSL;Hot/bad pixel filter
 TP_COLORAPP_BADPIXSL_TOOLTIP;Suppression of hot/bad (brightly colored) pixels.\n0 = No effect\n1 = Median\n2 = Gaussian.\nAlternatively, adjust the image to avoid very dark shadows.\n\nThese artifacts are due to limitations of CIECAM02.
 TP_COLORAPP_BRIGHT;Brightness (Q)
-TP_COLORAPP_BRIGHT_CUR_BLUE;Brightness - blue
-TP_COLORAPP_BRIGHT_CUR_GREEN;Brightness - green
+TP_COLORAPP_BRIGHTHRES;Threshold Brightness Curves
```
Lawrence37:
```suggestion
TP_COLORAPP_BRIGHTHRES;Brightness curves threshold
```

rtdata/languages/default
```diff
@@ -2516,9 +2534,11 @@ TP_COLORAPP_ALGO_TOOLTIP;Lets you choose between parameter subsets or all parame
...
+TP_COLORAPP_BRIGHT_TOOLTIP;Brightness in CIECAM is the amount of perceived light emanating from a stimulus. It differs from L*a*b* and RGB brightness.
```
Lawrence37: This is awkwardly worded and too scientific. I suggest using more plain language.

```suggestion
TP_COLORAPP_BRIGHT_TOOLTIP;CIECAM brightness closely follows human perception of the amount of light. It is not exactly the same as L*a*b* and RGB brightness.
```

rtdata/languages/default
```diff
@@ -3133,7 +3154,7 @@ TP_LOCALLAB_ADJ;Equalizer Color
 TP_LOCALLAB_AMOUNT;Amount
 TP_LOCALLAB_APSATUR;Saturation
 TP_LOCALLAB_ARTIF;ΔE Shape detection
-TP_LOCALLAB_ARTIF_TOOLTIP;ΔE scope threshold increases the range of ΔE scope. High values are for very wide gamut images.\nIncreasing ΔE decay can improve shape detection, but can also reduce the scope.
+TP_LOCALLAB_ARTIF_TOOLTIP;ΔE range: increases the range of ΔE scope. High values are for very wide gamut images.\nIncreasing ΔE decay can improve shape detection, high values ​​can reduce artifacts, but can also reduce the scope.
```
Lawrence37: Combine the two statements about large delta E.
```suggestion
TP_LOCALLAB_ARTIF_TOOLTIP;ΔE range: increases the range of ΔE scope. High values are for very wide gamut images.\nIncreasing ΔE decay can improve shape detection and reduce artifacts, but can also reduce the scope.
```

rtdata/languages/default
```diff
@@ -3228,6 +3253,7 @@ TP_LOCALLAB_CIEMODE_DR;Dynamic Range
 TP_LOCALLAB_CIEMODE_TM;Tone-Mapping
 TP_LOCALLAB_CIEMODE_TOOLTIP;In Default mode, CIECAM is added at the end of the process. 'Mask And Modifications' and 'Recovery Based On Luminance Mask' are available for'CAM16 and JzCzHz' at your disposal .\nYou can also integrate CIECAM into other tools if you wish (TM, Wavelet, Dynamic Range, Log encoding). The results for these tools will be different to those without CIECAM. In this mode, you can also use 'Mask And Modifications' and 'Recovery Based On Luminance Mask'.
 TP_LOCALLAB_CIEMODE_WAV;Wavelet
+TP_LOCALLAB_CIEREDGREENBLUEFRA;Red Green Blue
```
Lawrence37: The key name has an unclear abbreviation.
```suggestion
TP_LOCALLAB_CIEREDGREENBLUE_FRAME;Red Green Blue
```

rtdata/languages/default
```diff
@@ -3250,7 +3276,7 @@ TP_LOCALLAB_CLARISOFT;Soft radius
 TP_LOCALLAB_CLARISOFTJZ_TOOLTIP;The 'Soft radius' slider (guided filter algorithm) reduces halos and irregularities for Clarity, Sharp Mask and Local contrast wavelets Jz.
 TP_LOCALLAB_CLARISOFT_TOOLTIP;The 'Soft radius' slider (guided filter algorithm) reduces halos and irregularities for Clarity, Sharp Mask and all wavelet pyramid processes. To deactivate, set slider to zero.
 TP_LOCALLAB_CLARITYML;Clarity
-TP_LOCALLAB_CLARI_TOOLTIP;Levels 0 to 4 (included): 'Sharp mask' is enabled\nLevels 5 and above: 'Clarity' is enabled.\nUseful if you use 'Wavelet level tone mapping'.
+TP_LOCALLAB_CLARI_TOOLTIP;This algorithm combines two decompositions of the same image and recombines them using different processes, It involves significant system resources.\n\n From levels 0 to Bottom-right = 4 (included) 'Sharp mask' is enabled.\n Starting from levels 5 – Top-right or Bottom-right and above: 'Clarity' is enabled.\n\n In the case of the 'Sharp Mask' the change of level is similar to changing the radius in the 'Unsharp Mask'. For 'Clarity' a level change results in a greater or lesser three-dimensional volume effect on the image.
```
Lawrence37: - The `,` in the first sentence should be a `.`
- Change `It involves significant system resources.` to `It requires significant system resources.` (`involves` -> `requires`).
- It is unclear what the description of the levels refers to. Nothing in the clarity frame is obviously related to "Bottom-right", "Top-right", or any of the levels. Image
- In the last sentence, "volume" can be removed. "Three-dimensional" already implies volume and "three-dimensional effect" is a common phrase.

rtdata/languages/default
```diff
@@ -3406,6 +3432,7 @@ TP_LOCALLAB_GAMMASKCOL;Gamma
...
 TP_LOCALLAB_GAMSH;Gamma
 TP_LOCALLAB_GAMUTLABRELA;Lab
+TP_LOCALLAB_GAMUTW_TOOLTIP; Placed at the end of Selective Editing > Color Appearance (CAM16 & JzCzHz), takes into account, only for CAM16 & JzCzHz, everything preceding the current Spot, this tool allows you to compress the data to fit into the output profile.\n\n This output gamut compression is much simpler than at the beginning of the processing pipeline because unless the user has exaggerated the settings, the gamut will already be close to the Working Profile, hence the simplified graphical interface.\n\n More often than not, you will not see any changes to the image or the histogram when using Gamut Compression. This is because it does not convert the data from Working profile (often Rec2020) to Target Gamut (often sRGB), but compresses it to fit the Target gamut (often sRGB) while remaining in Working profile (often Rec2020), and the upstream treatments must have preserved the gamut. The effect is visible on Led’s, flowers with high saturation. On a normal landscape or portrait, there is most often no change.
```
Lawrence37: - The first sentence is a run-on sentence. It's also not clear to me what `only for CAM16 & JzCzHz` means. Ignoring that part, I suggest `Compresses the colors into the chosen profile. It is applied after all previous spots and the other adjustments in this tool.`
- The second sentence does not flow well. I suggest `Output gamut compression is much simpler than input gamut compression. Colors near the output stage of processing are rarely outside the working profile. A reduced set of adjustments is sufficient.`
- The third and fourth sentences are wordy. Listing the typical color spaces does not add value, and listing them twice is definitely unnecessary. Also, "working profile" and "target gamut" do not need to be capitalized because it makes at least as much sense for us to refer to the actual profile and gamut rather than the adjustments. I suggest `Often, there is little to no change to the image when using Gamut Compression. This is because it does not convert the color profile, but compresses colors to fit the target gamut. Adjustments applied before compression usually keep colors in the target gamut.`
- "LED" should be in full uppercase because it is an initialism and there should be no apostrophe because it is not possessive. The sentence is also ungrammatical. I suggest `The effect is visible on LEDs and vibrant flowers.`
- A singular "landscape" and "portrait" is awkward. "Most often" is wordy. Use `On normal landscapes and portraits, there is often no change.`

rtdata/languages/default
```diff
@@ -3492,6 +3519,8 @@ TP_LOCALLAB_GRADSTRHUE;Hue gradient strength
 TP_LOCALLAB_GRADSTRHUE2;Hue gradient strength
 TP_LOCALLAB_GRADSTRHUE_TOOLTIP;Adjusts hue gradient strength.
 TP_LOCALLAB_GRADSTRLUM;Luma gradient strength
+TP_LOCALLAB_GRADWAV;Gradient levels
+TP_LOCALLAB_GRADWAV_TOOLTIP;Adjust local contrast for high and medium decomposition levels.\n Medium corresponds to a Top-right value greater than 3. High corresponds to a Top-right value greater than 5.
```
Lawrence37: Are "high" and "medium" standard language to refer to the levels? Typically, they are used to indicate intensity. "High" and "middle" are often used for height and levels.

What does "Top-right" refer to?

rtdata/languages/default
```diff
@@ -3578,6 +3607,8 @@ TP_LOCALLAB_LEVFRA;Levels
 TP_LOCALLAB_LIGHTNESS;Lightness
 TP_LOCALLAB_LIGHTN_TOOLTIP;In inverse mode: selection = -100 forces luminance to zero.
 TP_LOCALLAB_LIGHTRETI;Lightness
+TP_LOCALLAB_LIMITWAV;Prevent decomposition: contrast high & low light
```
Lawrence37: If I understand the tooltip correctly, I think it's clearer to write
```suggestion
TP_LOCALLAB_LIMITWAV;Protect extreme shadows and highlights
```

rtdata/languages/default
```diff
@@ -3578,6 +3607,8 @@ TP_LOCALLAB_LEVFRA;Levels
...
+TP_LOCALLAB_LIMITWAV_TOOLTIP;Allows you to avoid, for all 'wavelet' functions which implement contrast, modifying the very low lights and the very high lights.
```
Lawrence37: Injecting the qualifier in the middle of the sentence breaks the sentence flow. Rearrange or make it an essential part of the sentence. "Wavelet" does not need to be quoted.
```suggestion
TP_LOCALLAB_LIMITWAV_TOOLTIP;Prevents the wavelet functions from modifying the contrast in the extreme shadows and highlights.
```

Lawrence37:
I am reviewing the `default` file because I know English is not your expertise.

rtdata/languages/default
```diff
@@ -3216,6 +3237,10 @@ TP_LOCALLAB_CHROMASK_TOOLTIP;Changes the chroma of the mask if one exists (i.e.
 TP_LOCALLAB_CHROML;Chroma (C)
 TP_LOCALLAB_CHRRT;Chroma
 TP_LOCALLAB_CIE;Color Appearance (CAM16 & JzCzHz)
+TP_LOCALLAB_CIE_FINALRGB;Final RGB Max=%1
+TP_LOCALLAB_CIE_FINALRGB_NO;---
+TP_LOCALLAB_CIE_FINALSAT;Final Saturation Max=%1
+TP_LOCALLAB_CIE_FINALSAT_NO;---
```
Lawrence37: `---` doesn't need to be translated. It can be hard-coded. `TP_LOCALLAB_CIE_FINALRGB_NO` and `TP_LOCALLAB_CIE_FINALSAT_NO` can then be removed to avoid cluttering the `default` file and translations.

rtdata/languages/default
```diff
@@ -3216,6 +3237,10 @@ TP_LOCALLAB_CHROMASK_TOOLTIP;Changes the chroma of the mask if one exists (i.e.
 TP_LOCALLAB_CHROML;Chroma (C)
 TP_LOCALLAB_CHRRT;Chroma
 TP_LOCALLAB_CIE;Color Appearance (CAM16 & JzCzHz)
+TP_LOCALLAB_CIE_FINALRGB;Final RGB Max=%1
+TP_LOCALLAB_CIE_FINALRGB_NO;---
+TP_LOCALLAB_CIE_FINALSAT;Final Saturation Max=%1
```
Lawrence37: "<something> Max=<number>" doesn't look polished. Also, it can easily be misread as "(<something> Max)=<number>" instead of "<something> (Max=<number>)". I suggest rewording it like this:
`Final max RGB: %1`
`Final max saturation: %1`
If changing, make sure to update `TP_LOCALLAB_RGBMAXDATA_TOOLTIP` too.

rtgui/tools/locallabtools2.cc
```diff
@@ -9878,7 +10077,9 @@ void Locallabcie::updateAdviceTooltips(const bool showTooltips)
         sourceabscie->set_tooltip_text(M("TP_COLORAPP_ADAPSCEN_TOOLTIP"));
         cie1Frame->set_tooltip_text(M("TP_LOCALLAB_LOGIMAGE_TOOLTIP"));
         smoothFrame->set_tooltip_text(M("TP_LOCALLAB_SMOOTHCIE_TOOLTIP"));
-
+        rgbmaxdata->set_tooltip_text(M("TP_LOCALLAB_RGBMAXDATA_TOOLTIP"));
```
Lawrence37, line 10080: The tooltip refers to both the maximum RGB and the saturation, but the tooltip only appears for the maximum RGB.

rtdata/languages/default
```diff
@@ -3871,6 +3902,7 @@ TP_LOCALLAB_RET_TOOLNAME;Dehaze & Retinex
...
...IP;Provides information on out-of-limit RGB values and RGB saturation. Values greater than 1 for ‘Final RGB Max’ and ‘Final Saturation Max’ clearly indicate that we are out of gamut whereas for values less than 1, we cannot say whether the image is within gamut or not because it depends, for example, on the luminance. This check is performed at the end of Selective Editing > Color Appearance (CAM16 & JzCzHz) and takes into account everything preceding the current Spot. Specifically, it considers 'Source Data Adjustments' and 'Tone Mapping Operators', as well as the CIECAM 'CAM16 Image Adjustments' settings. It obviously does not consider settings located later in the process, such as Abstract Profile or Color Appearance & Lighting. \n\nAdjusting the Gain (Ev) and Gamut Compression (Target Gamut and Power) settings will allow you to see the impact of these adjustments. You should also observe the histogram, which takes into account the output profile therefore the gamma. This data, which is directly related to the RGB values ​​and Saturation and is in the Working Profile, is in linear mode.\n\n The data is only displayed if Target Gamut is enabled, or if Gain (Ev) is not equal to 0.
```
Lawrence37: - "Provides information on" implies it show more than just the maximum RGB and saturation values.
- "We" and "you" does not match the pattern of other tooltips.
- The second sentence is a little long.
- The full name of the tool "Color Appearance..." is unnecessary.
- Calling out specific adjustments suggests that they are treated differently somehow, but this does not appear to be the case. It also does not provide any useful information.
- No need to write `obviously`. If it were obvious, the sentence wouldn't need to be written.
- The meaning of "allow you to see the impact of these adjustments" is unclear. I'm guessing it means that the max RGB and saturation values reflect the image after Final Gain & Gamut Compression is applied.
- It tells the user to observe the histogram, but does not explain why.
- `the output profile therefore the gamma` is not grammatical.
- Writing that the data is directly related to the RGB values and saturation is unnecessary because the UI shows "RGB" and "saturation".

Suggestion that does not refer to the histogram:
```suggestion
TP_LOCALLAB_RGBMAXDATA_TOOLTIP;Shows the maximum RGB value and saturation, which can be used to determine if any colors are out-of-gamut. Values greater than 1 for ‘Final RGB Max’ and ‘Final Saturation Max’ always indicate out-of-gamut colors. Values less than 1 are inconclusive because it depends on other factors like the luminance. This check is performed at the end of the tool and considers everything applied before the current Spot. It does not consider adjustments located later in the process, such as Abstract Profile or Color Appearance & Lighting. \n\nThe maximum RGB value and saturation reflect the state of the image after the Gain (Ev) and Gamut Compression (Target Gamut and Power) adjustments. The values are shown in linear space in the working profile.\n\nData is only displayed if Target Gamut is enabled, or if Gain (Ev) is not equal to 0.
```

rtdata/languages/default
```diff
@@ -4094,13 +4126,13 @@ TP_LOCALLAB_WAT_RESIDCOMP_TOOLTIP;Compresses the residual image to increase or r
...
-TP_LOCALLAB_WAT_THRESHOLDWAV_TOOLTIP;Range of wavelet levels used throughout the Wavelets module.
+TP_LOCALLAB_WAT_THRESHOLDWAV_TOOLTIP;Range of wavelet levels used throughout the Wavelets module.\n\n The numbers associated with Top-left and Bottom-left represent the minimum decomposition levels used. The numbers associated with Top-right and Bottom-right represent the maximum decomposition levels used.\n If the respective values ​​of Top-left and Bottom-left, or Top-right and Bottom-right, are different, a transition is applied.\n\n 0 - means a 2x2 pixels decomposition. \n 1 - means a 4x4 pixels decomposition.\n 2 - means a 8x8 pixels decomposition.\n 3 - means a 16x16 pixels decomposition.\n 4 - means a 32x32 pixels decomposition.\n 5 - means a 64x64 pixels decomposition.\n 6 - means a 128x128 pixels decomposition.\n 7 - means a 256x256 pixels decomposition.\n 8 - means a 512x512 pixels decomposition.\n 9 - means a 1024x1024 pixels decomposition.\n\n The values ​​8 and 9 assume, in order to be perceived in the Preview, that the size of the Preview is sufficient.
```
Lawrence37: - "Top-left", "bottom-left", "Top-right", and "bottom-right" are not names and do not need to be capitalized every time. Also add "the" before them.
- "Preview does not need to be capitalized.
- There are extra spaces after several newlines.

```suggestion
TP_LOCALLAB_WAT_THRESHOLDWAV_TOOLTIP;Range of wavelet levels used throughout the Wavelets module.\n\n The numbers associated with the top-left and bottom-left represent the minimum decomposition levels used. The numbers associated with the top-right and bottom-right represent the maximum decomposition levels used.\n If the respective values ​​of the top-left and bottom-left, or the top-right and bottom-right, are different, a transition is applied.\n\n0 - means a 2x2 pixels decomposition. \n1 - means a 4x4 pixels decomposition.\n2 - means a 8x8 pixels decomposition.\n3 - means a 16x16 pixels decomposition.\n4 - means a 32x32 pixels decomposition.\n5 - means a 64x64 pixels decomposition.\n6 - means a 128x128 pixels decomposition.\n7 - means a 256x256 pixels decomposition.\n8 - means a 512x512 pixels decomposition.\n9 - means a 1024x1024 pixels decomposition.\n\n The values ​​8 and 9 assume, in order to be perceived in the preview, that the size of the preview is sufficient.
```

rtdata/languages/default
```diff
@@ -4094,13 +4126,13 @@ TP_LOCALLAB_WAT_RESIDCOMP_TOOLTIP;Compresses the residual image to increase or r
...
 TP_LOCALLAB_WAT_WAVESHOW_TOOLTIP;Shows all of the 'Edge sharpness' tools. It is advisable to read the Wavelet Levels documentation.
 TP_LOCALLAB_WAT_WAVLEVELBLUR_TOOLTIP;Allows you to adjust the maximum effect of blurring on the levels.
-TP_LOCALLAB_WAT_WAVSHAPE_TOOLTIP;Low to high local contrast from left to right on the x-axis\nIncrease or decrease local contrast on the y-axis.
+TP_LOCALLAB_WAT_WAVSHAPE_TOOLTIP;Low to high local contrast from left to right on the x-axis\nIncrease or decrease local contrast on the y-axis.\n\n The system calculates, for each decomposed level, for each direction (horizontal, vertical, diagonal), and for positive or negative values, the value corresponding to the predominant signal. This corresponds approximately to the midpoint of the x-axis. The x-axis has no relation to luminance assessment, but only to the local contrast level. Manipulating the curve will amplify or reduce the signal: low to high local contrast from left to right on the x-axis.\n * ‘Attenuation response’ will reduce or increase the predominant area. \n * ‘Offset’ extends beyond the predominant area.
```
Lawrence37: - The first new sentence does not flow well.
- It is not clear what `luminance assessment` means. The first sentence already explains what the x-axis represents, so it's not even necessary. Same for the following sentence.
- Attenuation response and Offset are already explained in their respective tooltips. This tooltip does not add any new information.
- 
```suggestion
TP_LOCALLAB_WAT_WAVSHAPE_TOOLTIP;Low to high local contrast from left to right on the x-axis\nIncrease or decrease local contrast on the y-axis.\n\nFor each decomposed level, each direction (horizontal, vertical, and diagonal), and positive or negative values, the system calculates the value corresponding to the predominant signal. This corresponds approximately to the midpoint of the x-axis.
```

rtdata/languages/default
Desmis: ok

rtdata/languages/default
Desmis: OK, I change all with CIECAM

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: It's referring to the double-slider wavelet levels.

I chnage the tooltip... But is 'double' the right word ?

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: It is referring to the double-slider wavelet levels...

It’s a very old change, dating back almost to the creation of Local Adjustments :)

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: ok

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK

Desmis:
@Lawrence37 

Many thanks for this review. My English is definitely poor, and the translators aren't doing a good job.
Besides, English syntax is really different from French.
I think (or at least I tried) I’ve incorporated what you suggested.

Some comments date from that change (February, March, April 2026...) , but others go back much further - some at least 7 or 8 years.

This confirms for me that machine and/or AI translations are not (yet) sustainable solutions.

The other issue is whether what I meant is understood... Since no one makes any comments, I assume it is clear - though that isn't necessarily the case. Being practically on my own (and elderly, with very poor health) doesn't make things any easier.

I also harmonized CAL - for the 'red green blue' part - to make it consistent with the changes made to Selective Editing (lowercase, threshold, etc.)

If you have any further tweaks to make, or want to modify my latest changes, you can do so directly in the PR.

Thank you

Jacques

rtgui/tools/locallabtools2.cc
Desmis: OK

Desmis:
@Lawrence37 

I’ve made further additions based on your comment - specifically the shared 'max RGB' and 'max saturation' tooltips - and other changes for consistency in the history messages (Wavelets).

A few random remarks:
* There are inconsistencies - specifically regarding the "-" (and others) in the history entries - that have nothing to do with this change; however, changing everything would be a huge undertaking, and to what end?
* Drafting tooltips - especially for topics like wavelets, which few people fully master - is an incredibly complex task; this is particularly true for concepts such as vanishing moments, attenuation response, offset, clarity, etc., which sparked heated debates a few years ago (that's an understatement...). It is a choice between including nothing at all (as DrSlony once suggested) or providing an explanation... Certainly, there are - aside from my poor English - as many possible versions as there are developers and users. Much the same can be said for the concepts of AP (Abstract Profile), CIECAM, Game Changer, etc.
* And of course, the grass is always greener on the other side; the things I’ve heard about GIMP and wavelets being "simpler" than RT- it’s like comparing apples and oranges. It's a bit the same with DT...(Agx and GHS...)
* It is gratifying to see some users "discovering" features like those added to Capture Sharpening (pre-sharpening/denoise and post-sharpening/denoise). What we really lack is genuine communication. If I look at Rawpedia (the Hugo version), you certainly couldn't say - to use a French expression - that people are exactly queuing up at the gate (cela ne se presse pas au portillon). Likewise, videos are rare...

Of course, I remain open to any suggestions for general improvements, and I would like to thank you for the "administrative" work involved in tracking and building the RT versions (currently 5.13).

Thank you again.

Jacques

Desmis:
@Lawrence37 @digitalcarp 

I think I’ve finished two things.
* The first concerns Lawrence37’s comments - specifically regarding my poor English and the commit history... But nothing is perfect; beyond the syntax, there is the substance - specifically regarding tooltips (a vast subject).
* The second point concerns changes imposed by C++17; while unrelated to this PR, they had to be addressed - I wasn't about to let `iplocallab.cc` compile with over 100 warnings (or CodeQL/C++ alerts), as is the case in the `Dev` branch. It is a particularly tedious task.

Regarding your comments - and isn't it time to merge this (old) PR?

Jacques

digitalcarp:
I updated #7708 to fix the float/double conversion warnings on GCC 16 for `improcfun.cc` and `iplocallab.cc`. I did it slightly differently by using the `std::` version of functions instead of the `...f` version which detects the right type automatically.

Desmis:
@digitalcarp @Lawrence37 

**Does this mean that what I did isn't useful...** (the fix for Cpp17)? I had been using Windows with GCC 16 for a while, and the warnings weren't showing up...

With GCC16 in Windows11, no warning (or very little), with the recent introduction of Cpp17 about 100 warning ...which made the changes specific to this PR (secam16) unusable (or difficult to locate:  there are so many warnings, unrelated to the PR).
Note that I encountered the same Cpp17 warnings on Ubuntu 24.04 with GCC 14.

I am not a C++ expert (at all), but before making a change - whether by adding `std::` or `(float)` before the `double` function, or by using the dedicated `float` function - I checked online, and it was this third option (using the dedicated `float` function) that was recommended. But all of that is open to debate; I am not an expert at all, perhaps std:: is better ?Depending on the sites consulted, the answers vary.
**I did what I thought was best: using the float functions (e.g., sinf, logf, expf...).**

Another warning appears in iplocallab.cc at line 10127 in commit 510242d
    const auto dct_fw = fftwf_plan_r2r_2d(bfh, bfw, data_tmp, data_fft, FFTW_REDFT10, FFTW_REDFT10, FFTW_ESTIMATE | FFTW_DESTROY_INPUT);
I replaced it with
    fftwf_plan dct_fw;
    dct_fw = fftwf_plan_r2r_2d(bfh, bfw, data_tmp, data_fft, FFTW_REDFT10, FFTW_REDFT10, FFTW_ESTIMATE | FFTW_DESTROY_INPUT);

**If I had known you were going to make that change to `iplocallab.cc` for GCC16 (which was done after mine), I obviously wouldn't have done anything.**

So, what should I do? I'm not really sure how to undo it... Please let me know what I need to do. If a new PR needs to be recreated at commit 75d3b3d...or leave it as I did it.

If any 'reverses' need to be done, I suggest one of you does them, because otherwise I might mess things up.

Wouldn't an intermediate solution be to push this PR - which has the merit of eliminating the Cpp 17-related warnings in 'dev' - and then merge @digitalcarp PR relative to GCC16 ?


Jacques

digitalcarp:
I'll just revert my changes to `iplocallab.cc` since it's easier than making changes to ur branch.

Desmis:
@digitalcarp 
Thank you very much. It will still be possible to make your changes a little later.

Thanks again

Jacques

rtdata/languages/default
Lawrence37: image

Do you mean this thing? It's a threshold adjuster.
Also, the last bullet point has not been addressed.

rtdata/languages/default
```diff
@@ -3432,7 +3429,7 @@ TP_LOCALLAB_GAMMASKCOL;Gamma
...
+TP_LOCALLAB_GAMUTW_TOOLTIP; Placed at the end of Selective Editing > Color Appearance (CAM16 & JzCzHz), Compresses the colors into the chosen profile. It is applied after all previous spots and the other adjustments in this tool. \n\n Output gamut compression is much simpler than input gamut compression. Colors near the output stage of processing are rarely outside the working profile. A reduced set of adjustments is sufficient. .\n\n  Often, there is little to no change to the image when using Gamut Compression. This is because it does not convert the color profile, but compresses colors to fit the target gamut. Adjustments applied before compression usually keep colors in the target gamut.On normal landscapes and portraits, there is often no change.
```
Lawrence37:
```suggestion
TP_LOCALLAB_GAMUTW_TOOLTIP; Placed at the end of Selective Editing > Color Appearance (CAM16 & JzCzHz), compresses the colors into the chosen profile. It is applied after all previous spots and the other adjustments in this tool.\n\nOutput gamut compression is much simpler than input gamut compression. Colors near the output stage of processing are rarely outside the working profile. A reduced set of adjustments is sufficient.\n\nOften, there is little to no change to the image when using Gamut Compression. This is because it does not convert the color profile, but compresses colors to fit the target gamut. Adjustments applied before compression usually keep colors in the target gamut. The effect is visible on LEDs and vibrant flowers. On normal landscapes and portraits, there is often no change.
```

rtdata/languages/default
Lawrence37: For the first sentence, I would change "medium" to "middle" (if my understanding is correct).

Because there isn't a threshold adjuster visible to the user, it's better to just describe the logic directly. Even the current description is not clear to me. The top-right adjuster is where the effect starts tapering off. Why are there two top-rights? Does it mean local contrast is applied twice (once for all wavelet levels up to the middle ones and again for all wavelet levels up to the high ones)? Why does it say top-right values greater than a certain number as if it can be changed? There is no threshold adjuster, so the user cannot change the top-right value. Does RawTherapee internally change the threshold? Or do you mean to write something else?

rtengine/params/advanced.cc
```diff
@@ -0,0 +1,1557 @@
...
+    assignFromKeyfile(keyFile, "Color appearance", "M-Chroma", colorappearance.mchroma, pedited->colorappearance.mchroma);
+    assignFromKeyfile(keyFile, "Color appearance", "RSTProtection", colorappearance.rstprotection, pedited->colorappearance.rstprotection);
+    assignFromKeyfile(keyFile, "Color appearance", "J-Contrast", colorappearance.contrast, pedited->colorappearance.contrast);
+    assignFromKeyfile(keyFile, "Color appearance", "Q-Contrast", colorappearance.qcontrast, pedited->colorappearance.qcontrast);
+    assignFromKeyfile(keyFile, "Color appearance", "H-Hue", colorappearance.colorh, pedited->colorappearance.colorh);
+    assignFromKeyfile(keyFile, "Color appearance", "H-Hue-red", colorappearance.colorhred, pedited->colorappearance.colorhred);
+    assignFromKeyfile(keyFile, "Color appearance", "H-Hue-green", colorappearance.colorhgreen, pedited->colorappearance.colorhgreen);
+    assignFromKeyfile(keyFile, "Color appearance", "H-Hue-blue", colorappearance.colorhblue, pedited->colorappearance.colorhblue);
+    assignFromKeyfile(keyFile, "Color appearance", "H-Hue-blue", colorappearance.colorhblue, pedited->colorappearance.colorhblue);
```
Lawrence37: Duplicate line. Did you mean to change this to the brightness threshold?

Desmis: OK, Its a mistake, thank you

Medium, middle... why not, but it is the behavior that changes depending on the position of the decomposition levels.

This function change values in a function created about 2018 or 2019  (I don't remeber exactly...) in ipwavelet.cc

std::unique_ptr<LUTf> ImProcFunctions::buildMeaLut(const float inVals[11], const float mea[10], float& lutFactor)

This function complements the 'Attenuation response' but differs based on the levels of decomposition; the higher the 'Gradient levels' values, the broader the consideration of local contrast, and perhaps the generation of artifacts.
Depending on the choice of the maximum decomposition value, the levels affected by the function change.

The 'Contrast profile' slider in Abstract Profile (AP) > Contrast enhancement adjusts these parameters in a similar way, though naturally, with AP, the control is less fine-grained. 
It involves signal processing. I am taking advantage of having access to the decomposition levels (via a double slider) to allow the user to better isolate local contrast, depending on the specific images. The 'Selective Editing' version is similar to the one in AP, but offers more precise selection.

Your comments are of the same nature (this is not a criticism) as those made during the development of Wavelets a few years ago.

Jacques

rtdata/languages/default
Desmis: OK, but the TP_LOCALLAB_CLARI_TOOLTIP has been update after..If you use RawTherapee, instead of reading the code changes :)
You name it "threshold adjuster", but its is a 'double slider', at least, that is what it is often called.

TP_LOCALLAB_CLARI_TOOLTIP;This algorithm combines two decompositions of the same image and recombines them using different processes. It requires significant system resources.\n\nUsing the 'Wavelets levels' double slider as a reference:\n * from levels 0 to the bottom-right = 4 (included) 'Sharp mask' is enabled.\n * starting from levels 5 – the top-right or bottom-right and above: 'Clarity' is enabled.\n\nIn the case of the 'Sharp Mask' the change of level is similar to changing the radius in the 'Unsharp Mask'. For 'Clarity' a level change results in a greater or lesser three-dimensional volume effect on the image.

I just change (a little to sse a change) with a space before 'Using' and before 'In the case', to make it possible to see this change.
There has already been a debate - without a conclusion - regarding tooltips: should a space be added, an asterisk (*)...? Some advocate aligning everything, others staggering the lines, and others adding an asterisk or another symbol.

rtdata/languages/default
```diff
@@ -3250,7 +3273,7 @@ TP_LOCALLAB_CLARISOFT;Soft radius
...
 TP_LOCALLAB_CLARISOFT_TOOLTIP;The 'Soft radius' slider (guided filter algorithm) reduces halos and irregularities for Clarity, Sharp Mask and all wavelet pyramid processes. To deactivate, set slider to zero.
 TP_LOCALLAB_CLARITYML;Clarity
-TP_LOCALLAB_CLARI_TOOLTIP;Levels 0 to 4 (included): 'Sharp mask' is enabled\nLevels 5 and above: 'Clarity' is enabled.\nUseful if you use 'Wavelet level tone mapping'.
+TP_LOCALLAB_CLARI_TOOLTIP;This algorithm combines two decompositions of the same image and recombines them using different processes. It requires significant system resources.\n\n Using the 'Wavelets levels' double slider as a reference:\n * from levels 0 to the bottom-right = 4 (included) 'Sharp mask' is enabled.\n * starting from levels 5 – the top-right or bottom-right and above: 'Clarity' is enabled.\n\n In the case of the 'Sharp Mask' the change of level is similar to changing the radius in the 'Unsharp Mask'. For 'Clarity' a level change results in a greater or lesser three-dimensional volume effect on the image.
```
Lawrence37:
```suggestion
TP_LOCALLAB_CLARI_TOOLTIP;This algorithm combines two decompositions of the same image and recombines them using different processes. It requires significant system resources.\n\n Using the 'Wavelets levels' double slider as a reference:\n * from levels 0 to the bottom-right = 4 (included) 'Sharp mask' is enabled.\n * starting from levels 5 – the top-right or bottom-right and above: 'Clarity' is enabled.\n\n In the case of the 'Sharp Mask' the change of level is similar to changing the radius in the 'Unsharp Mask'. For 'Clarity' a level change results in a greater or lesser three-dimensional effect on the image.
```

rtdata/languages/default
```diff
@@ -3492,6 +3516,8 @@ TP_LOCALLAB_GRADSTRHUE;Hue gradient strength
 TP_LOCALLAB_GRADSTRHUE2;Hue gradient strength
 TP_LOCALLAB_GRADSTRHUE_TOOLTIP;Adjusts hue gradient strength.
 TP_LOCALLAB_GRADSTRLUM;Luma gradient strength
+TP_LOCALLAB_GRADWAV;Gradient levels
+TP_LOCALLAB_GRADWAV_TOOLTIP;Adjust local contrast for high and middle decomposition levels, referring to the double-slider wavelet levels.\n\nMiddle corresponds to a top-right value greater than 3 (larger than 16x16 pixels). High corresponds to a top-right value greater than 5 (larger than 64x64 pixels).\n\nThis function complements the 'Attenuation response' but differs based on the levels of decomposition; the higher the 'Gradient levels' values, the broader the consideration of local contrast.
```
Lawrence37: After reading the code and playing around with the "Gradient levels" adjuster to confirm my understanding, I am even more convinced that this tooltip is confusing and possibly incorrect.

My understanding is that Gradient levels:
- Adjusts the strength of the local contrast adjustment made by the "Local contrast" curve
- It is very roughly a factor. 1 means no adjustment. <1 means reduce the effect of the Local contrast curve. >1 means enhance the effect of the Local contrast curve.
- It only applies to wavelet levels 3 and higher (I'm assuming the numbers in `inva3` through `inva10` refers to the wavelet level)
- It only has an effect if the "Wavelet levels" threshold adjuster's top-right value is greater than 3
- It is only half as effective when the top-right value is less than 6 (but there is a built-in strength reduction when the top-right value is 6 or greater, which makes it hard to see this fact if you want to try it for yourself)

If I were to summarize, I would say
"Controls the strength of the Local contrast adjustment for wavelet levels 3 and higher. Has full effect when the Wavelet levels threshold adjuster's top-right value is at least 6, half the effect for values 4 and 5, and no effect for values 3 or lower."

rtdata/languages/default
Desmis: OK

rtdata/languages/default
Desmis: OK, It's clearer while preserving the meaning.

PR #7631, opened 2026-03-22
CLANG64 for Windows x86_64 workflows

Lawrence37:
MINGW64 is deprecated: https://www.msys2.org/news/#2026-03-15-soft-deprecating-the-mingw64-environment
This replaces MINGW64 with CLANG64.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -30,8 +30,8 @@ jobs:
         build_type: [release, debug]
         arch: [x86_64, arm64]
     env:
-      MSYS2_ENV_NAME: ${{ matrix.arch == 'arm64' && 'CLANGARM64' || 'MINGW64' }}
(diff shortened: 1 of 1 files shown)


Lawrence37:
UCRT64 is also a possibility. I'm not sure which one to go with. On one hand, CLANG64 makes the x86 build similar the the ARM build. On the other hand, UCRT64 is more similar to MINGW64 which we have been using for quite some time.

digitalcarp:
I've actually been using ucrt64 for a while now but I'm ambivalent.

Lawrence37:
The binary created by gcc is larger than the one created by clang, especially the debug build. However, the gcc build was roughly 5% faster on one image I tested. There's no clear advantage for CLANG64 or UCRT64, so I will play it safe by using UCRT64, which uses gcc.

PR #7640, opened 2026-04-01
Improve clarity of power-*-small icons

### LINKED ISSUE #7167: clearer on/off button state
A bit more contrast between the modules on/off buttons states icons would be welcome.

I have a small screen and cant tell if the modules are on/off easily so making it a bit bigger, color codding it or adding more contrast could help a lot!

digitalcarp:
- Decrease width of power-off icon
- Make power-inconsistent exclamation mark color consistent with power-on

icons

batch_edit

Closes #7167


kaesa:
Tested on few themes and looks good to my eyes. I don't know where the "inconsisent power" is used though.

digitalcarp:
It's used in the batch editor when you select two images with different settings.

Lawrence37:
The issue report did not show how the current icons look. We cannot be sure that this will solve the issue.

digitalcarp:
Maybe the "off" symbol can just be a circle? See https://en.wikipedia.org/wiki/Power_symbol

PR #7643, opened 2026-04-02
Add colors to crop guide presets

### LINKED ISSUE #6338: Color of the guide type
Hi,
I am talking about the module: Crop -> Guide Type.
When the image is clear sometimes the grid is not visible (or is very little visible), because it is confused with the color of the image.
I would suggest the possibility that the user can change the grid color in the same module (for example to dark, or green or red...).
I am using the Rawtherapee-Legacy theme.
Examples:
Captura1

Captura2

Arturo.

digitalcarp:
The previous change only added colors to aspect ratio crop guides. This adds it for presets as well to close #6338.

It also allows for changing the alpha of the crop guide color. The alpha is not displayed in the color preview.

### DIFF
--- a/rtengine/procparams.cc
+++ b/rtengine/procparams.cc
@@ -330,12 +330,38 @@ void loadCropGuideParams(
 
     assignFromKeyfile(keyFile, group, TOOL_ENABLED, params.enabled, edited.enabled);
 
+    auto parse_color = [&](const cJSON* obj, const char* key, double& out) {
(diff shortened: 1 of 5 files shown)


kaesa:
Seems to work. I was thinking about this in the original PR too. Agreed that this is important for making the new crop guides complete. It would be kind of funny and frustrating to the users perspective if the inscribed guides would have more features than the guides the user would actually like to use.

Lawrence37 (approved):
Thank you @digitalcarp for this improvement (and thank you @kaesa for testing)!

PR #7644, opened 2026-04-05
Canon EOS R6 Mark III camconst masked areas (and raw crop)

### LINKED ISSUE #7574: EOS R6 Mark III
Hello, 

RAW and C-Raw samples have been posted for the Canon EOS R6 Mark III. 

Please let me know if you need more samples

Lawrence37:
Adds masked areas to fix the black level.

Also adds raw crops, but they have no effect due to the more aggressive crop applied by the LibRaw function currently used by RawTherapee.

Closes #7574.

### DIFF
--- a/rtengine/camconst.json
+++ b/rtengine/camconst.json
@@ -1331,6 +1331,21 @@ Camera constants:
     },
 
     { // Quality C
+        "make_model": ["Canon EOS R6 Mark III"],
+        "raw_crop": [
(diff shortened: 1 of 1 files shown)


kmilos:
> due to the more aggressive crop applied by the LibRaw function currently used by RawTherapee

Note that LibRaw currently doesn't fully support the EOS R6 Mark III yet, so who knows what crop it is falling back to internally...

Issue #7564, opened 2025-12-18
Sony A7V support (ILCE-7M5)


wutscher:
I submitted sample images for the A7V to https://raw.pixls.us/ because I didn't find any existing samples. If you need any more infos, I'd be happy to help

Lawrence37:
Basic support will be provided by #7645. Colors already look similar to the DCP from Adobe DNG Converter.

PR #7645, opened 2026-04-05
Sony ILCE-7M5 raw crop and white level

Lawrence37:
Partially addresses #7564.


kmilos:
Note that LibRaw currently doesn't fully support the A7V yet, so who knows what color matrix it is falling back to internally...

PR #7646, opened 2026-04-05
Fix raw crop dimensions message

Lawrence37:
The crop width and height are `width` and `height`, not `iwidth` and `iheight` as evidenced in `compress_image()`.


digitalcarp:
LGTM

PR #7671, opened 2026-04-19
Add Hald CLUT LUT export button to the editor

NicoNex:
Adds a "LUT" button to the editor toolbar that exports the current colorimetric adjustments as a Hald 12 CLUT PNG file.                        
                                                                                                                                                 
A Hald identity image is generated in-engine using Imagefloat, processed through a colorimetric-only subset of the pipeline (white balance, tone curves, color grading, etc.), and saved as a 16-bit PNG. 
Spatially-dependent tools (sharpening, noise reduction, geometric transforms, local adjustments, etc.) are automatically disabled for the export.

The save dialog follows the same pattern as the existing Save As dialog (Gtk::Dialog + embedded Gtk::FileChooserWidget).

This feature is useful in two main scenarios:
- Film simulation presets: export a RawTherapee look (film curves, color grading, white balance) as a portable LUT that can be applied in other applications or shared with others.
- In-camera LUT support: some modern cameras (e.g. Panasonic Lumix S9) can load custom LUTs and apply them at capture time for JPEG output and live view. This allows photographers to develop a look in RawTherapee and carry it into their camera workflow.


NicoNex:
Moved to #7674.

PR #7689, opened 2026-05-09
Removes black point line from history when Automatic Black point and White point is enabled.

kaesa:
This changes the history reporting so that it will first report Preprocessing and just on the second round of processing Enabled. Previously the code printed the black point value to history to launch the second processing round. This was wrong in many ways: 1. The value was printed regardless if it changed or not. 2. The White point was never printed although it might have been the value that actually changed.

I tried to use existing labels to avoid opening translations. I'm not sure if Preprocessing fits for the purpose or not, but it sounded ok.

There is a case if the user clicks the checkbox to enabled and then changes another value fast enough, the history will have Preprocessing and the user changed value under it. There will be Enabled printed to the history eventually. If the user then clicks the row labeled Preprocessing, the enabled checkbox will be checked but the values are not the automatically determined values. I think the Preprocessing label may be enough for the user to forgive the inconsistency in values.

### DIFF
--- a/rtgui/tools/locallabtools.cc
+++ b/rtgui/tools/locallabtools.cc
@@ -5898,7 +5898,7 @@ void LocallabShadow::updateghsbw2(double ghsb, double ghsw, bool ghsaut)//auto G
                     ghs_BLP->setValue(ghsb);
                     enableListener();
                 }
-                listener->panelChanged (Evlocallabghs_BLP,ghs_BLP->getTextValue());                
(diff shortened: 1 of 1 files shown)


Desmis:
@kaesa 
It's not a very simple problem to solve for "history".

I think what you did using "TP_PREPROCESS_LABEL" is a clever workaround.

From my point of view, it's satisfactory.

Lawrence37 (approved):

Lawrence37:
Thank you @kaesa!

PR #7716, opened 2026-06-28
Color Appearance & Lighting RGB curve fix

Desmis:


Lawrence37:
@Desmis I think something else may have fixed the RGB curves issue. I tried this branch and it still has the issue. Then I tried the secam16 branch which is ok.

Desmis:
@Lawrence37 

I didn't know which 'brightness' curve was being discussed. If it's the one related to 'Red Blue Green', it is obvious that it isn't affected by the latest modification in 5.13.
In fact, the 'secam16' branch was originally created for this modification, which alters 'Q' brightness curves (in the CIECAM sense) rather than RGB curves (it's not at all a RGB curve)

In secam16, I also modified the equivalent setting under Selective Editing > Color Appearance CAM16, and others modifications to be equivalent to Abstract Profile.

In secam16, there is also the modification to make SE work with thumbnails.

I had proposed this branch for 5.13 because, in my view, it offered undeniable benefits, as outlined above.

I don't think there's any issue with merging 'secam16', though we could also take a different approach and only include specific parts.

Just so you know, I am not at home but in a remote mountain village. The internet connection is unreliable, and I only have my laptop with me.

jacques

Lawrence37:
It's the curves under Advanced tab > Color Appearance & Lighting > Image Adjustments > Red Green Blue > Brightness - red/green/blue
image
Notice the effect on the brick. It seems to change the brightness of all pixels with a green or close to green hue equally. There is no smooth transition as the hue gets further away from green.

I'm hesitant to include the secam16 branch because it is a large change (almost 2000 new lines and multiple enhancements).

Desmis:
@Lawrence37 

Yes, it's the red-green-blue modification. In some cases not all (often in sky)  when you adjust the curve, it introduces a change equivalent to a Delta E value, which is accounted for by two code modifications: a slider (threshold) and a measurement of the chroma variation (specifically Cpro).

Lawrence37:
I copied the changes to `improcfun.cc` from commit 94d731f (but hard-coded the threshold because the adjuster does not exist in 5.13-rc1) and it fixed the issue I described.

PR #7724, opened 2026-07-09
 testing rotation for local spots

jonathanBieler:
This was done by Claude Fable (it one-shot it), so it's maybe more a proof of principle than a mergeable PRbut I think it's a important missing functionality for the local edits, and if anything this can be used as inspiration. This PR adds a handle to local spots that allow to rotate the spot, which in my experience would be very useful for image with objects that are not aligned with the x/y axis of the image. The code looks reasonable (I probably wouldn't be able to write better code), although I didn't review it in great details. I didn't manage to find issues while testing, fall-off, delta and tools work.

Capture

Capture

This is some notes from Claude :

### Maths

Coordinate rotation (engine). The spot shape is kept axis-aligned in its own local frame; rotation is applied by transforming each pixel into that frame rather than rotating the shape. For a pixel p = (x, y), spot center c = (xc, yc) and angle θ:

d  = p − c                        // offset from center
d' = R(−θ) · d                    // inverse rotation
   = ( cosθ·dx + sinθ·dy,
      −sinθ·dx + cosθ·dy )

The existing tests then run unchanged on c + d': the quadrant is picked by the signs of d' (selecting the proper semi-axis pair from shlx/shlxL/shly/shlyT), the standard normalized distance (dx'/rx)² +(dy'/ry)² < 1 (inner ellipse scaled by ach = transit/100 for zone 2, annulus up to 1 for the transition zone). Rotating the point by −θ is equivalent to rotating the shape by +θ, and since the feather kernels also receive the rotated coordinates, the transition falloff rotates automatically.

Bounding box. Call sites pre-clip loops with axis-aligned bounds built from lp.lx/ly/..., so those fields now hold an AABB that encloses the rotated shape. With a = max(shlx, shlxL), b = max(shly, shlyT):                                
bbX = a·|cosθ| + b·|sinθ|
bbY = a·|sinθ| + b·|cosθ|                                                                                             
This is the exact AABB of the rotated rectangle [−a,a]×[−b,b], hence a safe (slightly conservative) superset for the ellipse too — the exact ellipse AABB would be √(a²cos²θ + b²sin²θ). Over-covering only costs a few extra skipped       pixels; under-covering would truncate the shct. At θ = 0 it degenerates to (a, b), andper-side asymmetry collapses to the max of each pair, preserving the θ = 0 fast path exactly (fields are copied 1:1 in that case).

GUI forward rotation. The preview draws in image space, so handle/corner offsets use the forward matrix R(+θ): (x, y)  → (cosθ·x − sinθ·y, sinθ·x + cosθ·y) around s the ellipse with the transform chain translate(center) → rotate(θ) → scale(rx, ryer quadrant.
                                                                                                                       Handle interactions. Resizing at angle θ pro) onto the rotated axes with dot products:along-X′ is dx·cosθ + dy·sinθ, along-Y′ is dy·cosθ − dx·sinθ (the unit vectors being (cosθ, sinθ) and (−sinθ, cosθ)). Rotation drag: the handle rests at direction (0, −1) ("up") from the center, so θ = atan2(dy, dx) + 90°, normalized to (−180°, 180°]. Y is inverted vs. math convenwnward — positive θ appears clockwise onscreen, consistently in both engine and GUI.

### Code changes

Adds a rotation angle to Local Adjustments spots (ellipse/rect), controlled by an on-preview handle.

Parameters
- LocallabSpot::spotangle (double, degrees, −180..180, default 0) in rtengine/params/locallab.{h,cc}: ctor default, operator==, keyfile I/O as Spotangle_<i>. Edited flag added in rtgui/paramsedited.{h,cc} (4 sites); SpotRow bridge in rtgui/tools/locallab.cc.
- New event EvLocallabSpotRotation registered dynamically via ProcEventMapper::newEvent(AUTOEXP, "HISTORY_MSG_LOCAL_SPOTANGLE") in ControlSpotPanel — not appended to the static refreshmap/procevents tables (those are misaligned upstream: 1149 initializers for 1150 events, so appended entries land one slot short and get action 0).

Engine (rtengine/iplocallab.cc)
- local_params gains true shape semi-axes shlx/shlxL/shly/shlyT and angcos/angsin. The legacy lx/lxL/ly/lyT are redefined as the rotated shape's axis-aligned bounding box (identical when angle = 0), so all ~20 call sites' bounding-box clipping stays valid untouched.
- calcTransition/calcTransitionrect inverse-rotate (x,y) around the spot center, then run the existing quadrant logic against the shape semi-axes; feather kernels excluding spots, and mask previews; deltaEscope is orthogonal.
- The two post-calcLocalParams mutation siteT-size shrink scales the semi-axes uniformlyto fit the reduced box; full-image override

GUI                                                                                                                 - Ellipse (editwidgets) gains angle; the 4× ks are factored into one drawEllipsePathhelper using translate→rotate→scale (also re).
- ControlSpotPanel: per-spot geometry goes 7→8 objects (kSpotGeomCount); the rectangle outline becomes a Polyline of rotated corners (EditRectangle is axis-aligned); resize handles rotate with the shape and drag deltas are projected onto the rotated axes; new rotation handle (-aroundnode, hicontrast variant on hover,CSMoveRotate cursor) placed above the top edge at decayYT/8, drag sets angle via atan2 (rest position = up, normalized to ±180°). No panel slider — the angle lives only in the spot row / procparams.

Compatibility: profiles without Spotangle_ render byte-identically (verified via CLI diff); 45° changes only the spot
mask footprint.


Desmis:
@jonathanBieler 
I am still working with limited resources (computer, highly unstable internet connection, etc.).

**I am surprised by the efficiency of Claude and your work**. It is an undeniable asset for Selective Editing. However, my code (this part dates back to around 2016) is not a model of clarity. Claude correctly identified the key elements of the system I designed: the values ​​needed for the 'calculation rectangle,' followed by the changes required for the 'transition' to work.

The initial tests I ran show no faults, but further checks are needed.

Perhaps, but I don't see a simple alternative; the rotation widget could be placed differently ?

**This proposal should be accepted**; for the record, it serves to:
* make the correspondence between the spot and the object more intuitive (it is essentially psychological)
* **Improve transitions for non-vertical and non-horizontal objects – this is the key improvement.** . That is the reason why the 'Symmetry X & Y' option exists in 'Transition gradient' (admittedly, it is not as good).
* It provides a slight improvement - through better matching of the object/spot to the Delta E - but the effect will be minimal. That is why I haven't really sought to improve this method.

Obviously, this only works for 'Normal Spot' and 'Excluding Spot'.

Thank you so much :)

Jacques

jonathanBieler:
Yeah I was surprised it did such a good job, specially since the local edit file is so large it cannot read the whole thing. I'm not sure if Sonnet would have done it so easily, but I haven't tested.

Desmis:
@jonathanBieler 

In any case, it works perfectly; I had given up on doing it. I didn't ask for a check with Copilot - after all, what other AI would correct it? Besides, the modifications - apart from the GUI - are simple... and mainly affect the transitions.

Thanks again.

kaesa:
I second this feature would be useful for me too. I have often wished there was an ability to rotate ellipses when masking eyes. I'll link a request from the forum https://discuss.pixls.us/t/mask-functionality-request/60727, in which to my interpretation of the text, the user would find even this relatively small modification a life improvement.

PR #7725, opened 2026-07-12
CAL RGB brightness curves threshold

Lawrence37:
Eliminate hard transitions when using the Color Appearance & Lighting RGB brightness curves.

Addresses the behavior seen in https://discuss.pixls.us/t/rawtherapee-5-13-release-candidate-1/58782/6 and is meant to be compatible with #7629


Desmis:
@Lawrence37 

For me it's OK

PR #7675, opened 2026-04-21
Add .cube LUT support to Film Simulation

NicoNex:
## Summary                                                                                                                                     
                                                                                                                                                 
  Adds support for Adobe/DaVinci Resolve `.cube` LUT files in the Film Simulation tool, alongside the existing Hald CLUT (PNG/TIFF) format and addresses issue #5704.      
   
  ### Engine changes (`rtengine/`)                                                                                                               
  - Introduces a `CLUT3D` abstract base class holding the shared flat `uint16` RGB buffer, trilinear `getRGB()` interpolation (including SSE
  path), and common metadata (`filename`, `profile`)                                                                                             
  - `HaldCLUT` and the new `CubeLUT` are concrete subclasses of `CLUT3D`
  - `CubeLUT::load()` parses text-based `.cube` files: `LUT_3D_SIZE`, `DOMAIN_MIN`/`DOMAIN_MAX`, and N³ RGB float triplets (R-fastest order)     
  - `CLUTStore::getClut()` now acts as a factory, selecting `HaldCLUT` or `CubeLUT` automatically based on the file extension                    
  - `improcfun.cc` updated to use `std::shared_ptr<CLUT3D>` instead of `std::shared_ptr<HaldCLUT>`                                               
                                                                                                                                                 
  ### GUI changes (`rtgui/`)                                                                                                                     
  - Film Simulation file browser now includes `.cube` in its extension filter alongside `.png` and `.tif`

Issue #5704, opened 2020-04-08
please add .cube LUT support.


PhilBaz:
Please add .cube LUT support. Darktable offers .cube in addition to HaldCLUT.

It is a common LUT format and would be a great addition to the program.

PhilBaz:
Is there anyone I can buy a couple of beers to get this one bumped up ;)

abnormally-distributed:
Is there any chance of this happening? It would be nice to be able to use the more space-efficient cube file format, and also to not have to have second copies of all my luts as pngs just to use in RT (everything else I have that can utilize luts allows cube files)

abnormally-distributed:
Bumping this again ;)

PR #7676, opened 2026-04-21
Add .cube LUT support and HALD/.cube LUT export

NicoNex:
## Summary                                                                                                                                     
  
  Depends on #7674 (Hald CLUT export) and #7675 (.cube Film Simulation                                                                           
  support), which are independent of each other. This PR combines both
  and extends the result.                                                                                                                        
                                                                                                                                                 
  ### What this PR adds                                                                                                                          
                                                                                                                                                 
  **`.cube` export**
  - `CubeLUT::createIdentityTempFile()` generates the identity image fed
    into the processing pipeline for `.cube` export                                                                                              
  - `CubeLUT::saveAsCubeFile()` reads the processed image and writes a                                                                           
    standard `.cube` text file (TITLE, LUT_3D_SIZE, DOMAIN, N³ entries)                                                                          
                                                                                                                                                 
  **LUT export dialog improvements**
  - Format selection via drop-down (Hald CLUT PNG / Cube LUT .cube),                                                                             
    mirroring the Save As dialog pattern                                                                                                         
  - Format-specific options (Hald level: 8–16; cube size: 17/33/65)
    show and hide based on the selected format, rather than being                                                                                
    permanently visible or just greyed out                                                                                                       
  - File filter and filename extension update automatically on format change                                                                     
  - Hald CLUT filter now includes `.tif`/`.tiff`                                                                                                 
  - Default filename derived from the source image name (`<name>_lut`)                                                                           
  - Tone curve excluded from the exported LUT by default                                                                                         
  - Configurable Hald level (previously hard-coded to 12)

PR #7726, opened 2026-07-14
Add .cube LUT support to Film Simulation

wilx:
## Summary

Adds 3D Adobe/DaVinci Resolve `.cube` LUT support to Film Simulation alongside the existing Hald CLUT formats, addressing #5704.

This work builds on the original implementation by Nicolò Santamaria (@NicoNex) in #7675. His original commit and authorship are preserved in this branch. The implementation has subsequently been extended to follow Cube semantics more closely, preserve valid intermediate colour values, and support technical LUTs whose transfer function or gamut does not directly match RawTherapee’s Film Simulation pipeline.

## Why Cube LUTs use float storage

Cube outputs are not restricted to `[0,1]`. Negative values and values above 1 are valid intermediate results used for gamut conversion, log transforms, and highlight headroom. Clamping them while loading changes the transform before RawTherapee can convert the result back into its working profile.

For example, a baked Fujifilm 65³ F-Log2-to-F-Log2 LUT contains 50 negative channel samples and 65,195 samples above 1, reaching 1.595. Reala Ace, ETERNA, and WDR remain within `[0,1]`, so retaining extended values does not alter those ordinary creative LUTs.

Cube vertices are therefore stored as aligned RGBX floats. Values inside `[0,1]` retain the existing inverse-sRGB table path; extended values use the corresponding analytic inverse transfer. Profile conversion remains unclipped, allowing intermediate values to return to the displayable range later in the pipeline.

The memory cost is approximately 0.55 MiB for a 33³ LUT and 4.2 MiB for a 65³ LUT; the specification maximum of 256³ requires about 256 MiB. Hald CLUTs remain on their original `uint16_t` path and produce unchanged output.

## Changes beyond the original implementation

- Applies `DOMAIN_MIN/MAX` to input coordinates and validates finite, ordered bounds.
- Enforces the 2–256 grid-size range with overflow-safe allocation and bounded parsing.
- Accepts leading whitespace on keyword lines.
- Uses Adobe-compatible tetrahedral interpolation for Cube LUTs while retaining Hald’s trilinear interpolation.
- Provides matching scalar and SSE2/SIMDe Cube interpolation paths.
- Preserves finite Cube outputs below 0 and above 1 through strength blending, inverse transfer, and profile conversion.
- Includes and installs `bake_cube_lut.py`, a standard-library Python tool for adapting technical LUTs to RawTherapee’s Film Simulation pipeline.
- The baker converts the source transfer function and gamut into RawTherapee’s `_Rec2020.cube` convention and tetrahedrally resamples the source LUT to the requested grid size. This allows LUTs designed for inputs such as Fujifilm F-Log2/F-Gamut to be used without adding vendor-specific transforms to the runtime processing pipeline.
- Supports automatic Fujifilm metadata recognition and explicit input/output gamut and transfer options for LUTs from other vendors.
- Includes usage documentation and standard-library tests for the baking workflow.

## Validation

Tested native, generic SSE2, and SIMDe builds. Synthetic LUTs cover all six tetrahedra, boundaries, endpoints, domains, extended values, and 0%, 50%, and 100% strength. Identity Cube output and 0% Film Simulation match their references, while existing Hald output remains pixel-identical.

The baker tests parsing, domains, transfer-function round trips, metadata recognition, tetrahedral interpolation, identity composition, resampling, and extended output. Representative Fujifilm Reala Ace, ETERNA, WDR, and F-Log2-preserving LUTs were also baked and processed through `rawtherapee-cli`.

Film Simulation input remains limited to `[0,1]`; 1D/shaper LUTs, tone mapping, and final-export clipping policy remain outside this PR.

---

AI assistance: This work was developed with assistance from OpenAI Codex / ChatGPT.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 .idea
 .env
 .venv
+__pycache__/
+*.pyc
 
 CMakeUserPresets.json
 CMakeCache.txt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -470,6 +470,8 @@ install(
     FILES RELEASE_NOTES.txt
     DESTINATION "${CREDITSDIR}"
     OPTIONAL)
+install(FILES tools/bake_cube_lut.py
+        DESTINATION "${DATADIR}/tools")
 
 # The standard location for man pages in Linux is /usr/share/man Use "manpath"
 # to see the search paths for man pages on your system.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -471,6 +471,7 @@ install(
     DESTINATION "${CREDITSDIR}"
     OPTIONAL)
 install(FILES tools/bake_cube_lut.py
+              tools/bake_cube_lut-README.md
         DESTINATION "${DATADIR}/tools")
 
 # The standard location for man pages in Linux is /usr/share/man Use "manpath"
--- a/rtengine/clutstore.cc
+++ b/rtengine/clutstore.cc
@@ -1,5 +1,6 @@
 #include <algorithm>
 #include <array>
+#include <cmath>
 #include <fstream>
 #include <sstream>
 #include <vector>
@@ -27,7 +28,7 @@ namespace
 // ---------------------------------------------------------------------------
 
 #if defined(__SSE2__) || defined(RT_SIMDE)
-vfloat2 getClutValues(const AlignedBuffer<std::uint16_t>& clut_image, size_t index)
+vfloat2 getHaldClutValues(const AlignedBuffer<std::uint16_t>& clut_image, size_t index)
 {
     const vint v_values = _mm_loadu_si128(reinterpret_cast<const vint*>(clut_image.data + index));
 #ifdef __SSE4_1__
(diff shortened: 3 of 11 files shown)


NicoNex:
Hey, thanks for the work you put into this, all much more solid than my original implementation.

Just wanted to give you some context: I have another PR open #7676 which builds on #7675 and adds .cube/Hald export. Might be worth coordinating so we don't duplicate work or end up with unnecessary conflicts, I could pull your commits into the #7676 branch so the export keeps working on top of the updated version. I'll likely just close #7675 in favor of #7676 and your #7726, since that's the state of the art for this feature at this point, rather than keeping three overlapping PRs around. Let me know if that works for you, or if you already had something else in mind.

wilx:
> Hey, thanks for the work you put into this, all much more solid than my original implementation.
> 
> Just wanted to give you some context: I have another PR open #7676 which builds on #7675 and adds .cube/Hald export. Might be worth coordinating so we don't duplicate work or end up with unnecessary conflicts, I could pull your commits into the #7676 branch so the export keeps working on top of the updated version. I'll likely just close #7675 in favor of #7676 and your #7726, since that's the state of the art for this feature at this point, rather than keeping three overlapping PRs around. Let me know if that works for you, or if you already had something else in mind.

I am fine with this. All I want is to eventually use the cube files for Fujifilm Eterna film sims properly.

PR #7739, opened 2026-08-16
rtengine: Remove rtgui includes

hfiguiere:
So that building rtengine rely less on rtgui.
Added rtengine/labgrid.h for the interesting symbols.

### DIFF
--- a/rtgui/labgrid.h
+++ b/rtgui/labgrid.h
@@ -41,16 +41,10 @@
 #include <gtkmm.h>
 
 #include "rtengine/improcfun.h"
+#include "rtengine/color.h"
 
 #include "eventmapper.h"
 #include "toolpanel.h"
-
-namespace rtengine
-{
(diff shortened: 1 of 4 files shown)


rtgui/labgrid.h
```diff
@@ -41,17 +41,11 @@
...
 #include "eventmapper.h"
 #include "toolpanel.h"
 
-namespace rtengine
-{
-    //OFFSET_MODIFIER and CIExy_MARGIN - Regarding the conversion of xy data to Labgrid coordinates
-    constexpr float OFFSET_MODIFIER = 1.81818f;//Scaling coefficient of primary data and CIExy diagram with that of Labgrid.
-    constexpr float CIExy_MARGIN = 0.1f;//corresponds to the left and bottom margin on the CIExy diagram
-}
-
```
digitalcarp, line 54: Instead of creating a new file, it might be better to place this in an existing one like `rtengine/color.h`

hfiguiere: done

PR #7749, opened 2026-09-01
Avoid batch queue preview update deadlocks

zacharyvmm:
## Summary

Batch queue preview generation can deadlock while restoring a populated queue. The updater worker enters GTK after taking the preview lock, while drawing and layout can take those locks in the opposite order.

This schedules preview application on the GTK idle loop in [`BatchQueueEntry::updateImage`](https://github.com/zacharyvmm/RawTherapee/blob/a534892d3d910ea950c6f01456dbbb7f66924080/rtgui/batchqueueentry.cc#L205-L216). [`BatchQueueEntry::_updateImage`](https://github.com/zacharyvmm/RawTherapee/blob/a534892d3d910ea950c6f01456dbbb7f66924080/rtgui/batchqueueentry.cc#L218-L243) holds the preview lock only while replacing image data and requests the redraw after releasing it.

Entry destruction cancels queued idle work. [`BatchQueueEntryUpdater::removeJobs`](https://github.com/zacharyvmm/RawTherapee/blob/a534892d3d910ea950c6f01456dbbb7f66924080/rtgui/bqentryupdater.cc#L190-L212) also waits for an active callback before the listener is destroyed, closing the race between scheduling the idle callback and removing the entry.

## Validation

Built the current commit in Release mode on Apple Silicon macOS. Restored a copied 63-item queue, waited for its previews to load, and scrolled through the queue. The GTK main loop remained responsive with no batch queue lock wait. This repository has no configured CTest tests.

### DIFF
--- a/rtgui/batchqueueentry.cc
+++ b/rtgui/batchqueueentry.cc
@@ -65,6 +65,7 @@ BatchQueueEntry::~BatchQueueEntry ()
 {
 
     batchQueueEntryUpdater.removeJobs (this);
+    idle_register.destroy ();
 
     if (opreview) {
         delete [] opreview;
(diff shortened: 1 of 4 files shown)


rtgui/batchqueueentry.cc
```diff
@@ -204,33 +205,41 @@ std::tuple<Glib::ustring, bool> BatchQueueEntry::getToolTip (int x, int y) const
...
                                    int origw, int origh, guint8* newOPreview)
 {
-    // since the update itself is already called in an async thread and there are problem with accessing opreview in thumbbrowserbase,
-    // it's safer to do this synchronously
-    {
-        GThreadLock lock;
-
-        _updateImage(img, size, deviceScale);
-    }
+    // Updating from the worker under GThreadLock can deadlock with drawing or
```
kaesa: Can you remove the comment please. GThreadLock is deprecated and will eventually get completely removed from the entire source. The comment is describing a resolved bug and removal of deprecated code, which is something that will just get ignored by future coders.

rtgui/bqentryupdater.cc
```diff
@@ -173,6 +176,11 @@ void BatchQueueEntryUpdater::processThread ()
             delete[] current.oimg;
             current.oimg = nullptr;
         }
+
+        qMutex->lock ();
+        activeListener = nullptr;
+        inactive.notify_all ();
+        qMutex->unlock ();
```
digitalcarp, line 183: Nit: Slight performance improvement if you notify outside the lock scope

rtgui/batchqueueentry.cc
```diff
@@ -204,33 +205,39 @@ std::tuple<Glib::ustring, bool> BatchQueueEntry::getToolTip (int x, int y) const
...
 {
-    // since the update itself is already called in an async thread and there are problem with accessing opreview in thumbbrowserbase,
-    // it's safer to do this synchronously
-    {
-        GThreadLock lock;
-
-        _updateImage(img, size, deviceScale);
-    }
+    idle_register.add(
+        [this, img, size, deviceScale] () -> bool
```
kaesa, line 209: IdleRegister can leak maybe few megabytes of memory if the callback is cancelled before execution. See the `_updateImage` function body, and how the `img` is deleted there. More precisely, if the job is being cancelled after it has finished but before the UI thread has finished with the update. The risk is increased by the choice of using `G_PRIORITY_LOW`, which can keep pushing the execution back while higher priority events are being processed. Using `std::unique_ptr` with `std::move` would solve the leakage here.

